PR #7, opened 2019-04-05
Initial commit of code for creating tenant and namespaces state in the

srampal:
controller and in the cluster  (basically works but further cleanup and
functionality to be added in subsequent commits/ PRs)

### DIFF
--- a/poc/tenant-controller/pkg/controllers/tenants/controller.go
+++ b/poc/tenant-controller/pkg/controllers/tenants/controller.go
@@ -20,21 +20,27 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8srt "k8s.io/apimachinery/pkg/runtime"
 	utilrt "k8s.io/apimachinery/pkg/util/runtime"
+	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	tenantsapi "sigs.k8s.io/multi-tenancy/poc/tenant-controller/pkg/apis/tenants/v1alpha1"
 	tenantsclient "sigs.k8s.io/multi-tenancy/poc/tenant-controller/pkg/clients/tenants/clientset/v1alpha1"
 	tenantsinformers "sigs.k8s.io/multi-tenancy/poc/tenant-controller/pkg/clients/tenants/informers/externalversions"
+	corev1 "k8s.io/api/core/v1"
 )
 
 // Controller is k8s controller managing Tenant CRDs.
 type Controller struct {
-	informer cache.SharedIndexInformer
+	informer      cache.SharedIndexInformer
+	tenantsclient tenantsclient.Interface
+	k8sclient     k8sclient.Interface
 }
 
 // NewController creates the controller.
-func NewController(client tenantsclient.Interface, informerFactory tenantsinformers.SharedInformerFactory) *Controller {
+func NewController(k8sclient k8sclient.Interface, tenantsclient tenantsclient.Interface, informerFactory tenantsinformers.SharedInformerFactory) *Controller {
 	c := &Controller{
-		informer: informerFactory.Tenants().V1alpha1().Tenants().Informer(),
+		informer:      informerFactory.Tenants().V1alpha1().Tenants().Informer(),
+		tenantsclient: tenantsclient,
+		k8sclient:     k8sclient,
 	}
 	c.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    func(o interface{}) { c.createTenant(o.(*tenantsapi.Tenant)) },
@@ -63,7 +69,20 @@ func (c *Controller) Run(ctx context.Context) error {
 func (c *Controller) createTenant(obj *tenantsapi.Tenant) {
 	glog.Info("createTenant: %#v", obj)
 	syncRBACForTenantCRD(obj)
-	// TODO create namespace
+
+	// TODO Add later ... sanity checks to ensure namespaces being requested are valid and not already assigned to another tenant
+
+	namespace := corev1.Namespace{}
+
+	for n := range obj.Spec.Namespaces {
+		namespace.ObjectMeta.Name = obj.Spec.Namespaces[n].Name
+		//Create namespace
+		c.k8sclient.CoreV1().Namespaces().Create(&namespace)
+
+		glog.Info("Created namespace: %s", obj.Spec.Namespaces[n].Name)
+	}
+
+	glog.Info("createTenant completed: %#v", obj)
 	// TODO create RBAC inside namespace
 }
 
@@ -77,7 +96,15 @@ func (c *Controller) updateTenant(old, obj *tenantsapi.Tenant) {
 func (c *Controller) deleteTenant(obj *tenantsapi.Tenant) {
 	glog.Info("deleteTenant: %#v", obj)
 	deleteRBACForTenantCRD(obj)
-	// TODO delete namespace
+	// TODO add a full set of sanity checks in future before deleting
+
+	for n := range obj.Spec.Namespaces {
+		glog.Info("Deleting namespace: %s", obj.Spec.Namespaces[n].Name)
+		// Delete namespace
+		c.k8sclient.CoreV1().Namespaces().Delete(obj.Spec.Namespaces[n].Name, nil)
+	}
+
+	glog.Info("deleteTenant completed: %#v", obj)
 }
 
 func rbacForTenantCRD(obj *tenantsapi.Tenant) []k8srt.Object {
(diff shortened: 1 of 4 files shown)


poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -14,12 +14,14 @@ package tenants
 import (
 	"context"
 	"fmt"
+	corev1 "k8s.io/api/core/v1"
```
easeway: could you move this import to next group? (away from standard packages)

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -28,13 +30,18 @@ import (
 
 // Controller is k8s controller managing Tenant CRDs.
 type Controller struct {
-	informer cache.SharedIndexInformer
+	informer      cache.SharedIndexInformer
+	tenantsclient tenantsclient.Interface
+	k8sclient     k8sclient.Interface
+	tenantsList   map[string]tenantsapi.Tenant
```
easeway: Regarding `tenantsList`, we will eventually need this map. However I don't see the need at this point.

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -63,7 +70,21 @@ func (c *Controller) Run(ctx context.Context) error {
 func (c *Controller) createTenant(obj *tenantsapi.Tenant) {
 	glog.Info("createTenant: %#v", obj)
 	syncRBACForTenantCRD(obj)
-	// TODO create namespace
+
+	// TODO Add later ... sanity checks to ensure namespaces being requested are valid and not already assigned to another tenant
+	c.tenantsList = make(map[string]tenantsapi.Tenant)
```
easeway: The map should be initialized in `NewController`. But I don't see the need of using this map, shall we add that later when we need it?

poc/tenant-controller/pkg/controllers/tenants/controller.go
easeway: For the value type of the map, it's better to use "*tenantsapi.Tenant" to avoid duplicating the object, as we only read from these objects.

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -63,7 +70,21 @@ func (c *Controller) Run(ctx context.Context) error {
...
+	c.tenantsList[obj.ObjectMeta.Name] = *obj
```
easeway: `DeepCopy` should be used instead of shallow copy of Kubernetes objects.

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -63,7 +70,21 @@ func (c *Controller) Run(ctx context.Context) error {
...
+
+	namespace := corev1.Namespace{}
+
+	for n, _ := range obj.Spec.Namespaces {
```
easeway: this line can be simplified as `for n := range ...`

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -63,7 +70,21 @@ func (c *Controller) Run(ctx context.Context) error {
...
+		namespace.ObjectMeta.Name = obj.Spec.Namespaces[n].Name
```
easeway: It's better to construct a namespace object inline, as there's not too much info to fill in. like:

c.k8sclient.CoreV1().Namespaces().Create(&corev1.Namespace{...})

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -63,7 +70,21 @@ func (c *Controller) Run(ctx context.Context) error {
 func (c *Controller) createTenant(obj *tenantsapi.Tenant) {
 	glog.Info("createTenant: %#v", obj)
 	syncRBACForTenantCRD(obj)
-	// TODO create namespace
+
+	// TODO Add later ... sanity checks to ensure namespaces being requested are valid and not already assigned to another tenant
+	c.tenantsList = make(map[string]tenantsapi.Tenant)
+	c.tenantsList[obj.ObjectMeta.Name] = *obj
+
+	namespace := corev1.Namespace{}
```
easeway: It's not a good idea to share a single object in multiple for loop iterations below.

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -63,7 +70,21 @@ func (c *Controller) Run(ctx context.Context) error {
...
+
+	for n, _ := range obj.Spec.Namespaces {
+		namespace.ObjectMeta.Name = obj.Spec.Namespaces[n].Name
+		//Create namespace
+		c.k8sclient.CoreV1().Namespaces().Create(&namespace)
```
easeway: We will need to handle errors on Create. I suggest a TODO comment here.

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -77,7 +98,16 @@ func (c *Controller) updateTenant(old, obj *tenantsapi.Tenant) {
 func (c *Controller) deleteTenant(obj *tenantsapi.Tenant) {
 	glog.Info("deleteTenant: %#v", obj)
 	deleteRBACForTenantCRD(obj)
-	// TODO delete namespace
+	// TODO add a full set of sanity checks in future before deleting
+
+	for n, _ := range c.tenantsList[obj.ObjectMeta.Name].Spec.Namespaces {
```
easeway: It's not necessary to get namespaces from c.tenantsList. The current `obj` provides all namespaces already. That's the reason I mentioned currently there's no need for tenantsList.

poc/tenant-controller/pkg/controllers/tenants/controller.go
easeway: I think it's important at this stage to experiment labels on namespaces to represent the tenant information. Could you try:

- add a label (or annotation) to link the namespace to the tenant?
- use `OwnerReference` in namespace metadata to create the ownership relation ship that garbage collector can kick in if deleting namespace failed in deleting tenant CRD.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: okay, will do. I had it in the next group originally but was moved up automatically by my vim-go plugin's built-in formatter, but will move it back now in any case.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: I agree its not critical to have tenantsList right now.  One use to have it right now is that when somene does a tenant delete, we do not need to rely on the namespace objects provided in the manifest for the delete call. We do not want to have any namespaces getting leaked if the user did not list the correct set of namespaces for the tenant during the deletion, hence ideally its safer to save the state in the tenantsList object. This makes it easier for the user to not specify the namespace list when deleting a tenant and also protects against errors if they provide an incorrect list of namespaces during the tenant deletion.  Let me know if you still prefer that we remove the tenantsList or you feel its better to keep it ? In future we can also use it to check against namespace duplication across tenants and so on.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: Ifwe decide to remove tenantsList, then I can remove this also, lets wait and see what we do with that first.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: Thanks.  If we decide to retain the tenantsList then I will change this one to deepcopy.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: Sure .. will change it.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: Also note that when we do namespace auto-allocation, we will anyway not have the namespace list in the obj.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: It seems to be a style issue mostly ... but okay I can change it to be inline as suggested.  One minor benefit of the current code is that we do not need to allocate the object during every loop. Anyway will change it as suggested.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: Okay, will change it per comment above.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: Sure, will add.

poc/tenant-controller/pkg/controllers/tenants/controller.go
srampal: As mentioned above, there is a benefit to doing it this way but will change it if we decide to remove the tenantsList

srampal:
Made all the suggested changes except one (moving the namespace struct initialization inline).

easeway:
I still have concerns sharing a single namespace instance with multiple creations (in this limited context, it should be OK), generally, this practice hides potential risks for future maintainers about unawareness of the sharing instance when writing code in the for loop. E.g. if updating some fields, it will affect all other places using the namespace object, though they should be different namespace objects intentionally. Performance is less a concern here compared to higher possibility for unintentional mistakes.

I'd like to merge this PR now, and we can fix that later.

PR #12, opened 2019-04-11
tenant-controller: add namespace/rbac sync on update of tenant

easeway:
Refactor the code and add the functionality to:

- synchronize namespaces when Tenant object is updated, including namespace deletion/creation;
- synchronize RBAC inside namespaces when Tenant object is updated, for tenant admins;
- add OwnerReferences to be managed by garbage collector.

### DIFF
--- a/poc/tenant-controller/pkg/controllers/tenants/controller.go
+++ b/poc/tenant-controller/pkg/controllers/tenants/controller.go
@@ -14,8 +14,11 @@ package tenants
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/golang/glog"
+	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8srt "k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -67,63 +74,174 @@ func (c *Controller) Run(ctx context.Context) error {
...
+		glog.Errorf("Tenant %q delete namespace %q error: %v", tenant.Name, nsName, err)
+		return err
+	}
+	glog.Infof("Tenant %q namespace %q deleted", tenant.Name, nsName)
+	return nil
+}
+
+func (c *Controller) ensureNamespaceExists(tenant *tenantsapi.Tenant, nsName string) error {
+	// TODO Add later ... sanity checks to ensure namespaces being requested are valid and not already assigned to another tenant
+	if err := newKubeCtl().addObjects(&corev1.Namespace{
```
srampal: Minor comment: In the same file we now have calls using k8sclient as well as newKubeCtl() ... we should discuss whether to make it consistent and use the same approach everywhere or whether it is okay to use different approaches.  The separate k8sclient approach was taken to natch the example in the sample-controller repo.

poc/tenant-controller/pkg/controllers/tenants/controller.go
```diff
@@ -67,63 +74,174 @@ func (c *Controller) Run(ctx context.Context) error {
...
+	glog.Infof("deleteTenant completed: %#v", obj)
 }
 
-func (c *Controller) updateTenant(old, obj *tenantsapi.Tenant) {
-	glog.Info("updateTenant: %#v", obj)
-	syncRBACForTenantCRD(obj)
-	// TODO sync namespace
-	// TODO sync RBAC inside namespace
+func (c *Controller) createNamespaceForTenant(tenant *tenantsapi.Tenant, nsReq *tenantsapi.TenantNamespace) error {
+	if err := c.ensureNamespaceExists(tenant, nsReq.Name); err != nil {
```
srampal: Minor comment: The names of the functions ensureNamespaceExists and syncNamespaceForTenant are slightly misleading.  The first one is not ensuring but actually creating the namespace (maybe a name like createNamespaceForTenant is something to consider ?).  The second one is attaching rbac for this namespace (so maybe a name like syncRbacForTenant or syncPropertiesForTenant could be considered ?).  Anyway its a minor comment.

Issue #14, opened 2019-04-11
Populate namespace with policies defined in NamespaceTemplate


easeway:
Create policy objects defined in NamespaceTemplate into namespaces created after Tenant CR.

PR #19, opened 2019-04-17
apply namespace template into tenant namespaces

easeway:
A simple implementation for applying policy objects defined in NamespaceTemplate into tenant namespaces:

- The NamespaceTemplate must exist if it's referenced by the Tenant, otherwise creating Tenant will fail;
- NamespaceTemplate is only applied when Tenant is created or updated;
- Updates on NamespaceTemplate won't be applied to Tenant namespaces which are already created.

Behavior changes since last commit:

-  The tenant namespace name is constructed with tenant name as prefix, e.g. `tenant-a` puts `ns-1` in Tenant CR, the final namespace name will be `tenant-a-ns-1`.


easeway:
Issue #14

PR #32, opened 2019-07-30
Kubebuilder scaffolding codes for tenant CRD

Fei-Guo:
This change pushes the  scaffolding codes generated by kubebuilder after running the following 
commands:

kubebuilder init --domain x-k8s.io --license apache2 --owner "The Kubernetes Authors"
kubebuilder create api --group tenancy  --namespaced false --version v1alpha1 --kind Tenant

No more changes.


rjbez17:
> --domain k8s.io

I recall a discussion with sig-arch where they were trying to prevent sigs from using this domain. I can't recall the result but we should double check we can use this.

tenant/config/crds/tenant_v1alpha1_tenant.yaml
```diff
@@ -0,0 +1,39 @@
...
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: tenants.tenant.k8s.io
+spec:
+  group: tenant.k8s.io
+  names:
+    kind: Tenant
+    plural: tenants
+  scope: Namespaced
```
kfox1111: Namespaced?

rjbez17:
we should remove the vendor directory from the repo

Fei-Guo:
> we should remove the vendor directory from the repo

We cannot compile the controller without the vendor, e.g., all the controller runtime codes are here
tenant/vendor/sigs.k8s.io/controller-runtime

rjbez17:
> > we should remove the vendor directory from the repo
> 
> We cannot compile the controller without the vendor, e.g., all the controller runtime codes are here
> tenant/vendor/sigs.k8s.io/controller-runtime

I guess I always just rely on `go mod vendor` done locally to make commits easier to follow. Seems like it's common enough in the k8s repos to keep vendor in the repo though.

rjbez17:
> > --domain k8s.io
> 
> I recall a discussion with sig-arch where they were trying to prevent sigs from using this domain. I can't recall the result but we should double check we can use this.

This doc was created for the outlining of the domains: https://docs.google.com/document/d/1nWLC_JISeuAAaMF0eRWftbGETef7u-52jzAK4PjedU0/edit

We should change to `multitenancy-k8s.io` is seems.

tenant/config/crds/tenant_v1alpha1_tenant.yaml
Fei-Guo: fixed

Fei-Guo:
Address all comments. 

> Change CRD domain and then I think we should be ok here.

done

Fei-Guo:
split the change to two commits for easy review.

tenant/config/samples/tenant_v1alpha1_tenant.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: tenant.multitenancy-k8s.io/v1alpha1
```
adrianludwin: CC @rjbez17 

I though the guidance was literally to use "x-k8s.io", where "x" is the character "x" and not a wildcard. See, for example, https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/919/files, which literally uses "x". 

My assumption was that this was modelled off of the golang.org/x/* packages, which (AIUI) are packages that are not yet considered to be part of the standard library, but are semi-official - kind of like Boost in C++. SIG-endorsed packages would seem to fall into a similar category.

If I'm correct, this would suggest that we call this group something like tenancy.x-k8s.io/v1alpha1

rjbez17: yes that is the intent, literal x.

Fei-Guo:
Change the domain to x-k8s.io and fix all unit tests.

tenant/config/crds/tenant_v1alpha1_tenant.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: tenants.tenant.x-k8s.io
+spec:
+  group: tenant.x-k8s.io
```
adrianludwin: Groups are often plural if it's a concrete noun (eg "apps", "extensions") or singular if it's a concept (eg "policy," "authentication," "networking"). Should this be "tenancy.x-k8s.io" or "tenants.x-k8s.io" instead?

adrianludwin: Other than this, lgtm

kfox1111: I kind of like tenancy.

Fei-Guo:
I also like tenancy, update the commits and the first comment. @kfox1111 @adrianludwin

tashimi:
@cjwagner can we remove the hold on this now?

cjwagner:
> @cjwagner can we remove the hold on this now?

I don't think the issue was addressed correctly?

adrianludwin:
> > @cjwagner can we remove the hold on this now?
> 
> I don't think the issue was addressed correctly?

I'm happy with it - was there another problem? What's still missing?

cjwagner:
@adrianludwin To be clear, the issue is not with this PR itself, but with merging any PR in the repo. Please see https://github.com/kubernetes-sigs/multi-tenancy/issues/33#issuecomment-517777731

adrianludwin:
Sorry I missed that - thanks.

cjwagner:
Lets see if this will merge now.
/hold cancel

PR #35, opened 2019-08-08
Add TenantAdminNamespaceName field and remove unused code

Fei-Guo:
This change adds a TenantAdminNamespaceName field in Tenant CR. Tenant controller will create a namespace using the specified name (future change). 

Later, all tenant related policy objects and custom resources will use the TenantAdminNamespace. 

This change also removes the unused example code generated by Kubebuilder.

### DIFF
--- a/tenant/config/crds/tenancy_v1alpha1_tenant.yaml
+++ b/tenant/config/crds/tenancy_v1alpha1_tenant.yaml
@@ -27,6 +27,11 @@ spec:
         metadata:
           type: object
         spec:
+          properties:
+            tenantAdminNamespaceName:
+              description: The name of tenant administration namespace, which is used
+                to create all tenant related policy objects and custom resources.
+              type: string
           type: object
         status:
           type: object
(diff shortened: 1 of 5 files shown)


tenant/config/crds/tenancy_v1alpha1_tenant.yaml
```diff
@@ -27,6 +27,11 @@ spec:
         metadata:
           type: object
         spec:
+          properties:
+            tenantAdminNamespaceName:
```
rjbez17: Thoughts on calling it `tenantParentNamespaceName` instead? I could go either way, just seeing what others think.

Fei-Guo: In a hierarchical structure, "parent" is ambiguous term which may indicate many nodes. Here this namespace acts more like a "root" so I think tenantAdminNamespaceName sounds better.

rjbez17: Right, but the work @adrianludwin is doing may mean we have nested tenants. It's not something we've really discussed.

PR #37, opened 2019-08-09
[incubator] Add virtualcluster project owners

Fei-Guo:
The incubator project has different owners.

### DIFF
--- a/incubator/virtualcluster/OWNERS
+++ b/incubator/virtualcluster/OWNERS
@@ -1,0 +1,11 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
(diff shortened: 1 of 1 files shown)


incubator/virtualcluster/OWNERS
```diff
@@ -0,0 +1,11 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
+approvers:
```
rjbez17: I think we probably also want the leads here don't we? I'm not sure we can have this 100% owned by one company per the guidelines. 

@tashimi @srampal ^

Fei-Guo: No problem at all, added Tasha.

PR #38, opened 2019-08-12
Reconcile for creating TenantAdminNamespace on Tenant creation

Fei-Guo:
This change implements creating TenantAdminNamespace upon tenant creation.
Controller adds tenant OwnerReference to TenantAdminNamespace. TenantAdminNamespace will be deleted when tenant CR is deleted. Reconcile fails if the TenantAdminNamespace has been created but owned by others. 

The reconcile logic for namespace deletion will be done in next change.

### DIFF
--- a/tenant/pkg/controller/tenant/tenant_controller.go
+++ b/tenant/pkg/controller/tenant/tenant_controller.go
@@ -18,9 +18,12 @@ package tenant
 
 import (
 	"context"
+	"fmt"
 
 	tenancyv1alpha1 "github.com/multi-tenancy/tenant/pkg/apis/tenancy/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -62,7 +65,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


tenant/pkg/controller/tenant/tenant_controller.go
```diff
@@ -95,5 +98,47 @@ func (r *ReconcileTenant) Reconcile(request reconcile.Request) (reconcile.Result
...
+				APIVersion: corev1.SchemeGroupVersion.String(),
+				Kind:       "Namespace",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            instance.Spec.TenantAdminNamespaceName,
+				OwnerReferences: []metav1.OwnerReference{expectedOwnerRef},
+			},
+		}
+		if err := r.Client.Create(context.TODO(), tenantAdminNs); err != nil {
+			return reconcile.Result{}, err
```
rjbez17: Wouldn't we want to return a requeue here so it attempts the create again on error? I've seen a few operators that don't properly requeue cause issues when the api server is in a bad state and returns an error during a create due simply to a network blip or similar. Note: I am not an expert on kubebuilder.

```
reuturn reconcile.Result{
        RequeueAfter: time.Second, // probably want some backoff here or longer than second wait
        Requeue:      true,
    }, nil
```

Fei-Guo: // RunInformersAndControllers the syncHandler, passing it the namespace/Name string of the
	// resource to be synced.
	if result, err := c.Do.Reconcile(req); err != nil {
		c.Queue.AddRateLimited(req)
		log.Error(err, "Reconciler error", "controller", c.Name, "request", req)
		ctrlmetrics.ReconcileErrors.WithLabelValues(c.Name).Inc()
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, "error").Inc()
		return false
	} else if result.RequeueAfter > 0 {
		c.Queue.AddAfter(req, result.RequeueAfter)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, "requeue_after").Inc()
		return true
	} else if result.Requeue {
		c.Queue.AddRateLimited(req)
		ctrlmetrics.ReconcileTotal.WithLabelValues(c.Name, "requeue").Inc()
		return true
	}

The controller framework will automatically requeue the request if err is not nil.

rjbez17: Missed that, thanks.

PR #39, opened 2019-08-13
Add watcher for tenantAdminNamespace

Fei-Guo:
This change adds watcher for tenantAdminNamespace. As a result, if tenantAdminNamespace is deleted for some reason, the tenant controller will create a new one with the same name.

### DIFF
--- a/tenant/pkg/controller/tenant/tenant_controller.go
+++ b/tenant/pkg/controller/tenant/tenant_controller.go
@@ -65,6 +65,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+
+	// Watch for changes to namespaces
+	err = c.Watch(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForOwner{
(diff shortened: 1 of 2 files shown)


tenant/pkg/controller/tenant/tenant_controller.go
```diff
@@ -65,6 +65,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
+
+	// Watch for changes to namespaces
+	err = c.Watch(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForOwner{
```
rjbez17: Does it make sense to scope this via labels/annotations so we don't have to process changes to namespaces not used by the tenant?

Fei-Guo: “EnqueueRequestForOwner” makes sure only the namespace that has a tenant as ownership will be put into reconcile request if it has CUD operations.

rjbez17: Thanks for the patience. You can tell I've spent most of my time in other controller frameworks.

PR #41, opened 2019-08-16
Add Name field in TenantNamespace CRD

Fei-Guo:
This change adds Name field in TenantNamespace CRD plus corresponding controller code for reconciling the tenant namespace CR. A few notes:

1) Controller will add OwnerReference to TenantNamespace CR for the newly created tenant namespace; 
2) Controller will add OwnerReference to TenantNamespace CR for existing namespace if the name matches instead of reporting error (not in this change);
3) One namespace can only be owned by one TenantNamespace CR. An admission Webhook will be added later to guarantee this;
4) Controller will not add prefix to the namespace name. It is up to user to resolve any naming conflict. 
5) TenantNamespace CR's namespace has be one tenant admin namespace. An admission Webhook will be added later to guarantee this;

### DIFF
--- a/tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
+++ b/tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
@@ -18,9 +18,12 @@ package tenantnamespace
 
 import (
 	"context"
+	"fmt"
 
 	tenancyv1alpha1 "github.com/multi-tenancy/tenant/pkg/apis/tenancy/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -58,6 +61,13 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to namespaces
+	err = c.Watch(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForOwner{
+		OwnerType: &tenancyv1alpha1.TenantNamespace{},
+	})
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
```diff
@@ -88,5 +99,50 @@ func (r *ReconcileTenantNamespace) Reconcile(request reconcile.Request) (reconci
...
+					// The namespace is owned by another TenantNamespace CR, fail the reconcile
+					err = fmt.Errorf("Namespace %v is owned by another %v TenantNamespace CR", each.Name, ownerRef)
+					return reconcile.Result{}, err
+				}
+			}
+			if !found {
+				log.Info("Namespace has been created without TenantNamespace owner", "namespace", each.Name)
+				// TODO: grab the namespace ownership
+			}
+			return reconcile.Result{}, nil
```
runyontr: Exiting early here will probably need to change once we start doing other things inside the namespace that's been created

Fei-Guo: Yes. This is going to be a small change though. I'd like to keep each change as concise as possible.

Fei-Guo:
> I expected the name of the Namespace created to match the name of the TenantNamespace. This would prevent (at creation time) any name conflicts that might arise from tenants already claiming the name.

This is one option. I want to hear others' options to see if there is a need to claim namespace name differently.

tashimi:
@Fei-Guo @rjbez17 I like the idea of avoiding namespace collisions at creation and tying it to tenant name, but one thing I've run into in working on other platforms is that ability to change the name always becomes a top requested feature

runyontr:
> that ability to change the name always becomes a top requested feature

I think the implementation of this would be problematic based on the current functionality of internal k8s.  Namespace rename isn't really supported unless you consider tools like Velero that can do a "restore" of a configuration.  For now I think we should have this be an immutable field if we're leaving it as part of the spec.

tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
rjbez17: > I expected the name of the Namespace created to match the name of the TenantNamespace. This would prevent (at creation time) any name conflicts that might arise from tenants already claiming the name.

A TenantNamespace will be namespace scoped while namespaces are cluster scoped. So you'd still have possible collisions. However, I tend to agree that that TenantNamespace name would map to the namespace name.

kfox1111:
"Controller will not add prefix to the namespace name. It is up to user to resolve any naming conflict."

I don't think we should bake this into the api. Some sites may want to do prefixing. The api should allow for it.

So, how about having the spec.name field, but also a status.name field. the status.name field is filled in with the chosen namespace name, influenced by or fully using the spec.name field?

kfox1111:
One other thing. Is there a reason to ever do .spec.name instead of reusing .metadata.name?

kfox1111:
What about allowing .spec.name to be optional? If it is not specified, use .metadata.name?

PR #42, opened 2019-08-20
Add scaffolding code for TenantNameSpace CRD Webhook

Fei-Guo:
This change adds kubebuilder scaffold codes for TenantNamespace CRD webhook. i.e., codes generated by following cmd

kubebuilder alpha webhook --group tenancy --version v1alpha1 --kind TenantNamespace --type=validating --operations=create,update 

In addition, I change controller manager namespace to tenant-system and 
add an all_in_one.yaml for easy installation. 
The image name will be changed later once we have a formal docker hub account.

### DIFF
--- a/tenant/config/manager/manager.yaml
+++ b/tenant/config/manager/manager.yaml
@@ -50,7 +50,7 @@ spec:
       containers:
       - command:
         - /manager
-        image: vrgf2003/tenant:latest
+        image: tenant-controller:latest
         imagePullPolicy: Always
         name: manager
         env:
--- a/tenant/config/manager/all_in_one.yaml
+++ b/tenant/config/manager/all_in_one.yaml
@@ -6,6 +6,37 @@ metadata:
     controller-tools.k8s.io: "1.0"
   name: tenant-system
 ---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    alpha.admissionwebhook.cert-manager.io: "true"
+  creationTimestamp: null
+  name: tenant-validating-webhook-cfg
+webhooks:
+- clientConfig:
+    caBundle: XG4=
+    service:
+      name: webhook-server-service
+      namespace: tenant-system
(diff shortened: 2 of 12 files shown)


tenant/config/manager/manager.yaml
```diff
@@ -44,7 +50,7 @@ spec:
       containers:
       - command:
         - /manager
-        image: controller:latest
+        image: vrgf2003/tenant:latest
```
rjbez17: We should be able to use the k8s gcr I'd think. @tashimi are you aware of the process to use it/configure full CI/CD of our repo? At the least we should create an issue to configure this asap.

rjbez17:
Waiting for @tashimi  input

tenant/config/manager/manager.yaml
Fei-Guo: I am going to change it to Multitenancy/tenant for now and I suggest @tashimi to create a docker hub account to save the image instead of using gcr.  Countries outside of US may have difficulty to access gcr.

tashimi:
@Fei-Guo @rjbez17 Docker Hub is good - I created a multitenancy organization I am happy to add you to as owners if that works https://cloud.docker.com/u/multitenancy/repository/list

kfox1111:
> @Fei-Guo @rjbez17 Docker Hub is good - I created a multitenancy organization I am happy to add you to as owners if that works https://cloud.docker.com/u/multitenancy/repository/list

Probably fine for the prototype. Feels kind of bad long term though. Probably should at least file a separate issue to deal with build/release related things.

tashimi:
@kfox1111 let me know what you'd rather do re: dockerhub (in another issue) and we can do it.

kfox1111:
Looks like #wg-k8s-infra can hook us up with gce.

"first step is to create a google group for the access control for the repo (please add one in https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml) and then an entry in the script for creating the staging repo (please update https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L41). Please see older PR(s) that were merged when in doubt."

PR #44, opened 2019-08-22
Add RequireNamespacePrefix field in Tenant CRD

Fei-Guo:
This change simply adds a new field in Tenant CRD. This field will be checked by TenantNamespaceCRD later. 

This change also adds all the missing autogenerated webhook code. 
For now, "config/manager/all_in_one.yaml" is going to be the maintained yaml file for installing the controller manager. It contains all small fixes (such as namespace name).

### DIFF
--- a/tenant/config/crds/tenancy_v1alpha1_tenant.yaml
+++ b/tenant/config/crds/tenancy_v1alpha1_tenant.yaml
@@ -28,6 +28,11 @@ spec:
           type: object
         spec:
           properties:
+            requireNamespacePrefix:
+              description: If set to True, all the namespaces belong to the tenant
+                are requried to have TenantAdminNamespaceName as name prefix. By default,
+                namespace prefix is not required.
+              type: boolean
             tenantAdminNamespaceName:
               description: The name of tenant administration namespace, which is used
                 to create all tenant related policy objects and custom resources.
--- a/tenant/config/rbac/rbac_role_binding.yaml
+++ b/tenant/config/rbac/rbac_role_binding.yaml
@@ -10,4 +10,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: default
-  namespace: tenant-system
+  namespace: system
--- a/tenant/config/default/manager_patch.yaml
+++ b/tenant/config/default/manager_patch.yaml
@@ -1,0 +1,25 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: controller-manager
+spec:
+  template:
+    metadata:
+      labels:
+        control-plane: controller-manager
+    spec:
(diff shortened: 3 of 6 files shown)


tenant/config/rbac/rbac_role_binding.yaml
```diff
@@ -10,4 +10,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: default
-  namespace: tenant-system
+  namespace: system
```
kfox1111: Was this intended?

tenant/config/crds/tenancy_v1alpha1_tenant.yaml
```diff
@@ -28,6 +28,11 @@ spec:
           type: object
         spec:
           properties:
+            requireNamespacePrefix:
+              description: If set to True, all the namespaces belong to the tenant
+                are requried to have TenantAdminNamespaceName as name prefix. By default,
+                namespace prefix is not required.
+              type: boolean
```
kfox1111: General question. Should this be an enum instead?

Maybe some site wants to use the namespace as a suffix instead of a prefix, or some other idea we haven't thought of yet.

namespacePattern: [ TenantNamespace, NamespaceTenant, Namespace ] ?

tenant/config/rbac/rbac_role_binding.yaml
Fei-Guo: Kubebuilder auto generated code always produces `system` namespace name. I was manually changing it to tenant-system in previous change. I'd rather not do that. The right approach is to use kustomize to add prefix to namespace name. Alternatively, I maintain an all-in-one.yaml to install all components and make namespace adjustment in that file, which is more convenient at this stage.

tenant/config/crds/tenancy_v1alpha1_tenant.yaml
Fei-Guo: I‘d rather not go with this far if a boolean can serve 90% use cases.

rjbez17: @Fei-Guo I agree with keeping it simple, but it does bring up a good point. Most generated items in k8s are suffixs rather than prefixes. A pro for a prefix is you get sorting when doing a list, however, a suffix does hide a redundant item and follows the precedent set by deployments -> pods, cronjobs -> jobs, etc.

tenant/config/rbac/rbac_role_binding.yaml
rjbez17: Any reason not to remove it all together so people that don't use kustomize can easily override via `kubectl apply -n tenant-system -f all-in-one.yaml`? I don't really have a preference here, just asking.

PR #47, opened 2019-08-29
Initial checkin of the HNC

adrianludwin:
Note: this is not set up for any CI yet. All testing has been manual, either by running "make test" or by testing on a local KIND cluster.


rjbez17:
🎉

incubator/hnc/config/default/manager_auth_proxy_patch.yaml
```diff
@@ -0,0 +1,25 @@
...
+  template:
+    spec:
+      containers:
+      - name: kube-rbac-proxy
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.4.0
+        args:
+        - "--secure-listen-address=0.0.0.0:8443"
+        - "--upstream=http://127.0.0.1:8080/"
+        - "--logtostderr=true"
+        - "--v=10"
```
rjbez17: probably want to change this before releasing

incubator/hnc/scripts/kind-watch
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+pod=`kubectl get pods -n hnc-system -ojsonpath='{.items[0].metadata.name}'`
```
rjbez17: Protip, you can use `kubectl logs` on a deployment so this step is easier.

incubator/hnc/pkg/forest/forest.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	// Unset the parent from all children, if any
+	for _, c := range ns.children {
+		affected = append(affected, c.name)
+		c.parent = nil
+	}
+
+	// Delete and return affected
+	delete(f.namespaces, nm)
```
rjbez17: If this fails here we are in a bad state because all children have been removed but the parent would still remain. The failure is also not caught/handled so a reconcile wont know to try again later.

Fei-Guo:
This is a fairly big change with mix of scaffolding codes and functional code. I am not sure if I need to consider the coordination with Tenant V2 CRD during the review. So for incubator project code, should we just let it go and tune later? 

/assign @rjbez17

incubator/hnc/config/default/manager_auth_proxy_patch.yaml
adrianludwin: These are simply the defaults from kubebuilder 2.0, and I don't know enough about kube-rbac-proxy to know what the correct values are. I'm happy to add a TODO here, do you have any suggestions of what I should write (other than just "revisit before initial release")?

incubator/hnc/pkg/forest/forest.go
adrianludwin: While I'm certain there are many problems with deletion, I don't really understand this one? The in-memory forest is only supposed to _reflect_ what's on the cluster - it's not the source of truth. On the cluster, all the "child" namespaces will continue to have pointers to the now-missing parent, so if we set the resync period to be something reasonable, we'll eventually find the problem.

I'm also not certain if there's a _specific_ failure that you had in mind here - is there?

incubator/hnc/scripts/kind-watch
adrianludwin: D'oh, fixed. Thanks.

adrianludwin:
> This is a fairly big change with mix of scaffolding codes and functional code.

Agreed - I _could_ try to break it up into pieces to recreate what I've done over the past ~month but I don't think it would have a lot of value, myself. (The actual history, of course, has taken lots of twists and turns but is very hard to merge into an existing repo).

> I am not sure if I need to consider the coordination with Tenant V2 CRD during the review. So for incubator project code, should we just let it go and tune later?

I don't think code-level coordination is required at this stage. As discussed at the last meeting, I think we should have some _design_ discussions about how the two should fit together, but (as you said) this is just in incubation for now so the code is very malleable.

incubator/hnc/pkg/forest/forest.go
rjbez17: Sorry, in my rush to review this before the weekend I missed that this was the built in and thought this called the controllers delete.

rjbez17:
> > This is a fairly big change with mix of scaffolding codes and functional code.
> 
> Agreed - I _could_ try to break it up into pieces to recreate what I've done over the past ~month but I don't think it would have a lot of value, myself. (The actual history, of course, has taken lots of twists and turns but is very hard to merge into an existing repo).
> 
> > I am not sure if I need to consider the coordination with Tenant V2 CRD during the review. So for incubator project code, should we just let it go and tune later?
> 
> I don't think code-level coordination is required at this stage. As discussed at the last meeting, I think we should have some _design_ discussions about how the two should fit together, but (as you said) this is just in incubation for now so the code is very malleable.

Agreed on both accounts.

PR #48, opened 2019-08-29
Import existing namespace by adding OwnerReference

Fei-Guo:
This change adds support to import existing namespace to tenant management system by adding proper ownerReference. 

Notes: I didn't check the RequireNamespacePrefix flag in this change because I think the logic is going to be messy if I did it. The below shows how RequireNamespacePrefix flag affects the reconcile result.

Case A:
Tenant: RequireNamespacePrefix == false
TenantNamespace: Name == t1
Namespace: t1
RESULT: t1 is imported by adding ownerReference to tenantNamespace object.

Case B:
Tenant: RequireNamespacePrefix == true
TenantNamespace: Name = t1
Namespace: t1
RESULT: t1 is not imported and a new namespace tenant-t1 is created

Case C:
Tenant: RequireNamespacePrefix == true
TenantNamespace: Name = t1
Namespace: tenant-t1
RESULT: tenant-t1 is going to be imported by adding ownerReference to tenantNamespace object.

### DIFF
--- a/tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
+++ b/tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
@@ -25,6 +25,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


rjbez17:
> This change adds support to import existing namespace to tenant management system by adding proper ownerReference.
> 
> Notes: I didn't check the RequireNamespacePrefix flag in this change because I think the logic is going to be messy if I did it. The below shows how RequireNamespacePrefix flag affects the reconcile result.
> 
> Case A:
> Tenant: RequireNamespacePrefix == false
> TenantNamespace: Name == t1
> Namespace: t1
> RESULT: t1 is imported by adding ownerReference to tenantNamespace object.
> 
> Case B:
> Tenant: RequireNamespacePrefix == true
> TenantNamespace: Name = t1
> Namespace: t1
> RESULT: t1 is not imported and a new namespace tenant-t1 is created
> 
> Case C:
> Tenant: RequireNamespacePrefix == true
> TenantNamespace: Name = t1
> Namespace: tenant-t1
> RESULT: tenant-t1 is going to be imported by adding ownerReference to tenantNamespace object.

Case B is the only one that needs discussion. I'd argue we fail the import/return errors in logs and not create a new namespace. The user clearly intended to import an existing namespace so creating something different will only cause confusion (in my opinion at least).

Fei-Guo:
> > This change adds support to import existing namespace to tenant management system by adding proper ownerReference.
> > Notes: I didn't check the RequireNamespacePrefix flag in this change because I think the logic is going to be messy if I did it. The below shows how RequireNamespacePrefix flag affects the reconcile result.
> > Case A:
> > Tenant: RequireNamespacePrefix == false
> > TenantNamespace: Name == t1
> > Namespace: t1
> > RESULT: t1 is imported by adding ownerReference to tenantNamespace object.
> > Case B:
> > Tenant: RequireNamespacePrefix == true
> > TenantNamespace: Name = t1
> > Namespace: t1
> > RESULT: t1 is not imported and a new namespace tenant-t1 is created
> > Case C:
> > Tenant: RequireNamespacePrefix == true
> > TenantNamespace: Name = t1
> > Namespace: tenant-t1
> > RESULT: tenant-t1 is going to be imported by adding ownerReference to tenantNamespace object.
> 
> Case B is the only one that needs discussion. I'd argue we fail the import/return errors in logs and not create a new namespace. The user clearly intended to import an existing namespace so creating something different will only cause confusion (in my opinion at least).

Fail the import may also be problematic. If someone accidentally creates a ns with a "common org" name say team1，tenant admin cannot create TeantX-team1 anymore. The problem is that import operation is implicit in the Spec and it causes ambiguity when someone actually wants to create a new one. Another choice to use another CRD to import existing namespace which I think is more appropriate or we can accept the less restrictive behavior for now until we have real use case and clear requirement.

kfox1111:
There needs to be a way for the cluster-admin to prevent this if desired. Is there an attribute on the Tenant CRD that could disable this?

kfox1111:
Or I guess it could be a config option to the operator too as a global setting.

PR #49, opened 2019-08-30
initial virtual node agent with briefly kubelet api proxy

zhuangqh:
This change introduce the vn-agent(virtual node agent).
vn-agent acts as a proxy between tenant api server and kubelet, translate some naming between super master namespaces and tenant master namespaces.

this PR basically initialize this component, support /pods, /run, /logs, /containerLogs proxy.


incubator/virtualcluster/pkg/vn-agent/server/route.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/emicklei/go-restful"
+	"k8s.io/apimachinery/pkg/util/proxy"
+	"k8s.io/klog"
+)
+
+// InstallHandlers set router and handlers.
+func (s *Server) InstallHandlers() {
+	ws := new(restful.WebService)
+	ws.
```
xiang90: can this paas gofmt? just make this one line?

incubator/virtualcluster/pkg/vn-agent/server/route.go
```diff
@@ -0,0 +1,100 @@
...
+		Path("/containerLogs")
+	ws.Route(ws.GET("/{podNamespace}/{podID}/{containerName}").
+		To(s.proxy).
+		Operation("getContainerLogs"))
+	s.restfulCont.Add(ws)
+}
+
+func (s *Server) proxy(req *restful.Request, resp *restful.Response) {
+	klog.Infof("request %+v", req)
```
xiang90: this logging should not be infof level. maybe at debugging level.

incubator/virtualcluster/pkg/vn-agent/server/translate.go
```diff
@@ -0,0 +1,37 @@
...
+package server
+
+import (
+	"strings"
+
+	"github.com/emicklei/go-restful"
+)
+
+// TranslatePath translate the naming between tenant and master cluster.
+func TranslatePath(req *restful.Request, tenantName string) string {
```
xiang90: PathFromTenantToMaster.

Let's make the func naming more descriptive.

xiang90: Also it is strange to take in an http request and translate it to a string type. the abstraction is not symmetric.

xiang90:
need some tests.

incubator/virtualcluster/pkg/vn-agent/server/route.go
zhuangqh: alright. But it really could pass the gofmt.

incubator/virtualcluster/pkg/vn-agent/server/translate.go
zhuangqh: good suggestion

zhuangqh: `restful.Request` here because it holds the route info it matched. I changed it to an object mutable method here.

incubator/virtualcluster/pkg/vn-agent/server/route.go
zhuangqh: changed verbose level here.

zhuangqh:
> need some tests.

I've created the test framework and a brief test. Completed tests would be submitted in another pr.

PR #52, opened 2019-09-05
vc: vn-agent: test for /containerLogs, /run api proxy

zhuangqh:
the principle job is to test the naming translation of each api.

### DIFF
--- a/incubator/virtualcluster/pkg/vn-agent/server/server_test.go
+++ b/incubator/virtualcluster/pkg/vn-agent/server/server_test.go
@@ -21,10 +21,12 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/http/httputil"
 	"net/url"
+	"reflect"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


incubator/virtualcluster/pkg/vn-agent/server/server_test.go
```diff
@@ -413,9 +420,13 @@ func getPodName(name, namespace string) string {
 	return name + "_" + namespace
 }
 
+func getTenantNamespace(tenantName, namespace string) string {
```
Fei-Guo: How about getTenantNamespace -> getEffectiveNamespace, also add a comment to this method.

incubator/virtualcluster/pkg/vn-agent/server/server_test.go
```diff
@@ -444,3 +455,167 @@ func TestServeLogs(t *testing.T) {
...
+	if result != output {
+		t.Errorf("expected %s, got %s", output, result)
+	}
+}
+
+func TestServeRunInContainerWithUID(t *testing.T) {
+	// TODO: support UID naming translate
+}
+
+func setPodByNameFunc(fv *serverTestFramework, namespace, pod, container string) {
```
Fei-Guo: Just curious, why do you extract this method out? the same for setGetContainerLogsFunc? 
(you did not do this for runFunc in TestServeRunInContainer)

zhuangqh: > Just curious, why do you extract this method out? the same for setGetContainerLogsFunc?
> (you did not do this for runFunc in TestServeRunInContainer)

there is a validation in the kubelet server, it gets the Pod spec from `GetPodByName` interface and check whether there exists the correspond container. We inject this info from `setPodByNameFunc` this method. However, there is no this validation in the `/run` api.  It's strange...

incubator/virtualcluster/pkg/vn-agent/server/server_test.go
zhuangqh: updated

PR #54, opened 2019-09-11
Fix wrong controller-tools version for 1.0.8 kubebuilder

Fei-Guo:
Based on the kubebuilder 1.0.8 Gopkg, the corresponding controller-tools version should be v0.1.1. 
Otherwise, the `make` cmd breaks.


incubator/virtualcluster/go.mod
```diff
@@ -24,8 +26,9 @@ require (
 	k8s.io/cri-api v0.0.0
 	k8s.io/klog v0.3.1
 	k8s.io/kubernetes v1.15.3
-	k8s.io/utils v0.0.0-20190829053155-3a4a5477acf8 // indirect
+	k8s.io/utils v0.0.0-20190829053155-3a4a5477acf8
 	sigs.k8s.io/controller-runtime v0.2.0
+	sigs.k8s.io/controller-tools v0.1.1 // indirect
```
zhuangqh: please do `go mod tidy` before submit the PR.  `sigs.k8s.io/controller-tools` is not directly imported by code. It would be removed after `go mod tidy`.
Actually, I think it is unnecessary to involve this dependency to go mod. controller-gen just a tool

PR #60, opened 2019-09-16
Add README.md for Tenant Operator

Fei-Guo:
Add a basic README.md for people to try the tenant controllers.

### DIFF
--- a/tenant/README.md
+++ b/tenant/README.md
@@ -1,0 +1,32 @@
+# Tenant
+
+This repo provides a set of CRDs used to manage tenant resources in a Kubernetes cluster.
+
+## Overview
+A cluster admin primarily leverages a tenant admin namespace to create, and thus manage all tenant related API objects including tenant namespaces (not directly though), arbitrary customized CRDs and policy objects like RBAC. This makes the tenant concept fully extendible instead of limiting the types of tenant objects specifically in TenantSpec. Basically, the tenant admin namespace name is specified in Tenant CR and tenant controller simply creates the namespace and configure the cluster RBAC rules for tenant admins to access various tenant related objects which are created on-demand (not implmented yet).
+
+The most common tenant objects are tenant namespace CRs which encapsulate the configurations of tenant namespaces including name and RBAC rules etc. Tenant namespace controller does actual namespace creation for tenant users to use. The primary reason why we introduce tenant namespace CRD is that now tenant namespace creation can be RBAC permissioned, and tenant admins can do `kubectl get tenantnamespace -n <tenant admin namespace>` to list all the tenant namespaces. Otherwise had K8s namespace list API been used, all namespaces in the cluster would be exposed to tenant admins. Tenant admin can do self-service namespace creation via creating tenant namespace CR or even import existing namespace to tenant management system.
+
+## Getting Started
+
+### Install CRDs
+```
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/crds/tenancy_v1alpha1_tenant.yaml
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/crds/tenancy_v1alpha1_tenantnamespace.yaml
+```
+### Install tenant-controller-managers
+```
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/manager/all_in_one.yaml
+```
+
+## Basic Usage
+
+Use the following cmd to create a tenant CR. A tenant admin namespace `tenant1admin` will be created.
+```
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/samples/tenancy_v1alpha1_tenant.yaml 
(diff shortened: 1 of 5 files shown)


tenant/README.md
```diff
@@ -0,0 +1,32 @@
+# Tenant
+
+This repo provides a set of CRDs used to manage tenant resources in a Kubernetes cluster.
+
+## Overview
+A cluster admin primarily leverages a tenant admin namespace to create, and thus manage all tenant related API objects including tenant namespaces (not directly though), arbitrary customized CRDs and policy objects like RBAC. This makes the tenant concept fully extendible instead of limiting the types of tenant objects specifically in TenantSpec. Basically, the tenant admin namespace name is specified in Tenant CR and tenant controller simply creates the namespace and configure the cluster RBAC rules for tenant admins to access various tenant related objects which are created on-demand (not implmented yet).
```
sb1975: extensible

tenant/README.md
```diff
@@ -0,0 +1,32 @@
...
+### Install tenant-controller-managers
+```
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/manager/all_in_one.yaml
+```
+
+## Basic Usage
+
+Use the following cmd to create a tenant CR. A tenant admin namespace `tenant1admin` will be created.
+```
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/samples/tenancy_v1alpha1_tenant.yaml
```
sb1975: So when I do kubectl apply -f  samples/tenancy_v1alpha1_tenant.yaml - It creates a tenant named "tenant-sample" with spec - tenantAdminNamespaceName: "tenant1admin" ; and when kubectl apply -f samples/tenancy_v1alpha1_tenantnamespace.yaml - that creates a TenantNamespace named "tenant1admin" with spec "t1-ns1" but does not actually create any tenantnamespace named"t1-ns1"
Now I am assuming tenant admin namespace is a type of tenantnamespace only
kubectl get tenantnamespace --all-namespaces
NAMESPACE   NAME                     AGE
default     tenant1admin             22s
default     tenantnamespace-sample   137m
but it doesn not correspond to an actual K8s namespace, so how do I use it to launch pods within a tenant namespace ?
the sample tenancy yaml only creates the tenant CR but no namespace really.. # cat tenant/config/samples/tenancy_v1alpha1_tenant.yaml
apiVersion: tenancy.x-k8s.io/v1alpha1
kind: Tenant
metadata:
 labels:
   controller-tools.k8s.io: "1.0"
 name: tenant-sample
spec:
 tenantAdminNamespaceName: "tenant1admin"

Are we missing something here ?

tenant/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+Use the following cmd to create a tenantnamespace CR. A tenant namespace `t1-ns1` will be created.
+```
+kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/blob/master/tenant/config/samples/tenancy_v1alpha1_tenantnamespace.yaml
+```
```
sb1975: hence the error :
kubectl apply -f tenant/config/samples/tenancy_v1alpha1_tenantnamespace.yaml
Error from server (NotFound): error when creating "tenant/config/samples/tenancy_v1alpha1_tenantnamespace.yaml": namespaces "tenant1admin" not found

tenant/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+The most common tenant objects are tenant namespace CRs which encapsulate the configurations of tenant namespaces including name and RBAC rules etc. Tenant namespace controller does actual namespace creation for tenant users to use. The primary reason why we introduce tenant namespace CRD is that now tenant namespace creation can be RBAC permissioned, and tenant admins can do `kubectl get tenantnamespace -n <tenant admin namespace>` to list all the tenant namespaces. Otherwise had K8s namespace list API been used, all namespaces in the cluster would be exposed to tenant admins. Tenant admin can do self-service namespace creation via creating tenant namespace CR or even import existing namespace to tenant management system.
+
+## Getting Started
+
```
sb1975: I think it would be nice to include the basic setup needed to try this : 
1. Install Go
2. Install docker
3. Install KIND and kind create cluster

Fei-Guo: I'd rather put these specific steps to a tutorial document instead of README.  People may try this successfully in their own environments without these steps.

Fei-Guo:
> Finally I would also appreciate if there is a basic test of creating a pod on a tenantnamespace.
> Later on we can expand further by adding resourcequota,network policy.
> The objective will be the simple fact that I would like my tenants be bounded by the basic resources (RAM,Disk,CPU - possible by using resourcequota) , Network (Network Policy) and other quota limits possible( https://kubernetes.io/docs/concepts/policy/resource-quotas/ )

I'd like to treat this as one of follow up changes. Anyone can submit ideas about readme.

Fei-Guo:
> Do you want to rename Tenant CRD to Tenant Operator per convo in Slack?

You meant the PR name, right? I already avoid using tenant CRD in readme text.

PR #62, opened 2019-09-18
Don't require or create unnecessary singletons

adrianludwin:
This change removes the need for each hierarchy to contain a singleton.
Instead, singletons are automatically created if needed (for example,
because the namespace has children), but otherwise, the controller
assumes that "no singleton" means "no special hierarchy requirements".

Tested: re-ran the demo by hand; updated integ tests.


incubator/hnc/controllers/hierarchy_controller.go
```diff
@@ -47,76 +49,144 @@ type HierarchyReconciler struct {
...
+			// The namespace is gone. Remove all traces of the namespace from  our data structures.
+			log.Info("Namespace was deleted", "ns", nm)
+			return r.onDelete(ctx, log, nm)
+		}
+
+		// If the namespace is being deleted, we should still sync any changes that we might have missed
+		// (ie, maybe the parent used to be set?) but then don't update anything.
+		if !ns.GetDeletionTimestamp().IsZero() {
+			log.Info("Namespace is being deleted", "ns", nm)
+			update = false
```
Fei-Guo: Why do we need this check? If ns is being deleted. A reconcile will eventually happen and go to r.onDelete path to clear everything.

incubator/hnc/controllers/hierarchy_controller.go
```diff
@@ -47,76 +49,144 @@ type HierarchyReconciler struct {
...
+	// Update the in-memory data structures
+	orig := inst.DeepCopy()
 	affected := r.syncWithForest(ctx, log, inst)
-	if err := r.Update(ctx, inst); err != nil {
-		log.Error(err, "while updating apiserver")
-		return err
+
+	// If appropriate, write back. We only write if there's been a *change* in the data structure, and
+	// if the caller wants us to update it (ie, if the namespace isn't being deleted).
+	if update && !reflect.DeepEqual(orig, inst) {
```
Fei-Guo: In create new ns case, will reflect.DeepEqual(orig, inst) return true since syncWithForest will not change the content of inst object? Particularly,  I am not sure if adding inst.Status.Children = nil will alter the DeepEqual() result.

adrianludwin:
> I am assuming you will add webhook to prevent hier object from being created by any user other than the controller.

No, users can create it if it doesn't exist - that's how you set the parent. However, the CRD validation schema will only allow a single name, so there can only be at most one per namespace.

incubator/hnc/controllers/hierarchy_controller.go
adrianludwin: TBH I wasn't sure what to do with this. I think this check showed up in the kubebuilder boilerplate code and I just filled it in with something that seemed reasonable. I should probably update the comments to indicate that I don't really know what I'm doing.

incubator/hnc/controllers/hierarchy_controller.go
adrianludwin: I set inst.Status.Children=nil to make sure that DeepEqual worked correctly.  So yes, for new namespaces, DeepEqual does return true, and the `hier` object is not written back as a result (which is the intended behaviour).

incubator/hnc/controllers/hierarchy_controller.go
Fei-Guo: Are you going to change this later?

incubator/hnc/controllers/hierarchy_controller.go
Fei-Guo: > > I am assuming you will add webhook to prevent hier object from being created by any user other than the controller.
> 
> No, users can create it if it doesn't exist - that's how you set the parent. However, the CRD validation schema will only allow a single name, so there can only be at most one per namespace.

Got you.

incubator/hnc/controllers/hierarchy_controller.go
adrianludwin: I'll update the comments in another commit, but I'll leave this as-is unless someone has a better idea.

While I understand that a reconcile will _eventually_ happen, my idea here was this could take a while (eg namespaces take a while to be deleted) and we should update everything else ASAP.

PR #63, opened 2019-09-19
[syncer part 1] add multi cluster controller runtime and entry syncer…

adohe:
… command

### DIFF
--- a/incubator/virtualcluster/pkg/resource-syncer/controller/controller.go
+++ b/incubator/virtualcluster/pkg/resource-syncer/controller/controller.go
@@ -1,0 +1,235 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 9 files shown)


incubator/virtualcluster/pkg/resource-syncer/controller/controller.go
```diff
@@ -0,0 +1,212 @@
...
+		klog.Warning("Work item is not a Request. Ignore it. Next.")
+		c.Queue.Forget(obj)
+		return true
+	}
+
+	if result, err := c.reconciler.Reconcile(req); err != nil {
+		klog.Error(err)
+		klog.Error("Could not reconcile Request. Stop working.")
+		c.Queue.AddRateLimited(req)
+		return false
```
Fei-Guo: Do you permanently lose one reconciler thread if Reconcile func returns false?

adohe: `go wait.Until(func() {
    for c.processNextWorkItem() {
    }
}, c.JitterPeriod, stop)`
IIUC this just means the for loop will break, the reconciler thread will not return, wait.Until will call worker func again after JitterPeriod.

Fei-Guo: Got you

Issue #59, opened 2019-09-16
HNC: in-memory tree may be out-of-sync with version on apiserver


adrianludwin:
See https://github.com/kubernetes-sigs/multi-tenancy/pull/58#pullrequestreview-288817351

adrianludwin:
CC @Fei-Guo 

/assign adrianludwin

adrianludwin:
Fixed via #64 and #66

PR #64, opened 2019-09-19
Idiomatic watches in the hierarchy controller

adrianludwin:
Instead of directly updating any affected namespaces (ie old parents,
children) when a namespace is updated, give the affected namespaces to
the controller-runtime directly. This means that even if there's a
problem saving the "original" namespace, the list of affected namespaces
is stored by controller-runtime and they'll all eventually get resolved.
This fixes issue #59.

At the same time, I've fully removed the namespace reconciller and
simply added a mapping function to the hierarchy reconciller to notice
changes to namespaces.

Tested: integ tests pass; demo works.


Fei-Guo:
Changes LG. Some thoughts for your consideration:

- You approach now assumes in memory Forest is source of truth because affected ns will be updated with what Forest presents (e.g., the child ns). If the controller manager is restarted and Forest in memory cache is gone,  there is a chance APIserver preserves a wrong hierarchy.

adrianludwin:
> Changes LG. Some thoughts for your consideration:
> 
> * You approach now assumes in memory Forest is source of truth because affected ns will be updated with what Forest presents (e.g., the child ns). If the controller manager is restarted and Forest in memory cache is gone,  there is a chance APIserver preserves a wrong hierarchy.

That certainly wasn't my intent. I thought everything was first syncing from the API objects into the forest, and then (as part of the same call to the reconciler) determining the other affected namespaces.

The only thing that was _supposed_ to change in this commit was that instead of the reconciler _directly_ updating other namespaces, it's now _enqueuing_ those other namespaces to be updated at a later time. But the SoT is still supposed to be the apiserver.

Can you give me a scenario where this wouldn't work?

Fei-Guo:
> Can you give me a scenario where this wouldn't work?

For example, somebody restarts the controller manager right before executing this method. 

if err := r.updateAffected(ctx, log, affected); err != nil 

at this point, the cur hier has been updated to APIserver with parent being changed. But the 
affected NS are not updated in APIServer because of the restart of controller. 

Now, once controller is restarted, I don't see a logic to fix the wrong hierarchy in APIServer.

adrianludwin:
> For example, somebody restarts the controller manager right before executing this method.
> 
> if err := r.updateAffected(ctx, log, affected); err != nil
> 
> at this point, the cur hier has been updated to APIserver with parent being changed. But the
> affected NS are not updated in APIServer because of the restart of controller.
> 
> Now, once controller is restarted, I don't see a logic to fix the wrong hierarchy in APIServer.

That's fine, because once the controller is restarted, it will reconcile _all_ the namespaces, including the ones that would have been queued up for reconciliation by `updateAffected`. Maybe I should change the name of that function to `enqueueAffected` instead to be clear that I'm just requesting a perfectly normal reconciliation. Does that make sense?

Fei-Guo:
"Because once the controller is restarted, it will reconcile _all_ the namespaces, including the ones that would have been queued up for reconciliation by `updateAffected`. " 

Don't you have an order problem here? If the controller restarts, the controller needs to reconcile the ns tree in bottom-up order.  Otherwise, how the parent ns know what the right childns are from Forest cache?

adrianludwin:
> "Because once the controller is restarted, it will reconcile _all_ the namespaces, including the ones that would have been queued up for reconciliation by `updateAffected`. "
> 
> Don't you have an order problem here? If the controller restarts, the controller needs to reconcile the ns tree in bottom-up order. Otherwise, how the parent ns know what the right childns are from Forest cache?

A parent being synced before its child isn't a problem - it will be synced and its status will be updated with "no children," then its child will be synced, and the child will force the parent to be _resynced_, at which point the parent will notice that it has children.

I do currently have a bug that if the child is synced *first*, it will think it has a missing parent because the forest won't contain an entry for the parent. But that bug existed before this change as well. Why don't I fix that on my next checkin.

Fei-Guo:
Adrian will fix start up problem in next change. This change LGTM.

PR #65, opened 2019-09-20
Report any conflicts in propagated objects

yiqigao217:
This handles conflicts in propagated objects. It will delete obsolete objects or set annotation to modified objects.

If there's any conflicts in the propagated objects, it will prevent further propagation.

### DIFF
--- a/incubator/hnc/controllers/object_controller.go
+++ b/incubator/hnc/controllers/object_controller.go
@@ -18,6 +18,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -34,6 +35,7 @@ import (
 const (
 	metaGroup          = "hnc.x-k8s.io"
 	labelInheritedFrom = metaGroup + "/inheritedFrom"
+	annotationModified = metaGroup + "/modified"
 )
 
 // ObjectReconciler reconciles generic propagated objects. You must create one for each
@@ -102,9 +104,10 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
 // update deletes this object if it's an obsolete copy, and otherwise ensures it's been propagated
 // to any child namespaces.
 func (r *ObjectReconciler) update(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) error {
-	// Make sure this object is actually supposed to exist. If not, delete it. This should
-	// trigger one more reconciliation cycle and we'll get the children then.
-	if deleted, err := r.deleteObsolete(ctx, log, inst); deleted || err != nil {
+	// Make sure this object is correct and supposed to propagate. If not, delete it or mark it as
+	// modified. This should trigger one more reconciliation cycle and we'll get the children then.
+	if correct, err := r.ensureCorrectAncestry(ctx, log, inst); !correct || err != nil {
+		log.Info("Incorrect will not propagate")
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


adrianludwin:
/hold
We should add at least one integ test, plus squash all the commits prior to merging.

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,96 @@ func (r *ObjectReconciler) ReconcileNamespace(ctx context.Context, log logr.Logg
 // update deletes this object if it's an obsolete copy, and otherwise ensures it's been propagated
 // to any child namespaces.
 func (r *ObjectReconciler) update(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) error {
-	// Make sure this object is actually supposed to exist. If not, delete it. This should
-	// trigger one more reconciliation cycle and we'll get the children then.
-	if deleted, err := r.deleteObsolete(ctx, log, inst); deleted || err != nil {
+	// Make sure this object is correct and supposed to propagate. If not, delete it or mark it as
+	// modified. This should trigger one more reconciliation cycle and we'll get the children then.
+	if success, err := r.ensureCorrectAncestry(ctx, log, inst); !success || err != nil {
```
adrianludwin: Nit: the function's called ensureCorrect so maybe call the resulting variable `correct` as opposed to `success`. We're not checking whether some operation succeeded or failed (that's what `err`is for), just whether this object was supposed to be here or not.

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,96 @@ func (r *ObjectReconciler) ReconcileNamespace(ctx context.Context, log logr.Logg
...
+		return ""
 	}
 	inheritedFrom, _ := labels[labelInheritedFrom]
-	if inheritedFrom == "" {
-		// this isn't a copy
-		return false, nil
+	return inheritedFrom
+}
+
+// getSourceInst gets source instance of this one. It returns nil if this instance isn't a copy.
```
adrianludwin: Update this comment to say what the `bool` return variable is

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,96 @@ func (r *ObjectReconciler) ReconcileNamespace(ctx context.Context, log logr.Logg
...
-			// delete). TODO: revisit this.
-			log.Error(err, "deleteObsolete")
-			return false, err
-		} else if yes {
-			// We found the object in an ancestor namespace; no need to delete it.
-			return false, nil
+	if err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("The source object no longer exists", "inheritedFrom", srcNS)
+			return true, nil, err
```
adrianludwin: Don't return `err`, just return `nil` here because this isn't really an error to the caller

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,96 @@ func (r *ObjectReconciler) ReconcileNamespace(ctx context.Context, log logr.Logg
...
+	return false, src, nil
+}
+
+// ensureCorrectAncestry checks to see if the given object is correct. It returns true if
+// it's correct and ready to be propagated. The incorrect object will be deleted
+// if it's obsolete; or marked as modified if it's changed.
+func (r *ObjectReconciler) ensureCorrectAncestry(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) (bool, error) {
+	// Find what we expect the source to be.
+	missing, src, err := r.getSourceInst(ctx, log, inst)
+	if err != nil && !missing {
```
adrianludwin: This should usually just be `if err != nil`, especially if you make the other change I've suggested above.

incubator/hnc/controllers/object_controller.go
```diff
@@ -253,6 +288,7 @@ func copyObject(inst *unstructured.Unstructured, ns string) *unstructured.Unstru
 	copied.SetUID("")
 
 	copied.SetNamespace(ns)
```
adrianludwin: I don't think this should be here anymore - move this to the caller.

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,96 @@ func (r *ObjectReconciler) ReconcileNamespace(ctx context.Context, log logr.Logg
...
+	// The source exists. Is it still in an ancestor namespace?
+	if yes, err := r.isAncestor(inst.GetNamespace(), src.GetNamespace()); err != nil {
+		// This can fail if the names are bad. For now, take no action (ie don't
+		// delete). TODO: revisit this.
+		log.Error(err, "deleteObsolete")
+		return false, err
+	} else if yes {
+		// We found the object in an ancestor namespace; compare the copied fields.
+		instCopied := copyObject(inst, "")
+		instCopied.SetLabels(map[string]string{})
```
adrianludwin: Explain why you're overwriting the labels. Also, why not just set this to `nil`? No-one's going to try to add another label.

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,96 @@ func (r *ObjectReconciler) ReconcileNamespace(ctx context.Context, log logr.Logg
...
+		return true, nil
+	}
+
+	// The source exists. Is it still in an ancestor namespace?
+	if yes, err := r.isAncestor(inst.GetNamespace(), src.GetNamespace()); err != nil {
+		// This can fail if the names are bad. For now, take no action (ie don't
+		// delete). TODO: revisit this.
+		log.Error(err, "deleteObsolete")
+		return false, err
+	} else if yes {
```
adrianludwin: I'd change this to `else if !yes { <delete and return> }`. Generally, the longest clause shouldn't be indented if possible.

Better yet, move the `r.isAncestor` check into `getSourceInst` so it returns `missing=true` and you only have one code path that actually deletes the object. However, you should maintain the log messages that have the different reasons for the source not existing.

incubator/hnc/go.sum
```diff
@@ -87,7 +87,7 @@ golang.org/x/text v0.3.0 h1:g61tztE5qeGQ89tm6NTjjM9VPIm088od1l6aSorWRWg=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 h1:+DCIGbF/swA92ohVg0//6X2IVY3KZs6p9mix0ziNYJM=
 golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2/go.mod h1:tRJNPiyCQ0inRvYxbN9jk5I+vvW/OXSQhTDSoE431IQ=
-gomodules.xyz/jsonpatch/v2 v2.0.0 h1:OyHbl+7IOECpPKfVK42oFr6N7+Y2dR+Jsb/IiDV3hOo=
+gomodules.xyz/jsonpatch/v2 v2.0.0 h1:lHNQverf0+Gm1TbSbVIDWVXOhZ2FpZopxRqpr2uIjs4=
```
adrianludwin: Revert?

incubator/hnc/controllers/object_controller.go
yiqigao217: done

incubator/hnc/controllers/object_controller.go
yiqigao217: done

incubator/hnc/controllers/object_controller.go
yiqigao217: done

incubator/hnc/controllers/object_controller.go
yiqigao217: The only possible known differences are 1) namespace and 2) "labelInheritedFrom", so I changed the code to reset these 2 fields before comparison and had a comment above it.
Here, I only reset the labelInheritedFrom label instead of the label map[string]string in case anyone modified the labels. Is it necessary?

incubator/hnc/controllers/object_controller.go
```diff
@@ -102,63 +104,110 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
+
+	// We found the object in an ancestor namespace. Is this object the same as the source?
+	log.Info("Compare with the source", "inheritedFrom", src.GetNamespace())
+	instCopied := copyObject(inst)
+	// Make the namespaces and inheritedFrom labels the same before comparison.
+	instCopied.SetNamespace(src.GetNamespace())
+	setLabel(instCopied, labelInheritedFrom, "")
+	srcCopied := copyObject(src)
+	setLabel(srcCopied, labelInheritedFrom, "")
+	equal := reflect.DeepEqual(instCopied, srcCopied)
```
yiqigao217: Please review the comparison:

The only possible known differences are 1) namespace and 2) "labelInheritedFrom", so I changed the code to reset these 2 fields before comparison and had a comment above it.
Here, I only reset the labelInheritedFrom label instead of the label map[string]string in case anyone modified the labels. Is it necessary?

incubator/hnc/controllers/object_controller.go
yiqigao217: I moved r.isAncestor into getSourceInst and updated the comment.

incubator/hnc/go.sum
yiqigao217: Removed this change file from this commit.

incubator/hnc/controllers/object_controller_test.go
```diff
@@ -52,6 +55,59 @@ var _ = Describe("Secret", func() {
...
+		setParent(ctx, quxName, bazName)
+		Eventually(hasSecret(ctx, bazName, "bar-sec")).Should(BeTrue())
+		Eventually(hasSecret(ctx, quxName, "bar-sec")).Should(BeTrue())
+
+		Eventually(hasAnnotationModified(ctx, bazName, "bar-sec")).Should(BeFalse())
+		Eventually(hasAnnotationModified(ctx, bazName, "baz-sec")).Should(BeFalse())
+
+		modifySecret(ctx, bazName, "bar-sec")
+
+		// Wait 3 more seconds for the annotation to be updated
```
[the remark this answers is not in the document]
yiqigao217: I tried even 4 seconds. It cannot always pass. It could occasionally get a timed out before the hasAnnotationModified set to be true. The default timeout for Eventually() is 2 seconds, so I tried 5s. Is it a concern that it took this long?
Also, is it better to set a custom timeout than use time.sleep()? Since it will hit and proceed before the full timeout.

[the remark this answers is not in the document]
yiqigao217: There may be another racing updates between modify and then setParent(). I will look.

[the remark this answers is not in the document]
adrianludwin: Ok we can leave it for now, but let's have a look again later. Maybe add a comment that it didn't work at 4s and you needed to increase it, and add a TODO to see why it's so slow.

incubator/hnc/controllers/object_controller_test.go
```diff
@@ -52,6 +55,59 @@ var _ = Describe("Secret", func() {
...
+		setParent(ctx, quxName, bazName)
+		Eventually(hasSecret(ctx, bazName, "bar-sec")).Should(BeTrue())
+		Eventually(hasSecret(ctx, quxName, "bar-sec")).Should(BeTrue())
+
+		Eventually(hasAnnotationModified(ctx, bazName, "bar-sec")).Should(BeFalse())
+		Eventually(hasAnnotationModified(ctx, bazName, "baz-sec")).Should(BeFalse())
+
+		modifySecret(ctx, bazName, "bar-sec")
+
+		// Wait 3 more seconds (default timeout is 2s) for the annotation to be updated.
```
adrianludwin: Update to "5s" and state that even 4s wasn't enough

incubator/hnc/controllers/object_controller_test.go
```diff
@@ -52,6 +55,59 @@ var _ = Describe("Secret", func() {
...
+		// Set tree as qux -> bar -> foo and make sure the propagation of foo-sec is
+		// *fully finished* before modifying the foo-sec in bar namespace
+		setParent(ctx, barName, fooName)
+		setParent(ctx, quxName, barName)
+		Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeTrue())
+		Eventually(hasSecret(ctx, quxName, "foo-sec")).Should(BeTrue())
+
+		modifySecret(ctx, barName, "foo-sec")
+
+		// Wait 3 more seconds (default timeout is 2s) for the annotation to be updated
```
adrianludwin: Ditto

adrianludwin:
@Fei-Guo - this lgtm. What do you think? (Yiqi is working with me on the HNC)

adrianludwin:
Hey @rjbez17 , Fei is busy, would have a chance to review this PR from my colleague? Thanks!

PR #66, opened 2019-09-20
Lets namespaces be created/reconciled out-of-order

adrianludwin:
Previously, if we hadn't actually seen a namespace on the apiserver, we
wouldn't create an entry for it in the forest. With this change, we
create an entry in the forest for any *referenced* namespace, even if it
doesn't exist, and mark that ns as existing or not existing as
appropriate. This allows us to catch state changes (ie, not existing to
existing) and therefore to update any other affected namespace.

More concretely, this allows namespaces to be created out of order, and
as a side effect elegantly solves reconciliation problems when the
controller is started (since the controller reconciles namespaces in an
arbitrary order). If a child is reconciled first, it may briefly be
given the "missing parent" condition, but when the parent is later
reconciled, it will notice that the child needs to be reconciled too,
and the condition will then be removed.

This fully fixed issue #59 .

Tested: new integ test, manual

### DIFF
--- a/incubator/hnc/controllers/hierarchy_controller.go
+++ b/incubator/hnc/controllers/hierarchy_controller.go
@@ -18,7 +18,6 @@ package controllers
 import (
 	"context"
 	"reflect"
-	"sort"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -95,7 +94,6 @@ func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm
 		}
 
 		// It doesn't exist - initialize it to a sane initial value.
-		log.V(1).Info("Missing singleton")
 		missing = true
 		inst.ObjectMeta.Name = tenancy.Singleton
 		inst.ObjectMeta.Namespace = nm
@@ -121,9 +119,9 @@ func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm
 				return err
 			}
 
-			// The namespace is gone. Remove all traces of the namespace from  our data structures.
-			log.Info("Namespace was deleted")
-			return r.onDelete(ctx, log, nm)
+			// The namespace is gone, update things accordingly.
+			log.Info("Namespace does not exist")
+			return r.removeNamespace(ctx, log, nm)
 		}
 
 		// If the namespace is being deleted, we should still sync any changes that we might have missed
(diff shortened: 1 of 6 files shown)


rjbez17:
/hold
/approve
/lgtm

@Fei-Guo  feel free to remove hold when satisfied

incubator/hnc/pkg/forest/forest.go
```diff
@@ -51,70 +51,76 @@ func (f *Forest) AddOrGet(nm string) *Namespace {
...
+	exists   bool
+}
+
+// Exists returns true if the namespace exists.
+func (ns *Namespace) Exists() bool {
+	return ns.exists
+}
+
+// Create marks this namespace as existing, returning true if didn't previously exist.
+func (ns *Namespace) Create() bool {
```
Fei-Guo: Can we add an SetExists() interface and replace Create with 
If !ns.Exist() {
   ns.SetExist()
}
I feel this improves readability.

incubator/hnc/controllers/hierarchy_controller.go
```diff
@@ -203,20 +209,26 @@ func (r *HierarchyReconciler) updateObjects(ctx context.Context, log logr.Logger
...
+	// reconciling this namespace, mark its relatives as being affected since they may have been
+	// waiting for this parent.
+	ns := r.Forest.Get(inst.ObjectMeta.Namespace)
+	if ns.Create() {
+		log.Info("Found new namespace")
+		affected = append(affected, ns.RelativesNames()...)
+	}
+
+	// Sync this namespace with its current parent.
+	var curParent *forest.Namespace
```
Fei-Guo: I am a little confused about ns deletion workflow.
If a ns is the parent of child namespaces, what will happen if this parent ns is deleted? Who is responsible to change the parent field of the singleton in all child namespaces to nil? or nothing will happen besides child ns always stays in missing parent condition.

incubator/hnc/pkg/forest/forest.go
adrianludwin: Done. Also changed `Remove` to `UnsetExists`.

incubator/hnc/controllers/hierarchy_controller.go
adrianludwin: Right now, the child will simply stay in the "missing parent" condition. We may add the ability to add an `ownersReference` from the child to the parent, which would automatically delete the child when the parent is deleted as well. However, by default, if the parent is deleted, the child just sticks around.

PR #72, opened 2019-09-23
Prevent cycles in validating webhook

adrianludwin:
This is the first nontrivial validating webhook. It determines if a
change would cause an error condition due to the structure of the
hierarchy and, if so, returns the error.

Tested: manually plus new unit tests


adrianludwin:
Will resubmit in a different order

PR #78, opened 2019-09-26
Prevent cycles in validating webhook

adrianludwin:
This is the first nontrivial validating webhook. It determines if a
change would cause an error condition due to the structure of the
hierarchy and, if so, returns the error.

Tested: manually plus new unit tests

### DIFF
--- a/incubator/hnc/pkg/forest/forest.go
+++ b/incubator/hnc/pkg/forest/forest.go
@@ -2,6 +2,7 @@
 package forest
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -100,16 +101,8 @@ func (ns *Namespace) clean() {
 // children of its own parent, if necessary. It may return an error if the parent is illegal, i.e.
 // if it causes a cycle. It cannot cause an error if the parent is being set to nil.
 func (ns *Namespace) SetParent(p *Namespace) error {
-	// Check for cycles
-	if p != nil {
-		// Simple case
-		if p == ns {
-			return fmt.Errorf("%q cannot be set as its own parent", p.name)
-		}
-		if chain := p.AncestoryNames(ns); chain != nil {
-			return fmt.Errorf("cycle when making %q the parent of %q: current ancestry is %q",
-				p.name, ns.name, strings.Join(chain, " <- "))
-		}
+	if reason := ns.CanSetParent(p); reason != "" {
+		return errors.New(reason)
 	}
 
 	// Remove old parent and cleans it up.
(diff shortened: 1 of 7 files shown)


incubator/hnc/pkg/forest/forest.go
```diff
@@ -100,16 +101,8 @@ func (ns *Namespace) clean() {
...
-	if p != nil {
-		// Simple case
-		if p == ns {
-			return fmt.Errorf("%q cannot be set as its own parent", p.name)
-		}
-		if chain := p.AncestoryNames(ns); chain != nil {
-			return fmt.Errorf("cycle when making %q the parent of %q: current ancestry is %q",
-				p.name, ns.name, strings.Join(chain, " <- "))
-		}
+	if reason := ns.CanSetParent(p); reason != "" {
```
Fei-Guo: reason -> err?

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -17,24 +21,77 @@ const (
...
 }
 
 func (v *Hierarchy) Handle(ctx context.Context, req admission.Request) admission.Response {
-	log := v.Log.WithValues("namespace", req.Namespace)
-	log.Info("Validating")
-	hier := &tenancy.Hierarchy{}
-	err := v.decoder.Decode(req, hier)
+	log := v.Log.WithValues("ns", req.Namespace)
+
+	if isHNCServiceAccount(req.AdmissionRequest.UserInfo) {
```
Fei-Guo: Why HNC SA can do anything even when he potentially can create a cycle?

incubator/hnc/pkg/forest/forest.go
adrianludwin: I didn't really want to return an error, since we're not really requesting an operation here. It's more like a dry run. I can change it if you feel strongly though.

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: Because the object might _already_ be in a bad state - that is, a state that would not pass validation. But the controller would like to update `hier.status` to _show_ that it's in a bad state.

Is there a better way to do this?

Fei-Guo: I'd suggest HNC SA delete/recreate hier CR if it is in bad state.

incubator/hnc/pkg/forest/forest.go
Fei-Guo: Effectively, a non empty reason means something is wrong, hence the suggestion. Not a strong opinion though.

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: But the bad state could be set by the user, so we can't simply delete and recreate it.

I feel like I've heard that there's a way to update the status without the spec, why don't I open an issue for that.

adrianludwin: Filed #80

PR #82, opened 2019-09-27
Reconcile namespace and singleton in tandem

adrianludwin:
Passes the namespace object to the `reconcile` function to allow it to
be modified at the same time as the singleton. This will enable more
sophisticated reconciliations including self-service namespace creation
and adding labels to the namespace to indicate ancestry.

Tested: existing integ tests; manual with subsequent commits.

_See https://github.com/adrianludwin/multi-tenancy/commits/self-serve for the full context of this PR_

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -81,60 +81,30 @@ func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 }
 
 func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm string) error {
-	update := true
-	missing := false
-
-	// Get the hierarchy itself.
-	nnm := types.NamespacedName{Namespace: nm, Name: tenancy.Singleton}
-	inst := &tenancy.Hierarchy{}
-	if err := r.Get(ctx, nnm, inst); err != nil {
-		if !errors.IsNotFound(err) {
-			log.Error(err, "Couldn't read singleton")
-			return err
-		}
-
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -81,60 +81,30 @@ func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-	// If the hierarchy object exists but is being deleted, we won't update it when we're finished
-	// syncing (we should sync our internal data structure anyway just in case something's changed).
-	// I'm not sure if this is the right thing to do but the kubebuilder boilerplate included this
-	// case explicitly.
-	if !inst.GetDeletionTimestamp().IsZero() {
-		log.Info("Singleton is being deleted")
-		update = false
+	ns, err := r.getNamespace(ctx, nm)
+	if err != nil {
+		log.Error(err, "Couldn't read namespace")
```
rjbez17: If we get here is means the singleton exists but the namespace doesn't, wouldn't we want to cleanup the singleton before returning?

adrianludwin: Firstly, if the singleton doesn't exist, `getSingleton` doesn't return an error - it just returns an object that hasn't been committed to the apiserver. `getNamespace` works the same way.

Secondly, the singleton is _inside_ the namespace, so K8s won't allow it to exist if the namespace doesn't.

Does that answer your question?

rjbez17: Sorry, I missed the `!` in this check below:

```
		if !errors.IsNotFound(err) {
			return nil, err
		}
```

Issue #90, opened 2019-10-03
HNC: Self-service namespace creation


adrianludwin:
Allow tenant admins who do not have cluster-wide namespace creation privileges to still create subnamespaces. Webhooks will be addressed by subsequent issues; this issue only tracks the core functionality (the reconciler).

These changes are already underway with #84; the remaining commits in https://github.com/adrianludwin/multi-tenancy/tree/self-serve will follow once this is merged.

adrianludwin:
Completed as a part of #95

PR #95, opened 2019-10-03
Self-service subnamespace creation (reconciler only)

adrianludwin:
Instead of creating subnamespaces as a two-step process, this adds an
method to create them directly. This is the first time we also modify
namespaces, as opposed to simply hierarchy singletons. This is part of
#90.

This change does *not* include a validating webhook to prevent most
failure cases that are identified by the conditions; this will come in a
future commit. It also doesn't attempt to set up ownersReferences to
automatically delete child namespaces when parents are deleted, although
we should probably do that too.

Tested: manual plus new integ tests


adrianludwin:
/assign @Fei-Guo 

Last one for now then I'll leave you alone for a bit :)

PR #114, opened 2019-10-05
Rename 'patches' to 'patchesStrategicMerge' and update README with missing dependencies install intructions

danielSbastos:
Hey there delilah,

This PR is being done due to the difficulties I encountered when setting up the HNC project. I have added them to the README. 

There is one change, though, which is the renaming `patches` to `patchesStrategicMerge` on the config yml files due to this issue https://github.com/kubernetes-sigs/kustomize/issues/1429

### DIFF
--- a/incubator/hnc/README.md
+++ b/incubator/hnc/README.md
@@ -26,6 +26,11 @@ or just join a MTWG meeting.
 
 ## Usage
 
+Make sure you have downloaded the following libraries/packages:
+  - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+  - [kustomize](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md)
+  - [kubebuilder](https://github.com/kubernetes-sigs/controller-runtime/issues/90#issuecomment-494878527) (_Github issue_). You will most likely encounter this issue when running the tests or any other command.
+
 Install the operator as you would any other kubebuilder controller (eg `make
 install`, `make deploy`). No prebuilt images exist yet.
 
@@ -69,7 +74,7 @@ I do most of my testing in [KIND](https://kind.sigs.k8s.io).
 
 ### Testing on KIND (Kubernetes IN Docker)
 
-* Run `. devenv` to setup your `KUBECONFIG` and `PATH` env vars correctly.
+* Run `. devenv` (or `source devenv`) to setup your `KUBECONFIG` and `PATH` env vars correctly.
 * Run `./kind-reset` to stop any existing KIND cluster and setup a new one,
   including the cert manager required to run the webhooks.
 * Run `make test` to run the controller (excluding the validating webhook)
--- a/incubator/hnc/config/default/kustomization.yaml
+++ b/incubator/hnc/config/default/kustomization.yaml
@@ -19,7 +19,7 @@ bases:
 - ../webhook
 - ../certmanager
 
-patches:
+patchesStrategicMerge::
 - manager_image_patch.yaml
   # Protect the /metrics endpoint by putting it behind auth.
   # Only one of manager_auth_proxy_patch.yaml and
--- a/incubator/hnc/config/crd/kustomization.yaml
+++ b/incubator/hnc/config/crd/kustomization.yaml
@@ -5,7 +5,7 @@ resources:
 - bases/hnc.x-k8s.io_hierarchies.yaml
 # +kubebuilder:scaffold:crdkustomizeresource
 
-patches:
+patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
 #- patches/webhook_in_hierarchies.yaml


adrianludwin:
/hold

Hey Daniel, can you explain what you did a bit more? One of the [comments in the issue you referenced](https://github.com/kubernetes-sigs/kustomize/issues/1429#issuecomment-520061233) suggested that `kustomize edit fix` would fix the problem, while another described this as a "workaround." Is this actually just a workaround (in which case, you should add comments to the lines you're changing) or a permanent change in the way `kustomize` works?

cc @lafh and @yiqigao217 whom I think were running into the same problem yesterday.

incubator/hnc/README.md
```diff
@@ -69,7 +74,7 @@ I do most of my testing in [KIND](https://kind.sigs.k8s.io).
 
 ### Testing on KIND (Kubernetes IN Docker)
 
-* Run `. devenv` to setup your `KUBECONFIG` and `PATH` env vars correctly.
+* Run `source devenv` to setup your `KUBECONFIG` and `PATH` env vars correctly.
```
adrianludwin: Why this change?

incubator/virtualcluster/config/default/kustomization.yaml
```diff
@@ -28,7 +28,7 @@ resources:
 - ../rbac/auth_proxy_role.yaml
 - ../rbac/auth_proxy_role_binding.yaml
 
-patches:
+patchesStrategicMerge:
```
adrianludwin: Let's not make changes across _both_ HNC and non-HNC code. Can you please revert the changes to `virtualcluster`?

tenant/config/default/kustomization.yaml
```diff
@@ -28,7 +28,7 @@ resources:
 - ../rbac/auth_proxy_role.yaml
 - ../rbac/auth_proxy_role_binding.yaml
 
-patches:
+patchesStrategicMerge:
```
adrianludwin: Ditto, please revert the change to `tenant`

tenant/vendor/sigs.k8s.io/controller-tools/testData/config/default/kustomization.yaml
```diff
@@ -21,5 +21,5 @@ resources:
 - ../rbac/*.yaml
 - ../manager/*.yaml
 
-patches:
+patchesStrategicMerge:
```
adrianludwin: In general, you should never make a change inside a `/vendor` directory in Go. `/vendor` is the old way of importing dependencies (the new way is called [modules](https://blog.golang.org/using-go-modules)) and they should always reflect the original source code as much as possible.

In this case, it's also unnecessary - the only point to putting something in `/vendor` is to let the Go compiler find `.go` files, so no-one will ever run `kustomize` in that directory too.

incubator/virtualcluster/config/default/kustomization.yaml
danielSbastos: Ah, sure. You right, I changed all the locations that had `patches` and forgot about the virtualcluster files. My bad! I'll change this.

tenant/vendor/sigs.k8s.io/controller-tools/testData/config/default/kustomization.yaml
danielSbastos: :man_facepalming: I'm, sorry. I did a `sed` on my repo and this passed my eyes. I'll revert it.

incubator/hnc/README.md
danielSbastos: The previous command (`. dotenv`) didn't work for me, but `source dotenv`. I think I'll add this as another option rather than removing the first one.

adrianludwin: Ah, what shell are you using? I think `.` only works in bash.

danielSbastos:
> /hold
> 
> Hey Daniel, can you explain what you did a bit more? One of the [comments in the issue you referenced](https://github.com/kubernetes-sigs/kustomize/issues/1429#issuecomment-520061233) suggested that `kustomize edit fix` would fix the problem, while another described this as a "workaround." Is this actually just a workaround (in which case, you should add comments to the lines you're changing) or a permanent change in the way `kustomize` works?
> 
> cc @lafh and @yiqigao217 whom I think were running into the same problem yesterday.

Hey Adrian. `kustomize edit fix` does what I just did and more. 

```sh
» kustomize edit fix --help                                                                                                            
Fix the missing fields in kustomization file

Usage:
  kustomize edit fix [flags]

Examples:

	# Fix the missing and deprecated fields in kustomization file
	kustomize edit fix
```

It fixes the deprecated fields (`patches` field included). E.i., it would do this renaming for me instead of my manual intervention.

The question If it is a workaround, based on other issues, is that doesn't seem to be. `patches` seems deprecated since kustomize [v1.0.9](https://github.com/kubernetes-sigs/kustomize/blob/v1.0.9/pkg/types/kustomization.go#L129). 

I didn't run `kustomize edit fix`, but rather altered manually and just one item in the file. I'll run it in order to do a full deprecation review and changing. WDYT?

incubator/hnc/README.md
danielSbastos: I use zsh :upside_down_face:

adrianludwin:
Ok, that sounds good. Here's what I'd do:

- First, on your current branch (`improve-readme-and-fix-yml-config-syntax`), throw away all your changes except to the readme file:
  - Run `git reset master`, which will move your branch pointer back to the unmodified master, but with your changes still on disk. It will look as though you'd just edited the files but hadn't committed them yet.
  - Run `git restore <modified files>` on the files you _don't_ want to keep
- Then run `kustomize edit fix`
- Commit your changes locally
- Push the branch. It will fail if you don't use the `--force` option because it will think you're mistakenly overwriting the commits you'd previously made, so make sure you use `--force`.

Make sense?

danielSbastos:
Yup, makes sense.

incubator/hnc/config/default/kustomization.yaml
```diff
@@ -12,55 +12,59 @@ namePrefix: hnc-
...
   # endpoint w/o any authn/z, uncomment the following line and
   # comment manager_auth_proxy_patch.yaml.
   # Only one of manager_auth_proxy_patch.yaml and
   # manager_prometheus_metrics_patch.yaml should be enabled.
 #- manager_prometheus_metrics_patch.yaml
 
-- manager_webhook_patch.yaml
 
+patchesStrategicMerge:
+- manager_image_patch.yaml
```
adrianludwin: Can you move the comments as well?

incubator/hnc/config/default/kustomization.yaml
```diff
@@ -12,55 +12,59 @@ namePrefix: hnc-
...
+    version: v1
+- fieldref: {}
+  name: SERVICE_NAME
   objref:
     kind: Service
-    version: v1
     name: webhook-service
+    version: v1
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
adrianludwin: Move these two lines to the top of the file?

incubator/hnc/config/default/kustomization.yaml
```diff
@@ -12,55 +12,59 @@ namePrefix: hnc-
...
+resources:
+- ../crd
+- ../rbac
+- ../manager
+- ../webhook
+- ../certmanager
```
adrianludwin: Move these up to ~line 14 so that they line up with where `bases:` used to be?

incubator/hnc/config/default/kustomization.yaml
```diff
@@ -12,55 +12,59 @@ namePrefix: hnc-
...
-- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
+- fieldref:
+    fieldPath: metadata.namespace
+  name: CERTIFICATE_NAMESPACE
   objref:
-    kind: Certificate
     group: certmanager.k8s.io
+    kind: Certificate
+    name: serving-cert
     version: v1alpha1
```
adrianludwin: Can you reorder these so they match the original order? Note that the first line in each group must have the dash, and maybe try to preserve the comments as well. So the first line should be:

`  - name: CERTIFICATE_NAMESPACE # namespace of the certificate CR`

I'm not sure why `kustomize edit fix` reordered these lines :/ It was probably easier to reorder them than to preserve the existing order. Just make sure that if it _added, changed or removed_ something, you maintain preserve that change.

danielSbastos: Sure, makes total sense! Not sure why it reordered too.

danielSbastos:
@adrianludwin, it seems like `kustomize` reordered things and just altered the `patches` to `patchesStrategicMerge`. I will reset that file and do this single change manually, WDYT?

adrianludwin:
> @adrianludwin, like we saw, `kustomize` reordered things and just altered the `patches` to `patchesStrategicMerge`. I will reset that file and do this single change manually, WDYT? It's easier than re-reordering everything 😛

sgtm!

(In case you don't know, that's Googlish for "Sounds Good To Me"; "lgtm" = "Looks Good To Me" which means "I approve this PR")

incubator/hnc/config/default/kustomization.yaml
adrianludwin: My guess is:

1. It reads the YAML into a Go map
2. It makes the necessary changes
3. It writes the values from the Go maps back as YAML, but with the keys sorted in alphabetical order

This is way easier, but less user-friendly, than trying to preserve the existing line ordering _except_ for the bits that have changed.

Issue #76, opened 2019-09-26
HNC: Add proper conditions to Hierarchy status


adrianludwin:
Currently, the [conditions](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/api/v1alpha1/hierarchy_types.go#L68) in the HNC status are simply a set of strings. In addition, conditions don't propagate between namespaces. This is a proposal to make them more structured, machine-readable, and useful.

# Error codes and references

We should expand the `Condition` struct to include the following new fields:

- `code`: a machine-readable string value (values shown below) summarizing the condition.
- `affects`: a list of group-version-kind-namespace-name that uniquely identifies the object(s) affected by the condition.

## Initial list of conditions.

Some condition codes will start with the prefix `CRIT_` which indicates that this is a critical error; until the error is resolved, HNC will no longer make any changes in this namespace (or its descendants).

I would recommend the following _initial_ set of codes (clients should expect that the list can gro over time):

- `CRIT_PARENT_MISSING`: The specified parent is missing.
  - `affected` and `msg` are empty.
- `CRIT_PARENT_INVALID`: The specified parent is invalid (ie would cause a cycle).
  - `affected` is empty and `msg` describes the cycle.
- `CRIT_ANCESTOR`: A critical error exists in an ancestor namespace, so this namespace is no longer being updated.
  - `affected[0].name` identifies the most recent ancestor with a critical condition. `msg` is empty.
- `OBJECT_OVERRIDDEN`: An object in this namespace has been overridden from its parent and will no longer be updated.
  - `affected[x].{group,version,kind,namespace,name}` is set for every such object. `msg` _may_ summarize the errors in a human-readable form.
- `OBJECT_DESCENDANT_OVERRIDDEN`: An object in this namespace is no longer being propagated because a propagated copy has been modified.
  - `affected[x].{group,version,kind,namespace,name}` identifies the modified copy, and `msg` _may_ summarize the errors.
  - Note that `OBJECT_OVERRIDDEN` will be set for the same object in its own namespace
  - This condition should be set in the namespace that contains the source object. If the source is in a grandparent, and the modification is in the child, `DESCENDANT_OVERRIDDEN` will _not_ be set on the intermediate namespace.

## Implementation

[...]

Issue #112, opened 2019-10-05
HNC: stop modifying source files whenever we call `make docker-build`


adrianludwin:
See the following fairly terrifying lines in `Makefile` (added by `kubebuilder` when it first scaffolds the project):

```
        docker build . -t ${IMG}
        @echo "updating kustomize image patch file for manager resource"
        sed -i'' -e 's@image: .*@image: '"${IMG}"'@' ./config/default/manager_image_patch.yaml
```

This means that every time we run `make docker-build`, _our source files get modified_! We should not do this. I'm not sure what we _should_ do. Maybe have a non-default kustomization that we can `.gitignore`?

adrianludwin:
/assign @adrianludwin 

See https://github.com/adrianludwin/multi-tenancy/commit/194bef94a9653f894f0d0c05bf42df750ca9e112

adrianludwin:
Fixed via #161 and #164

PR #115, opened 2019-10-05
Add proper conditions to Hierarchy status

yiqigao217:
Part of issue #76 

Store locally detectable CRIT_ conditions into in-memory forest while reconciling Hierarchy and update all conditions from forest into Hierarchy object.


adrianludwin:
Can you rebase to remove my two commits from this PR? Thanks!

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -19,6 +19,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
```
adrianludwin: Maybe add a comment here to say that:

- Any changes here need to also be reflected in the kubebuilder:validation:Enum comment, below.
- The meanings of all these constants are defined in the comment to `Condition.Code`, below.

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -70,8 +81,35 @@ type HierarchyList struct {
...
+// Code is a machine-readable string value summarizing the condition.
+// +kubebuilder:validation:Enum=CRIT_PARENT_MISSING;CRIT_PARENT_INVALID;CRIT_REQUIRED_CHILD_CONFLICT;CRIT_ANCESTOR;OBJECT_OVERRIDDEN;OBJECT_DESCENDANT_OVERRIDDEN
+type Code string
+
+// Condition specifies the condition and the affected objects.
 type Condition struct {
-	Msg string `json:"msg"`
+	// Defines the conditions in a machine-readable string value.
+	// Valid values are:
+	// - "CRIT_PARENT_MISSING": the specified parent is missing;
```
adrianludwin: Nit: I'd drop the semicolon at the end of each line since these are all independent bullet points.

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchies.yaml
```diff
@@ -52,17 +52,56 @@ spec:
...
+                          type: string
+                        namespace:
+                          type: string
+                        version:
+                          type: string
+                      type: object
+                    type: array
+                  code:
+                    description: 'Defines the conditions in a machine-readable string
+                      value. Valid values are: - "CRIT_PARENT_MISSING": the specified
```
adrianludwin: I'm noticing that all your newlines seem to be removed here. Perhaps adding a blank line between each of the bullet points in the comment will help?

incubator/hnc/config/default/manager_image_patch.yaml
```diff
@@ -11,5 +11,5 @@ spec:
       # based on the current value of the IMG envvar. Please try not to check
       # in changes; leave the default value (currently "controller:latest"
       # unless/until we choose something better).
-      - image: controller:latest
+      - image: controller:kind-local
```
adrianludwin: Please don't check this in (it's not your fault, see #112)

incubator/hnc/pkg/forest/forest.go
```diff
@@ -46,12 +48,16 @@ func (f *Forest) Get(nm string) *Namespace {
 	if ok {
 		return ns
 	}
-	ns = &Namespace{forest: f, name: nm, children: namedNamespaces{}}
+	ns = &Namespace{forest: f, name: nm, children: namedNamespaces{}, conditions: conditions{}}
```
adrianludwin: Maybe this is getting long enough to split into multiple lines?

```
ns = &Namespace{
  forest: f,
  name: nm,
  children: namedNamespaces{},
  conditions: conditions{},
}
```

incubator/hnc/pkg/forest/forest.go
```diff
@@ -205,3 +220,83 @@ func (ns *Namespace) IsAncestor(other *Namespace) bool {
...
 	return ns.parent.IsAncestor(other)
 }
+
+// ClearConditions clears local conditions in the namespace.
+func (ns *Namespace) ClearConditions(key string) {
+	delete(ns.conditions, key)
+}
+
+// SetCondition adds a condition into the list of conditions for key string, returning
+// true if it does not exist previously.
```
adrianludwin: I don't think we ever use this return value? If so, maybe take it out?

incubator/hnc/pkg/forest/forest.go
```diff
@@ -205,3 +220,83 @@ func (ns *Namespace) IsAncestor(other *Namespace) bool {
...
+	if len(flattened) > 0 {
+		return flattened
+	}
+	return nil
+}
+
+// getAffectedObject gets AffectedObject from a namespace or a string of form group/version/kind/namespace/name
+func getAffectedObject(gvknn string) tenancy.AffectedObject {
+	if gvknn == Local {
+		return tenancy.AffectedObject{}
```
adrianludwin: I think `return nil` would be better here?

adrianludwin:
Looking good so far!

/assign @rjbez17

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: Done

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: done

rjbez17:
lgtm after @adrianludwin's comments are addressed.

/hold
/approve

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchies.yaml
yiqigao217: I added blank lines in the comments and did see the newlines "/n" here. However, I checked my previous commits, there were no newlines either. https://github.com/yiqigao217/multi-tenancy/commit/d252f90663a768661b7ac4f470b13bd3bfa36bc4

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: done

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: done

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: done

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchies.yaml
adrianludwin: That's fine if this wasn't looking good before, let's just make it better now :) Thanks!

incubator/hnc/config/default/manager_image_patch.yaml
yiqigao217: done

incubator/hnc/pkg/forest/forest.go
yiqigao217: done

incubator/hnc/pkg/forest/forest.go
yiqigao217: done

incubator/hnc/pkg/forest/forest.go
yiqigao217: Currently this func returns a copy of the AffectedObject struct. I changed it into returning a pointer and an error (*tenancy.AffectedObject, error). It returns error if the key is not valid to get an AffectedObject.

adrianludwin: As discussed offline, don't return an error here since we don't have any way to handle it. Instead, maybe return an `AffectedObject` with the namespace set to "INVALID OBJECT" (which is not a valid namespace name) and the name set to the key. Also pass in the `log` so you can log an error.

Also, don't return a pointer like I suggested - I misread this and thought it was a slice, but it's a single object.

adrianludwin:
lgtm, let's just get rid of those zombie commits then we're good to go.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -468,6 +465,7 @@ func (r *HierarchyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				}},
 			}
 		})
+	r.namespaceLocks = map[string]*sync.Mutex{}
```
adrianludwin: This looks like it crept in from one of the squashed commits? I think this shouldn't be here.

yiqigao217: yes I was working on it. It should be good now!

Issue #91, opened 2019-10-03
HNC: Finalize the initial list of HNC-controlled objects


adrianludwin:
Currently, the HNC will propagate `Roles`, `RoleBindings` and `Secrets`. I'd like to propose the following additional Kinds that get propagated in HNC 0.1:

- `ConfigMap`
- `NetworkPolicy`
- `PodSecurityPolicy`
- `ResourceQuota`
- `LimitRange`

That is, all the common policy objects that an admin might want to set and then forbid subadmins from modifying.

In HNC 0.2 (aka Phase 2 in the design doc), we can look at making this set configurable.

adrianludwin:
/good-first-issue
/unassign

Let's go with the list shown above. This should be an easy PR for someone.

entro-pi:
I'd like to take this on, I studied the problem during my downtime recently and think I have a good idea of how to solve it, or at least the questions I need to ask.

/assign @entro-pi

entro-pi:
Here's a question, should I close this issue? (if I can)?

adrianludwin:
Yup! If you can't, I will.

entro-pi:
Fixed by #116

entro-pi:
Looks like I can't, would you mind?

PR #116, opened 2019-10-07
Passes new kinds to be inherited

entro-pi:
This commit adds five Kinds to be inherited via their GVK
values.

Test: manually via make manifests then make test.

### DIFF
--- a/incubator/hnc/pkg/controllers/setup.go
+++ b/incubator/hnc/pkg/controllers/setup.go
@@ -21,6 +21,11 @@ func Create(mgr ctrl.Manager, f *forest.Forest) error {
 		{Group: "", Version: "v1", Kind: "Secret"},
 		{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "Role"},
 		{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "RoleBinding"},
+		{Group: "networking.k8s.io", Version: "v1", Kind: "NetworkPolicy"},
+		{Group: "policy", Version: "v1beta1", Kind: "PodSecurityPolicy"},
+		{Group: "", Version: "v1", Kind: "ResourceQuota"},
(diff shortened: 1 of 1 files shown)


entro-pi:
addresses #91

adrianludwin:
For a change like this, I'd like to see manual testing that you can actually create all the new types in a parent namespace and see them show up in a child namespace too. Did you do this?

entro-pi:
I have not! This will require multiple clusters, right? If so, I will have
to wait until after Tuesday to test. I can only run one minikube on this
laptop.

adrianludwin:
I'm not sure why it would need multiple clusters - just multiple namespaces
in a cluster. If you can run the demo (referenced here
<https://github.com/kubernetes-sigs/multi-tenancy/tree/master/incubator/hnc#the-hierarchical-namespace-controller-hnc>),
you should be able to create Network Policies, PSPs, etc just like you
would have created secrets.

entro-pi:
Excellent! I'm still new to kubernetes, this is all so fun!

/approve cancel
/w.i.p.

adrianludwin:
You shouldn't have to set these yourself since you don't have permission to
approve or lgtm a change. The "WIP" label is useful just to communicate to
the other reviewers but it's usually not necessary either (you could also
just close the PR and reopen a new one when you're ready, unless there's a
large discussion you'd like to be part of the _same_ PR).

entro-pi:
NIce, thanks!

With this, I'll probably get to running these tests in the next couple of hours. Is that enough time to warrant closing the PR and re-opening it after?

adrianludwin:
Nah, I'd just leave it open.

entro-pi:
/hold cancel
/approve

Testing completed, commit log updated.
All types propagate! Woohoo!

adrianludwin:
lgtm but I'll let @rjbez17 do the honours

PR #118, opened 2019-10-07
Create bash script for CI

danielSbastos:
Hey there delilah,

Here we create a bash script to run the tests and do some prechecking, like `make kubectl`. This script will be called on the Prow config file.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -50,6 +50,9 @@ fmt:
 vet:
 	go vet ./...
 
+test-kustomize:
+	kustomize build config/default
+
 # Generate code
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
--- a/incubator/hnc/hack/ci-test.sh
+++ b/incubator/hnc/hack/ci-test.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -e
+
+echo "Installing 'kubebuilder'"
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.0.0-alpha.1/kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
+tar -zxvf kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
+mv kubebuilder_2.0.0-alpha.1_linux_amd64 /usr/local/kubebuilder
+
+echo "Installing 'kustomize'"
+GO111MODULE=on go get sigs.k8s.io/kustomize/kustomize/v3@v3.2.1
+
+hack_dir=$(dirname ${BASH_SOURCE})
+
+echo "Running 'make kubectl'"
+make kubectl -C "$hack_dir/.."
+
+echo "Running 'make test-kustomize'"
+make test-kustomize -C "$hack_dir/.."
+
+echo "Running 'make tests'"
+make test -C "$hack_dir/.."


incubator/hnc/hack/ci-test.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -e
+
+export PATH=$(go env GOPATH)/bin:$PATH
```
adrianludwin: Why is this part necessary?

incubator/hnc/hack/ci-test.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir -p $(go env GOPATH)/bin
+
+echo "Installing kubebuilder"
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.0.0-alpha.1/kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
+tar -zxvf kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
+mv kubebuilder_2.0.0-alpha.1_linux_amd64 /usr/local/kubebuilder
+echo "Finished installing kubebuilder"
+
+make test -C "${PWD}/incubator/hnc"
```
adrianludwin: Also run `make kubectl`and `make docker-build` since those are different targets not covered by `make test`.

danielSbastos: WDYT of creating a different file for this? One for the tests, which would be this one and another for these sort of preconditions, perhaps a `ci-check-preconditions.sh`. Then at the prow job we call these two files in two separate commands.

incubator/hnc/hack/ci-test.sh
danielSbastos: I based this script on this script: https://github.com/kubernetes-sigs/controller-runtime/blob/master/hack/ci-check-everything.sh#L22, however, I haven't tried without the `export ...` to see if it fails. I'll try it (shouldn't have added this without knowing for sure why it is used)

adrianludwin: You should be able to test this whole thing in the `golang` container. Do you know how to run this script in the golang Docker container?

incubator/hnc/hack/ci-test.sh
adrianludwin: I know that controller-runtime has a bunch of different files but I'm actually fine with leaving it as a single file for now. It's trivial to split it into multiple files in the future, so let's keep it easier to read for now.

incubator/hnc/hack/ci-test.sh
danielSbastos: Yup, I need to mount the dir in order to access this file inside the container. 

In Prow, this script will be run inside a docker container with the golang image. I'll send a commit removing this export and test it locally with Prow. I need to send a commit because Prow uses a PR number and its current state to run locally.

adrianludwin: I'm not sure I fully understand that - you mean that when you pushed this commit, you're somehow getting Prow to run? Where? On your forked repo? Are you running it locally? Where are the instructions you're following? Inquiring minds want to know! :)

danielSbastos: Yup until now I thought that [Prow locally](https://github.com/kubernetes/test-infra/blob/master/prow/build_test_update.md#example) used the current state of the opened PR (it asks for the PR to clone when running locally), which would be this one to run the job, but I just noticed that this is not the case, hahahahhaa :sweat_smile:. It clones the repo and uses the master branch to test it locally. I did an `ls -R` on the job commands and this file, `ci-test.sh`, was not present on the output.

So summing up, I think this PR needs to merged in order for me to test it locally with Prow. Not sure I if explained myself correctly...

danielSbastos: From the clone log, I don't really understand why this file is not present, because it merges this PR with the master. I'm trying to understand.

danielSbastos: But anyways, I can test this file alone and then when this PR is merged. I can test locally with Prow. The code for the test-infra repo is done already, I'll open there after this one is merged :)

incubator/hnc/hack/ci-test.sh
danielSbastos: I'm not sure if I can run `make docker-build` inside a docker container :thinking:

incubator/hnc/hack/ci-test.sh
adrianludwin: Hmm. Have you been able to at least run this script manually in a docker container? i.e., without Prow? If it doesn't work w/o Prow, it definitely won't work _with_ it.

```
docker run -it golang /bin/bash
<manually get the code>
hack/ci-test.sh
<see if it works>
```

adrianludwin: Also, maybe comment out these `PATH` modifying lines for now if you're not sure they're needed. Or leave them in if the script doesn't run without them!

danielSbastos: Yup, I was able to run inside a container. I removed those PATH lines and everything worked.

incubator/hnc/hack/ci-test.sh
danielSbastos: Can I just keep the `make test` and `make kubectl`?

adrianludwin: The section in that blog post called "The Solution" seems to be a reasonable solution - bind the Docker socket so that the other container gets created as a sibling. I'm assuming that the [`containers` list](https://github.com/kubernetes/test-infra/blob/4b879e817d005138c9775102a171f46054cb4303/config/jobs/kubernetes-sigs/controller-runtime/controller-runtime-presubmits.yaml#L12) is a list of [`Container`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#container-v1-core), which means you should be able to access the Docker socket like this: https://estl.tech/accessing-docker-from-a-kubernetes-pod-68996709c04b

incubator/hnc/hack/ci-test.sh
adrianludwin: Ok, either delete them or comment them out for now. And update the commit message (if you haven't already) to point out how you tested this script.

incubator/hnc/hack/ci-test.sh
adrianludwin: On the one hand, `make docker-build` is important because it tests the code inside of `Dockerfile`. Without that test, you could completely break the dockerfile and no-one would know.

On the other hand, it's fine to delay that to a later PR :) Maybe check in this "basic" PR now, then set up the Prow job (with access to the Docker socket), and if that's working, add `make docker-build` back to this script then.

incubator/hnc/hack/ci-test.sh
danielSbastos: Okidoki :smile:

incubator/hnc/hack/ci-test.sh
danielSbastos: Oh, okay. I read the link you gave and it seems reasonable. I prefer to first do this simpler PR and then create another one with the docker build functionality.

adrianludwin:
@rjbez17 Any thoughts on this? It lgtm once the commits have been squashed and a longer commit message goes in, at least as an initial attempt.

incubator/hnc/hack/ci-test.sh
adrianludwin: sgtm

incubator/hnc/hack/ci-test.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -e
+
+echo "Installing 'kubebuilder'"
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.0.0-alpha.1/kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
+tar -zxvf kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
+mv kubebuilder_2.0.0-alpha.1_linux_amd64 /usr/local/kubebuilder
+
+echo "Running 'make kubectl'"
+make kubectl -C "${PWD}/incubator/hnc"
```
adrianludwin: Hmm instead of `${PWD}` which can't control, how about `$BASH/..`? We know exactly where this file is located inside the repo, this is how controller-runtime does it IIRC

lafh:
Could you please add a test to run `kustomize` in this new CI flow? That will be super helpful to catch errors like [these](https://github.com/kubernetes-sigs/multi-tenancy/pull/123)! 

Thanks in advance! 😊

danielSbastos:
> Could you please add a test to run `kustomize` in this new CI flow? That will be super helpful to catch errors like [these](https://github.com/kubernetes-sigs/multi-tenancy/pull/123)!
> 
> Thanks in advance!

Nicee. I think this is good idea. I'll do it

incubator/hnc/Makefile
```diff
@@ -50,6 +50,9 @@ fmt:
 vet:
 	go vet ./...
 
+kustomize:
```
adrianludwin: Maybe test-kustomize? It doesn't actually do anything useful other than testing.

adrianludwin:
@rjbez17 Please remove the hold if you're happy. This the script that Prow will call.

PR #126, opened 2019-10-08
Refactor hierarchy controller

### LINKED ISSUE #109: HNC: Refactor hierarchy controller into smaller methods
The hierarchy controller has the following rough structure:

- `reconcile` loads the singleton and namespace from the apiserver, determines if updates should be written back to the apiserver, calls `updateTree`, and calls `updateObjects` if we're writing updates.
- `updateTree` makes a copy of the singleton and namespace, calls `syncWithForest`, and (if we're updating) writes the objects back to the apiserver.
- `syncWithForest` is very long, but roughly does the following things, mostly inline:
  - Locks the forest
  - If the namespace is being reconciled for the first time, enqueues all relatives
  - If the namespace is a required child, ensure the parent is set correctly
  - Ensures the parent exists
  - If the parent has changed, ensures the new one is legal and enqueues the old/new parent
  - Determine if any required children are missing and handle them
  - Set the conditions

There are two problems with this. Firstly, `syncWithForest` is far too long at about 80 lines and growing continuously. Fortunately, once #76 is done, there will be _no_ local variables that must be maintained throughout the function. Therefore, it will be easy to break up. _I recommend that no one function be over 30 lines if possible_; it should look roughly like the sublist of tasks shown above.

Secondly, the boundary between `reconcile` and `updateTree` is very fuzzy - the former reads the objects, but the latter writes them? What gives? We should probably merge the two of them together, and factor out anything that would push the combined function over 30 lines. For example, we could load _both_ objects in one function, copy them both in another, and write them back in a third; the logic in `reconcile` to determine whether we should write back to the apiserver can probably also be factored out. Therefore, the revamped `reconcile` should probably look something like:

- load the singleton and namespace
- copy the singleton and namespace
- call `syncWithForest`
- early-exit if we shouldn't write back to the apiserver because something's being deleted
- write the singleton and namespace back to the apiserver
- call `updateObjects`

This should result in a far easier-to-understand and easier-to-extend flow.

### LINKED ISSUE #122: HNC: s/syncRequiredChildren/syncChildren/
The `syncRequiredChildren` function in `pkg/controller/hierarchy_controller.go` is a bit weird: it relies on someone calling `inst.Status.Children = ns.ChildNames()` even though it has both `inst` and `ns`.

We should change the method name to `syncChildren` and move the assignment mentioned above to be the first line of that function.

/good-first-issue

yiqigao217:
Fixes #109
Fixes #122

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -99,53 +99,26 @@ func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 }
 
 func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm string) error {
-	// Get the singleton and namespace
-	inst, err := r.getSingleton(ctx, nm)
+	// Get the singleton and namespace.
+	inst, nsInst, err := r.getInstances(ctx, log, nm)
 	if err != nil {
-		log.Error(err, "Couldn't read singleton")
 		return err
 	}
-	ns, err := r.getNamespace(ctx, nm)
-	if err != nil {
-		log.Error(err, "Couldn't read namespace")
-		return err
-	}
+
+	origHC := inst.DeepCopy()
+	origNS := nsInst.DeepCopy()
 
 	// If either object exists but is being deleted, we won't update them when we're finished syncing
 	// (we should sync our internal data structure anyway just in case something's changed).  I'm not
 	// sure if this is the right thing to do but the kubebuilder boilerplate included this case
 	// explicitly.
 	update := true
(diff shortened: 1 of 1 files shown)


adrianludwin:
This all lgtm

/assign @rjbez17 

@rjbez17 , I recommend reviewing the two commits in this PR individually: the first tries to minimize diffs while refactoring, the second moves around entire functions (but doesn't make other changes).

yiqigao217:
Rebased on upstream/master and resolved the merge conflict of the newly added lines of code in ```syncWithForest()```. I factored out that piece of code into ```syncLabel()``` function and move the func call below ```r.syncChildren()``` because otherwise ```r.syncChildren()``` could set new ```CRIT_``` conditions after ```syncLabel()```.

adrianludwin:
Still lgtm, nice merge.

rjbez17:
Sorry, started this yesterday and got side tracked.

/lgtm
/approve

PR #127, opened 2019-10-08
Add GOPATH/bin to PATH env var in ci-test.sh

danielSbastos:
Hey there delilah,

Here we set GOPATH/bin in PATH env var (because is not included or existing by default in Prow) in the `incubator/hnc/hack/ci-test.sh`. This is part of an ongoing discussion present on the issue https://github.com/kubernetes/test-infra/issues/9469.

### DIFF
--- a/incubator/hnc/hack/ci-test.sh
+++ b/incubator/hnc/hack/ci-test.sh
@@ -14,6 +14,17 @@
 
 set -e
 
+# Not included or existing by default in Prow
+#  - This is part of an ongoing discussion present on the issue:
+#    https://github.com/kubernetes/test-infra/issues/9469;
+#  - Other projects that use Prow as the CI, e.g. kubernetes-sigs/controller-runtime,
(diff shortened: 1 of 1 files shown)


incubator/hnc/hack/ci-test.sh
```diff
@@ -14,6 +14,14 @@
 
 set -e
 
+# Not included or existing by default in Prow
+#  - This is part of an ongoing discussion present on the issue:
+#    https://github.com/kubernetes/test-infra/issues/9469,
+#  - In Prow, the GOPATH is set to /home/prow/go, whereas in
+#    the Docker container is /go, which is the default one.
```
adrianludwin: Maybe add the fact that other K8s projects like `controller-runtime` do the same thing.

adrianludwin:
lgtm other than one minor point

/assign @Fei-Guo

incubator/hnc/hack/ci-test.sh
danielSbastos: Noice. Will do :)

PR #128, opened 2019-10-09
[incubator][vc] vn-agent support streaming api

zhuangqh:
## What this PR do
vn-agent support exec, attach, portForward proxy.

## How does the test work?
First, we generate the testing certification chain. Each certification is signed by a root CA.
Second, vn-agent will require the client providing the tenant cert, make the connection goes on HTTPS.
streaming server is inside the fakeKubelet, kubelet communicate with local streaming server through http.

image

### DIFF
--- a/incubator/virtualcluster/pkg/vn-agent/server/route.go
+++ b/incubator/virtualcluster/pkg/vn-agent/server/route.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 
 	"github.com/emicklei/go-restful"
+	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/apimachinery/pkg/util/proxy"
 	"k8s.io/klog"
 )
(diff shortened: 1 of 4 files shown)


incubator/virtualcluster/pkg/vn-agent/server/route.go
```diff
@@ -91,6 +140,33 @@ func (s *Server) proxy(req *restful.Request, resp *restful.Response) {
 	handler.ServeHTTP(resp.ResponseWriter, req.Request)
 }
 
+func (s *Server) proxyStream(req *restful.Request, resp *restful.Response) {
```
Fei-Guo: Is it possible to merge proxyStream and proxy method?  They are very similar except upgradeRequired flag. 
Will httpstream.IsUpgradeRequest(req.Request) always return false for non-stream APIs?

zhuangqh: good catch!

PR #133, opened 2019-10-10
Update test to verify specific condition codes

yiqigao217:


incubator/hnc/pkg/controllers/hierarchy_controller_test.go
```diff
@@ -88,16 +88,22 @@ var _ = Describe("Hierarchy", func() {
...
-		Eventually(func() []tenancy.Condition {
-			return getHierarchy(ctx, fooName).Status.Conditions
-		}).ShouldNot(BeNil())
+		Eventually(func() bool {
+			fooHier = getHierarchy(ctx, fooName)
+			for _, cond := range fooHier.Status.Conditions {
+				if cond.Code == tenancy.CritParentInvalid {
+					return true
+				}
+			}
```
adrianludwin: Should this be factored out into a function?

`Eventually(hasCondition(fooName, tenancy.CritParentInvalid)).Should(Equal(true))`

yiqigao217: I tried using ```hasCondition()``` and somehow the tests just failed (timed out).

Here's what I have. The ```geHierarchy()``` is called each time the ```hasCondition()``` is called. So the hier instance should be up to date eventually.
```
func hasCondition(ctx context.Context, nm string, code tenancy.Code) bool {
	for _, cond := range getHierarchy(ctx, nm).Status.Conditions {
		if cond.Code == code {
			return true
		}
	}
	return false
}
```

yiqigao217: The issue is resolved offline with the help of @adrianludwin.

It should return a func() instead of the value from a func call.

PR #134, opened 2019-10-10
Add CRIT_ANCESTOR condition to hier controller

yiqigao217:
a part of #76

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -154,6 +154,7 @@ func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Nam
 	}
 
 	// Clear locally-set conditions in the forest so we can set them to the latest.
+	hadCrit := ns.HasCritCondition()
 	ns.ClearConditions(forest.Local)
 
 	r.markExisting(log, ns)
@@ -161,15 +162,13 @@ func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Nam
 	r.syncRequiredChildOf(log, inst, ns)
 	r.syncParent(log, inst, ns)
 
-	// TODO add CRIT_ANCESTOR condition
-
 	// Update the list of actual children, then resolve it versus the list of required children.
 	r.syncChildren(log, inst, ns)
 
 	r.syncLabel(log, nsInst, ns)
 
-	// Convert and pass in-memory conditions to Hierarchy object.
-	inst.Status.Conditions = ns.Conditions(log)
+	// Sync all conditions. This should be placed at the end after all conditions are updated.
+	r.syncConditions(log, inst, ns, hadCrit)
 }
 
 func (r *HierarchyReconciler) onMissingNamespace(log logr.Logger, ns *forest.Namespace, nsInst *corev1.Namespace) bool {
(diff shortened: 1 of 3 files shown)


incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -154,22 +154,25 @@ func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Nam
...
 
 	r.syncLabel(log, nsInst, ns)
 
-	// Convert and pass in-memory conditions to Hierarchy object.
+	// Sync critical conditions after all locally-set conditions are updated.
+	r.syncCritConditions(log, ns, oldCritExists)
+
+	// Convert and pass in-memory conditions to HierarchyConfiguration object.
 	inst.Status.Conditions = ns.Conditions(log)
+	setCritAncestorCondition(log, inst, ns)
```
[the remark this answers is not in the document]
adrianludwin: Also, maybe add a comment that this should be the _last_ part of this function, because any earlier function could add a condition.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -341,6 +344,42 @@ func (r *HierarchyReconciler) syncChildren(log logr.Logger, inst *tenancy.Hierar
...
+	newCritExists := ns.HasCritCondition()
+
+	// Early exit if there's no need to enqueue relatives.
+	if oldCritExists == newCritExists {
+		return
+	}
+
+	msg := "found"
+	if oldCritExists == true {
+		msg = "gone"
```
adrianludwin: Nit: "added/removed" might be a bit easier to understand?

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -341,6 +344,42 @@ func (r *HierarchyReconciler) syncChildren(log logr.Logger, inst *tenancy.Hierar
...
+	}
+	log.Info("Critical conditions are " + msg)
+	r.enqueueAffected(log, "children of a namespace with critical conditions "+msg, ns.ChildNames()...)
```
adrianludwin: This message should be singular because it's printed once for every enqueued child: `"child of a namespace with ..."`

Also, should this be _children_ or _descendants_? If having `CRIT_ANCESTOR` doesn't qualify as being a critical condition, the children will get enqueued, will _not_ appear to have a critical condition themselves, and therefore will not enqueue _their_ children - right?

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -341,6 +344,42 @@ func (r *HierarchyReconciler) syncChildren(log logr.Logger, inst *tenancy.Hierar
...
+}
+
+func setCritAncestorCondition(log logr.Logger, inst *tenancy.HierarchyConfiguration, ns *forest.Namespace) {
+	ans := ns.Parent()
+	for ans != nil {
+		if ans.HasCritCondition() {
```
adrianludwin: Golang prefers early-exits, so better to say:

```
if !ans.HasCritCondition() {
   ans = ans.Parent()
   continue
}

File: incubator/hnc/pkg/forest/forest.go [hunk 5]
```diff
@@ -233,11 +233,25 @@ func (ns *Namespace) HasCondition() bool {
 	return len(ns.conditions) > 0
 }
 
+// HasCritCondition returns if the namespace has any critical condition.
+func (ns *Namespace) HasCritCondition() bool {
+	return ns.GetCondition(Local) != nil
```
adrianludwin: You should document this very clearly.

incubator/hnc/pkg/forest/forest.go
```diff
@@ -233,11 +233,25 @@ func (ns *Namespace) HasCondition() bool {
...
+
 // ClearConditions clears local conditions in the namespace.
 func (ns *Namespace) ClearConditions(key string) {
 	delete(ns.conditions, key)
 }
 
+// GetCondition gets a condition list from a key string. It returns nil, if the key doesn't exist.
+func (ns *Namespace) GetCondition(key string) []condition {
+	conditions, ok := ns.conditions[key]
+	if !ok {
```
adrianludwin: Do we need this? If `ok==false`, won't `conditions` be nil anyway? So you could say:

```
c, _ := ns.conditions[key]
return c
```

Note that very short variable names are actually preferred (https://github.com/golang/go/wiki/CodeReviewComments#variable-names).

yiqigao217: Good to know! Thanks

incubator/hnc/pkg/controllers/hierarchy_controller.go
yiqigao217: Right, here I should enqueue all descendants, not just children.

Replaced the log msg to "descendant" and func call to ```ns.DescendantNames()```

yiqigao217: Right, here I should enqueue descendants, not just children.

I replaced the log msg to say "descendant" and func call to ```ns.DescendantNames()```.

yiqigao217:
Fixed the merge conflict in the hierarchy_controller_test.go file (after the #133 is merged)

PR #142, opened 2019-10-11
extends capabilities and tests for finalizers

entro-pi:
Tested: via `make test`

Let me know if I totally borked up by making copyObject part of the reconciler interface.
It seemed like the best way to get access to the namespace in the same way that
other functions were doing it.

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -28,6 +28,7 @@ const (
 	CritAncestor              Code = "CRIT_ANCESTOR"
 	ObjectOverridden          Code = "OBJECT_OVERRIDDEN"
 	ObjectDescendantOverriden Code = "OBJECT_DESCENDANT_OVERRIDDEN"
+	ObjectMetadata            Code = "OBJECT_METADATA"
 )
 
 var (
@@ -84,7 +85,7 @@ type HierarchyConfigurationList struct {
 }
 
 // Code is a machine-readable string value summarizing the condition.
-// +kubebuilder:validation:Enum=CRIT_PARENT_MISSING;CRIT_PARENT_INVALID;CRIT_REQUIRED_CHILD_CONFLICT;CRIT_ANCESTOR;OBJECT_OVERRIDDEN;OBJECT_DESCENDANT_OVERRIDDEN
+// +kubebuilder:validation:Enum=CRIT_PARENT_MISSING;CRIT_PARENT_INVALID;CRIT_REQUIRED_CHILD_CONFLICT;CRIT_ANCESTOR;OBJECT_OVERRIDDEN;OBJECT_DESCENDANT_OVERRIDDEN;OBJECT_METADATA
 type Code string
 
 // Condition specifies the condition and the affected objects.
@@ -103,6 +104,9 @@ type Condition struct {
 	// - "OBJECT_OVERRIDDEN": an object in this namespace has been overridden from its parent and will no longer be updated
 	//
 	// - "OBJECT_DESCENDANT_OVERRIDDEN": an object in this namespace is no longer being propagated because a propagated copy has been modified
+	//
+	// - "OBJECT_METADATA": The object has metadata set that prevents it from being propagated to desencendant namespaces
+
 	Code Code   `json:"code,omitempty"`
 	Msg  string `json:"msg,omitempty"`
 
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -19,8 +19,10 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/go-logr/logr"
+	tenancy "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -104,6 +106,12 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
 // update deletes this object if it's an obsolete copy, and otherwise ensures it's been propagated
 // to any child namespaces.
 func (r *ObjectReconciler) update(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) error {
+	//Checks to see if the object is correct and should propagate
+	if !r.ensureCanPropagate(inst.GetNamespace(), inst) {
+		log.Info("Object cannot propagate")
+		return nil
+	}
+
 	// Make sure this object is correct and supposed to propagate. If not, delete it or mark it as
 	// modified. This should trigger one more reconciliation cycle and we'll get the children then.
 	if correct, err := r.ensureCorrectAncestry(ctx, log, inst); !correct || err != nil {
@@ -284,14 +292,38 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
 	return nil
 }
 
+func (r *ObjectReconciler) ensureCanPropagate(nsNm string, inst *unstructured.Unstructured) bool {
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+	ns := r.Forest.Get(nsNm)
+	if !ns.Exists() {
+		return false
+	}
+	//TODO TAYLOR
+	var conditionString string
+	if len(strings.Split(r.GVK.String(), ".")) > 1 {
+		affected := strings.Split(r.GVK.String(), ".")
+		conditionString = fmt.Sprint(affected[0], "/", affected[1], "/", affected[2], "/", nsNm, "/", ns.Name())
+	}
+
+	if inst.GetDeletionGracePeriodSeconds() != nil {
+		ns.SetCondition(conditionString, tenancy.ObjectMetadata, "Deletion grace period set")
+		return false
+	}
+	if inst.GetDeletionTimestamp() != nil {
+		ns.SetCondition(conditionString, tenancy.ObjectMetadata, "deletion timestamp set")
+		return false
+	}
+	if inst.GetFinalizers() != nil {
+		ns.SetCondition(conditionString, tenancy.ObjectMetadata, "finalizer is set")
+		return false
+	}
+	return true
+}
+
 func copyObject(inst *unstructured.Unstructured) *unstructured.Unstructured {
 	copied := inst.DeepCopy()
-
-	// Clear all irrelevant fields. cf https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta
 	copied.SetCreationTimestamp(metav1.Time{})
-	copied.SetDeletionGracePeriodSeconds(nil)
-	copied.SetDeletionTimestamp(nil)
-	copied.SetFinalizers(nil) // TODO: double-check this is the right thing to do?
 	copied.SetGenerateName("")
 	copied.SetGeneration(0)
 	copied.SetInitializers(nil) // TODO: is this correct?
--- a/incubator/hnc/pkg/controllers/object_controller_test.go
+++ b/incubator/hnc/pkg/controllers/object_controller_test.go
@@ -4,13 +4,12 @@ import (
 	"context"
 	"time"
 
+	tenancy "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-
-	tenancy "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
 var _ = Describe("Secret", func() {
@@ -100,6 +99,15 @@ var _ = Describe("Secret", func() {
 		Eventually(hasSecret(ctx, bazName, "foo-sec")).Should(BeFalse())
 	})
 
+	It("should not be propagated if it has a condition", func() {
+
+		setParent(ctx, barName, fooName)
+		setParent(ctx, quxName, barName)
+		if hasAnyCondition(ctx, barName) {
+			panic("CONDITION SET")
+		}
+	})
+
 	It("should be removed if the source no longer exists", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, barName)
@@ -111,6 +119,15 @@ var _ = Describe("Secret", func() {
 		Eventually(hasSecret(ctx, bazName, "foo-sec")).Should(BeFalse())
 	})
 })
+
+func hasAnyCondition(ctx context.Context, nm string) bool {
+	hier := newOrGetHierarchy(ctx, nm)
+	conds := hier.Status.Conditions
+	if len(conds) >= 1 {
+		return true
+	}
+	return false
+}
 
 func makeSecret(ctx context.Context, nsName, secretName string) {
 	sec := &corev1.Secret{}


entro-pi:
/assign @srampal 
And because we were talking about this, I'd like their input.
/assign @yiqigao217

rjbez17:
Is this still a WIP?

entro-pi:
I'd like to get input before I mark it done. So...ready for review?

entro-pi:
I'll mark that as ready for review in a second.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +285,25 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
 	return nil
 }
 
-func copyObject(inst *unstructured.Unstructured) *unstructured.Unstructured {
+func (r *ObjectReconciler) copyObject(nsNm string, inst *unstructured.Unstructured) *unstructured.Unstructured {
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+	ns := r.Forest.Get(nsNm)
+	if !ns.Exists() {
+		return nil
```
adrianludwin: If you look at the callers for copyObject, you'll see that a lot of them use the returned value immediately after it comes out of this function, so returning nil will result in a segfault. So for this solution to work, you'd have to modify all the call sites. That's a good "code smell" that this functionality doesn't belong here :)

entro-pi:
Thank you!

I'll hunt down where the decision process is made and make those changes
tonight.

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -28,6 +28,7 @@ const (
 	CritAncestor              Code = "CRIT_ANCESTOR"
 	ObjectOverridden          Code = "OBJECT_OVERRIDDEN"
 	ObjectDescendantOverriden Code = "OBJECT_DESCENDANT_OVERRIDDEN"
+	ObjectHasFinalizer        Code = "OBJECT_HAS_FINALIZER"
```
yiqigao217: We can probably rename the condition to ```ObjectInternal```, considering we are checking 3 fields for now ```DeletionGracePeriodSeconds```, ```DeletionTimestamp```,```Finalizers```. We may also check more object internal issues in the future in the ```ensureCanPropagate()``` function suggested by Adrian below.

entro-pi:
Cool! I was unsure of the wording too.

I'm on it!

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +285,25 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+	}
 	copied := inst.DeepCopy()
-
 	// Clear all irrelevant fields. cf https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta
 	copied.SetCreationTimestamp(metav1.Time{})
-	copied.SetDeletionGracePeriodSeconds(nil)
-	copied.SetDeletionTimestamp(nil)
-	copied.SetFinalizers(nil) // TODO: double-check this is the right thing to do?
+	if copied.GetDeletionGracePeriodSeconds() != nil {
+		ns.SetCondition(forest.Local, tenancy.ObjectHasFinalizer, "finalizer set")
```
yiqigao217: For ```OBJECT_``` conditions, the key should be a string in the format of ```group/version/kind/namespace/name```. ```forest.Local``` is the key for locally-set namespace conditions, not object conditions.

You can find the documentation at #76

entro-pi:
Excellent! Thank you!

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -103,6 +104,9 @@ type Condition struct {
 	// - "OBJECT_OVERRIDDEN": an object in this namespace has been overridden from its parent and will no longer be updated
 	//
 	// - "OBJECT_DESCENDANT_OVERRIDDEN": an object in this namespace is no longer being propagated because a propagated copy has been modified
+	//
+	// - "OBJECT_HAS_FINALIZER": The object has a finalizer type set, and should not be propagated.
```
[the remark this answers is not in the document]
entro-pi: Done

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -137,12 +138,12 @@ func (r *ObjectReconciler) ensureCorrectAncestry(ctx context.Context, log logr.L
 	}
 
 	// We found the object in an ancestor namespace. Is this object the same as the source?
-	instCopied := copyObject(inst)
+	instCopied := r.copyObject(src.GetNamespace(), inst)
```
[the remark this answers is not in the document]
entro-pi: Added the method mentioned and called it before the call to ensureCorrectAncestry, as well, undid the changes to the copyObject references.

incubator/hnc/pkg/controllers/object_controller.go
entro-pi: My nose is training!  Hahaha.

I should have listened to my gut when I started by adding the receiver.

Live and learn!

incubator/hnc/api/v1alpha1/hierarchy_types.go
entro-pi: Done!

incubator/hnc/pkg/controllers/object_controller.go
entro-pi: I THINK I referenced this correctly.

entro-pi:
Ok! Made all changes requested, passes make test.

Happy Friday!

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -103,6 +104,9 @@ type Condition struct {
 	// - "OBJECT_OVERRIDDEN": an object in this namespace has been overridden from its parent and will no longer be updated
 	//
 	// - "OBJECT_DESCENDANT_OVERRIDDEN": an object in this namespace is no longer being propagated because a propagated copy has been modified
+	//
+	// - "OBJECT_INTERNAL": The object has a finalizer set, and should not be propagated.
```
adrianludwin: @yiqigao217 , what's the rationale behind this name? I'm not sure what "internal" means. Maybe something like OBJECT_METADATA with a description "This object has metadata that prevents it from being propagated to descendant namespaces" and the `msg` field will identify the actual problems?

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -104,6 +105,12 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
 // update deletes this object if it's an obsolete copy, and otherwise ensures it's been propagated
 // to any child namespaces.
 func (r *ObjectReconciler) update(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) error {
+	//Checks to see if the object is correct and should propagate
+	if canPropagate := r.ensureCanPropagate(inst.GetNamespace(), inst); !canPropagate {
```
adrianludwin: No need to introduce a new variable here:

`if !r.ensureCanPropagate(inst.GetNamespace(), inst) {`

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -104,6 +105,12 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
+		log.Info("Object cannot propagate")
+		return fmt.Errorf("Object has finalizers set")
```
adrianludwin: Returning an error means "the reconciler failed and needs to be retried." That's not what's happened here (the reconciler successfully determined that the object couldn't be propagated) so this should be `return nil`.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +291,31 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
 
+func (r *ObjectReconciler) ensureCanPropagate(nsNm string, inst *unstructured.Unstructured) bool {
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+	ns := r.Forest.Get(nsNm)
+	if !ns.Exists() {
+		return false
+	}
+	if inst.GetDeletionGracePeriodSeconds() != nil {
+		ns.SetCondition(fmt.Sprint(r.GVK, ns, nsNm), tenancy.ObjectInternal, "finalizer set")
```
adrianludwin: A few things here:

- The key needs to be separated by slashes (`group/version/kind/namespace/name`), does your code do this? My reading of `fmt.Sprint` is that is separates the args using spaces.
- Not `"finalizer set"` - more like `"object is being deleted (deletionGracePeriodSeconds is set)"`

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +291,31 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
 }
 
+func (r *ObjectReconciler) ensureCanPropagate(nsNm string, inst *unstructured.Unstructured) bool {
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+	ns := r.Forest.Get(nsNm)
+	if !ns.Exists() {
+		return false
+	}
+	if inst.GetDeletionGracePeriodSeconds() != nil {
```
adrianludwin: I don't think this will compile - this doesn't return a pointer so it can't be compared to `nil`. See https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/pkg/controllers/hierarchy_controller.go#L116 for the correct way to check this.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +291,31 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+		ns.SetCondition(fmt.Sprint(r.GVK, ns, nsNm), tenancy.ObjectInternal, "finalizer set")
+		return false
+	}
+	if inst.GetDeletionTimestamp() != nil {
+		ns.SetCondition(fmt.Sprint(r.GVK, ns, nsNm), tenancy.ObjectInternal, "finalizer set")
```
adrianludwin: Ditto. Also, rather than cutting-and-pasting the complex `fmt.Sprint` line, can you factor that out somehow?

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +291,31 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+		return false
+	}
+	if inst.GetFinalizers() != nil {
+		ns.SetCondition(fmt.Sprint(r.GVK, ns, nsNm), tenancy.ObjectInternal, "finalizer set")
```
adrianludwin: In this case `"finalizer set"` is technically correct, but the message is going to a human so a longer message is probably friendlier (eg `"object has finalizers set and can't be propagated"`)

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +291,31 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+		return false
+	}
+	return true
+}
+
 func copyObject(inst *unstructured.Unstructured) *unstructured.Unstructured {
 	copied := inst.DeepCopy()
-
-	// Clear all irrelevant fields. cf https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta
```
adrianludwin: Why delete this comment?

entro-pi:
Fyewf! That'll keep me busy for a while. Thank you for your input and
guidance!

(The comment deletion was accidental and I missed it in review.)

incubator/hnc/pkg/controllers/object_controller.go
entro-pi: I KNEW something was off when I had to learn a new way of sending an error!

I think I'm starting to understand code smell, now I just need to spot it more.

incubator/hnc/pkg/controllers/object_controller.go
entro-pi: It does not. Again, should have asked when I thought of it.

incubator/hnc/pkg/controllers/object_controller.go
entro-pi: Factor that out meaning declare it as a local variable and use that as opposed to calling it all on one line and making it hard to read? Sure!

incubator/hnc/pkg/controllers/object_controller.go
entro-pi: Accidental, was writing lines around it and missed it in review.

incubator/hnc/pkg/controllers/object_controller.go
adrianludwin: Yup, that works!

incubator/hnc/pkg/controllers/object_controller.go
adrianludwin: thx

entro-pi:
Having a bit of trouble with the unit test, reading documentation. Will report more as I finish this today.

incubator/hnc/pkg/controllers/object_controller_test.go
```diff
@@ -4,13 +4,12 @@ import (
...
 	"time"
 
+	tenancy "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-
-	tenancy "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
```
adrianludwin: Can you restore this please? Usually we have (up to) three blocks of imports:

- Core libraries (eg `context`)
- Packages from other repos (eg `k8s.io/...`)
- Other packages from this repo (eg `v1alpha1`)

incubator/hnc/pkg/controllers/object_controller_test.go
```diff
@@ -100,6 +99,15 @@ var _ = Describe("Secret", func() {
 		Eventually(hasSecret(ctx, bazName, "foo-sec")).Should(BeFalse())
 	})
 
+	It("should not be propagated if it has a condition", func() {
+
+		setParent(ctx, barName, fooName)
+		setParent(ctx, quxName, barName)
+		if hasAnyCondition(ctx, barName) {
+			panic("CONDITION SET")
+		}
```
adrianludwin: Based on https://onsi.github.io/gomega/#making-assertions, I'd say something like:

`Expect(hasAnyCondition(ctx, barName)).ToNot(Equal(true))`

You shouldn't panic since it stops the rest of the tests from running.

Usually, you'll use `Eventually()` instead of `Expect()` since the latter runs immediately while the former can retry for up to two seconds (configurable in `suite_test.go` IIRC, but please don't change it). This is necessary because the controller is running asynchronously alongside the test code, so when you make a change (like updating the test apiserver), it takes a few moments to respond.

entro-pi:
Nice! Thank you. I figured Eventually had something to do with asynchronous
calls.

I'll restore the import and rewrite this test.

Do you have any ideas on testing for propagation as well? Maybe set a
condition, then a parent and then grab the object and test for a condition?
The step I'm having the most trouble with is grabbing the correct namespace
so I can test for a condition.

adrianludwin:
It should be relatively easy to set a condition in a test - just set the parent of a namespace to something that doesn't exist. There should be lots of examples in hierarchy_controller_test.go.

I'm not sure what you mean by "grabbing the correct namespace." Perhaps have a live chat or video meeting with @yiqigao217 and then pull me in if you're still having trouble?

entro-pi:
I think I've got it figured out. I'll get back at it tomorrow after work.

Cheers,
Thanks for all your help!

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: SGTM. Yes, metadata is better.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +292,38 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+	defer r.Forest.Unlock()
+	ns := r.Forest.Get(nsNm)
+	if !ns.Exists() {
+		return false
+	}
+	//TODO TAYLOR
+	var conditionString string
+	if len(strings.Split(r.GVK.String(), ".")) > 1 {
+		affected := strings.Split(r.GVK.String(), ".")
+		conditionString = fmt.Sprint(affected[0], "/", affected[1], "/", affected[2], "/", nsNm, "/", ns.Name())
```
yiqigao217: The last piece in the string should be ```inst.GetName()``` since it's the name of the object not namespace name. ```ns.Name()``` is the same as ```nsNm``` here.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +292,38 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
 
+func (r *ObjectReconciler) ensureCanPropagate(nsNm string, inst *unstructured.Unstructured) bool {
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+	ns := r.Forest.Get(nsNm)
+	if !ns.Exists() {
+		return false
+	}
+	//TODO TAYLOR
+	var conditionString string
```
adrianludwin: `String` isn't a very descriptive suffix for a variable of type `string`. Perhaps `conditionKey` is more along the lines of what you meant?

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +292,38 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+	if len(strings.Split(r.GVK.String(), ".")) > 1 {
```
adrianludwin: I don't understand this line? There will _always_ be a GVK for _every_ object. However, for legacy reason, the "core" objects (like `Pod` or `Secret`) have an empty string as the group. Eg, a pod "bar" in namespace "foo" should have the key `/v1/pod/foo/bar`, while a replica set "rs" in the same namespace would have the key `apps/v1/pod/foo/rs`.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -284,14 +292,38 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
+		affected := strings.Split(r.GVK.String(), ".")
```
adrianludwin: Instead of splitting and combining, why not just use the components directly?

yiqigao217:
> Do you have any ideas on testing for propagation as well? Maybe set a condition, then a parent and then grab the object and test for a condition? The step I'm having the most trouble with is grabbing the correct namespace so I can test for a condition.

I don't think you need to set a condition to test the propagation. The ```OBJECT_METADATA``` condition should be set by your code in object controller if anything's wrong with the object metadata. You can just test in 2 steps:
* Set finalizer (or deletion meta) to a src object and see if ```OBJECT_METADATA``` condition is set.
* Set a child namespace to the src object namespace and see if the src object is propagated to this child namespace.

entro-pi:
Thank you!

That actually answers all of my hanging questions.

I'll update tonight with more details and my rational.

entro-pi:
Ooh, I was confused about that for sure. I'll change the calls as it isn't
necessary to pass the namespace if it's passed with the instance.

entro-pi:
I will continue work on this tomorrow night. Had some stuff come up.

adrianludwin:
Closing as @entro-pi told me they aren't ready to work on this for the time being.

PR #147, opened 2019-10-12
HNC: hardcoded value for resource is replaced by constant

### LINKED ISSUE #146: HNC: create constants for API types
Issue #144 exposed a bug there the string `"hierarchies"` was hardcoded, but should have been changed to `"hierarchyconfigurations"` as a part of #98 . PR #145 fixes the immediate problem by changing the hardcoded string, but I think that in addition to the `tenancy.Singleton` const that defines the singleton _name_, we should also have a `tenancy.Resource` const with the correct string, rather than hardcoding it in several places.

/assign @shivi28

shivi28:
**What this PR does / why we need it**:
This PR added a constant for resource to replace hardcoded value of resource `hierarchyconfigurations`

**Which issue(s) this PR fixes**:
Fixes #146

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -32,6 +32,7 @@ const (
 
 var (
 	Singleton = "hierarchy"
+	Resource  = "hierarchyconfigurations"
 )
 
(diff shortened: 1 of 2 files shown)


adrianludwin:
Simple change so I'll just approve
/lgtm
/approved

PR #148, opened 2019-10-13
kubectl hnc tree shows all namespaces

### LINKED ISSUE #130: HNC: allow `kubectl hnc tree` to show _all_ namespaces if no argument is given
(Filed as a result of #93)

Currently `kubectl hnc tree` requires an argument, which is the root namespace of the tree to display. We should also allow the user to omit the root argument and print _every_ namespace (and their subnamespaces) instead.

/good-first-issue

shivi28:
This PR adds a feature that allow `kubectl hnc tree` to show all namespaces if no argument is given. This feature enables users to list all namespaces along with their heirarchies. 

Which issue(s) this PR fixes:
Fixes #130


incubator/hnc/pkg/kubectl/root.go
```diff
@@ -90,6 +90,18 @@ func homeDir() string {
 	}
 	return os.Getenv("USERPROFILE") // windows
 }
+func getAllNamespaces() []string {
```
adrianludwin: IMO let's put this in tree.go until it's needed somewhere else?

incubator/hnc/pkg/kubectl/root.go
```diff
@@ -90,6 +90,18 @@ func homeDir() string {
...
+	nsList, err := k8sClient.CoreV1().Namespaces().List(metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("Error reading namespaces List %s\n", err)
```
adrianludwin: Suggested: `"Could not list namespaces: %s\n"`

incubator/hnc/pkg/kubectl/root.go
```diff
@@ -90,6 +90,18 @@ func homeDir() string {
...
+		os.Exit(1)
+	}
+	result := make([]string, 0)
```
adrianludwin: Preferred style is usually `result := []string{}`

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,19 +34,32 @@ var (
...
 	Use:   "tree",
 	Short: "Displays the hierarchy tree rooted at the given namespace",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		nnm := args[0]
+		var nnm string
+		if len(args) != 0 {
+			nnm = args[0]
+		}
```
adrianludwin: You don't really need this - below, where you would have said `if nnm != ""`, just say `if len(args) != 0` instead

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,19 +34,32 @@ var (
...
-			for i, n := range footnotes {
-				fmt.Printf("%d) %s\n", i+1, n)
+		nsList := make([]string, 0)
+		if nnm != "" {
+			nsList = append(nsList, nnm)
+		} else {
+			allNamespaces := getAllNamespaces()
+			nsList = append(nsList, allNamespaces...)
+		}
+		for _, each := range nsList {
```
adrianludwin: Nit: instead of `each`, maybe call this `nnm` ("Namespace NaMe"), especially if you get rid of the existing `nnm` variable, above. The variable `each` makes sense on this line, but `getHierarchy(each)` doesn't read as well (to me) as `getHierarchy(nnm)`.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,19 +34,32 @@ var (
...
+			hier := getHierarchy(each)
+			if (hier.Spec.Parent == "" && nnm == "") || nnm != "" {
+				fmt.Println(nameAndFootnotes(hier))
+				printSubtree("", hier)
+				if len(footnotes) > 0 {
```
adrianludwin: These should be printed out after the entire loop is complete, IMO.

incubator/hnc/pkg/kubectl/root.go
shivi28: Actually I am putting this function inside root.go because this function uses a kubernetes client that is already created in root.go file.

shivi28:
**Input 1** :  ```kubectl hnc tree```
Output : 
```acme-org
└── teama
cert-manager
default
hnc-system
kube-node-lease
kube-public
kube-system
teamb
```

**Input 2** : ```kubectl hnc tree acme-org```
output : 
```acme-org
└── teama
```

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,15 +36,24 @@ var (
 var treeCmd = &cobra.Command{
 	Use:   "tree",
 	Short: "Displays the hierarchy tree rooted at the given namespace",
```
adrianludwin: Can you please add some help text here? E.g., "If no namespace is provided, displays the tree starting at every root namespace on the cluster."

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,15 +36,24 @@ var (
...
 	Run: func(cmd *cobra.Command, args []string) {
-		nnm := args[0]
 		footnotesByMsg = map[string]int{}
 		footnotes = []string{}
-		hier := getHierarchy(nnm)
-		fmt.Println(nameAndFootnotes(hier))
-		printSubtree("", hier)
-
+		nsList := make([]string, 0)
+		if len(args) != 0 {
```
adrianludwin: Now I'm wondering if we should allow _multiple_ namespaces to be provided by the user? Eg, I could say `kubectl hnc tree foo bar baz` and it would show me the trees rooted at each of these namespaces. There could be some duplication but I think that's fine. Would you please make this change (and update the help text)?

adrianludwin: You use the test `len(args) == 0` several times; you should probably factor this out into a variable to make it more obvious what you're doing. For example:

```
defaultList := len(args) == 0
if defaultList {
  ...
}
```

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,15 +36,24 @@ var (
...
+			nsList = append(nsList, args[0])
+		} else {
+			allNamespaces := getAllNamespaces()
+			nsList = append(nsList, allNamespaces...)
+		}
+		for _, nnm := range nsList {
+			hier := getHierarchy(nnm)
+			if (hier.Spec.Parent == "" && len(args) == 0) || len(args) != 0 {
```
adrianludwin: In Go, it's usually considered better style to early exit. If you use the `defaultList` suggestion above, this then becomes:

```
if defaultList && hier.Spec.Parent != "" {
  // By default, skip non-root namespaces since they'll be displayed as part of
  // another namespace's tree.
  continue
}
```

which is a much easier flow of logic to follow, IMO.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -34,15 +36,24 @@ var (
...
+	Args:  cobra.MaximumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		nnm := args[0]
 		footnotesByMsg = map[string]int{}
 		footnotes = []string{}
-		hier := getHierarchy(nnm)
-		fmt.Println(nameAndFootnotes(hier))
-		printSubtree("", hier)
-
+		nsList := make([]string, 0)
```
adrianludwin: I think you could just say: `var nsList []string`, and then directly assign it below. Eg, `nsList = args` if any args are provided, or `nsList = getAllNamespaces()` if they're not. No need to append or manipulate slices since in all cases, you're just taking the entire list.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -93,3 +104,16 @@ func nameAndFootnotes(hier *tenancy.HierarchyConfiguration) string {
...
+	nsList, err := k8sClient.CoreV1().Namespaces().List(metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("Could not list namespaces: %s\n", err)
+		os.Exit(1)
+	}
+	result := []string{}
+	for _, each := range nsList.Items {
+		result = append(result, each.Name)
+	}
+	return result
```
adrianludwin: Perhaps sort them before returning?

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -35,27 +35,28 @@ var (
...
-			nsList = append(nsList, allNamespaces...)
+			nsList = args
 		}
 		for _, nnm := range nsList {
 			hier := getHierarchy(nnm)
-			if (hier.Spec.Parent == "" && len(args) == 0) || len(args) != 0 {
-				fmt.Println(nameAndFootnotes(hier))
-				printSubtree("", hier)
+			if defaultList && hier.Spec.Parent != "" {
+				continue
```
adrianludwin: Maybe a comment here is worthwhile: "If we're showing the default list, skip all non-root namespaces since they'll be displayed as part of another namespace's tree."

adrianludwin:
/assign @rjbez17 
Hey Ryan, can you ptal and remove the hold if you're happy?

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -33,18 +35,29 @@ var (
...
+		defaultList := len(args) == 0
 		footnotesByMsg = map[string]int{}
 		footnotes = []string{}
-		hier := getHierarchy(nnm)
-		fmt.Println(nameAndFootnotes(hier))
-		printSubtree("", hier)
-
+		var nsList []string
+		if defaultList {
+			nsList = getAllNamespaces()
```
rjbez17: This seems different from most other k8s CLIs where if you don't provice a default one is chosen for you and you need to specify `--all-namespaces` to get them all. This is to prevent unnecessary strain on the api server (I suspect). 

@adrianludwin thoughts?

adrianludwin: So you suggestion would be:

```
kubectl hnc tree foo # displays foo
kubectl hnc tree # error
kubectl hnc tree --all-namespaces # displays the tree of all namespaces
```

Is that correct?

adrianludwin: That seems pretty reasonable to me. @shivi28 can you please make this change?  See [here](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/pkg/kubectl/set.go#L77) for an example of how to add flags.

adrianludwin: And see https://github.com/spf13/cobra#persistent-flags to show how to add short aliases (eg `-A` as a synonym for `--all-namespaces`)

shivi28: yaa @rjbez17 @adrianludwin I will do it

adrianludwin:
Checking to see if our new Prow config is working:

/retest

adrianludwin:
Checking to see if our new Prow config is working:

/test all

adrianludwin:
Replaced by #180

PR #149, opened 2019-10-13
Add better not-found handling to `kubectl hnc tree`

adrianludwin:


adrianludwin:
Whoops, noticed that this depends on a change in #150 . Will close and refactor.

Issue #131, opened 2019-10-10
HNC: Clean up kubectl plugin hierarchy modification functions


adrianludwin:
(Filed as a part of #93; may supplant #108)

We currently have three commands to change the structure of the hierarchy: `set-parent`, `unset-parent` and `subnamespace`. These commands don't work consistently and aren't grouped well.

As a result, I think we should have a single `set <ns>` command with multiple flags:

- `--parent <parent>`: sets the parent
- `--root`: unsets the parent (i.e., turns this into a root namespace)
- `--requiredChild <child>`: specifies that `<child>` is a required child
- `--unrequiredChild <child>`: **FUTURE** removes the currently unrequired child

I think this is more clear. For example, let's say we have a namespace called `middle` and we wanted to set its parent `top` and its required child `bottom`. Today, this would be expressed as:

```
kubectl hnc set-parent mid top
kubectl hnc subnamespace mid bottom
```

It's not clear, in this setting, what's the meaning of the first or second parameters without looking at the documentation. Instead, the new format will be:

```
kubectl hnc set mid --parent top
kubectl hnc set mid --requiredChild bottom
```
or even:

```
kubectl hnc set mid --parent top --requiredChild bottom --requiredChild otherChild
```

This makes it more clear (I think) which namespace is being acted on and what the resulting relationship will look like. As this example shows, you should be able to set multiple properties in one line as long as they don't conflict (ie `--parent` and `--root`).

Finally, where there's a `set`, there should probably be a `get` as well. Let's change `kubectl hnc show` to `kubectl hnc get` instead.

adrianludwin:
@vjdhama looks like you're not a member of `kubernetes-sigs` yet, can you please add yourself as per https://github.com/kubernetes-sigs/multi-tenancy#join-this-repo? tashimi and I can sponsor you, and it's ok not to have contributions _yet_, just say you'll be working on HNC and reference this issue.

vjdhama:
Thanks. I raised membership request. 🤞

PR #152, opened 2019-10-14
Add set command

vjdhama:
Adds support for the unified `set` command to enable setting parent namespace of other namespace and root namespace.

Which issue(s) this PR fixes:
Fixes: #131


incubator/hnc/pkg/kubectl/set.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+var setCmd = &cobra.Command{
+	Use:     "set",
+	Short:   "Allows setting namespace hierarchy",
```
adrianludwin: Suggestion: "Sets properties of the hierarchy configuration."

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -0,0 +1,82 @@
...
+	Aliases: []string{"set"},
+	Args:    cobra.MinimumNArgs(1),
```
adrianludwin: Shouldn't this be "exact args?"

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -0,0 +1,82 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		nnm := args[0]
+		hier := getHierarchy(nnm)
+		oldpnm := hier.Spec.Parent
+
+		if cmd.Flags().Changed("root") {
```
adrianludwin: What does "Changed" mean? Is it the same as "Set"?

Also we need to make sure that `--root` and `--parent` aren't set at the same time.

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -0,0 +1,82 @@
...
+			if oldpnm == "" {
+				fmt.Printf("%s is already a root namespace; unchanged", nnm)
+				return
```
adrianludwin: This will break if you try to set multiple things at the same time (eg `--root --requiredChild foo`). I think if you factor handling each flag out into its own function, this will work better.

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+			hier.Spec.Parent = ""
+			updateHierarchy(hier, fmt.Sprintf("unsetting the parent of %s", nnm))
+			fmt.Printf("Unset the parent of %s (was previously %s)\n", nnm, oldpnm)
+			return
```
adrianludwin: This is inconsistent with how `--parent` works

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -0,0 +1,82 @@
...
+			rcns, _ := cmd.Flags().GetStringArray("requiredChild")
+
+			for _, rcn := range rcns {
+				ns := getHierarchy(nnm)
+				if childNamespaceExists(ns, rcn) {
+					fmt.Printf("Required child %s already present in %s\n", rcn, nnm)
+					continue
+				}
+				ns.Spec.RequiredChildren = append(ns.Spec.RequiredChildren, rcn)
+				updateHierarchy(ns, fmt.Sprintf("adding required child %s", rcn))
```
adrianludwin: Should only call `updateHierarchy` once at the end of the loop IMO. Actually, should only call it once after _all_ flags (`--root` or `--parent`; `--requiredChild`) have been applied.

incubator/hnc/pkg/kubectl/set.go
vjdhama: Yeah, actually that's correct. I'll change it.

PR #153, opened 2019-10-15
implement the functionality for constructing the PKI for virtual cluster

charleszheng44:
1. implement the functionality for constructing the PKI
2. create sample yaml for the ClusterVersion

### DIFF
--- a/incubator/virtualcluster/config/rbac/auth_proxy_service.yaml
+++ b/incubator/virtualcluster/config/rbac/auth_proxy_service.yaml
@@ -9,7 +9,7 @@ metadata:
     control-plane: controller-manager
     controller-tools.k8s.io: "1.0"
   name: controller-manager-metrics-service
-  namespace: system
+  namespace: kube-system
 spec:
   ports:
   - name: https
--- a/incubator/virtualcluster/pkg/controller/add_clusterversion.go
+++ b/incubator/virtualcluster/pkg/controller/add_clusterversion.go
@@ -16,11 +16,11 @@ limitations under the License.
 
 package controller
 
-// import (
-// 	"github.com/multi-tenancy/incubator/virtualcluster/pkg/controller/clusterversion"
-// )
-//
-// func init() {
-// 	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
-// 	AddToManagerFuncs = append(AddToManagerFuncs, clusterversion.Add)
-// }
+import (
+	"github.com/multi-tenancy/incubator/virtualcluster/pkg/controller/clusterversion"
+)
+
+func init() {
+	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
+	AddToManagerFuncs = append(AddToManagerFuncs, clusterversion.Add)
+}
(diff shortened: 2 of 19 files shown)


incubator/virtualcluster/config/rbac/auth_proxy_service.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     control-plane: controller-manager
     controller-tools.k8s.io: "1.0"
   name: controller-manager-metrics-service
-  namespace: system
+  namespace: kube-system
```
Fei-Guo: change to virtualcluster-system

incubator/virtualcluster/pkg/controller/add_clusterversion.go
```diff
@@ -16,11 +16,11 @@ limitations under the License.
...
 
-import (
-	"github.com/multi-tenancy/incubator/virtualcluster/pkg/controller/clusterversion"
-)
-
-func init() {
-	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
-	AddToManagerFuncs = append(AddToManagerFuncs, clusterversion.Add)
-}
+// import (
```
Fei-Guo: why do you comment out these?

incubator/virtualcluster/pkg/controller/virtualcluster/virtualcluster_controller.go
```diff
@@ -68,13 +111,266 @@ var _ reconcile.Reconciler = &ReconcileVirtualcluster{}
 // ReconcileVirtualcluster reconciles a Virtualcluster object
 type ReconcileVirtualcluster struct {
 	client.Client
-	scheme *runtime.Scheme
+	scheme          *runtime.Scheme
+	clusterVersions map[string]*tenancyv1alpha1.ClusterVersion
+	virtualClusters map[string]*tenancyv1alpha1.Virtualcluster
+}
+
+func (r *ReconcileVirtualcluster) checkCRKind(request *reconcile.Request) (runtime.Object, CRKind, error) {
```
Fei-Guo: You cannot prevent vc and clusterversion from having same name which breaks this method. I suggest to use another reconcile loop to reconcile for clusterversion.

charleszheng44: when a new vc cr is created, we have to refer to existing cluster versions. This cannot be achieved by using two separate controllers. In other words, without a client set, we can’t query apiserver when we want to get all available clusterversions.

incubator/virtualcluster/pkg/controller/add_clusterversion.go
charleszheng44: I have uncommented these lines

incubator/virtualcluster/config/rbac/auth_proxy_service.yaml
charleszheng44: I think we should keep it this way. This service is for meta cluster.

Fei-Guo: ok

PR #155, opened 2019-10-16
Unbreak webhook config

adrianludwin:


adrianludwin:
Self-approving as this is pretty straightforward.

/lgtm

adrianludwin:
Or not, as the case may be :)

/assign @tashimi

PR #156, opened 2019-10-16
Check authz when changing a namespace's parent

adrianludwin:
This implements the logic described in the design doc to prevent a
subadmin from depriving a superadmin of access to a namespace by
assigning it away, and from gaining access to a new namespace's
propagated objects by joining it as a child.

Tested: new unit tests. Also manually verified that by default, acting
as a service account did not allow me to change the hierarchy of a
subnamespace even if I had get/update permissions on its
hierarchyconfig. Verified that as I added rolebindings to the parents,
the error messages changed accordingly as the reasons for failures
changed until I had sufficient permissions, at which point it passed. I
have not tested this exhaustively, apart from the unit tests.


adrianludwin:
/hold

I'll rebase and remove the hold when #155 has been checked in.

adrianludwin:
/hold cancel
#155 has been merged

adrianludwin:
/assign @Fei-Guo 

Sanjeev says he can't look at this one soon. Fei, can you PTAL? Or are you swamped too?

PR #160, opened 2019-10-17
Restructure update() to consolidate calls with forest lock

yiqigao217:
This is the 1st step of https://github.com/kubernetes-sigs/multi-tenancy/issues/76#issuecomment-541188994

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -17,7 +17,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"reflect"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


adrianludwin:
/assign @Fei-Guo 

I'll also give this a look though

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -104,68 +103,93 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
-// it's correct and ready to be propagated. The incorrect object will be deleted
-// if it's obsolete; or marked as modified if it's changed.
-func (r *ObjectReconciler) ensureCorrectAncestry(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) (bool, error) {
-	// Find what we expect the source to be.
-	missing, src, err := r.getSourceInst(ctx, log, inst)
-	if err != nil {
-		return false, err
+// syncWithForest syncs the object instance with the in-memory forest. This is the only place we
+// should lock the forest, so this fn needs to return everything relevant for the rest of the
+// reconciler. It returns true if the object instance is good and shouldn't be deleted. It will
```
adrianludwin: We should also add that this method doesn't (or more accurately, shouldn't) contact the apiserver since that's a slow operation and everything will block on the lock being held.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -104,68 +103,93 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
+// also return a list of destinations for propagation.
+func (r *ObjectReconciler) syncWithForest(ctx context.Context, log logr.Logger, inst, srcInst *unstructured.Unstructured) ([]string, bool) {
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+
+	if !r.hasCorrectAncestry(ctx, log, inst, srcInst) {
```
adrianludwin: Maybe: `// If srcInst isn't an ancestor of inst anymore, it should be deleted`

adrianludwin: In general, I'd add a top-level comment for each of the three stages in this function. The other two might look something like:

```
// If the object has been modified, alert the users and don't propagate it further
...
// The object looks good; it should be propagated to its namespaces' children
```

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -104,68 +103,93 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
+	if !curNS.IsAncestor(srcNS) {
+		log.Info("Will delete because the source namespace is no longer an ancestor", "inheritedFrom", srcNS)
+		return false
+	}
+
+	return true
+}
+
+// isUnmodified checks to see if this object is the same as the source. It returns true if it's
+// correct and ready to be propagated.
```
adrianludwin: I don't think you need the second sentence anymore. The _caller_ decides that this is ready to be propagated, not this function - this function literally just decides whether it's modified or not.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -104,68 +103,93 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
-	return equal, nil
+	return reflect.DeepEqual(instCopied, srcCopied)
 }
 
-// getSourceInst gets source instance of this one. It returns nil if this instance isn't a copy.
-// It returns true if the source instance is missing. It could be that the source instance no
-// longer exists or the source namespace is no longer an ancestor.
-func (r *ObjectReconciler) getSourceInst(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) (bool, *unstructured.Unstructured, error) {
+// getSourceInst gets source instance of this one. It returns itself if it's the source. It returns
+// nil if the source instance no longer exists in the source namespace or there's an error.
```
adrianludwin: I think this method needs to return an error - we want to return an error when we want the reconciler to try again, and "I temporarily couldn't access the apiserver" is probably a very good example of that.

If there's a temporary interruption accessing the apiserver and we delete this object, we'll never revisit it!

So this should return (*unstructured, error), if if the caller gets an error, it should pass that on to the top level.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -174,27 +198,13 @@ func (r *ObjectReconciler) getSourceInst(ctx context.Context, log logr.Logger, i
 	src.SetGroupVersionKind(inst.GroupVersionKind())
 	err := r.Get(ctx, srcNm, src)
 	if err != nil {
-		if errors.IsNotFound(err) {
-			log.Info("The source object no longer exists", "inheritedFrom", srcNS)
-			return true, nil, nil
-		} else {
+		if !errors.IsNotFound(err) {
 			log.Info("Couldn't read", "source", srcNm)
```
adrianludwin: Use `log.Error(err, "Couldn't read", "source", srcNm)`. Right now, the error is lost completely.

yiqigao217: I will keep this log as info since I changed the func back to return error.

incubator/hnc/pkg/controllers/object_controller.go
yiqigao217: Done. Changed it back to return error when it's not `IsNotFound` error.

PR #161, opened 2019-10-17
Separate intermediate files from Git files

adrianludwin:
This change removes the need to run `sed` on kustomization files that
get checked into Git. Instead, it creates a new, temporary
kustomization.yaml file in a directory ignored by Git (`/manifests`),
using standard `kustomize edit` features. It also builds the resulting
manifests into this directory, which can be used in the future to build
releasable artifacts.

In addition, it removes the `kind-*` scripts and turns them into `make`
targets instead (i.e., `make kind-reset`, `make kind-deploy` and `make
kind-watch`). Since these scripts were calling into the makefile anyway,
putting everything in one place is a little easier to follow, for now at
least.

Tested: ran `make deploy` and `make manifests` with and without
`CONFIG=kind`. Verified that `make kind-deploy` was a synonym for
`CONFIG=kind make deploy`. Also tried
`IMG=gcr.io/hnc-x-k8s-io/controller:v0.1` (a potential released image
name, not one that's actually set up yet!). Manually inspected the
resulting files in `/manifests` and verified that they contained the
expected results in all cases.


adrianludwin:
/assign @rjbez17 

wdyt about this?

PR #164, opened 2019-10-18
Make HNC easily deployable onto non-Kind clusters

adrianludwin:
/assign @rjbez17 
/assign @yiqigao217 

Also fixes a bug that `deploy` doesn't delete the old deployment first.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -19,29 +19,27 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-.PHONY: all test manager kubectl run deploy manifests fmt vet generate docker-build docker-push controller-gen kind-reset kind-deploy kind-watch
+all: test docker-build
 
-all: manager
+###################### LOCAL ARTIFACTS #########################
 
 # Run tests
-test: manager kubectl manifests
+test: build
 	go test ./api/... ./cmd/... ./pkg/... -coverprofile cover.out
 
-# Build manager binary
-manager: generate fmt vet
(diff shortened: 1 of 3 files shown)


incubator/hnc/Makefile
```diff
@@ -91,22 +76,50 @@ else
...
-# Creates a local kind cluster, destroying the old one if necessary. It's not
-# *necessary* to call this wih CONFIG=kind but it's not a bad idea either so
-# the correct manifests get created.
-kind-reset: manifests
-	@echo "Warning: you must have sourced 'devenv' so kubectl points to the Kind cluster"
+###################### DEPLOYABLE ARTIFACTS AND ACTIONS #########################
+
+# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: deploy-prereq docker-push kubectl manifests
+	-kubectl -n hnc-system delete deployment hnc-controller-manager
```
rjbez17: Any reason not to use `kubectl delete -f manifests/hnc-manager.yaml` here?

adrianludwin: Good call, fixed.

adrianludwin: Actually, this deletes the CRDs so any existing hierarchy configs are wiped out too. Oops.

adrianludwin: I'm reverting this back to the way it was.

adrianludwin: I updated the comment now, lgty?

rjbez17: 👍

PR #165, opened 2019-10-18
Remove ObjectReconciler on PodSecurityPolicy kind

yiqigao217:
Tested: make test & manual test on Kind.

Edit (@adrianludwin): PodSecurityPolicy isn't namespaced and shouldn't be part of the HNC


adrianludwin:
This change has a few issues; since @yiqigao217 is away for the weekend and I have an alternative commit, I'll just submit that instead.

PR #167, opened 2019-10-19
Implement the function for creating HA etcd services

charleszheng44:
Allow users to specify the number of etcd servers by changing the value of `replicas` field of the etcd StatefulSet, for example,  if `replicas: 3`, a 3 nodes etcd service will be created.


incubator/virtualcluster/config/sampleswithspec/clusterversion_v1.yaml
```diff
@@ -9,64 +9,67 @@ spec:
...
     metadata:
       name: etcd 
-      namespace: vc-test
     statefulset:
       metadata:
-        name: vc-test/statefulset/etcd
+        name: etcd
       spec:
-        replicas: 1
+        replicas: 3
```
rjbez17: This will definitely cause you issues in production. If you have an issue with the first or second replica you have no way to remove/re-add a corrupted member easily. But as this is a sample I'm sure it's ok. Just something I've ran into in the past!

charleszheng44: Won't statefulset controller restart fail pod?

charleszheng44: Did you mean it will be difficult to control each individual pod (eg. reschedule failed pod to another node) of the statefulset?

charleszheng44:
Will use one etcd server in the first draft-version for testing

PR #169, opened 2019-10-20
Improvements to generation and display of conditions

adrianludwin:
Changes:

- Better human-readable message for critical conditions in an ancestor.
- In `kubectl-hnc show`, pretty-prints all information from the
conditions, including the code and affected namespaces.
- In `kubectl-hnc tree`, the condition codes are now shown.
- Verbose log messages removed.

reported correctly.


adrianludwin:
/cc @rjbez17 

Sorry can you please re-lgtm this? I had a merge conflict

PR #171, opened 2019-10-20
Better comparison of propagated objects

adrianludwin:
Supposedly-identical objects were being marked as "modified," I think
because the HNC annotations themselves were out-of-sync. This could also
affect objects (like ResourceQuota) whose status is automatically
updated.

To fix this, copyObject() is renamed to canonical() and it now clears
any non-canonical fields, including the Status and all labels and
annotations (this was partially implemented before, but is now fully
correct).

Tested: on GKE, automatically populated and identical objects like
gke-resource-quotas are no longer marked as being modified.

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -18,10 +18,10 @@ package controllers
 import (
 	"context"
 	"reflect"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -283,53 +265,81 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
...
-	copied.SetGenerateName("")
-	copied.SetGeneration(0)
-	copied.SetInitializers(nil) // TODO: is this correct?
+// canonical returns a canonicalized version of the object - that is, one that has the same name and
+// spec, but with automatically generated fields, the Status, and any HNC-added labels or
+// annotations cleared. The resulting object is suitable to be copied into a new namespace, or two
+// canonicalized objects are suitable for being compared via reflect.DeepEqual.
+//
+// As a side effect, the label and annotation maps are always initialized in the returned value.
+func canonical(inst *unstructured.Unstructured) *unstructured.Unstructured {
```
rjbez17: This seems like a lot of things to keep track of throughout k8s releases. Could we possibly just shove the canoical state that was applied last into an annotation similar to how `kubectl apply` works with the `kubectl.kubernetes.io/last-applied-configuration` annotation? My main concern tracking/locking HNC versions based on k8s version. Although, the concern is probably overstated as this likely wont change that often.

adrianludwin: I'm a bit irrationally weirded out by having objects contain copies of themselves 😐 More practically, some objects could have _two_ copies of themselves - one from `kubectl apply` and one from HNC. Also, it's hard to tell when we should update it...

How about this instead: basically, we only want to keep the spec, name, and non-HNC labels and annotations. How about we document that this is the formal behaviour and _all_ other ObjectMetadata fields will be cleared? Then instead of a brittle blacklist, we have a much smaller and robust whitelist?

rjbez17: That works for me. If you'd rather take that on later in a separate issue lmk and i'll lgtm this.

adrianludwin:
Nah I'll go make the change today, thanks!

incubator/hnc/pkg/controllers/object_controller.go
adrianludwin: Fixed

PR #174, opened 2019-10-21
Add --optionalChild flag and assorted changes to kubectl-hnc-set

adrianludwin:
Other changes:

- Fixed initialization segfault when the init() functions stopped
being called in the order I wanted (oops).
- Only all `getHierarchy` and `updateHierarchy` once during `set` with
better messages explaining what's going on.

Tested: manually confirmed that illegal names or valid, existing
required children were removed. Also retried --parent, --root and
--requiredChild in various combinations and everythign seems to work as
expected. Filed
https://github.com/kubernetes-sigs/multi-tenancy/issues/173 to add
better unit testing in the future.

/cc @vjdhama 
/assign @rjbez17

### DIFF
--- a/incubator/hnc/pkg/kubectl/root.go
+++ b/incubator/hnc/pkg/kubectl/root.go
@@ -72,6 +72,10 @@ func init() {
 		},
 	}
 	kubecfgFlags.AddFlags(rootCmd.PersistentFlags())
+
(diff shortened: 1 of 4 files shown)


adrianludwin:
Seeing if https://github.com/kubernetes/test-infra/pull/14660 worked

/ok-to-test

PR #178, opened 2019-10-21
Code refactoring: move constants declared in `controller` to be in `api`

lafh:
The constants all are part of the API so there's no point for them to be in the `controller` package.

Tested by running `make test`

/assign @adrianludwin

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller_test.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller_test.go
@@ -13,10 +13,6 @@ import (
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
-const (
-	metaGroup = "hnc.x-k8s.io"
-)
-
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/controllers/hierarchy_controller_test.go
```diff
@@ -190,38 +186,38 @@ var _ = Describe("Hierarchy", func() {
 		// Verify that bar has a tree label related to foo
 		Eventually(func() bool {
 			barNS := getNamespace(ctx, barName)
-			_, ok := barNS.GetLabels()[fooName+".tree."+metaGroup+"/depth"]
+			_, ok := barNS.GetLabels()[fooName+".tree."+api.MetaGroup+"/depth"]
```
adrianludwin: At some point (not in this PR), maybe consider turning ".tree.x-k8s.io/depth" into a const, even if it's just in this package.

lafh: Sounds good!

PR #180, opened 2019-10-22
Flag added to show tree of all namespaces

### LINKED ISSUE #130: HNC: allow `kubectl hnc tree` to show _all_ namespaces if no argument is given

shivi28:
This PR adds a feature that allow kubectl hnc tree --all-namespaces to show all namespaces. This feature enables users to list all namespaces along with their heirarchies.

Suggested changes :
`kubectl hnc tree foo # displays foo`
`kubectl hnc tree # error`
`kubectl hnc tree --all-namespaces # displays the tree of all namespaces`
`kubectl hnc tree -A # synonym for --all-namespaces`

Which issue(s) this PR fixes:
Fixes #130

### DIFF
--- a/incubator/hnc/pkg/kubectl/tree.go
+++ b/incubator/hnc/pkg/kubectl/tree.go
@@ -17,11 +17,13 @@ package kubectl
 
 import (
 	"fmt"
+	"os"
 	"sort"
 	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
@@ -33,18 +35,32 @@ var (
 
 var treeCmd = &cobra.Command{
 	Use:   "tree",
-	Short: "Displays the hierarchy tree rooted at the given namespace",
-	Args:  cobra.ExactArgs(1),
+	Short: "Display one or more hierarchy trees",
 	Run: func(cmd *cobra.Command, args []string) {
-		nnm := args[0]
+		flags := cmd.Flags()
 		footnotesByMsg = map[string]int{}
 		footnotes = []string{}
-		hier := getHierarchy(nnm)
-		fmt.Println(nameAndFootnotes(hier))
-		printSubtree("", hier)
-
+		defaultList := len(args) == 0
+		nsList := args
+		if flags.Changed("all-namespaces") {
+			nsList = getAllNamespaces()
+		}
+		if defaultList && !flags.Changed("all-namespaces") {
+			fmt.Printf("Error: Must specify the root of the tree(s) to display or else specify --all-namespaces\n")
+			os.Exit(1)
+		}
+		for _, nnm := range nsList {
+			hier := getHierarchy(nnm)
+			//If we're showing the default list, skip all non-root namespaces since they'll be displayed as part of another namespace's tree.
+			if defaultList && hier.Spec.Parent != "" {
+				continue
+			}
+			fmt.Println(nameAndFootnotes(hier))
+			printSubtree("", hier)
+		}
 		if len(footnotes) > 0 {
 			fmt.Printf("\nConditions:\n")
+
 			for i, n := range footnotes {
 				fmt.Printf("%d) %s\n", i+1, n)
 			}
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -17,13 +17,14 @@ package kubectl
...
+	"os"
 	"sort"
 	"strconv"
 	"strings"
 
-	"github.com/spf13/cobra"
-
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
adrianludwin: Usually we have up to three "groups" of imports:

- Standard library ("os", "strings", etc)
- Packages from other projects ("cobra", "metav1", etc)
- Packages from this project ("api")

I won't hold this PR up for this change, but I'll probably change it back myself.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -32,19 +33,35 @@ var (
 )
 
 var treeCmd = &cobra.Command{
-	Use:   "tree",
-	Short: "Displays the hierarchy tree rooted at the given namespace",
-	Args:  cobra.ExactArgs(1),
+	Use:     "tree",
+	Short:   "Displays the hierarchy tree rooted at the given namespaces separated by space and if no namespace is provided then throws an error, and in case of --all-namespace/-A flag shows tree for all root namespaces",
```
adrianludwin: This isn't really a short description any more - I'd change this to "Display one or more hierarchy trees." If you add help text to `-A`, then saying `kubectl hnc help tree` _should_ give enough information to make it clear.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -32,19 +33,35 @@ var (
...
+	Aliases: []string{"tree"},
```
adrianludwin: I don't think you need this? The command is already called "tree," why did you add this line?

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -93,5 +110,20 @@ func nameAndFootnotes(hier *api.HierarchyConfiguration) string {
 }
 
 func newTreeCmd() *cobra.Command {
+	treeCmd.Flags().CountP("all-namespaces", "A", "Show tree for all namespaces")
```
adrianludwin: Maybe "Displays all trees on the cluster"?

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -32,19 +33,35 @@ var (
...
-		printSubtree("", hier)
-
+		var nsList []string
+		defaultList := len(args) == 0
+		nsList = args
+		if flags.Changed("all-namespaces") {
+			nsList = getAllNamespaces()
+		}
+		if defaultList && !flags.Changed("all-namespaces") {
+			fmt.Printf("Error: requires some argument or flag but provided none\n")
```
adrianludwin: You should give more guidance than this. Maybe "you must specify the root of the tree(s) to display or else specify --all-namespaces"?

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -93,5 +110,20 @@ func nameAndFootnotes(hier *api.HierarchyConfiguration) string {
 }
 
 func newTreeCmd() *cobra.Command {
+	treeCmd.Flags().BoolP("all-namespaces", "A", true, "Displays all trees on the cluster")
```
rjbez17: Doesn't this default the flag to true?

shivi28: Yes default value is true, Should I make it false @rjbez17 ? Actually I am not clear with its context :(

adrianludwin: Good catch Ryan. Yes, the default should be false. If the default is true, then `kubectl hnc tree` will be the same as `kubectl hnc tree -A` which is _not_ what we want.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -33,18 +35,33 @@ var (
...
 	Run: func(cmd *cobra.Command, args []string) {
-		nnm := args[0]
+		flags := cmd.Flags()
 		footnotesByMsg = map[string]int{}
 		footnotes = []string{}
-		hier := getHierarchy(nnm)
-		fmt.Println(nameAndFootnotes(hier))
-		printSubtree("", hier)
-
+		var nsList []string
```
adrianludwin: You don't need this line; on line 45 you could just say:

`nsList := args`

which initializes the new variable to the same type as `args`.

incubator/hnc/pkg/kubectl/tree.go
```diff
@@ -33,18 +35,33 @@ var (
...
+		defaultList := len(args) == 0
+		nsList = args
+		if flags.Changed("all-namespaces") {
+			nsList = getAllNamespaces()
+		}
+		if defaultList && !flags.Changed("all-namespaces") {
+			fmt.Printf("Error: Must specify the root of the tree(s) to display or else specify --all-namespaces\n")
+			return
```
adrianludwin: This should probably be `os.Exit(1)` to indicate to any calling script that there was an error.

adrianludwin:
I think we resolved all of Ryan's concerns so I'm going to approve, we can always have another go at this if there are more problems.

/approve

PR #181, opened 2019-10-22
Prevent users from applying the inheritedFrom label

### LINKED ISSUE #104: HNC: prevent users from applying the inheritedFrom label
`hnc.x-k8s.io/inheritedFrom` is used by the controller to determine which objects to manage. We should add a webhook to prevent users from applying it directly.

lafh:
`hnc.x-k8s.io/inheritedFrom` is used by the controller to determine which
objects to manage, so we decide to add a webhook to prevent users from
applying it directly.

Minor Code refactoring: move customized GVKs to `pkg/config` as the usage is not
limited to `controller` only.

/resolve #104

### DIFF
--- a/incubator/hnc/pkg/config/gvk.go
+++ b/incubator/hnc/pkg/config/gvk.go
@@ -1,0 +1,16 @@
+package config
+
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+// GVKs is currently hardcoded to the set of GVKs handled by the HNC - namely, Secrets,
+// Roles and RoleBindings - but in the future we should get this from a
+// configuration object.
+var GVKs = []schema.GroupVersionKind{
+	{Group: "", Version: "v1", Kind: "Secret"},
+	{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "Role"},
+	{Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "RoleBinding"},
+	{Group: "networking.k8s.io", Version: "v1", Kind: "NetworkPolicy"},
+	{Group: "", Version: "v1", Kind: "ResourceQuota"},
+	{Group: "", Version: "v1", Kind: "LimitRange"},
+	{Group: "", Version: "v1", Kind: "ConfigMap"},
+}
--- a/incubator/hnc/pkg/validators/object.go
+++ b/incubator/hnc/pkg/validators/object.go
@@ -1,0 +1,119 @@
+package validators
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
+)
+
+// ObjectsServingPath is where the validator will run. Must be kept in sync with the
+// kubebuilder markers below.
+const (
+	ObjectsServingPath = "/validate-objects"
+)
+
+// Note: the validating webhook FAILS OPEN. This means that if the webhook goes down, all further
+// changes to the objects are allowed.
+//
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="",resources=secrets,verbs=create;update,versions=v1,name=secrets
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="rbac.authorization.k8s.io",resources=rols,verbs=create;update,versions=v1,name=roles.rbac.authorization.k8s.io
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=create;update,versions=v1,name=rolebindings.rbac.authorization.k8s.io
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="networking.k8s.io",resources=networkpolicies,verbs=create;update,versions=v1,name=networkpolicies.networking.k8s.io
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="",resources=resourcequotas,verbs=create;update,versions=v1,name=resourcesquotas
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="",resources=limitranges,verbs=create;update,versions=v1,name=limitranges
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="",resources=configmaps,verbs=create;update,versions=v1,name=configmaps
+
+type Object struct {
+	Log     logr.Logger
+	Forest  *forest.Forest
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+func (o *Object) Handle(ctx context.Context, req admission.Request) admission.Response {
+	log := o.Log.WithValues("nm", req.Name, "nnm", req.Namespace)
+	if isHNCServiceAccount(&req.AdmissionRequest.UserInfo) {
+		log.Info("Allowed change by HNC SA")
+		return allow("HNC SA")
+	}
+
+	inst := &unstructured.Unstructured{}
+	if err := o.decoder.Decode(req, inst); err != nil {
+		log.Error(err, "Couldn't decode req.Object", "raw", req.Object)
+		return deny(metav1.StatusReasonBadRequest, err.Error())
+	}
+	log = log.WithValues("object", inst.GetName())
+
+	oldInst := &unstructured.Unstructured{}
+	// req.OldObject is the existing object. DecodeRaw will return an error if it's empty, so we should skip the decoding here.
+	if len(req.OldObject.Raw) > 0 {
+		if err := o.decoder.DecodeRaw(req.OldObject, oldInst); err != nil {
+			log.Error(err, "Couldn't decode req.OldObject", "raw", req.OldObject)
+			return deny(metav1.StatusReasonBadRequest, err.Error())
+		}
+	}
+
+	resp := o.handle(ctx, log, inst, oldInst)
+	log.Info("Handled", "allowed", resp.Allowed, "code", resp.Result.Code, "reason", resp.Result.Reason, "message", resp.Result.Message)
+	return resp
+}
+
+// handle implements the non-webhook-y businesss logic of this validator, allowing it to be more
+// easily unit tested (ie without constructing an admission.Request, setting up user infos, etc).
+func (o *Object) handle(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured, oldInst *unstructured.Unstructured) admission.Response {
+	oldValue, oldExists := getLabel(oldInst, api.LabelInheritedFrom)
+	newValue, newExists := getLabel(inst, api.LabelInheritedFrom)
+	// If old object holds the label but the new one doesn't, reject it. Vice versa.
+	if oldExists != newExists {
+		verb := "add"
+		if !newExists {
+			verb = "remove"
+		}
+		return deny(metav1.StatusReasonForbidden, "Users should not "+verb+" the label "+api.LabelInheritedFrom)
+	}
+	// If both old and new objects hold the label but with different values, reject it.
+	if newExists && newValue != oldValue {
+		return deny(metav1.StatusReasonForbidden, "Users should not change the value of label "+api.LabelInheritedFrom)
+	}
+	return allow("")
+}
+
+func (o *Object) InjectClient(c client.Client) error {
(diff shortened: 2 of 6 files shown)


incubator/hnc/pkg/config/gvk.go
```diff
@@ -0,0 +1,16 @@
+package config
+
+import "k8s.io/apimachinery/pkg/runtime/schema"
+
+// For now, it hardcodes the list of GVKs handled by the HNC - namely, Secrets,
```
adrianludwin: Comments for exported names should start with that name so we can run `godoc` on it.

```
// GVKs is currently hardcoded to the set of GVKs
// handled by the HNC...
```

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,117 @@
...
+
+type Object struct {
+	Log     logr.Logger
+	Forest  *forest.Forest
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+func (o *Object) Handle(ctx context.Context, req admission.Request) admission.Response {
+	log := o.Log
```
adrianludwin: You can also say:

`log := o.log.WithValues("object", req.Name)`

You don't need to wait until line 47. Or better yet:

`log := o.log.WithValues("nm", req.Name, "nnm", req.Namespace)`

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,117 @@
...
+	inst := &unstructured.Unstructured{}
+	if err := o.decoder.Decode(req, inst); err != nil {
+		log.Error(err, "Couldn't decode request")
+		return deny(metav1.StatusReasonBadRequest, err.Error())
+	}
+	log = log.WithValues("object", inst.GetName())
+
+	// oldInst is the existing object. It is null for CREATE and CONNECT.
```
adrianludwin: s/null/nil/

What happens if `req.OldObject is nil`? Will this cause an error or panic? Maybe nice to specify that in the comment.

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,117 @@
...
+func (o *Object) handle(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured, oldInst *unstructured.Unstructured) admission.Response {
+	oldValue, oldOK := getLabel(oldInst, api.LabelInheritedFrom)
+	newValue, newOK := getLabel(inst, api.LabelInheritedFrom)
+	// If old object holds the label but the new one doesn't, reject it. Vice versa.
+	if oldOK != newOK {
+		verb := "add"
+		if !newOK {
+			verb = "remove"
+		}
+		return deny(metav1.StatusReasonUnauthorized, "Users should not"+verb+"label"+api.LabelInheritedFrom)
```
adrianludwin: "Unauthorized" requests can be retried with different authorization, but that isn't what we want here - _no_ user should be able to set this. So I'd go with "Forbidden" (403) instead.

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	// If both old and new objects hold the label but with different values, reject it.
+	if newOK && newValue != oldValue {
+		return deny(metav1.StatusReasonUnauthorized, "Users should not change the value of label"+api.LabelInheritedFrom)
```
adrianludwin: Ditto

incubator/hnc/pkg/validators/object.go
adrianludwin: Also, you need spaces in the message, I think:

```
"Users should not "+verb+" the label "+api.LabelInheritedFrom
                 ^        ^^^^^     ^
```

incubator/hnc/pkg/validators/object_test.go
```diff
@@ -0,0 +1,64 @@
...
+		newValue string
+		fail     bool
+	}{
+		{name: "LabelInheritedFrom label doesn't exist",
+			oldLabel: "oldLabel", oldValue: "foo",
+			newLabel: "newLabel", newValue: "bar"},
+		{name: "LabelInheritedFrom label stays the same",
+			oldLabel: api.LabelInheritedFrom, oldValue: "foo",
+			newLabel: api.LabelInheritedFrom, newValue: "foo"},
+		{name: "LabelInheritedFrom label's value is changed",
```
adrianludwin: Nit: this is entirely up to you but I sometimes like this syntax instead (less but more consistent indentation):

```
}{{
   name: "...",
   oldLabel: "...", oldValue: "...",
   ...
}, {
   ... next testcase ...
}}
```

Entirely up to you though.

incubator/hnc/pkg/validators/object_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
+)
+
+func TestObject(t *testing.T) {
```
adrianludwin: I'd call this TestInheritedFromLabel since you're not testing _arbitrary_ properties of the object. Then each subtest name can be shorter (ie "label doesn't exist").

incubator/hnc/pkg/validators/object_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	tests := []struct {
+		name     string
+		oldLabel string
+		oldValue string
+		newLabel string
+		newValue string
+		fail     bool
+	}{
+		{name: "LabelInheritedFrom label doesn't exist",
```
adrianludwin: I'd say something like "regular labels can be changed"

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
+)
+
+// ObjectsServingPath is where the validator will run. Must be kept in sync with the
+// kubebuilder markers below.
+const (
+	ObjectsServingPath = "/validate-objects"
+)
+
+// Note: the validating webhook FAILS OPEN. This means that if the webhook goes down, all further
```
rjbez17: Do we want this? I'd think failing closed would be more ideal here (for better or worse)?

incubator/hnc/pkg/validators/object_test.go
rjbez17: +1

incubator/hnc/pkg/validators/object.go
rjbez17: @adrianludwin  thoughts?

adrianludwin: For now, yes, I think so - if the HNC (which is pre-alpha!) crashes, and the webhook fails closed, _all_ modifications to any of these objects will be rejected. Maybe one day we can change this. When I spoke to @liggitt he felt quite strongly about this.

By contrast, changes to the HierarchyConfiguration _do_ fail closed because the worst that can happen is that you can no longer modify your (existing, presumably valid) configuration.

rjbez17: > if the HNC (which is pre-alpha!) crashes, and the webhook fails closed, all modifications to any of these objects will be rejected.

I'd argue if you expect a pre-alpha thing not to affect your cluster it's on you. My main concern is the controller going down for a short time and then uses being allowed to circumvent the HNC. This is less a concern for HNC but more when we wrap a Tenant controller around it.

If you're deploying HNC to  your cluster to have this hierarchy, you're likely relying on it to provide some form access controls for your users. If HNC goes down, then maybe it's ok if all objects are rejected? Or at least it's configurable. It really depends on the usecase. If HNC is just to help sync across namespaces it won't matter that much, but if you are trying to achieve some form of isolation from this could have negative consequences.

Regardless, we don't have to figure this out now. I would like to revisit this at some point though.

rjbez17:
This lgtm once @adrianludwin comments are closed and tests pass.

incubator/hnc/pkg/validators/object.go
adrianludwin: I mean, it is configurable by editing the manifest 😁  This is just the default.

adrianludwin: With that said, I filed #188 to revisit this default in v0.2. Thanks!

lafh: Good point. I will leave it for a future PR. This PR will focus on implementation of the validator.

incubator/hnc/pkg/config/gvk.go
lafh: Done.

incubator/hnc/pkg/validators/object.go
lafh: Done.

incubator/hnc/pkg/validators/object.go
lafh: Done.

incubator/hnc/pkg/validators/object.go
lafh: Done.

incubator/hnc/pkg/validators/object.go
lafh: Done.

incubator/hnc/pkg/validators/object_test.go
lafh: Done.

incubator/hnc/pkg/validators/object_test.go
lafh: Done.

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,119 @@
...
+	Forest  *forest.Forest
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+func (o *Object) Handle(ctx context.Context, req admission.Request) admission.Response {
+	log := o.Log.WithValues("nm", req.Name, "nnm", req.Namespace)
+	inst := &unstructured.Unstructured{}
+	if err := o.decoder.Decode(req, inst); err != nil {
+		log.Error(err, "Couldn't decode request")
```
adrianludwin: It might be more informative to say "Couldn't decode req.Object" and (for line 53) "Couldn't decode req.OldObject"? Also, it might be helpful to actually have the raw contents so we can see what went wrong:

```
// This line:
log.Error(err, "Couldn't decode req.Object", "raw", req.Object)
...
// Line 53:
log.Error(err, "Couldn't decode req.OldObject", "raw", req.OldObject)
```

That way, if something goes wrong, we'll be able to debug. Right now, we have no information we can use.

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,119 @@
...
+	oldInst := &unstructured.Unstructured{}
+	// req.OldObject is the existing object. DecodeRaw will return an error if it's empty, so we should skip the decoding here.
+	if len(req.OldObject.Raw) > 0 {
+		if err := o.decoder.DecodeRaw(req.OldObject, oldInst); err != nil {
+			log.Error(err, "Couldn't decode RAW request")
+			return deny(metav1.StatusReasonBadRequest, err.Error())
+		}
+	}
+
+	if isHNCServiceAccount(&req.AdmissionRequest.UserInfo) {
```
adrianludwin: Maybe put this before we try to decode the objects? Otherwise, HNC will never be able to fix any problem if there's something wrong.

incubator/hnc/pkg/validators/object.go
```diff
@@ -0,0 +1,119 @@
...
+
+	resp := o.handle(ctx, log, inst, oldInst)
+	log.Info("Handled", "allowed", resp.Allowed, "code", resp.Result.Code, "reason", resp.Result.Reason, "message", resp.Result.Message)
+	return resp
+}
+
+// handle implements the non-webhook-y businesss logic of this validator, allowing it to be more
+// easily unit tested (ie without constructing an admission.Request, setting up user infos, etc).
+func (o *Object) handle(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured, oldInst *unstructured.Unstructured) admission.Response {
+	oldValue, oldOK := getLabel(oldInst, api.LabelInheritedFrom)
```
adrianludwin: Nit: I wouldn't call this "OK", what it really means is if the label _exists_ or not. So this should be `oldExists` and `newExists`, that'll make the code below much easier to read.

E.g., instead of:

```
if !newOK {
  verb = "remove"
}
```

we get:

```
if !newExists {
  verb = "remove"
}
```

which reads more clearly (at least to me), because "!exists... remove" are more closely related than "!ok... remove".

incubator/hnc/pkg/validators/object_test.go
```diff
@@ -0,0 +1,64 @@
...
+	l := zap.Logger(false)
+
+	tests := []struct {
+		name     string
+		oldLabel string
+		oldValue string
+		newLabel string
+		newValue string
+		fail     bool
+	}{{name: "Regular labels can be changed",
```
adrianludwin: Nit: `name:` should be on its own line, so that it lines up with the other fields.

```
}{{
  name: "Regular labels can be changed",
  oldLabel: ...
  newLabel: ...
  ...
}, {
  name: "Label stays the same",
  ...
```

adrianludwin:
/lgtm
/approve
/hold
/assign @rjbez17 

To Ryan for a last review

Issue #89, opened 2019-10-03
HNC: forbid propagated objects from being modified


adrianludwin:
Thanks to #65 , we can now _report_ when a user has incorrectly modified a propagated object, but ideally we should _prevent_ it instead. So we should add a check in `pkg/validators` to ensure that propagated objects are never modified by anyone other than the HNC itself.

PR #183, opened 2019-10-22
Set object conditions to replace modified annotation

yiqigao217:
This is a part of #76 and it's related to #89 .

This replaces modified annotation with object conditions when object is
detected as modified from source object. It will set ObjectOverridden
Condition on the modified object and ObjectDescendantOverridden
Condition on the source object. Before propagation, if any ancestor of
the object has ObjectOverridden Condition, it will not be propagated.

Tested: make test and manual test in Kind.

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"strings"
 
@@ -40,6 +41,9 @@ type ObjectReconciler struct {
 
 	// Forest is the in-memory forest managed by the HierarchyReconciler.
 	Forest *forest.Forest
+
+	// HierarchyReconciler is for reconciling namespaces when object conditions are detected.
+	HierarchyReconciler *HierarchyReconciler
 
 	// GVK is the group/version/kind handled by this reconciler.
 	GVK schema.GroupVersionKind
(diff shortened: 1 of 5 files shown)


adrianludwin:
/assign @Fei-Guo 

I'll give a review as well.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -177,6 +180,60 @@ func (r *ObjectReconciler) hasCorrectAncestry(ctx context.Context, log logr.Logg
...
+	log.Info("Instance modified, setting ObjectOverridden", "affectedKey", ak)
+	msg := fmt.Sprintf("object overridden from source %s", sak)
+	ns.SetCondition(ak, api.ObjectOverridden, msg)
+	r.HierarchyReconciler.enqueueAffected(log, msg, nnm)
+
+	snnm := src.GetNamespace()
+	sns := r.Forest.Get(snnm)
+	log.Info("Descendent modified, setting ObjectDescendentOverridden", "affectedKey", sak)
+	msg = fmt.Sprintf("descendent object %s overridden", ak)
+	sns.SetCondition(sak, api.ObjectDescendantOverridden, msg)
```
adrianludwin: I think the affected object here should be the _modified_ object, not the source object, because knowing the modified object _and_ the fact that it's in a descendant of the current ns tells us exactly where the current one is.

adrianludwin:
/hold

@yiqigao217 and I discussed this and we've agreed we need to change the approach.

yiqigao217:
As discussed off-line with @adrianludwin , Object conditions may be rescheduled to 0.2 since we will mostly rely on webhook to prevent manual modification to objects. We will reconsider how controllers start up and deal with existing namespaces & objects.

As a record: 
* In this PR, we fixed these concurrency issues when detecting Object conditions:
  * Don't propagate to child with `ObjectOverridden` condition - prevent object being overridden and thus Object conditions being wiped out. 
  * Only call HierarchyReconciler (`enqueueAffected`) when Object conditions are set/unset - prevent infinite loop between Hier & Object controllers calling each other.
* What we didn't fix in this PR is racing between unfinished propagation of a parent and modification of a child, which may be ok to just leave it. Currently, I added `time.Sleep(2*time.Second)` before each manual object modification to avoid this racing case and make sure that all the test cases are not flaky.
* Future approaches/thoughts we may take to reduce the number of times we copy objects for no good reason:
  * Don't `updateObjects()` when namespace hier is reconciled if no hier structure changes
  * Don't propagate objects again if the source doesn't change.

I will close this PR for now, just keeping the code and discussions for future reference.
/close

Issue #172, opened 2019-10-21
HNC: validator should reject illegal namespace names


adrianludwin:
Currently `kubectl hnc set foo --requiredChild %^&^%` goes through just fine, though of course the namespace can never be created. We should reject this in the webhook.

frbimo:
hi @adrianludwin , this is my very first PR here, can i pick this one?

adrianludwin:
@frbimo - do you know how validating admission controller webhooks work in K8s? If so, please go right ahead! You're not a member of the kubernetes-sigs organization so I can't assign this issue to you but please feel free to send a PR.

We already have a fair number of new contributors so unfortunately we won't have a lot of time to respond to questions. However, you can try to ping us on the [K8s Slack](https://slack.k8s.io) - @rjbez17 ("Ryan Bez" on Slack) has offered to help out if you get stuck.

frbimo:
@adrianludwin Got it. I will try to ping @rjbez17  first to confirm about this issue. Thank you

adrianludwin:
Fixed by #184

PR #184, opened 2019-10-23
Validator reject illegal namespace names

frbimo:
This PR fix validation on webhook to reject illegal character on namespace when execute kubectl with `--requiredChild` parameter.

Which issue(s) this PR fixes:
#172

### DIFF
--- a/incubator/hnc/pkg/validators/hierarchy.go
+++ b/incubator/hnc/pkg/validators/hierarchy.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/go-logr/logr"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	authnv1 "k8s.io/api/authentication/v1"
 	authzv1 "k8s.io/api/authorization/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -101,6 +103,12 @@ func (v *Hierarchy) handle(ctx context.Context, log logr.Logger, req *request) a
 	if isHNCServiceAccount(req.ui) {
 		log.Info("Allowed change by HNC SA")
 		return allow("HNC SA")
+	}
+
+	// Verify the HC is legal in isolation (i.e., before checking the rest of the forest)
+	resp := checkConfig(req.hc)
+	if !resp.Allowed {
+		return resp
 	}
 
 	// Do all checks that require holding the in-memory lock. Generate a list of authz checks we
@@ -289,6 +297,31 @@ func isHNCServiceAccount(user *authnv1.UserInfo) bool {
 		}
 	}
 	return false
+}
+
+// checkConfig is checking whether namespaces in the hierarchy configuration meet kubernetes requirements.
+// if required children's field contains invalid name, it returns admission response that reject namespaces creation.
+func checkConfig(hc *api.HierarchyConfiguration) admission.Response {
+
+	// Check if children names in requiredChildren field obey kubernetes namespace regex format.
+	// invalidRCs accomodates illegal required child(RC) name(s).
+	invalidRCs := []string{}
+	for _, rc := range hc.Spec.RequiredChildren {
+		if resp := validateNamespace(rc); resp != nil {
+			invalidRCs = append(invalidRCs, rc)
+		}
+	}
+
+	if len(invalidRCs) > 0 {
+		return deny(metav1.StatusReasonBadRequest, fmt.Sprintf("The following required children are not valid namespace names: %s", strings.Join(invalidRCs, ", ")))
+	}
+	return allow("")
+}
+
+// validateNamespace validates a string is a valid namespace using apimachinery.
+// https://godoc.org/k8s.io/apimachinery/pkg/util/validation#IsDNS1123Label
+func validateNamespace(s string) []string {
+	return validation.IsDNS1123Label(s)
 }
 
 func (v *Hierarchy) InjectClient(c client.Client) error {
--- a/incubator/hnc/pkg/validators/hierarchy_test.go
+++ b/incubator/hnc/pkg/validators/hierarchy_test.go
@@ -26,18 +26,26 @@ func TestStructure(t *testing.T) {
 		name string
 		nnm  string
 		pnm  string
+		rcm  []string
 		fail bool
 	}{
 		{name: "ok", nnm: "foo", pnm: "baz"},
 		{name: "missing parent", nnm: "foo", pnm: "brumpf"},
 		{name: "self-cycle", nnm: "foo", pnm: "foo", fail: true},
 		{name: "other cycle", nnm: "foo", pnm: "bar", fail: true},
+		{name: "rc ok", rcm: []string{"bar-baz"}, nnm: "foo"},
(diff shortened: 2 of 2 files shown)


frbimo:
"I signed it"

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -127,6 +132,20 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
 	curParent := ns.Parent()
 	newParent := v.Forest.Get(hc.Spec.Parent)
 
+	// Check if requiredChild names in HierarchyConfiguration.Spec.RequiredChildren obey
```
adrianludwin: This check isn't actually checking the in-memory forest, so it shouldn't be in a function called `checkForest`. Maybe make a new function called `checkConfig`?

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -114,6 +116,9 @@ func (v *Hierarchy) handle(ctx context.Context, log logr.Logger, req *request) a
 	return v.checkAuthz(ctx, req.ui, authzReqs)
 }
 
+const validRCRexFmt string = "^[a-z0-9]([-a-z0-9]*[a-z0-9])?[a-z0-9]$"
+const validRCMaxLength int = 63
+
```
adrianludwin: Can you please comment:

- What the names mean ("RC?")
- Where you got these regexes from (i.e., if you didn't cut-and-paste the code, where's the documentation that told you how to write them)?

Also, what you're really doing here is validating that the namespace name is legal. Did you check whether any Kubernetes golang package can do this check for you?

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -127,6 +132,20 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
+	invalidRCs := []string{}
+	validRCRex := regexp.MustCompile(validRCRexFmt)
+	for _, rc := range hc.Spec.RequiredChildren {
+		if !validRCRex.MatchString(rc) || len(rc) > validRCMaxLength {
+			invalidRCs = append(invalidRCs, rc)
+		}
+	}
+
+	if len(invalidRCs) > 0 {
+		return nil, deny(metav1.StatusReasonBadRequest, fmt.Sprintf("Invalid requiredChild name: %v", strings.Join(invalidRCs, " ")))
```
adrianludwin: Better: maybe something like "The following required children are not valid namespace names: "

incubator/hnc/pkg/validators/hierarchy.go
frbimo: - Required Child (RC)

-When creating a namespace with invalid character, it will return info about regex used for validation.i referenced from there and modify it since it's not working when I implement and test it. I found the "real" regexes from kubernetes package. I will change according to it. 

- There is a kubernetes package for validating name but the function name (validation.IsDNS1123Label) seems not suitable here IMO. Should I go with that one?

incubator/hnc/pkg/validators/hierarchy.go
frbimo: Ok. I will change it according to your suggestion.

incubator/hnc/pkg/validators/hierarchy.go
frbimo: I see. WIll revised it soon. :)

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: Oh, I wasn't asking for you to comment _here_ - I meant you should add a comment to the _code_, using [godoc](https://blog.golang.org/godoc-documenting-go-code) style. Eg:

```
// validRCRexFmt is the regex used to validate namespaces for the required children
// field in the hierarchy spec. It's based on this: <link>
```

However, I _would_ use `validation.IsDNS1123Label` if that's what K8s uses, and just leave a comment in the code _why_ you did that.

adrianludwin:
Looks like we have a flaky presubmit test (these tests are quite new). Please ignore it for now.

adrianludwin:
@frbimo , it looks like you're trying to merge from your master branch, so you've picked up a bunch of commits that aren't part of your change. Do you know how to use Git to set up a feature branch and _only_ submit changes from that branch?

frbimo:
@adrianludwin usually i create branch rather than using master. But i forgot this time. 
Should i squash them?

adrianludwin:
Yes, can you please squash everything into one commit?

If it makes it easier, you can always close this PR, create a new feature branch, and start pushing on that. Whatever works for you.

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -103,6 +105,12 @@ func (v *Hierarchy) handle(ctx context.Context, log logr.Logger, req *request) a
 		return allow("HNC SA")
 	}
 
+	// Do pre-check namespaces in the request configuration, make sure
```
adrianludwin: Finish this comment? Maybe something like:

`// Verify the HC is legal in isolation (i.e., before checking the rest of the forest)`

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -291,6 +299,31 @@ func isHNCServiceAccount(user *authnv1.UserInfo) bool {
...
+	// invalidRCs accomodates illegal required child(RC) name(s).
+	invalidRCs := []string{}
+	for _, rc := range hc.Spec.RequiredChildren {
+		if resp := validateNamespace(rc); resp != nil {
+			invalidRCs = append(invalidRCs, rc)
+		}
+	}
+
+	if len(invalidRCs) > 0 {
+		return deny(metav1.StatusReasonBadRequest, fmt.Sprintf("The following required children are not valid namespace names: %v", strings.Join(invalidRCs, " ")))
```
adrianludwin: Please use `%s` instead of `%v`; in this case they're identical but we should use the strongest type we can.

Also, maybe consider `strings.Join(invalidRCs, ", ")` - the commas will probably make the list easier to read.

incubator/hnc/pkg/validators/hierarchy_test.go
```diff
@@ -143,6 +152,17 @@ func createNS(f *forest.Forest, nnm string, ue []string) *forest.Namespace {
 	return ns
 }
 
+// createRandString generate string for namespace based on given length value
```
adrianludwin: Can you please take out the randomness and just hardcode one 63-char and one 64-char string? I know it _shouldn't_ make a difference but we shouldn't be testing anything random (we only use randomness on the namespace names to get around the fact that we can't delete them properly in the test context).

frbimo: for this part, i only considered for the purpose that it is easy to check the length of the string exactly 63. rather than counting it manually.

adrianludwin: Better to just have 64 `a`s IMO then go through this more complicated logic.

incubator/hnc/pkg/validators/hierarchy.go
frbimo: ok

incubator/hnc/pkg/validators/hierarchy.go
frbimo: ok

incubator/hnc/pkg/validators/hierarchy_test.go
frbimo: noted.

adrianludwin:
/approve
/lgtm

Thanks!

PR #187, opened 2019-10-25
HNC: Added Client Interface to support test cases for kubectl commands

shivi28:
Changes:
Create Client interface in kubectl package and some code refactoring. This change is done to make the kubectl package functions testable in future.

Tested:  manually run kubectl tree, show and set commands and verified that they were
reported correctly.

### DIFF
--- a/incubator/hnc/pkg/kubectl/root.go
+++ b/incubator/hnc/pkg/kubectl/root.go
@@ -36,9 +36,19 @@ import (
 var k8sClient *kubernetes.Clientset
 var hncClient *rest.RESTClient
 var rootCmd *cobra.Command
+var client Client
+
+type realClient struct{}
+
+type Client interface {
+	getHierarchy(nnm string) *api.HierarchyConfiguration
+	updateHierarchy(hier *api.HierarchyConfiguration, reason string)
+}
 
 func init() {
 	api.AddToScheme(scheme.Scheme)
+
+	client = &realClient{}
 
 	kubecfgFlags := genericclioptions.NewConfigFlags(false)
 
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/kubectl/root.go
```diff
@@ -36,10 +36,24 @@ import (
...
+type APIServer struct{}
+
+type hncAPIServer interface {
+	getHierarchy(nnm string) *api.HierarchyConfiguration
+	updateHierarchy(hier *api.HierarchyConfiguration, reason string)
+}
+
+func NewAPIServer() *APIServer {
+	return &APIServer{}
+}
```
adrianludwin: A few things here:

- I think I suggested calling this type "Server" but it's really a client (see `k8sClient` and `hncClient`, above), so let's change the name to client.
- The variable should have the _interface_ type, not the _concrete_ type. Otherwise we can't replace the real server with a fake one.
- I'd suggest calling the interface type `Client` (or maybe `APIClient`) and the real implementation `realClient`. The interface is the type everyone will be using, so give it the better name.
- There's no need to have a method called `NewClient`. In `init()`, simply say `client = &realClient{}`. As far as I've seen, that's the preferred Go style (avoid a `New` function) unless there's nontrivial processing required to create an object.

adrianludwin: Finally, if `k8sClient` and `hncClient` are no longer used anywhere _except_ in `realClient`, maybe put them into that struct so they're not globals anymore?

rjbez17:
This lgtm, hold for @adrianludwin  to look at

/lgtm
/approve
/hold

shivi28:
@rjbez17 @adrianludwin can you please help in resolving this error.

adrianludwin:
Not sure why this test is so flaky today. A PR is coming to remove it.

/retest

adrianludwin:
Flaky test removed in #190

PR #192, opened 2019-10-26
fix golint errors

bysph:
Signed-off-by: sph <shepenghui1@huawei.com>


rjbez17:
I’d prefer the changes to hnc and virtual cluster be separate PRs. They really should be separate repos but it’s a limitation of the current org structure.

/hold

rjbez17:
@bysph any reason now to close this and open seperate PRs? I think I already saw another go through.

Issue #185, opened 2019-10-24
HNC: ns.tree.hnc.x-k8s.io labels are not removed when the hierarchy changes


adrianludwin:
I moved a namespace from one parent to another and noticed that the old namespace label was not removed. The reconciler should remove all existing hnc.x-k8s.io labels (and annotations) when it reconciles a namespace to ensure that old data is cleaned up.

/assign @lafh

frbimo:
@adrianludwin can i pick this issue?

adrianludwin:
Go for it. Note that writing unit tests in hierarchy_controller_test.go
will be an essential part of this task.

frbimo:
@adrianludwin Ok will try to sync with you or @rjbez17 later. 
One question, you said remove "all", means remove even if there are labels other than hnc's label?

adrianludwin:
Fixed via #193

PR #193, opened 2019-10-27
remove label when hierarchy changes

frbimo:
This PR fix label on namespace by removing all existing hnc.x-k8s.io labels (and annotations) when the hierarchy changes.

Which issue(s) this PR fixes:
#185

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -262,12 +263,22 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
 		return
 	}
 
+	// Pre-define label depth suffix
+	labelDepthSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
+
+	// Remove all existing depth labels.
+	for k := range nsInst.Labels {
+		if strings.HasSuffix(k, labelDepthSuffix) {
+			delete(nsInst.Labels, k)
+		}
+	}
+
 	// AncestoryNames includes the namespace itself.
 	ancestors := ns.AncestoryNames(nil)
 	for i, ancestor := range ancestors {
-		labelDepthSuffix := ancestor + ".tree." + api.MetaGroup + "/depth"
+		l := ancestor + labelDepthSuffix
 		dist := strconv.Itoa(len(ancestors) - i - 1)
-		setLabel(nsInst, labelDepthSuffix, dist)
+		setLabel(nsInst, l, dist)
 	}
 
 	// All namespaces in its subtree should update the labels as well.
--- a/incubator/hnc/pkg/controllers/hierarchy_controller_test.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller_test.go
@@ -221,6 +221,69 @@ var _ = Describe("Hierarchy", func() {
 			return val
 		}).Should(Equal("0"))
 	})
+
+	It("should update labels when parent is changed", func() {
+		// Set up key-value pair for non-HNC label
+		const keyName = "key"
+		const valueName = "value"
+
+		// Set up initial hierarchy
+		bazName := createNSWithLabel(ctx, "baz", map[string]string{keyName: valueName})
+		bazHier := newHierarchy(bazName)
+		depthSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
+		Eventually(getLabel(ctx, bazName, bazName+depthSuffix)).Should(Equal("0"))
+		Eventually(getLabel(ctx, bazName, keyName)).Should(Equal(valueName))
+
+		// Make baz as a child of bar
+		bazHier.Spec.Parent = barName
+		updateHierarchy(ctx, bazHier)
+
+		// Verify all labels on baz after set bar as parent
+		Eventually(getLabel(ctx, bazName, bazName+depthSuffix)).Should(Equal("0"))
+		Eventually(getLabel(ctx, bazName, barName+depthSuffix)).Should(Equal("1"))
+		Eventually(getLabel(ctx, bazName, keyName)).Should(Equal(valueName))
+
+		// Change parent to foo
+		bazHier.Spec.Parent = fooName
+		updateHierarchy(ctx, bazHier)
+
+		// Verify all labels on baz after change parent to foo
+		Eventually(getLabel(ctx, bazName, bazName+depthSuffix)).Should(Equal("0"))
+		Eventually(getLabel(ctx, bazName, fooName+depthSuffix)).Should(Equal("1"))
+		Eventually(getLabel(ctx, bazName, barName+depthSuffix)).Should(Equal(""))
+		Eventually(getLabel(ctx, bazName, keyName)).Should(Equal(valueName))
+	})
+
+	It("should update labels when parent is removed", func() {
+		// Set up key-value pair for non-HNC label
+		const keyName = "key"
+		const valueName = "value"
+
+		// Set up initial hierarchy
+		bazName := createNSWithLabel(ctx, "baz", map[string]string{keyName: valueName})
+		bazHier := newHierarchy(bazName)
+		depthSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
+		Eventually(getLabel(ctx, bazName, bazName+depthSuffix)).Should(Equal("0"))
+		Eventually(getLabel(ctx, bazName, keyName)).Should(Equal(valueName))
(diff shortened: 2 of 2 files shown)


incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -262,10 +263,27 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
...
+	labelSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
+
+	// Remove all existing HNC labels.
+	for k := range nsInst.Labels {
+		if strings.HasSuffix(k, labelSuffix) {
+			delete(nsInst.Labels, k)
+		}
+	}
+
+	// Remove all existing HNC annotations.
```
adrianludwin: I know I said "annotations" in the issue, but this is the only function that puts anything on the namespace and there don't appear to be any annotations (yet). If we do add any, they won't end in `/depth`. So we should remove this bit.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -262,10 +263,27 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
 		return
 	}
 
+	// pre-define HNC label suffix
```
adrianludwin: Nit: I'd call the the "depth label suffix," not the "HNC label suffix," since HNC has lots of suffixes.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -262,10 +263,27 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
...
+		if strings.HasSuffix(k, labelSuffix) {
+			delete(nsInst.Annotations, k)
+		}
+	}
+
 	// AncestoryNames includes the namespace itself.
 	ancestors := ns.AncestoryNames(nil)
 	for i, ancestor := range ancestors {
-		labelDepthSuffix := ancestor + ".tree." + api.MetaGroup + "/depth"
+		labelDepthSuffix := ancestor + labelSuffix
```
adrianludwin: We should fix the variable name (I missed it last time) - this isn't the suffix, it's the entire label. So instead of `labelDepthSuffix`, let's change it to `depthLabel` or even just `label` or `l`, to be honest (see [here](https://github.com/golang/go/wiki/CodeReviewComments#variable-names)).

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
```diff
@@ -19,6 +19,7 @@ var _ = Describe("Hierarchy", func() {
 	var (
 		fooName string
 		barName string
+		bazName string
```
adrianludwin: If you're creating a new global variable, you should initialize it in `BeforeEach` as well.

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
```diff
@@ -221,6 +222,164 @@ var _ = Describe("Hierarchy", func() {
 			return val
 		}).Should(Equal("0"))
 	})
+
+	It("should sync tree label with current hierarchy", func() {
```
adrianludwin: This test is very, very complicated (almost 200 lines). If it fails, it will take me a long time to figure out what the test was trying to do, which part of it broke, and why. For example, you change the parent of baz three times - why is the third time needed? Could you turn that into two tests? And can you factor out some of the code into smaller functions? Eg:

```
getDepth(ctx context.Context, from, to string) func() string {
  return func() string {
    ns := getNamespace(ctx, from)
    val, ok := ns.GetLabels[to+depthSuffix]
    return val
  }
}
```

You can use that to test the existence, nonexistence and value of the same label. Note that you need to return a _function_, not a value, since `Eventually()` will repeatedly try to call the function again.

frbimo: the third one is the case where baz (children from a parent) become root. Previously, i also thought that is similar concept with moving from parent a to parent b. But then decided to cover all possible case even though they have are similar. 

I will refactor the function as you suggest.

adrianludwin: In that case, perhaps factor it into two different testcases, ie "depth labels get updated when parent is changed" and "depth labels get updated when the parent is removed"?

incubator/hnc/pkg/controllers/hierarchy_controller.go
frbimo: removed

incubator/hnc/pkg/controllers/hierarchy_controller.go
frbimo: updated

incubator/hnc/pkg/controllers/hierarchy_controller.go
frbimo: Nice suggestion, i have update it

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
frbimo: since i use different function to create ns, i remove bazName from global and declare locally.

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
frbimo: Got it. Updated as you suggested. PTAL

incubator/hnc/pkg/controllers/hierarchy_controller.go
frbimo: sorry, the previous variable name is "labelDepthSuffix". Currently i change according to it. Not sure about the naming here, should i change to "depthLabelSuffix" ?

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -262,12 +263,22 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
 		return
 	}
 
+	// Pre-define depth label suffix
+	labelDepthSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
+
+	// Remove all existing HNC labels.
```
adrianludwin: We're not removing "all existing HNC labels," we're removing "all existing _depth_ labels."

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
```diff
@@ -221,6 +221,68 @@ var _ = Describe("Hierarchy", func() {
 			return val
 		}).Should(Equal("0"))
 	})
+
+	It("should update labels when parent is changed", func() {
+		// Set up key-value pair for non-HNC label
+		keyName := "key"
+		valueName := "value"
```
adrianludwin: Nit: maybe `const keyName = "key"`?

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
```diff
@@ -221,6 +221,68 @@ var _ = Describe("Hierarchy", func() {
...
+		bazHier.Spec.Parent = fooName
+		updateHierarchy(ctx, bazHier)
+
+		// Verify all labels on baz after change parent to foo
+		Eventually(getLabel(ctx, bazName, bazName+depthSuffix)).Should(Equal("0"))
+		Eventually(getLabel(ctx, bazName, fooName+depthSuffix)).Should(Equal("1"))
+		Eventually(getLabel(ctx, bazName, barName+depthSuffix)).Should(Equal(""))
+		Eventually(getLabel(ctx, bazName, keyName)).Should(Equal(valueName))
+	})
+	It("should update labels when parent is removed", func() {
```
adrianludwin: Add newline between tests

incubator/hnc/pkg/controllers/hierarchy_controller.go
adrianludwin: I just just talking about the comment, the code lgtm. I left a more precise comment in my later review.

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
adrianludwin: lgtm, thanks!

incubator/hnc/pkg/controllers/hierarchy_controller.go
frbimo: Ok. Done

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
frbimo: Done

incubator/hnc/pkg/controllers/hierarchy_controller_test.go
frbimo: Done

PR #196, opened 2019-10-28
HNC: changes to make set command function testable

shivi28:
changes:
Separate out set command runnable function and introduce dependency function to make function testable.

Tested: manually run kubectl set commands and verified that they were
reported correctly.

### DIFF
--- a/incubator/hnc/pkg/kubectl/set.go
+++ b/incubator/hnc/pkg/kubectl/set.go
@@ -16,11 +16,22 @@ limitations under the License.
 package kubectl
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
+
+//flagValues struct stores name of namespaces against type of flag passed
+type flagValues struct {
+	root                bool
+	parentString        string
+	requiredChildArray  []string
+	optionalChildString string
+}
 
 var setCmd = &cobra.Command{
 	Use:     "set",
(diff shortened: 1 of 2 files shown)


incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,153 @@ limitations under the License.
...
 )
 
+type flagStatus struct {
+	root                bool
+	parent              bool
+	parentString        string
+	requiredChild       bool
+	requiredChildArray  []string
+	optionalChild       bool
+	optionalChildString string
```
adrianludwin: Why do we need _both_ the `bool` and `[]string` versions? Don't we just need one?

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,153 @@ limitations under the License.
...
+	"errors"
 	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
+type flagStatus struct {
```
adrianludwin: Needs comment

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,153 @@ limitations under the License.
...
-			if oldpnm == "" {
-				fmt.Printf("%s is already a root namespace; unchanged", nnm)
-			} else {
-				hc.Spec.Parent = ""
-				fmt.Printf("Unsetting the parent of %s (was previously %s)\n", nnm, oldpnm)
-				numChanges++
-			}
+		for _, each := range msgs {
+			fmt.Print(each)
 		}
```
adrianludwin: What's the benefit from delaying all these messages from being printed out? I don't think we should be unit testing the messages themselves, is that the only reason?

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,153 @@ limitations under the License.
...
+			parentString:        parentString,
+			requiredChildArray:  requiredChildArray,
+			optionalChildString: optionalChildString,
+		}
 
-		if flags.Changed("root") && flags.Changed("parent") {
-			fmt.Println("Cannot set both --root and --parent at the same time")
+		msgs, err := setCmdFunc(client, flagStatus, nnm)
+		if err != nil {
+			fmt.Printf("%+v", err)
```
adrianludwin: We should just print the string. If err is any kind of complex struct, we don't want to show all the internal details. So this should just be:

`fmt.Println(err.Error())`

incubator/hnc/pkg/kubectl/set.go
shivi28: Yes I am thinking to make our main testable function to return proper messages so that I can test them properly.

incubator/hnc/pkg/kubectl/set.go
shivi28: ok

incubator/hnc/pkg/kubectl/set.go
shivi28: Yes I needed them separately because when `bool` values check the flag.Changed(values) is true or not. I think I can optimise it with one so I will try that approach once.

shivi28:
closing this PR due to some git blunder

incubator/hnc/pkg/kubectl/set.go
adrianludwin: I usually don't like testing human-readable messages because it discourages people from making changes or improvements. Maybe if kubectl-hnc gets much more complex we could do something smart (ie return a struct that's easy to compare, where one value in the struct is a pointer to a Printf format string) but I think that's overkill for now.

PR #198, opened 2019-10-28
HNC: Code refactoring to make set command function testable

shivi28:
Changes:
Separate out set command runnable function and introduce dependency function to make function testable.

Tested: manually run kubectl set commands and verified that they were
reported correctly.

### DIFF
--- a/incubator/hnc/pkg/kubectl/set.go
+++ b/incubator/hnc/pkg/kubectl/set.go
@@ -16,11 +16,22 @@ limitations under the License.
 package kubectl
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
+
+//flagValues struct stores name of namespaces against type of flag passed
+type flagValues struct {
+	root             bool
+	parent           string
+	requiredChildren []string
+	optionalChild    string
+}
 
 var setCmd = &cobra.Command{
 	Use:     "set",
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,148 @@ limitations under the License.
...
-		if flags.Changed("root") {
-			if oldpnm == "" {
-				fmt.Printf("%s is already a root namespace; unchanged", nnm)
-			} else {
-				hc.Spec.Parent = ""
-				fmt.Printf("Unsetting the parent of %s (was previously %s)\n", nnm, oldpnm)
-				numChanges++
-			}
+		for _, each := range msgs {
+			fmt.Print(each)
```
shivi28: @adrianludwin  added a comment : What's the benefit from delaying all these messages from being printed out? I don't think we should be unit testing the messages themselves, is that the only reason?

shivi28: Yes I am thinking to make our main testable function to return proper messages so that I can test them properly.

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,148 @@ limitations under the License.
...
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
+//flagValues struct stores name of namespaces against type of flag passed
+type flagValues struct {
+	root                bool
+	parentString        string
+	requiredChildArray  []string
+	optionalChildString string
```
adrianludwin: Don't append the type of a variable to its name. I'd just say:

```
parent string
requiredChildren []string
optionalChild string
```

incubator/hnc/pkg/kubectl/set.go
adrianludwin: I usually don't like testing human-readable messages because it discourages people from making changes or improvements. Maybe if kubectl-hnc gets much more complex we could do something smart (ie return a struct that's easy to compare, where one value in the struct is a pointer to a Printf format string) but I think that's overkill for now.

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,148 @@ limitations under the License.
...
+
+	if flagValues.root {
+		setRoot(hc, oldpnm, nnm, &msg, &numChanges)
+	}
+
+	if flagValues.parentString != "" {
+		setParent(hc, oldpnm, flagValues.parentString, nnm, &msg, &numChanges)
+	}
+
+	if flagValues.requiredChildArray != nil {
```
adrianludwin: It's preferred to say `if len(foo) != 0`, although in this particular case, `setRequiredChildren` should probably work even if you pass in an empty array.

incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,100 +16,148 @@ limitations under the License.
...
+		setRequiredChild(hc, flagValues.requiredChildArray, nnm, &msg, &numChanges)
```
adrianludwin: This can affect more than one child so `setRequiredChildren`.

rjbez17:
/lgtm
/approve
/ok-to-test
/hold

@adrianludwin release hold when happy

adrianludwin:
/hold cancel

Thanks!

PR #206, opened 2019-10-30
[VC] Fix project path

Fei-Guo:
This change adds "kubernetes-sigs" in the project path.


incubator/virtualcluster/pkg/apis/tenancy/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -15,7 +15,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-// Code generated by deepcopy-gen. DO NOT EDIT.
+// Code generated by main. DO NOT EDIT.
```
resouer: Not sure why this is changed?

Fei-Guo: The original file was generated from controller-gen in vendor directory. We use go mod now so the controller-gen@v.0.1.1 is used which makes the difference.

PR #210, opened 2019-10-30
Forbid users from setting a parent that doesn't yet exist

shivi28:
Changes:
-Added a logic to prevent setting a parent namespace that does not exist.

reported correctly.


incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -140,6 +140,9 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
 		return nil, allow("parent unchanged")
 	}
 
+	if newParent != nil && !newParent.Exists() {
+		return nil, deny(metav1.StatusReasonForbidden, "The requested parent "+hc.Spec.Parent+" does not yet exist")
+	}
 	// Is this change structurally legal? Note that this can "leak" information about the hierarchy
```
adrianludwin: Add newline after new "block" of code

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -140,6 +140,9 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
 		return nil, allow("parent unchanged")
 	}
 
+	if newParent != nil && !newParent.Exists() {
```
adrianludwin: Add comment (very short comment is fine, see line 138)

PR #213, opened 2019-10-31
Logic added to not propagate objects with nonempty finalizer list

### LINKED ISSUE #106: HNC: don't propagate objects with initializers or finalizers
We wipe away the initializers and finalizers on any propagated objects. Instead, we should probably refuse to propagate any object that has an initializer or finalizer. See #201 for a continuation of this feature.

shivi28:
Changes-
Logic added to not propagate objects with nonempty finalizer list

reported correctly.

Fixes #106

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"strings"
 
@@ -228,6 +229,11 @@ func (r *ObjectReconciler) propagate(ctx context.Context, log logr.Logger, inst 
 		}
 	}
 
+	// Object with nonempty finalizer list is not propagated
+	if !isObjectFinalizerEmpty(inst) {
+		return fmt.Errorf("Object %+v has nonempty finalizer list", *inst)
+	}
+
 	// Propagate the object to all destination namespaces.
 	for _, dst := range dests {
 		// Create an in-memory canonical version, and set the new properties.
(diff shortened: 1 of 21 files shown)


incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -228,6 +229,11 @@ func (r *ObjectReconciler) propagate(ctx context.Context, log logr.Logger, inst
 		}
 	}
 
+	// Object with nonempty finalizer list is not propagated
+	if !isObjectFinalizerEmpty(inst) {
+		return fmt.Errorf("Object %+v has nonempty finalizer list", *inst)
```
adrianludwin: FYI - returning an error from a reconciler simply asks the library to reconcile this object again, which isn't what we want. See the `return nil` above after the call to `isExcluded`.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -252,6 +258,14 @@ func (r *ObjectReconciler) propagate(ctx context.Context, log logr.Logger, inst
 	return nil
 }
 
+// isObjectFinalizerEmpty checks if object has nil value in finalizer or not
+func isObjectFinalizerEmpty(inst *unstructured.Unstructured) bool {
+	if inst != nil && inst.GetFinalizers() != nil {
```
adrianludwin: You should never compare a Go slice to `nil` to check to see whether it's empty, because it can be allocated (non-nil) but still have zero objects in it. Instead, say:

`len(inst.GetFinalizers()) != 0`

since `len()` works for both `nil` and non-nil slices.

rjbez17:
This lgtm after @adrianludwin's comments are resolved

shivi28:
closing this pull request because of branch name and git blunder

PR #216, opened 2019-11-01
Package kubectl plugin for krew

### LINKED ISSUE #132: HNC: Package the kubectl plugin for `krew`
(Filed as part of #93)

[Krew](https://github.com/kubernetes-sigs/krew/blob/master/docs/DEVELOPER_GUIDE.md) seems to be the recommended way to distribute kubectl plugins. We should package up `kubectl-hnc` as a `krew` plugin.

Also, right now, `make kubectl` puts the binary in `<hnc>/bin` and `<hnc>/devenv` adds it to the path. This means that any other executables in that directory (currently only `manager`) get added to the path as well, which isn't what we want. So we should update `make kubectl` to put the executable in `<hnc>/bin/kubectl` instead and add that to the path when developing.

/good-first-issue

frbimo:
Tested: Require to have archive `hnc.tar.gz` in the `<hnc>/plugin` , do `sha256sum` on it and copy the checksum value into `<hnc>/plugin/hnc.yaml`. 
Then run `make krew-local`

Fixes #132

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -19,6 +19,15 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Set default version tag for krew build (unless version is set)
+ifeq (,${VERSION})
+VERSION=v0.1.0
+endif
+
+# Get check sum value of krew archive
+KREW_CKSM=$(shell sha256sum bin/kubectl-hnc.tar.gz | cut -d " " -f 1)
+
+
 all: test docker-build
 
 ###################### LOCAL ARTIFACTS #########################
@@ -30,16 +39,37 @@ test: build
 # Builds all binaries (manager and kubectl) and manifests
 build: generate fmt vet manifests
 	go build -o bin/manager ./cmd/manager/main.go
-	go build -o bin/kubectl-hnc ./cmd/kubectl/main.go
+	go build -o bin/kubectl/kubectl-hnc ./cmd/kubectl/main.go
 
+# Clean all binaries (manager and kubectl)
+clean:
+	rm -r bin/kubectl/kubectl-hnc
+	rm -f bin/manager
+	
 # Install kubectl plugin
 kubectl: build
-	go build -o ${GOPATH}/bin/kubectl-hnc ./cmd/kubectl/main.go
+	cp bin/kubectl/kubectl-hnc ${GOPATH}/bin/kubectl-hnc
 	@echo "Installed kubectl-hnc to GOPATH/bin"
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: build
 	go run ./cmd/manager/main.go --novalidation
+
+# Install kubectl plugin locally using krew.
+krew-build: build tar
+	sed -i 's/^\(\s*sha256\s*:\s*\).*/\1"$(KREW_CKSM)"/' \
+		hack/hnc.yaml
+	sed -i 's/^\(\s*version\s*:\s*\).*/\1"$(VERSION)"/' \
+		hack/hnc.yaml
+	kubectl krew install --manifest=hack/hnc.yaml --archive=bin/kubectl-hnc.tar.gz
+
+# Make krew archive and put into /hack
+tar:
+	tar -zcvf bin/kubectl-hnc.tar.gz bin/kubectl
+
+# Uninstall krew
+krew-uninstall: 
+	kubectl krew uninstall hnc
 
 # Generate manifests e.g. CRD, RBAC etc. This can both update the generated
 # files in /config (which should be checked into Git) as well as the kustomized
--- a/incubator/hnc/hack/hnc.yaml
+++ b/incubator/hnc/hack/hnc.yaml
@@ -1,0 +1,23 @@
+# For PoC purpose only.
+# Will be moved to https://github.com/kubernetes-sigs/krew-index after hnc has it's release.
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: hnc
+spec:
+  shortDescription: Implement hierarchy concept on namespaces.
+  description: |
+    Hierarchical Namespace Controller (HNC).
+  version: --- will be populated by make krew-build ---
+  homepage: https://github.com/kubernetes-sigs/multi-tenancy/incubator/hnc
+  platforms:
+    - uri: https://github.com/example/foo/releases/v1.0.tar.gz
+      selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      sha256: --- will be populated by make krew-build ---
+      files:
+        - from: "bin/kubectl/kubectl-hnc"
+          to: "."
+      bin: "./kubectl-hnc"


incubator/hnc/Makefile
```diff
@@ -30,17 +30,32 @@ test: build
...
+	cp bin/kubectl/kubectl-hnc ${GOPATH}/bin/kubectl-hnc
 	@echo "Installed kubectl-hnc to GOPATH/bin"
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: build
 	go run ./cmd/manager/main.go --novalidation
 
+# Install kubectl plugin locally using krew.
+# Require to archive bin/kubectl/kubectl-hnc binary with name hnc.tar.gz, then 
+# do sha256sum plugin/hnc.tar.gz. Copy the checksum into plugin/hnc.yaml
```
adrianludwin: This should all be automated. Can you add a krew-build that does all this?

incubator/hnc/plugin/hnc.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: hnc
+spec:
+  version: "v0.0.1"
```
adrianludwin: We're calling the next release "v0.1" so let's keep them in sync. Alternatively, you can put some bogus word in here like VERSION and then use something like `sed` to replace it from the makefile (where the makefile itself gets the value from an env var, something like `VERSION=v0.1 make krew-build`).

incubator/hnc/plugin/hnc.yaml
```diff
@@ -0,0 +1,33 @@
...
+          arch: amd64
+      uri: https://github.com/kubernetes-sigs/multi-tenancy/releases/v0.0.1.tar.gz
+      sha256: "5c7b88db640b12f0fb4794392e7b9574115a3814e9783966d56144449d459c49"
+      files:
+        - from: "bin/kubectl/kubectl-hnc"
+          to: "."
+      bin: "./kubectl-hnc"
+  shortDescription: namespaces hierarchical administration.
+  homepage: https://github.com/kubernetes-sigs/multi-tenancy
+  caveats: |
```
adrianludwin: This seems more like "documentation" than "caveats?"

incubator/hnc/plugin/hnc.yaml
```diff
@@ -0,0 +1,33 @@
...
+    Usage:
+       kubectl hnc [command]
+
+     HNC documentation: 
+       https://github.com/kubernetes-sigs/multi-tenancy
+
+  description: |
```
adrianludwin: I'm not sure what this adds? Pointing people to the github repo seems enough to me.

incubator/hnc/plugin/hnc.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: hnc
+spec:
+  version: "v0.0.1"
+  platforms:
+    - selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      uri: https://github.com/kubernetes-sigs/multi-tenancy/releases/v0.0.1.tar.gz
```
adrianludwin: Ditto

incubator/hnc/plugin/hnc.yaml
```diff
@@ -0,0 +1,33 @@
...
+      sha256: "5c7b88db640b12f0fb4794392e7b9574115a3814e9783966d56144449d459c49"
```
adrianludwin: Ditto, this should be some dummy value like `SHA256` and be replaced by the _actual_ value during the build process.

incubator/hnc/plugin/hnc.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
```
adrianludwin: Do you happen to know where other projects put their krew YAMLs? I was wondering if it should go in `cmd/kubectl` or maybe `config/kubectl` instead?

incubator/hnc/plugin/hnc.yaml
rjbez17: Any reason we aren't using semver like most other k8s versions.

frbimo: > Any reason we aren't using semver like most other k8s versions.

+1

incubator/hnc/plugin/hnc.yaml
frbimo: I went through first 20 repos on [krew-index](https://github.com/kubernetes-sigs/krew-index) list and found only 3 put krew.yml into their repos (and there is no exact path). AFAIK, once we have releases, we dont need krew yml in the repo instead put it into [krew-index](https://github.com/kubernetes-sigs/krew-index) submitted as a PR. Any changes made e.g version release will revised on other PR.

incubator/hnc/plugin/hnc.yaml
frbimo: Caveats is optional, if you agree, i will remove it later. Require to have shortDescription and description. From [ingress-nginx](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/ingress-nginx.yaml) it seems only a few words. Any thoughts on that?

incubator/hnc/Makefile
frbimo: Ok. Will push new one after solve other nits.

incubator/hnc/plugin/hnc.yaml
adrianludwin: Yup you basically just need a pointer to the github repo and people can read up on it more there.

incubator/hnc/plugin/hnc.yaml
adrianludwin: Ok in that case (if this typically isn't included in the repo), let's put it in `hack/` and add a comment saying that it will eventually move to `krew-index` instead - sgty?

incubator/hnc/Makefile
```diff
@@ -30,17 +39,38 @@ test: build
...
+	go build -o bin/kubectl/kubectl-hnc ./cmd/kubectl/main.go
 
+# Clean all binaries (manager and kubectl)
+clean:
+	rm -r bin/kubectl/kubectl-hnc
+	rm -f bin/manager
+	
 # Install kubectl plugin
 kubectl: build
-	go build -o ${GOPATH}/bin/kubectl-hnc ./cmd/kubectl/main.go
```
frbimo: Any reason need to rebuild binary rather than copy it? I see no purpose to initiate build here.  Therefore i replace it.

incubator/hnc/plugin/hnc.yaml
frbimo: sgtm. Updated

rjbez17:
/ok-to-test
/lgtm
/approve
/hold

@adrianludwin  feel free to release hold when satisfied

incubator/hnc/Makefile
```diff
@@ -30,17 +39,38 @@ test: build
...
+krew-build: build tar
+	sed -i 's/^\(\s*sha256\s*:\s*\).*/\1"$(KREW_CKSM)"/' \
+		hack/hnc.yaml
+	sed -i 's/^\(\s*version\s*:\s*\).*/\1"$(VERSION)"/' \
+		hack/hnc.yaml
+	kubectl krew install --manifest=hack/hnc.yaml --archive=hack/hnc.tar.gz
+
+# Make krew archive and put into /hack
+tar:
+	tar -zcvf hack/hnc.tar.gz bin/kubectl
```
adrianludwin: `hack` is a Git folder, whereas `bin` is not. Can we create the tar file in `bin` instead of `hack`?

Also, `hnc.tar.gz` isn't very descriptive because _everything_ in this (part of the) repo is HNC. So how about `kubectl-hnc.tar.gz` instead?

incubator/hnc/Makefile
```diff
@@ -19,6 +19,15 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Set default version tag for krew build (unless version is set)
+ifeq (,${VERSION})
+VERSION=v0.1.0
+endif
+
+# Get check sum value of krew archive
+KREW_CKSM=$(shell sha256sum hack/hnc.tar.gz | cut -d " " -f 1)
```
adrianludwin: I would have thought that this got evaluated once when the makefile is first loaded, which means that if you call `make krew-build` and the archive is _modified_, then this would return the sha256 of the _old_ version, not the new one. You could test this by deleting `hack/hnc.tar.gz` (or as I suggest below, `bin/kubectl-hnc.tar.gz`) and seeing what this variable evaluates to - I would _guess_ it's an empty string, even if it gets built while `make` is running.

incubator/hnc/Makefile
adrianludwin: sgtm

incubator/hnc/hack/hnc.yaml
```diff
@@ -0,0 +1,23 @@
...
+# Will be moved to https://github.com/kubernetes-sigs/krew-index after hnc has it's release.
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
+kind: Plugin
+metadata:
+  name: hnc
+spec:
+  shortDescription: Implement hierarchy concept on namespaces.
+  description: |
+    Hierarchical Namespace Controller (HNC).
+  version: "v0.1.0"
```
adrianludwin: Suggest: replace `"v0.1.0"` with something obviously illegal like ` --- will be populated by make krew-build --- `

incubator/hnc/hack/hnc.yaml
```diff
@@ -0,0 +1,23 @@
...
+  homepage: https://github.com/kubernetes-sigs/multi-tenancy/incubator/hnc
+  platforms:
+    - uri: https://github.com/kubernetes-sigs/multi-tenancy/releases/v0.1.0.tar.gz
+      selector:
+        matchLabels:
+          os: linux
+          arch: amd64
+      sha256: "f492450c79b36c711e0d21645f874107139bc3d1e0ab23b246eea349b647eb0c"
```
adrianludwin: Ditto. If something isn't "real," make it look as fake as possible otherwise a future developer will scratch her head and say, "where did this come from? What does it mean?"

incubator/hnc/Makefile
frbimo: done

incubator/hnc/Makefile
frbimo: tested and work

incubator/hnc/hack/hnc.yaml
frbimo: i see your point. i change `version`, `platform.uri`, and `sha256 `. ptal

PR #221, opened 2019-11-02
[incubator][vc] add service sync controller

adohe:
/assign @Fei-Guo

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/helper.go
@@ -145,3 +145,10 @@ func MutatePod(namespace string, pod *corev1.Pod) {
 	}
 	pod.Spec.Volumes = volumes
 }
+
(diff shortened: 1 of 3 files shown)


incubator/virtualcluster/pkg/syncer/conversion/helper.go
```diff
@@ -145,3 +145,11 @@ func MutatePod(namespace string, pod *corev1.Pod) {
 	}
 	pod.Spec.Volumes = volumes
 }
+
+func MutateService(newService *corev1.Service) {
+	newService.Spec.Selector = nil
```
Fei-Guo: I thought we should keep the selector from service in tenant master?

adohe: yeap, we can keep the selector, since service selector works in namespace scope.

adohe:
@Fei-Guo ptal~

PR #223, opened 2019-11-03
Logic added to not propagate objects with nonempty finalizer list

### LINKED ISSUE #106: HNC: don't propagate objects with initializers or finalizers

shivi28:
Changes-
isExcluded() function handles the case of object with non-empty finalizer field

Tested-
observed and verified that object with non empty finalizer is not propagated.

Fixes #106

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -335,8 +335,14 @@ func (r *ObjectReconciler) getChildNamespaces(nm string) []string {
 }
 
 // isExcluded returns true if the object shouldn't be handled by the HNC. Eventually, this may be
-// user-configurable, but right now it's only used for Service Account token secrets.
+// user-configurable, but right now it's used for Service Account token secrets and to decide object
+// propagation based on finalizer field.
 func (r *ObjectReconciler) isExcluded(log logr.Logger, inst *unstructured.Unstructured) bool {
+	// Object with nonempty finalizer list is not propagated
+	if len(inst.GetFinalizers()) != 0 {
+		return true
+	}
+
 	switch {
 	case r.GVK.Group == "" && r.GVK.Kind == "Secret":
 		// These are reaped by a builtin K8s controller so there's no point copying them.


incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -351,6 +351,10 @@ func (r *ObjectReconciler) isExcluded(log logr.Logger, inst *unstructured.Unstru
 		}
 		return false
 
+	// Object with nonempty finalizer list is not propagated
```
adrianludwin: Can you also please update the comment on line 338?

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -351,6 +351,10 @@ func (r *ObjectReconciler) isExcluded(log logr.Logger, inst *unstructured.Unstru
...
+	case inst != nil && len(inst.GetFinalizers()) != 0:
```
adrianludwin: Two things:

1. This code will never be called on Secrets - a Secret with finalizers will `return false` on line 352.
2. The rest of the code in this function assumes that `inst != nil` so you can probably assume that here as well.

To solve issue #1, I'd recommend moving this logic _above_ the switch statement, since this issue applies to all types.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -335,8 +335,14 @@ func (r *ObjectReconciler) getChildNamespaces(nm string) []string {
 }
 
 // isExcluded returns true if the object shouldn't be handled by the HNC. Eventually, this may be
-// user-configurable, but right now it's only used for Service Account token secrets.
+// user-configurable, but right now it's used for Service Account token secrets and to decide object
+// propagation based on finalizer field.
 func (r *ObjectReconciler) isExcluded(log logr.Logger, inst *unstructured.Unstructured) bool {
+	// Object with nonempty finalizer list is not propagated
+	if inst != nil && len(inst.GetFinalizers()) != 0 {
```
adrianludwin: Take out `inst != nil` as mentioned?

shivi28: sorry I missed it done now

PR #234, opened 2019-11-06
Update generated CRD manifest file

adrianludwin:
A prior commit (0f10f39) updated the allowed codes and their
documentation but not the generated files. This commit simply includes
the modified files generated by the make process.


rjbez17:
/ok-to-test
/lgtm
/approve
/hold

@adrianludwin  @yiqigao217 

feel free to remove hold

PR #236, opened 2019-11-06
Enable one-line installation and smooth onboarding experience

lafh:
Now users can try out HNC by saying `kubectl apply -f <some single
file>`.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -158,3 +158,10 @@ kind-reset: kind-reboot deploy-prereq
 # Convenience target to deploy specifically for kind
 kind-deploy:
 	CONFIG=kind $(MAKE) deploy
+
+###################### RELEASE ACTIONS #########################
+
+# Generate release YAML that can easily install HNC via `kubectl apply -f`
+release: manifests
+	mkdir -p releases
+	kustomize build config/default > releases/manifests.yaml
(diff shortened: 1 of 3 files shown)


incubator/hnc/Makefile
```diff
@@ -158,3 +158,10 @@ kind-reset: kind-reboot deploy-prereq
 # Convenience target to deploy specifically for kind
 kind-deploy:
 	CONFIG=kind $(MAKE) deploy
+
+###################### RELEASE ACTIONS #########################
+
+# Generate release YAML that can easily install HNC via `kubectl apply -f`
+release: manifests
+	mkdir -p releases
+	kustomize build config/default > releases/manifests.yaml
```
adrianludwin: Why create this new directory? We already have manifests/hnc-manager.yaml, right?

rjbez17: I was under the impression the `manifests` dir was more of kustomize templates and other manifests/templates whereas this is a 'final' built yaml. Maybe `manifests.yaml` is the best name, but I see some advantage to separating 'built/rendered' outside of manifests. That said, I don't really have a strong enough opinion not to change it to use the `manifests` dir.

adrianludwin: No, the templates are in `config`. Only the _fully rendered_ files to in `manifests` - just like `releases`. We should only have one of them.

rjbez17: That’s totally my fault then, I should have caught that. I’ll open a PR to fix it

adrianludwin:
No worries, I spoke to Laetitia about it and she'll look at a way to
combine them.

PR #237, opened 2019-11-06
Ensure hierarchy doesn't violate required children

### LINKED ISSUE #99: HNC: ensure hierarchy doesn't violate required children
Once #90 is complete, we should add a validating webhook to ensure that required children are not reassigned to the wrong parents.

/good-first-issue

yiqigao217:
This validates required child and parent of a required children. It will close #99 .

Tested manually first by setting parent of a required child and setting
different NS as required child. Then unit tests are added.

### DIFF
--- a/incubator/hnc/pkg/validators/hierarchy.go
+++ b/incubator/hnc/pkg/validators/hierarchy.go
@@ -135,14 +135,29 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
 	curParent := ns.Parent()
 	newParent := v.Forest.Get(hc.Spec.Parent)
 
-	// No change -> no problem
+	resp := v.checkParent(ns, curParent, newParent)
+	if !resp.Allowed {
+		return nil, resp
+	}
+
+	resp = v.checkRequiredChildren(ns, hc.Spec.RequiredChildren)
+	if !resp.Allowed {
+		return nil, resp
+	}
+
+	// The structure looks good. Get the list of namespaces we need authz checks on.
+	return v.needAuthzOn(curParent, newParent), allow("")
+}
+
+// checkParent validates if the parent is legal based on the current in-memory state of the forest.
+func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admission.Response {
 	if curParent == newParent {
-		return nil, allow("parent unchanged")
+		return allow("parent unchanged")
 	}
 
 	// non existence of parent namespace -> not allowed
 	if newParent != nil && !newParent.Exists() {
-		return nil, deny(metav1.StatusReasonForbidden, "The requested parent "+hc.Spec.Parent+" does not yet exist")
+		return deny(metav1.StatusReasonForbidden, "The requested parent "+newParent.Name()+" does not exist")
 	}
 
 	// Is this change structurally legal? Note that this can "leak" information about the hierarchy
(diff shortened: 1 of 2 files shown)


incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -151,13 +166,40 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
 	// parent conflicts with something in the _existing_ hierarchy.
 	if reason := ns.CanSetParent(newParent); reason != "" {
-		return nil, deny(metav1.StatusReasonConflict, "Illegal parent: "+reason)
+		return deny(metav1.StatusReasonConflict, "Illegal parent: "+reason)
 	}
 
-	// TODO: requiredChild check: https://github.com/kubernetes-sigs/multi-tenancy/issues/99
+	// Prevent changing parent of a required child
+	if ns.RequiredChildOf != "" && ns.RequiredChildOf != newParent.Name() {
+		reason := fmt.Sprintf("Cannot set the parent of %q to %q because it's required child of %q", ns.Name(), newParent.Name(), ns.RequiredChildOf)
```
adrianludwin: because it's **a** required child

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -151,13 +166,40 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
+			return deny(metav1.StatusReasonConflict, "Illegal requiredChild: "+reason)
+		}
+		if cns.Parent() != nil && cns.Parent() != ns {
+			reason := fmt.Sprintf("Cannot set %q as the required child of %q because it's already a child of %q", cns.Name(), ns.Name(), cns.Parent().Name())
+			return deny(metav1.StatusReasonConflict, "Illegal requiredChild: "+reason)
+		}
+		if cns.Parent() == nil {
+			reason := fmt.Sprintf("Cannot set %q as the required child of %q because it's a root", cns.Name(), ns.Name())
+			return deny(metav1.StatusReasonConflict, "Illegal requiredChild: "+reason)
+		}
```
adrianludwin: Should we replace all of these with "because it already exists and is not a child of X"?

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -151,13 +166,40 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
-	return v.needAuthzOn(curParent, newParent), allow("")
+	return allow("")
+}
+
+func (v *Hierarchy) checkRequiredChildren(ns *forest.Namespace, requiredChildren []string) admission.Response {
+	for _, child := range requiredChildren {
+		cns := v.Forest.Get(child)
+		// Early exit if we find the newly-added requiredChild
+		if !cns.Exists() {
+			return allow("RequiredChild namesapce is just created")
```
adrianludwin: s/namesapce/namespace/

yiqigao217: Do you mean specify the name of the RequiredChild here?

adrianludwin:
No just that there's a typo 🙂

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: No just that there was a typo :)

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -151,13 +166,32 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
-	return v.needAuthzOn(curParent, newParent), allow("")
+	return allow("")
+}
+
+func (v *Hierarchy) checkRequiredChildren(ns *forest.Namespace, requiredChildren []string) admission.Response {
+	for _, child := range requiredChildren {
+		cns := v.Forest.Get(child)
+		// Early exit if we find the newly-added requiredChild
+		if !cns.Exists() {
+			return allow("RequiredChild namespace is just created")
```
adrianludwin: Whoops this should just be a `continue`, we don't want to short-circuit the entire list of required children!

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -151,13 +166,32 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
+		}
+		if cns.Parent() != ns && (cns.Parent() != nil || cns.RequiredChildOf != ns.Name()) {
```
adrianludwin: This might be easier to read in the reverse:

```
// If this is already a child, or is about to be, no problem
if cns.Parent() == ns || (cns.Parent() == nil && cns.RequiredChildOf == ns.Name()) {
  continue
}
```

incubator/hnc/pkg/validators/hierarchy.go
yiqigao217: Is it for in casing the concurrency issue when more than ONE requiredChild is added to the list?

adrianludwin: I was just thinking "we should always make sure the entire list is valid" - that's always the way to think when writing a validator (or reconciler). But a concrete problem here might be if you manually edited the hc.yaml to add two children, one valid and one not.

yiqigao217: sg.

Issue #238, opened 2019-11-07
HNC: Canonical function will fail for tree labels 


lafh:
The current approach only catches HNC-related labels/annotations with prefix `api.MetaGroup`, so the tree labels (i.e. `namespace.tree.hnc.x-k8s.io/depth` doesn't have `api.MetaGroup` as prefix) will remain in the canocial copy of an object, which is not the expected behaviour.

PR #239, opened 2019-11-07
HNC: forbid propagated objects from being modified

### LINKED ISSUE #89: HNC: forbid propagated objects from being modified

lafh:
Manual test:
org1
|
- team1

(1) Create a secret in org1. Modification to the origin secret is
allowed, but modification to the secret in team1 is not allowed.
(2) Create ResourceQuota in org1. Create a pod in org1 and see
ResourceQuota has been partially used, but the propagated ResourceQuota
for team1 is not affected.

P.S. #238 will be resolved in another PR. This PR will focus on the refactoring the canonical function.

/resolve #89

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -18,7 +18,6 @@ package controllers
 import (
 	"context"
 	"reflect"
-	"strings"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


adrianludwin:
/lgtm
/approve
/hold

@srampal can you ptal and remove the hold if you're happy?

PR #240, opened 2019-11-08
fix Canonical function

### LINKED ISSUE #238: HNC: Canonical function will fail for tree labels

lafh:
Currently the tree label will stay in the canocial version of the object

/resolve #238

### DIFF
--- a/incubator/hnc/pkg/object/object.go
+++ b/incubator/hnc/pkg/object/object.go
@@ -8,6 +8,16 @@ import (
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
+// metaPrefix returns the prefix (if any) of a label key.
+// Reference https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set
+func metaPrefix(s string) string {
+	p := strings.Split(s, "/")
+	if len(p) == 1 {
+		return ""
+	}
+	return p[0]
+}
+
 // Canonical returns a canonicalized version of the object - that is, one that has the same name,
 // spec and non-HNC labels and annotations, but with the status and all other metadata cleared
 // (including, notably, the namespace). The resulting object is suitable to be copied into a new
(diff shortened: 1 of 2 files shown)


lafh:
assign @adrianludwin

incubator/hnc/pkg/object/object.go
```diff
@@ -26,7 +26,7 @@ func Canonical(inst *unstructured.Unstructured) *unstructured.Unstructured {
 	// Non-HNC annotations:
 	newAnnots := map[string]string{}
 	for k, v := range inst.GetAnnotations() {
-		if !strings.HasPrefix(k, api.MetaGroup) {
+		if !strings.Contains(k, api.MetaGroup) {
```
adrianludwin: I don't think this is the right logic. While it's unlikely, `hnc.x-k8s.io.foo/bar` is actually fine, and shouldn't be stripped out. Maybe add this as a test case too?

lafh: sg.

incubator/hnc/pkg/object/object.go
```diff
@@ -35,7 +36,9 @@ func Canonical(inst *unstructured.Unstructured) *unstructured.Unstructured {
 	// Non-HNC labels:
 	newLabels := map[string]string{}
 	for k, v := range inst.GetLabels() {
-		if !strings.HasPrefix(k, api.MetaGroup) {
+		if !strings.HasPrefix(k, api.MetaGroup) &&
```
adrianludwin: This will break if we add any other prefixes like `tree`. How about something like:

```
if !strings.HasSuffix(metaPrefix(k), api.MetaGroup) { ... }

...

func metaPrefix(s string) string {
  p := strings.Split(s, "/")
  if len(p) == 1 {
    return ""
  }
  return p[0]
}
```

adrianludwin: Reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

Issue #43, opened 2019-08-22
Get official container registry from #wg-k8s-infra


rjbez17:
We need to move from docker hub to an official k8s sig registry. Info from @kfox1111  here:

https://github.com/kubernetes-sigs/multi-tenancy/pull/42#issuecomment-524025857

tashimi:
From @kfox1111: 

> Looks like #wg-k8s-infra can hook us up with gce.
> 
> "first step is to create a google group for the access control for the repo (please add one in https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml) and then an entry in the script for creating the staging repo (please update https://github.com/kubernetes/k8s.io/blob/master/infra/gcp/ensure-staging-storage.sh#L41). Please see older PR(s) that were merged when in doubt."

@kfox1111 : @Fei-Guo indicated that GCE is not a good tool for non-US based developers, which is why he was asking for a docker hub account. Can we discuss this before we go ahead and get this account?

rjbez17:
Aren't all kubernetes core images pushed to k8s.grc.io? Do they have a mirror somewhere? I have to image this issue has been resolved by other sigs? I'll reach out to #wg-k8s-infra to see what our options are.

kfox1111:
Hmm... part of the setup of the repo mentions the repo syncing into:
- name: asia.gcr.io/k8s-artifacts-prod/multitenancy

Is that accessible @Fei-Guo?

Fei-Guo:
> Hmm... part of the setup of the repo mentions the repo syncing into:
> 
> * name: asia.gcr.io/k8s-artifacts-prod/multitenancy
> 
> Is that accessible @Fei-Guo?

gcr.io is completely blocked in china as far as I know.

kfox1111:
I don't know how accurate it is, but:
https://viewdns.info/chinesefirewall/?domain=asia.gcr.io

Says it is accessible. Have you tried it?

rjbez17:
I reached out to #wg-k8s-infra to confirm whether or not asia.gcr.io is accessible from china. If they don't know we'll have to find someone within the community to test it for us.

kfox1111:
https://github.com/kubernetes/k8s.io/pull/343 Just merged.

adrianludwin:
Is this done?

rjbez17:
Completed

PR #243, opened 2019-11-11
Config Cloud Build YAML file for HNC releases

### LINKED ISSUE #228: HNC: build release candidates using Cloud Build
This is the approach taken by kubebuilder; let's try it out too.

/assign @lafh

lafh:
partially resolve #228 

Need to manually upload images to Docker Hub later. 

/assign @adrianludwin

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -1,3 +1,6 @@
+# Adding .PHONY to hold command shortcuts.
+.PHONY: release
+
 # If CONFIG is `kind`, various defaults will be optimized for deploying locally to Kind
 CONFIG ?= "default"
 # Image URL to use all building/pushing image targets
@@ -160,8 +163,6 @@ kind-deploy:
 	CONFIG=kind $(MAKE) deploy
 
 ###################### RELEASE ACTIONS #########################
-
-# Generate release YAML that can easily install HNC via `kubectl apply -f`
+# Build the container image by Cloud Build and build YAMLs locally
 release: manifests
-	mkdir -p releases
-	kustomize build config/default > releases/manifests.yaml
+	gcloud builds submit --config cloudbuild.yaml .
--- a/incubator/hnc/README.md
+++ b/incubator/hnc/README.md
@@ -129,3 +129,25 @@ milestones are defined:
 
 Non-coding tasks are also tracked in the [HNC
 project](https://github.com/kubernetes-sigs/multi-tenancy/projects/4).
+
+## Releasing
+
+[comment]: # (TODO: pull the code directly from Github)
+1.Check out the code to the version that you want to release and run the 
+following command in `incubator/hnc`.
+```bash
+make release
+```
+It will use Cloud Build (on your GCP account)
+to build the container image remotely while building YAMLs locally.
+
+2.Make sure you have [permisison]((https://github.com/kubernetes/k8s.io/blob/master/groups/groups.yaml#L566)) to access `k8s-staging-multitenancy`'s GCR.
+Then upload the container image via the following commands:
+```bash
+docker pull gcr.io/$PROJECT_ID/hnc/controller
+docker tag gcr.io/$PROJECT_ID/hnc/controller gcr.io/k8s-staging-multitenancy/hnc/controller:$VERSION_TAG
+docker push gcr.io/k8s-staging-multitenancy/hnc/controller:$VERSION_TAG
+```
+
+3.Upload generated YAMLs (located in `incubator/hnc/manifests/hnc-manager.yaml`) 
+to Github. See instructions [here](https://help.github.com/en/github/administering-a-repository/creating-releases).
--- a/incubator/hnc/cloudbuild.yaml
+++ b/incubator/hnc/cloudbuild.yaml
@@ -1,0 +1,7 @@
+steps:
+# TODO: pull the code directly from Github
+# Build release container image
+- name: gcr.io/cloud-builders/docker
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/hnc/controller', '.']
+
+images: ['gcr.io/$PROJECT_ID/hnc/controller']


incubator/hnc/cloudbuild.yaml
```diff
@@ -0,0 +1,16 @@
...
+- name: google/cloud-sdk
+  entrypoint: "bash"
+  args:
+  - -c
+  - kubectl kustomize config/default/ > release.yaml
+
+images: ['gcr.io/$PROJECT_ID/hnc/controller']
+artifacts:
+  objects:
+    location: 'gs://artifacts.$PROJECT_ID.appspot.com/hnc/'
```
adrianludwin: What's appspot.com? I thought that was an AppEngine thing?

incubator/hnc/cloudbuild.yaml
```diff
@@ -0,0 +1,16 @@
+steps:
```
adrianludwin: Is there a relatively easy way to add a step to pull the code directly from Github? Otherwise, I assume what happens is that you must check out the code locally on your workstation and invoke `make release` from there? Which is fine as a short-term solution but if it's easy to grab the source directly, that's even better.

If this is hard, can you add a TODO comment in both the makefile and the cloudbuild.yaml?

incubator/hnc/cloudbuild.yaml
lafh: It's just the a Default Cloud Storage Bucket in the App Engine. It is in the form <project-id>.appspot.com.
Reference
https://cloud.google.com/appengine/docs/standard/python/googlecloudstorageclient/setting-up-cloud-storage

incubator/hnc/Makefile
```diff
@@ -160,8 +163,6 @@ kind-deploy:
...
 
 ###################### RELEASE ACTIONS #########################
-
-# Generate release YAML that can easily install HNC via `kubectl apply -f`
-release: manifests
-	mkdir -p releases
-	kustomize build config/default > releases/manifests.yaml
+# Build the release artifacts by Cloud Build
+release:
+	gcloud builds submit --config cloudbuild.yaml .
```
rjbez17: Who 'owns' the official gcloud account for this? If I run this, it seems it'll use the current gcloud account I'm logged in on which is different from yours. Does this tie into the official k8s sig way of releasing (if there is one?).

incubator/hnc/cloudbuild.yaml
lafh: There's a (relatively more common) way to use Cloud Build to build from Github repo for each commit/PR, but it doesn't seem common to manually trigger only one release at a certain point of time. I will add the TODO and leave it for a later discussion.

incubator/hnc/Makefile
lafh: Thanks for pointing it out!

- I am trying to follow https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md here
- Currently I am playing around with it on a personal GCP project. It doesn't matter which `gcloud` account is used, as long as that it's built by Cloud Build (a trusted service that is transparent). Later manual work is required to download those artifacts and to upload the images to MTWG's Docker Hub and Github release page for YAMLs. (This is also included in the updated README.)

Hopefully it answers your questions.

incubator/hnc/cloudbuild.yaml
adrianludwin: Can't we use GCS separately from App Engine?

incubator/hnc/README.md
```diff
@@ -129,3 +129,12 @@ milestones are defined:
...
 project](https://github.com/kubernetes-sigs/multi-tenancy/projects/4).
+
+## Releasing
+
+[comment]: # (TODO: pull the code directly from Github)
+
+Check out the code to the version that you want to release and invoke
+`make release` in `incubator/hnc`. It will use Cloud Build (on your GCP account)
+to build the container image remotely while building YAMLs locally (in `manifests/hnc-manager.yaml`).
+The container image will be upload to MTWG's Docker Hub and release YAMLs should be hosted on Github.
```
adrianludwin: Can you explain this a bit more? At the end of the Cloud Build process, the image is in GCR, not Docker Hub. Can you add the _specific_ instructions about how to get it to the right place? Eg, something like:

- Make sure you have write access to <docker hub ID>
- Pull the image via `docker pull gcr.io/YOUR_PROJECT_ID/etc`
- Push the image via `docker push ETC`

You should basically have instructions I can cut-and-paste.

rjbez17:
@lafh and @adrianludwin  I think this PR will allow us to proceed a little easier: https://github.com/kubernetes/k8s.io/pull/467

adrianludwin:
/approve
/lgtm
/hold
/assign @rjbez17 

Ryan, does this look reasonable as a first cut of the release process to you?

incubator/hnc/README.md
```diff
@@ -129,3 +129,25 @@ milestones are defined:
...
+[comment]: # (TODO: pull the code directly from Github)
+1.Check out the code to the version that you want to release and run the 
+following command in `incubator/hnc`.
+```bash
+make release
+```
+It will use Cloud Build (on your GCP account)
+to build the container image remotely while building YAMLs locally.
+
+2.Make sure you have logged into your Docker ID which has the access to MTWG [Docker Hub](https://cloud.docker.com/u/multitenancy/repository/docker/multitenancy/tenant).
```
rjbez17: With the access granted, we should point to gcr.io/k8s-staging-multitenancy

lafh: As discussed in the Github [issue](https://github.com/kubernetes-sigs/multi-tenancy/issues/43), gcr.io is blocked in China, so we decided to use the Docker Hub repo owned by MTWG for better availability. This is the link to that Docker Hub.

lafh:
@rjbez17  thoughts?

incubator/hnc/README.md
rjbez17: We were able to confirm that other asia.gcr.io were accessible and we'd prefer to have the authentication handled for us via k8s hooks. I took a look at the other sigs and almost all are using gcr.io/k8s-staging-* so it'd seem like an anti-pattern not to IMO.

@adrianludwin thoughts?

adrianludwin: Given that asia.gcr.io is available in China, yes, we should go with that. Thanks Ryan.

PR #262, opened 2019-11-18
[incubator][VC] add work queue for upward syncer

### LINKED ISSUE #251: VC: Add requeue mechanism if backpopulate fails
In case backpopulate fails for any synced object, there is no retry now. This may lead to inconsistency between super master and tenant master. We should retry whenever we found mismatch may occur,

zhuangqh:
distinguish StartUWS and StartDWS method in controller interface

fixes #251 

notes for reviewer:
1. refactor controller interface
```go
type Controller interface {
	listener.ClusterChangeListener
	StartUWS(stopCh <-chan struct{}) error
	StartDWS(stopCh <-chan struct{}) error
}
```
Actually, every controller should implement this interface. And we register the listener when adding the controller to controller manager.

**How about start UWS and DWS in a method named `Start(stopCh <-chan struct{}) error`.**
 Both looks good for me. Distinguish these two method will more clear to control the concurrent problem.

2. the workqueue code for each controller is mostly the same.
Yes, the work queue part is mostly the same. But we may have different business logic for each controller later. I left this part the same now.

3. this PR just add the work queue mechanism. I do not want to correct any business logic

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/controllers/node/controller.go
+++ b/incubator/virtualcluster/pkg/syncer/controllers/node/controller.go
@@ -20,18 +20,15 @@ import (
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	coreinformers "k8s.io/client-go/informers/core/v1"
-	clientset "k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
-	restclient "k8s.io/client-go/rest"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/syncer/cluster"
 	sc "github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/syncer/controller"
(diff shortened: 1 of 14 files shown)


incubator/virtualcluster/pkg/syncer/controllers/node/controller.go
```diff
@@ -42,6 +39,11 @@ type controller struct {
 	nodeNameToCluster          map[string]map[string]struct{}
 	nodeClient                 v1core.NodesGetter
 	multiClusterNodeController *sc.MultiClusterController
+
+	workers    int
```
zhuangqh: TODO: make this optional. Is it necessary?

incubator/virtualcluster/pkg/syncer/controllers/pod/uws.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	<-stopCh
+	klog.V(1).Infof("shutting down")
+
+	return nil
+}
+
+// worker runs a worker thread that just dequeues items, processes them, and marks them done.
+// It enforces that the syncHandler is never invoked concurrently with the same key.
+func (c *controller) worker() {
```
Fei-Guo: Can we rename worker to run() for all uws?

incubator/virtualcluster/pkg/syncer/controllers/pod/controller.go
```diff
@@ -52,7 +56,9 @@ func Register(
 	controllerManager *manager.ControllerManager,
 ) {
 	c := &controller{
-		client: client,
+		client:  client,
+		queue:   workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "super_master_pod"),
+		workers: 1,
```
Fei-Guo: If we use workers, let us make default to a value > 1 . 
check http://openkruise.io/en-us/blog/blog2.html

PR #269, opened 2019-11-19
[incubator][VC] add work queue for virtual cluster controller

zhuangqh:
use `<namespace>-<name>` as the unique if for cluster

everything still works fine 😄

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/syncer.go
+++ b/incubator/virtualcluster/pkg/syncer/syncer.go
@@ -17,34 +17,52 @@ limitations under the License.
 package syncer
 
 import (
+	"fmt"
+	"sync"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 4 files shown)


incubator/virtualcluster/pkg/syncer/syncer.go
```diff
@@ -78,122 +108,164 @@ func New(
...
 	}
+
+	go func() {
+		if err = innerCluster.Start(s.stopChan); err != nil {
+			s.removeCluster(key)
+		}
+	}()
+
+	s.Lock()
+	s.clusterSet.Insert(key)
```
Fei-Guo: This looks problematic, if inclustercluser.start fails at line 262, should we insert the key? 
We probably should use error chan (check the innersCluster.Start method) to check the go routine error.

incubator/virtualcluster/pkg/syncer/syncer.go
```diff
@@ -17,34 +17,52 @@ limitations under the License.
...
+	sync.Mutex
 	secretClient      v1core.SecretsGetter
 	controllerManager *manager.ControllerManager
+	// lister that can list virtual clusters from a shared cache
+	lister vclisters.VirtualclusterLister
+	// returns true when the namespace cache is ready
+	virtualClusterSynced cache.InformerSynced
+	// virtual cluster that have been queued up for processing by workers
+	queue   workqueue.RateLimitingInterface
+	workers int
```
Fei-Guo: For this controller, I think we should force to do sequential reconcile as of now.

type MultiClusterController struct {
	// name is used to uniquely identify a Controller in tracing, logging and monitoring.  Name is required.
	name string

	// objectType is the type of object to watch.  e.g. &v1.Pod{}
	objectType runtime.Object

	// clusters is the internal cluster set this controller watches.
	clusters map[Cluster]struct{}

	Options
}
is not thread safe.

Issue #195, opened 2019-10-28
HNC: restructure object controller to allow explicit enqueuing


adrianludwin:
The Object controller doesn't allow for manual enqueing; either we can try to update every object of a certain type as a single, serial operation (`SyncNamespace`); if any _one_ fails, we need to redo _all_ of them. By contrast, the HierarchyConfig reconciler allows for individual enqueues and retries of individual failures.

Once we can enqueue a reconcile for a specific object, this unlocks many other improvements such as enqueuing a reconcile for objects that don't yet exist as a batch operation.

yiqigao217:
@rjbez17  Since the new object controller is implemented and tested, can we switch it as the default? We may want to keep the old controller code for the coming performance test. After the performance test is done, we can remove the code. WDYT?

I also manually tested the new object controller on Kind. It behaved just as we expected, e.g. manually-modified propagated objects got overwritten by the controller and got propagated to the newly-added children. In addition, everything worked fine when the controller is manually restarted.

rjbez17:
Yeah I'm all for it. Since we are still pre-aplpha I don't see any reason to do a typical deprecation policy.

rjbez17:
I would be interested in some metrics on the performance differences between the old vs new. I'm also happy to help with that testing, just let me know.

PR #270, opened 2019-11-19
Set flags to enable new object controller

yiqigao217:
This is setting flags to enable/disable the new object controller. Part of #195 

Tested manually by disabling the object propagation in the new object
controller and see if the propagation is enabled/disabled by
unsetting/setting the flag "--enable-new-object-controller" in
manager_auth_proxy_patch.yaml.

The test flag is tested by "make test NOC=1"

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -24,6 +24,11 @@ ifeq (,${VERSION})
 VERSION=v0.1.0
 endif
 
+# Set default object controller to the old one
+ifeq (,${NOC})
+NOC=0
+endif
+
 # Get check sum value of krew archive
 KREW_CKSM=$(shell sha256sum bin/kubectl-hnc.tar.gz | cut -d " " -f 1)
 
@@ -34,7 +39,12 @@ all: test docker-build
 
 # Run tests
 test: build
-	go test ./api/... ./cmd/... ./pkg/... -coverprofile cover.out
+	# run tests in all directories except ./pkg/controllers/
+	go test ./api/... ./cmd/... `go list ./pkg/... | grep -v controllers` -coverprofile cover.out
+	# separately run tests in ./pkg/controllers/
+	# by default it will use the old object controller
+	# with NOC=1 it will use the new object controller
+	go test ./pkg/controllers/... -enable-new-object-controller=${NOC} -coverprofile cover.out
 
 # Builds all binaries (manager and kubectl) and manifests
 build: generate fmt vet manifests
--- a/incubator/hnc/pkg/controllers/setup.go
+++ b/incubator/hnc/pkg/controllers/setup.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
@@ -13,17 +14,12 @@ import (
 // Create creates all reconcilers.
 //
 // This function is called both from main.go as well as from the integ tests.
-func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int) error {
+func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int, newObjectController bool) error {
 	// Create all object reconcillers
 	objReconcilers := []NamespaceSyncer{}
 	for _, gvk := range config.GVKs {
-		or := &ObjectReconciler{
-			Client: mgr.GetClient(),
-			Log:    ctrl.Log.WithName("controllers").WithName(gvk.Kind),
-			Forest: f,
-			GVK:    gvk,
-		}
-		if err := or.SetupWithManager(mgr); err != nil {
+		or, err := createObjectReconciler(newObjectController, mgr, f, gvk)
+		if err != nil {
 			return fmt.Errorf("cannot create %v controller: %s", gvk, err.Error())
 		}
 		objReconcilers = append(objReconcilers, or)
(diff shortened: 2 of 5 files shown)


incubator/hnc/pkg/controllers/suite_test.go
```diff
@@ -41,12 +42,17 @@ import (
...
-	k8sManager ctrl.Manager
+	cfg                 *rest.Config
+	k8sClient           client.Client
+	testEnv             *envtest.Environment
+	k8sManager          ctrl.Manager
+	newObjectController bool
 )
 
+func init() {
+	flag.BoolVar(&newObjectController, "enable-new-object-controller", false, "Enables new object controller.")
```
adrianludwin: Can you also add something to the makefile so I can say:

```
NOC=1 make test
```

or something like that? ("NOC" = "New Object Controller")

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -13,20 +13,35 @@ import (
...
+func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int, newObjectController bool) error {
 	// Create all object reconcillers
 	objReconcilers := []NamespaceSyncer{}
-	for _, gvk := range config.GVKs {
-		or := &ObjectReconciler{
-			Client: mgr.GetClient(),
-			Log:    ctrl.Log.WithName("controllers").WithName(gvk.Kind),
-			Forest: f,
-			GVK:    gvk,
+	if newObjectController {
```
adrianludwin: I feel like this might be easier to read if it were something like:

```
or := createObjectReconciler(newObjectController, mgr, f, gvk)
```

because it's not immediately clear from reading this that the _only_ thing that changes in these two code blocks is the type of the reconciler.

yiqigao217:
@adrianludwin The 2 changes are done and tested. Please take another look, thanks!

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -3,30 +3,22 @@ package controllers
...
-			Client: mgr.GetClient(),
-			Log:    ctrl.Log.WithName("controllers").WithName(gvk.Kind),
-			Forest: f,
-			GVK:    gvk,
-		}
-		if err := or.SetupWithManager(mgr); err != nil {
-			return fmt.Errorf("cannot create %v controller: %s", gvk, err.Error())
-		}
-		objReconcilers = append(objReconcilers, or)
+	objReconcilers, err := createObjectReconcilers(newObjectController, mgr, f, config.GVKs)
```
adrianludwin: This isn't what I meant - the loop is exactly the same in both cases so that should be in the top level function. It's only creating an _individual_ reconciler that's different so that's the part should be in the new function, or at least that's what I though - do you disagree?

incubator/hnc/Makefile
```diff
@@ -34,7 +39,8 @@ all: test docker-build
 
 # Run tests
 test: build
-	go test ./api/... ./cmd/... ./pkg/... -coverprofile cover.out
+	go test ./api/... ./cmd/... `go list ./pkg/... | grep -v controllers` -coverprofile cover.out
```
adrianludwin: Cool :) Maybe document (comment) this logic a bit?

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -82,6 +82,7 @@ type HierarchyReconciler struct {
 // namespace).
 type NamespaceSyncer interface {
 	SyncNamespace(context.Context, logr.Logger, string) error
+	SetupWithManager(manager ctrl.Manager) error
```
yiqigao217: Is it ok to put `SetupWithManager()` in the `NamespaceSyncer` interface too? It's needed because the return type of `createObjectReconciler()` is `NamesapceSyncer`.

yiqigao217: It can be removed after the new object controller replaces the old one.

incubator/hnc/pkg/controllers/setup.go
yiqigao217: Right, agreed!

incubator/hnc/pkg/controllers/hierarchy_controller.go
adrianludwin: Technically, you should probably have a new interface called CommonObjectReconciler that _composes_ NamespaceSyncer and also adds SetupWithManager. Or you could duplicate the call to SetupWithManager twice in the function, that's not _too_ bad. 

See https://golang.org/pkg/io/#ReadWriter for how `ReadWriter` composes the `Reader` and `Writer` interfaces.

yiqigao217: Thanks for explaining and sharing the ReadWriter example! I added the `SetupWithManager()` calls into the new func for now, since I will change the new object controller `SetupWithManager()` to take an additional parameter of `maxReconciles` in the next commit.

adrianludwin: +1

incubator/hnc/pkg/controllers/suite_test.go
rjbez17: I assume this is a temporary item until we fulling migrate everything to the new controller? Can we add some inline comments indicating the roadmap and some sort of target of when we can remove this as a boolean? The ticket or code don't really indicate the long term plan for this.

incubator/hnc/Makefile
```diff
@@ -34,7 +39,10 @@ all: test docker-build
 
 # Run tests
 test: build
-	go test ./api/... ./cmd/... ./pkg/... -coverprofile cover.out
+	# run tests in all directories except ./pkg/controllers/
+	go test ./api/... ./cmd/... `go list ./pkg/... | grep -v controllers` -coverprofile cover.out
+	# separately run tests in ./pkg/controllers/ with NOC flag to enable new object controller
```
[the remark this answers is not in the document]
adrianludwin: By default everything will still run with the old controller, we only test the new controller if you say `NOC=1 make test`. But yes, a better comment might make that more obvious.

incubator/hnc/pkg/controllers/suite_test.go
adrianludwin: Right - @yiqigao217 can you mention the GH issue?

yiqigao217: The GH issue is added in the comments.

incubator/hnc/Makefile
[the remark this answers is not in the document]
yiqigao217: The comment is updated :)

PR #280, opened 2019-11-27
[incubator][VC] reconcile pod deletion event

zhuangqh:
1. delete the pod in super master with grace period after observing
delete timestamp of virtual pod
2. in uws, delete virtual pod with 0 grace period second after the pod
in super master has been deleted
3. virtual pod actually deleted by tenant controller

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/controllers/pod/uws.go
+++ b/incubator/virtualcluster/pkg/syncer/controllers/pod/uws.go
@@ -27,12 +27,10 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
-	restclient "k8s.io/client-go/rest"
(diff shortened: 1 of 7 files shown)


incubator/virtualcluster/pkg/syncer/controllers/pod/uws.go
```diff
@@ -82,43 +90,60 @@ func (c *controller) processNextWorkItem() bool {
...
+		return fmt.Errorf("could not find pod %s/%s pod in controller cache %v", podInfo.vNamespace, pod.Name, err)
 	}
-	var client *clientset.Clientset
-	innerCluster := c.multiClusterPodController.GetCluster(clusterName)
-	if innerCluster == nil {
-		// virtual cluster is gone.
+	vPod := vPodObj.(*v1.Pod)
+
+	// pod is in deleting.
+	if pod.DeletionTimestamp != nil {
```
zhuangqh: if `DeletionTimestamp != nil`, `DeletionGracePeriodSeconds` must be non-nil

PR #299, opened 2019-12-06
Store CritAncestor condition in the forest

yiqigao217:
This is part of #282 and #195 

CritAncestor condition is set when any ancestor has local critical
conditions. Currently, it's only set in the HierarchyConfiguration
object. This commit sets the CritAncestor condition in the forest first
with ancestor namespace as the key. This change enables object
controller to just look in the forest to find all critical conditions in
the namespaces, so that it can stop propagation if any condition
detected.

Tested with "make test" and "make test NOC=1" and all tests passed.

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -157,8 +157,8 @@ func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Nam
 	}
 
 	// Clear locally-set conditions in the forest so we can set them to the latest.
-	hadCrit := ns.HasCritCondition()
-	ns.ClearConditions(forest.Local)
+	hadCrit := ns.HasLocalCritCondition()
+	ns.ClearCritConditions()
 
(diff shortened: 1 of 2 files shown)


adrianludwin:
I'm rethinking this idea now - why do we need to store the ancestors' status? Can't we just do a quick lookup on the ancestors when we need to know if any of them have a problem?

adrianludwin:
... I know I thought this was needed when we spoke yesterday but now I kind of forget why I thought that!

yiqigao217:
> I'm rethinking this idea now - why do we need to store the ancestors' status? Can't we just do a quick lookup on the ancestors when we need to know if any of them have a problem?

Yes, we can do that in the object controller looking in the forest for 1) its own local conditions; 2) ancestors' local conditions. ***


> ... I know I thought this was needed when we spoke yesterday but now I kind of forget why I thought that!

We thought it's needed because we wanted to avoid the API call to get the CritAncestor condition which is not set in the forest, but since we can look up the forest to find ancestor local conditions, we can do that instead of this PR. 

I will close this PR and implement the *** above. WDYT?

PR #301, opened 2019-12-09
Check Crit condition in object controller

yiqigao217:
This is part of #282 and #195

This commit stops the object controller for further action if any
critical condition is found in the namespace. The object controller will
propagate/delete objects after the critical conditions are gone.

Tested with "make test NOC=1" with additional tests. It passed both
"make test" and "make test NOC=1". When commenting out the "create
missing parent" code in the test and some of the following expectations,
the test failed at expectations that we expect it to fail.

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller_new.go
+++ b/incubator/hnc/pkg/controllers/object_controller_new.go
@@ -117,8 +117,8 @@ func (r *ObjectReconcilerNew) syncWithForest(ctx context.Context, log logr.Logge
 	r.Forest.Lock()
 	defer r.Forest.Unlock()
 
-	// If it's about to be deleted, do nothing, just wait for it to be actually deleted.
-	if !inst.GetDeletionTimestamp().IsZero() {
+	// Early exit if no action is required.
+	if r.ignore(ctx, log, inst) {
 		return ignore, nil
 	}
 
@@ -139,6 +139,27 @@ func (r *ObjectReconcilerNew) syncWithForest(ctx context.Context, log logr.Logge
 	}
 
 	return r.syncObject(ctx, log, inst)
+}
+
+func (r *ObjectReconcilerNew) ignore(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) bool {
+	// If it's about to be deleted, do nothing, just wait for it to be actually deleted.
+	if !inst.GetDeletionTimestamp().IsZero() {
+		return true
+	}
+
+	ns := r.Forest.Get(inst.GetNamespace())
+	// If the object is reconciled before the namespace is synced (when start-up), do nothing.
+	if !ns.Exists() {
+		log.Info("Containing namespace hasn't been synced yet")
+		return true
+	}
+	// If the namespace has critical condition, do nothing.
+	if ns.HasCritCondition() {
+		log.Info("Containing namespace has critical condition(s)")
+		return true
+	}
+
+	return false
 }
 
 // syncObject handles a source object and a propagated object differently. If a source object changes,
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/controllers/object_controller_new.go
```diff
@@ -141,6 +141,22 @@ func (r *ObjectReconcilerNew) syncWithForest(ctx context.Context, log logr.Logge
...
 }
 
+func (r *ObjectReconcilerNew) ignore(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) bool {
+	// If it's about to be deleted, do nothing, just wait for it to be actually deleted.
+	if !inst.GetDeletionTimestamp().IsZero() {
+		return true
+	}
+
+	// If the namespace has critical condition, do nothing.
+	ns := r.Forest.Get(inst.GetNamespace())
```
adrianludwin: Need another line here for when we're starting up:

```
if !ns.Exists() {
  log.Info("Containing namespace hasn't been synced yet")
  return true
}
```

incubator/hnc/pkg/forest/forest.go
```diff
@@ -317,11 +317,26 @@ func (ns *Namespace) HasCondition() bool {
...
+	return false
+}
+
 // HasCritCondition returns if the namespace has any critical condition.
 func (ns *Namespace) HasCritCondition() bool {
-	// For now, all the critical conditions are set locally. It may not be true for
-	// future critical conditions. We may not want to just use local conditions.
-	return ns.GetCondition(Local) != nil
+	return ns.HasLocalCritCondition() || ns.hasCritAncestorCondition()
 }
```
adrianludwin: Perhaps this can be written more concisely as:

```
func (ns *Namespace) HasCritCondition() bool {
  if ns.HasLocalCritCondition() {
    return true
  }
  if ns.Parent() == nil {
    return false
  }
  return ns.Parent().HasCritCondition()
}
```

incubator/hnc/pkg/controllers/object_controller_test.go
```diff
@@ -120,6 +120,65 @@ var _ = Describe("Secret", func() {
...
+
+		// Set parent to a non-existent namespace.
+		brumpfName := createNSName("brumpf")
+		barHier := newOrGetHierarchy(ctx, barName)
+		barHier.Spec.Parent = brumpfName
+		updateHierarchy(ctx, barHier)
+		Eventually(hasCondition(ctx, barName, api.CritParentMissing)).Should(Equal(true))
+		Eventually(hasCondition(ctx, bazName, api.CritAncestor)).Should(Equal(true))
+
+		// Everything should remain the same in the namespaces with Crit condition.
```
adrianludwin: I wonder if we can make this a bit more compelling and also more realistic for a controller restart case.

We start with baz -> bar -> foo, with baz inheriting a secret from both bar and foo. Then set the parent of foo to brumph; nothing should change. But then again, even if we _were_ syncing, nothing would change anyway, so this isn't a very strong test.

Now set the parent of baz to foo. If we were syncing, baz would lose its secret from bar, but because syncing is frozen, it should still stick around. We could also maybe add a second secret to foo and show that it's not propagated either. (It's hard to prove that something _doesn't_ happen. Maybe wait one second before an Expect() clause. And then manually comment out your new code and show that the test now fails).

Then, finally, create the brumpth namespace and make sure everything gets resolved.

Make sense?

yiqigao217: Yes, I see. Just confirming, your point is to make a real missing parent (foo -> "brumph") just like in controller restart phase instead of changing current parent to a missing namespace (from baz -> foo to baz -> "brumph")?

The whole test sgtm and I will go ahead to change it. And adding a new foo object sg. It's more explicit and therefore, we test both no-deleting and no-propagating.

adrianludwin: yup sgtm

Issue #302, opened 2019-12-10
Update HNC CLI to matching naming doc


adrianludwin:
Implement bit.ly/hnc-cli-naming

/assign @adrianludwin

VineethReddy02:
Hi @adrianludwin 
If you are okay. I can pick this up and update ```kubectl hnc``` plugin as mentioned in hnc-cli-naming doc.

adrianludwin:
Thanks @VineethReddy02 but there's a fair amount of context that was needed to implement this stuff, so it was easier to do it myself.

PR #305, opened 2019-12-12
Change name of kubectl plugin and fix krew.

adrianludwin:
See http://bit.ly/hnc-cli-naming for details on the rename. I also fixed
various Krew issues, most importantly that the source file
(hack/krew-hierarchical-namespaces.yaml) was being modified by the build
process.

Tested: manually verified that `kubectl hierarchical-namespaces tree`
and `kubectl hns tree` both work as expected after this change. Ran both
krew-install and krew-uninstall.

Addresses #302 but doesn't change the subcommands.

/assign @srampal 
/assign @rjbez17

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -33,7 +33,7 @@ NOC=0
 endif
 
 # Get check sum value of krew archive
-KREW_CKSM=$(shell sha256sum bin/kubectl-hnc.tar.gz | cut -d " " -f 1)
+KREW_CKSM=$(shell sha256sum bin/kubectl-hierarchical_namespaces.tar.gz | cut -d " " -f 1)
 
 
 all: test docker-build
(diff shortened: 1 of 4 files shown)


incubator/hnc/Makefile
```diff
@@ -52,37 +52,44 @@ test: build
...
+	-rm -rf bin/*
+	-rm -rf manifests/*
+	-rm -f ${GOPATH}/bin/kubectl-hierarchical_namespaces
+	-rm -f ${GOPATH}/bin/kubectl-hns
+
 # Install kubectl plugin
 kubectl: build
-	cp bin/kubectl/kubectl-hnc ${GOPATH}/bin/kubectl-hnc
-	@echo "Installed kubectl-hnc to GOPATH/bin"
+	cp bin/kubectl/kubectl-hierarchical_namespaces ${GOPATH}/bin/kubectl-hierarchical_namespaces
```
rjbez17: Would we want to delete this and the sym link when uninstalling via krew?

adrianludwin: I don't _think_ so because `make krew-install` doesn't depend on `make kubectl` so these typically won't be here. `make clean` deletes them both.

What would be the rationale for adding that here? "Just in case?"

rjbez17: Yeah, just more or less asking the question to see if it makes sense.

PR #308, opened 2019-12-12
Improve kubectl subcommands to match naming doc

### LINKED ISSUE #302: Update HNC CLI to matching naming doc

adrianludwin:
See http://bit.ly/hnc-cli-naming for details. The two major changes here are
to rename `show` to `describe`, and to add a `create` subcommand as a
synonym for `set --requiredChild`.

In addition, this change allows both `set --requiredChild` and `set
--optionalChild` to take multiple values, either comma-delimited or with
repeated flags. It also slightly improves error messages.

Tested:
* Tried both --requiredChild and --optionalChild with a combination of
comma-delimited and repeated flags.
* Manually tried the `create` command.
* Manually tried the `describe` command.

Fixes #302

### DIFF
--- a/incubator/hnc/pkg/kubectl/set.go
+++ b/incubator/hnc/pkg/kubectl/set.go
@@ -16,77 +16,76 @@ limitations under the License.
 package kubectl
 
 import (
-	"errors"
 	"fmt"
 	"os"
+	"sort"
+	"strings"
 
 	"github.com/spf13/cobra"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
-//flagValues struct stores name of namespaces against type of flag passed
-type flagValues struct {
+//hcUpdates struct stores name of namespaces against type of flag passed
+type hcUpdates struct {
 	root             bool
 	parent           string
 	requiredChildren []string
-	optionalChild    string
+	optionalChildren []string
 }
 
 var setCmd = &cobra.Command{
-	Use:     "set",
-	Short:   "Allows setting namespace hierarchy",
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/kubectl/set.go
```diff
@@ -16,77 +16,76 @@ limitations under the License.
...
+	Short: "Sets hierarchical properties of the given namespace",
+	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		nnm := args[0]
 		flags := cmd.Flags()
 		parent, _ := flags.GetString("parent")
 		requiredChildren, _ := flags.GetStringArray("requiredChild")
-		optionalChild, _ := flags.GetString("optionalChild")
+		optionalChildren, _ := flags.GetStringArray("optionalChild")
+		requiredChildren = normalizeStringArray(requiredChildren)
```
rjbez17: Side question: If I'm reading this properly, this is an acceptable use case: `--requiredChild foo --requiredChild bar,baz` and turns into a single list. Any reason to not just use `setCmd.Flags().String` to have only common separated? Just curious if there is a particular use case I haven't thought of?

adrianludwin: Yup, that's exactly what we support. I've always _understood_ that some people prefer repeated flags while others prefer comma-delimited lists, so I just figured we'd support both. Nothing really deeper than that.

rjbez17: Cool, that’s what I thought.

PR #310, opened 2019-12-12
Set v0.2 object controller as default

yiqigao217:
Part of #195 

Set v0.2 object controller as default in both controller manager and
tests.

Tested with "make test" by commenting out the if/else statement in the
object controller test for different behaviours of new/old object
controller handling a modified propagated object and leaving only the
expected behaviour of the new object controller. The "make test" passed
and "make test NOC=0" failed.

Then manually removed "-enable-new-object-controller=${NOC}" in go test
in Makefile and run "make test" again. It passed/failed with the default
value of newObjectController set to true/false in suite_test.go.

Finally tested manually to make sure the controller manager is using the
new object controller. The results were as expected.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -27,9 +27,9 @@ ifeq (,${VERSION})
 VERSION=v0.1.0
 endif
 
-# Set default object controller to the old one
+# Set default object controller to the new one
 ifeq (,${NOC})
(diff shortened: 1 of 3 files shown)


adrianludwin:
@rjbez17 already approved this in principle and the code looks pretty clear, so

/lgtm
/approve

Issue #235, opened 2019-11-06
HNC: conditions shouldn't be an enum


adrianludwin:
I believe that right now, our conditions are an enum, which means that adding any _new_ enums (which we intend to do) [is _not_ a  backwards compatible change](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas). These should simply be a string with a set of _recommended_ values, but clients need to be prepared to handle arbitrary values as well.

adrianludwin:
Note that our API version is currently v1alpha1 and we're not making any guarantees about compatibility between HNC v0.1 and v0.2, so we can make breaking changes at this time. But we'll have to stop at some point in the future.

PR #312, opened 2019-12-16
Turn Condition.code into a string, not an enum

adrianludwin:
See https://github.com/kubernetes-sigs/multi-tenancy/issues/235.
Enums are not backwards-compatible, but enum-like strings are just fine.
Since these are only generated by the controller, validation during
input isn't a concern anyway.

Also restructures some global vars and consts and improves
documentation.

Testing: builds and passes unit tests

Fixes: #235

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -19,20 +19,24 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Any changes here need to also be reflected in the kubebuilder:validation:Enum comment, below.
-// The meanings of all these constants are defined in the comment to Condition.Code, below.
+// Constants for types and well-known names
+const (
+	Singleton = "hierarchy"
+	Resource  = "hierarchyconfigurations"
+)
+
+// Constants for labels and annotations
+const (
+	MetaGroup          = "hnc.x-k8s.io"
+	LabelInheritedFrom = MetaGroup + "/inheritedFrom"
+)
+
+// Condition codes. *All* codes must also be documented in the comment to Condition.Code.
 const (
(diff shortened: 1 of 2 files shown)


rjbez17:
/ok-to-test
/approve
/lgtm
/hold

@yiqigao217  feel free to release hold when satisfied.

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -84,23 +88,37 @@ type HierarchyConfigurationList struct {
...
-	// - "CRIT_PARENT_INVALID": the specified parent is invalid (ie would cause a cycle)
+	// - "CRIT_PARENT_INVALID": the specified parent is invalid (e.g., would cause a cycle)
 	//
 	// - "CRIT_REQUIRED_CHILD_CONFLICT": there's a conflict (ie in between parent's RequiredChildren spec and child's Parent spec)
 	//
-	// - "CRIT_ANCESTOR": a critical error exists in an ancestor namespace, so this namespace is no longer being updated
-	Code Code   `json:"code,omitempty"`
+	// - "CRIT_ANCESTOR": a critical error exists in an ancestor namespace, so this namespace is no
+	// longer being updated either.
+	Code Code   `json:"code"`
```
yiqigao217: what's the difference of omitempty?

yiqigao217: is it because it's now required?

adrianludwin: There's not really any practical difference because HNC will _always_ emit a code. It's almost more like documentation to other developers that this field will always be here.

It could help a bit if there's a bug, though. Without this change, if we forgot to add a code, the YAML would just not have a `code` field at all. With this change, it will say:

```
code: ""
```

Which is more obviously wrong (and is therefore better, IMO).

PR #317, opened 2019-12-17
Improve RequiredChildConflict condition

### LINKED ISSUE #307: HNC: A missing required child shouldn't be a critical condition
Critical conditions on a namespace prevent any objects from being propagated in the namespace subtree. If we can't create a required child for some reason, this shouldn't block propagation in the rest of the subtree.

/assign @yiqigao217

adrianludwin:
This change downgrades RequiredChildConflict to be a non-critical
condition. It also only annotates it on the namespace *requesting* the
required child, not the child itself, so that one tenant cannot cause
another tenant's namespace to be marked as having a condition (critical
or otherwise).

Testing: manual with the validating webhook disabled. Found one bug and
added the unit test.

Fixed #307 
/assign @yiqigao217

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -158,7 +158,7 @@ func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Nam
 
 	// Clear locally-set conditions in the forest so we can set them to the latest.
 	hadCrit := ns.HasLocalCritCondition()
-	ns.ClearConditions(forest.Local)
+	ns.ClearConditions(forest.Local, "")
 
 	r.markExisting(log, ns)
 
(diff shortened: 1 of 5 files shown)


incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -208,24 +208,29 @@ func (r *HierarchyReconciler) markExisting(log logr.Logger, ns *forest.Namespace
...
 		log.Info("Required subnamespace: initializing", "parent", ns.RequiredChildOf)
 		inst.Spec.Parent = ns.RequiredChildOf
-	case inst.Spec.Parent == ns.RequiredChildOf:
+	case ns.RequiredChildOf:
 		// ok
 	default:
-		log.Info("Required subnamespace: assigned to wrong parent", "intended", ns.RequiredChildOf, "actual", inst.Spec.Parent)
-		r.enqueueAffected(log, "incorrect parent of the subnamespace", inst.Spec.Parent)
-		msg := fmt.Sprintf("required child of %s but parent is set to %s", ns.RequiredChildOf, inst.Spec.Parent)
-		r.enqueueAffected(log, "wrong parent set as a parent", inst.Spec.Parent)
```
yiqigao217: The updated code in green on the right side lgtm, but I got confused why we had `enqueueAffected inst.Spec.Parent` twice here?

adrianludwin: Um. Almost certainly a bug 😄

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -296,64 +301,82 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
...
+		if cns.RequiredChildOf != "" && cns.RequiredChildOf != ns.Name() {
+			// Since the list of children of this namespace came from the forest, this implies that the
+			// in-memory forest is out of sync with itself: requiredChildOf != parent. Obviously, this
+			// should never happen.
+			//
+			// Let's just log an error and enqueue the other namespace so it can report the condition.
+			// Also, reset the forest to match reality.
+			cns.RequiredChildOf = ""
+			log.Error(forest.OutOfSync, "While syncing children", "child", cn, "requiredChildOf", cns.RequiredChildOf)
+			r.enqueueAffected(log, "forest out-of-sync: requiredChildOf != parent", cns.RequiredChildOf)
```
yiqigao217: Line 327 just set it to "". Delete line 327?

adrianludwin: Great catch! I need to move line 327 to after here.

adrianludwin: Fixed

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -296,64 +301,82 @@ func (r *HierarchyReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespac
...
+		if cns.RequiredChildOf != "" && cns.RequiredChildOf != ns.Name() {
+			// Since the list of children of this namespace came from the forest, this implies that the
+			// in-memory forest is out of sync with itself: requiredChildOf != parent. Obviously, this
+			// should never happen.
+			//
+			// Let's just log an error and enqueue the other namespace so it can report the condition.
+			// Also, reset the forest to match reality.
+			log.Error(forest.OutOfSync, "While syncing children", "child", cn, "requiredChildOf", cns.RequiredChildOf)
+			r.enqueueAffected(log, "forest out-of-sync: requiredChildOf != parent", cns.RequiredChildOf)
+			cns.RequiredChildOf = ""
```
yiqigao217: We may not need this line? Since it will be set in the following `if/else`?

adrianludwin: Good point. This code used to be in a different location.

Issue #111, opened 2019-10-05
HNC: publish basic metrics


adrianludwin:
See https://kubebuilder.io/reference/metrics.html. We need to also write a mini-design doc (possibly write in this issue?) stating which metrics we need to collect.

/assign @lafh

lafh:
update: we decide to move forward with [OpenTelemetry](https://opentelemetry.io/), the merged version of OpenCensus and OpenTracing. The goal of OpenTelemetry is to standardize metric collection and distribution, and it is now the defacto standard for observability instrumentation.

rjbez17:
OpenTelemtry is pretty broad and requires you to implement the exporters within your code. Have we decided on the initial exports we plan to support? I suggest Prometheus as it is the only natively supported metrics implementation for core kubernetes (that I'm aware of). The currently supported metrics exporters are here: 

https://github.com/open-telemetry/opentelemetry-go/tree/master/exporter/metric

TL;DR: A few statsd variants, prometheus, and stdout.

adrianludwin:
/assign @yiqigao217 

Fixed by #396, #410 and #418.

/close

PR #321, opened 2019-12-17
Create reconcile counters and a flag to enable it

yiqigao217:
This commit created a reconcile counter to count total hier and object
reconciles and current ongoing object reconciles. It generates logs with
these numbers so we can get better idea of the controller performance,
for example, how long it takes to propagate objects. It would be when
the current ongoing number of object reconciles becomes zero.

Manually tested on GKE cluster and looking through the logs.

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -85,6 +85,12 @@ type NamespaceSyncer interface {
 // Reconcile simply calls SyncNamespace, which can also be called if a namespace is created or
 // deleted.
 func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	if !ex[req.Namespace] {
+		atomic.AddInt32(&hcTot, 1)
+		atomic.AddInt32(&hcCur, 1)
+		defer atomic.AddInt32(&hcCur, -1)
+	}
+
 	ctx := context.Background()
 	ns := req.NamespacedName.Namespace
 
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -18,6 +18,7 @@ package controllers
 import (
 	"context"
 	"reflect"
+	"sync/atomic"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -49,6 +50,12 @@ type ObjectReconciler struct {
 // +kubebuilder:rbac:groups=*,resources=*,verbs=get;list;watch;create;update;patch;delete;deletecollection;use;impersonate
 
 func (r *ObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	if !ex[req.Namespace] {
+		atomic.AddInt32(&obTot, 1)
+		atomic.AddInt32(&obCur, 1)
+		defer atomic.AddInt32(&obCur, -1)
+	}
+
 	resp := ctrl.Result{}
 	ctx := context.Background()
 	log := r.Log.WithValues("trigger", req.NamespacedName)
--- a/incubator/hnc/pkg/controllers/setup.go
+++ b/incubator/hnc/pkg/controllers/setup.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -10,6 +11,21 @@ import (
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 )
+
+// Counters of total/current number of hier/object reconciliation attempts.
+var (
+	hcTot int32
+	hcCur int32
+	obTot int32
+	obCur int32
+)
+
+// The ex map is only used by reconcile counters for performance testing. We explicitly
+// exclude some default namespaces with constantly changing objects.
+var ex = map[string]bool{
+	"kube-system": true,
+	"hnc-system":  true,
+}
 
 // Create creates all reconcilers.
 //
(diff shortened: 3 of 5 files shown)


incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -545,6 +546,14 @@ func (r *HierarchyReconciler) getNamespace(ctx context.Context, nm string) (*cor
 	return ns, nil
 }
 
+func (r *HierarchyReconciler) updateReconcileCounter(log logr.Logger, ns string) {
+	// Only count number of hier reconciles on user-created namespaces.
+	if !isExcludedNS(ns) {
+		hierTotal <- 1
+		log.Info("hierTotalReconciles increment")
```
adrianludwin: Can we take out the log message? Then we also wouldn't have to pass in the log object.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -52,6 +52,7 @@ func (r *ObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	resp := ctrl.Result{}
 	ctx := context.Background()
 	log := r.Log.WithValues("trigger", req.NamespacedName)
+	r.updateReconcileCounter(log, req.Namespace, true)
```
adrianludwin: Do you know about `defer`? Saying:

```
defer r.updateReconcileCounter(req.Namespace, false)
```

will call this function when `Reconcile` exits, which should allow you to get rid of _all_ the other calls, I think?

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -325,6 +337,23 @@ func (r *ObjectReconciler) isExcluded(log logr.Logger, inst *unstructured.Unstru
 	}
 }
 
+// updateReconcileCounter increments both total and current number of reconciles before
+// Reconcile and decrements the current reconciles after Reconcile.
+func (r *ObjectReconciler) updateReconcileCounter(log logr.Logger, ns string, before bool) {
```
adrianludwin: Does this need to be part of `ObjectReconciler`? It doesn't appear to access any of its members.

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -11,10 +12,27 @@ import (
...
 // Create creates all reconcilers.
 //
 // This function is called both from main.go as well as from the integ tests.
 func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int, newObjectController bool) error {
+	// Create reconcile counter
+	counter := &ReconcileCounter{
+		Log:         ctrl.Log.WithName("reconcileCounter"),
+		ObjectTotal: 0,
+		ObjectCur:   0,
+		HierTotal:   0,
```
adrianludwin: Why not have a `HierCur` as well? Not saying you should, just wondering why we don't.

If you did, we could just have a single `updateReconcileCounter` function across both (temporarily: all three) controllers that just indicated which totals we were updating.

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -61,3 +79,53 @@ func createObjectReconciler(newObjectController bool, mgr ctrl.Manager, f *fores
...
+	for t := range hierTotal {
+		counter.HierTotal += t
+		log.Info("Total number of hier reconciles", "Total", counter.HierTotal)
+	}
+}
+
+// This function is only used by reconcile counters for performance testing. We explicitly exclude the
+// default namespaces, so that only reconciles on objects in the user-created namespaces are counted.
+func isExcludedNS(ns string) bool {
+	ex := []string{"kube-system", "default", "cert-manager", "hnc-system", "kube-node-lease", "kube-public"}
```
adrianludwin: Can we make this a bit more minimal? Eg if we just exclude `kube-system` and `hnc-system`, will it still work?

Also, you could make this into a map:

```
var ex = map[string]bool{
  "kube-system": true,
  "hnc-system": true,
}

func isExcludedNS(ns string) bool {
  return ex[ns]
}
```

Although at that point you don't even really need `isExcludedNS` anymore :)

incubator/hnc/pkg/controllers/object_controller.go
yiqigao217: ah right!

incubator/hnc/pkg/controllers/setup.go
yiqigao217: Yes, it worked. Other namespaces don't have constantly changing objects.

incubator/hnc/pkg/controllers/setup.go
yiqigao217: I added it now just to be more informative. (Added as atomic counter)

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -11,6 +12,21 @@ import (
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 )
 
+// Counters of total/current number of hier/object reconciliation attempts.
+var (
+	hcTot int32
```
rjbez17: Are we not worried about overflowing here? Any reason to not use uint32 so we at least overflow to 0? I mean, it is highly unlikely we hit these limits during a perf test so we probably just don't care? How many reconciles to we anticipate per minute during a test? Will a long running soak test get near these limits?

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -61,3 +77,21 @@ func createObjectReconciler(newObjectController bool, mgr ctrl.Manager, f *fores
 	}
 	return or, or.SetupWithManager(mgr)
 }
+
+// LogActivity generates logs for performance testing.
+func LogActivity() {
```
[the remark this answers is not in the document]
adrianludwin: We opened issue #111 to publish metrics, though we were thinking to use OpenTelemetry instead of Prometheus, since OT's all about publishing metrics to various consumers, while Prometheus (I thought) was mainly a single implementation. If you have thoughts on OT vs Prom, maybe leave them in that issue?

But in the case of this one change, I think it will be faster to just read the logs while running performance tests - for starters, we don't have to install Prometheus, which is especially useful if we're testing on KinD or a small cluster. Does that sound reasonable? We can always update this later.

incubator/hnc/pkg/controllers/setup.go
adrianludwin: I doubt we're going to run anywhere near to 2B reconciles - even at 100 reconciles a second (which is way above what I think we can support), that's 240 days, which is a pretty long soak test!

I dunno - is uint32 better because it overflows to a possibly-valid value, or is int32 better because it overflows to an obviously-bad value? :)

incubator/hnc/pkg/controllers/setup.go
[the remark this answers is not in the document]
rjbez17: Yeah, nothing to hold up this PR. We can move the convo to the ticket.

incubator/hnc/pkg/controllers/setup.go
rjbez17: > I dunno - is uint32 better because it overflows to a possibly-valid value, or is int32 better because it overflows to an obviously-bad value? :)

Haha fair. I have a feeling this is like asking if you prefer emacs over vim...

> I doubt we're going to run anywhere near to 2B reconciles - even at 100 reconciles a second (which is way above what I think we can support), that's 240 days, which is a pretty long soak test!

Agreed, but I can't just let an unhandled overflow through without asking :)

PR #322, opened 2019-12-17
Add conditions for object problems

### LINKED ISSUE #201: HNC: warn about unpropagated objects
An object may not be propagated for several reasons:

- It's been manually modified by the user (bypassing the webhook)
- It has finalizers or other HNC-incompatible metadata set on it per #106 

In such cases, we should add a condition to the object's namespace as a warning.

adrianludwin:
Also prevent HC controller from invoking object controllers; since
object controllers can now invoke the HC controller, this was causing an
infinite loop.

Tested: manually create a namespace both with an uncopyable rolebinding
to cause a problem during the "write" stage and a secret with a
finalizer to cause a problem during the "sync" stage. Observed that all
expected conditions were applied to all namespaces, albeit with some
funny transient states that I should dig into. I also added a unit test
for objects with finalizers, but I can't stub out the "write" function
yet. This will come in a later PR.

Fixes #201

### DIFF
--- a/incubator/hnc/pkg/controllers/object_controller_new.go
+++ b/incubator/hnc/pkg/controllers/object_controller_new.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"sync/atomic"
 
@@ -66,6 +67,9 @@ type ObjectReconcilerNew struct {
 	// https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches.html) that is used to
 	// enqueue additional objects that need updating.
 	Affected chan event.GenericEvent
+
+	// AffectedNamespace is a channel of events used to update namespaces.
+	AffectedNamespace chan event.GenericEvent
 }
 
 // +kubebuilder:rbac:groups=*,resources=*,verbs=get;list;watch;create;update;patch;delete
@@ -128,6 +132,11 @@ func (r *ObjectReconcilerNew) syncWithForest(ctx context.Context, log logr.Logge
 	if r.ignore(ctx, log, inst) {
 		return ignore, nil
 	}
+
+	// Clear any existing conditions associated with this object. TODO: this will never be called if
+	// the source namespace holding this object goes away (i.e., if this namespace gets a new parent).
+	// See https://github.com/kubernetes-sigs/multi-tenancy/issues/328.
+	r.clearConditions(log, inst)
 
 	// If an object doesn't exist, assume it's been deleted or not yet created.
 	// inst.GetCreationTimestamp().IsZero() has compile time errors, so we manually check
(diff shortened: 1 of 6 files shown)


incubator/hnc/pkg/controllers/object_controller_new.go
```diff
@@ -299,31 +317,66 @@ func (r *ObjectReconcilerNew) delete(ctx context.Context, log logr.Logger, inst
...
+	defer r.Forest.Unlock()
+
+	gvk := inst.GetObjectKind().GroupVersionKind()
+	key := fmt.Sprintf("%s/%s/%s/%s/%s", gvk.Group, gvk.Version, gvk.Kind, inst.GetNamespace(), inst.GetName())
+	msg := fmt.Sprintf("Could not %s: %s", op, err.Error())
+	r.Forest.Get(inst.GetNamespace()).SetCondition(key, api.CannotUpdate, msg)
+	r.enqueueNamespace(log, inst.GetNamespace(), "Updating condition on propagated namespace: "+msg)
+	if srcInst != nil {
+		r.Forest.Get(inst.GetNamespace()).SetCondition(key, api.CannotPropagate, msg)
+		r.enqueueNamespace(log, inst.GetNamespace(), "Updating condition on source namespace: "+msg)
```
yiqigao217: Enqueue the `srcInst.GetNamespace()` here?

yiqigao217:
Other than one minor issue, the rest all lgtm.

adrianludwin:
/assign @rjbez17 

Hi @yiqigao217 , @rjbez17  - I think this is ready for a review now. Thanks!

incubator/hnc/pkg/controllers/object_controller_new.go
```diff
@@ -313,17 +341,61 @@ func (r *ObjectReconcilerNew) write(ctx context.Context, log logr.Logger, inst,
...
+	msg := fmt.Sprintf("Could not %s: %s", op, err.Error())
+	r.setCondition(log, api.CannotUpdate, inst.GetNamespace(), key, msg)
+	if srcInst != nil {
+		r.setCondition(log, api.CannotPropagate, srcInst.GetNamespace(), key, msg)
+	}
+}
+
+func (r *ObjectReconcilerNew) setLocalCondition(log logr.Logger, inst *unstructured.Unstructured, code api.Code, msg string) {
+	r.setCondition(log, code, inst.GetNamespace(), getObjectKey(inst), msg)
+}
```
yiqigao217: Why do we need this func?

adrianludwin: Good point, removed

incubator/hnc/pkg/controllers/object_controller_new.go
adrianludwin: Fixed

yiqigao217:
lgtm

adrianludwin:
Thanks @yiqigao217 !

/hold cancel
@rjbez17 just waiting for your ok

Issue #250, opened 2019-11-12
VC: kubernetes conformance test for tenant


zhuangqh:
set up a k8s e2e test for tenant to figure out whether the virtual cluster looks like a real k8s cluster.

Summarizing 35 Failures:

- [x]   [Fail] [k8s.io] InitContainer [NodeConformance] [It] should not start app containers if init containers fail on a RestartAlways pod [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/common/init_container.go:292

- [x] [Fail] [sig-storage] ConfigMap [It] updates should be reflected in volume [NodeConformance] [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/common/configmap_volume.go:188

- [x]  [Fail] [sig-node] Downward API [It] should provide pod name, namespace and IP address as env vars [NodeConformance] [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/framework/util.go:2342

- [x] [Fail] [k8s.io] Pods [It] should allow activeDeadlineSeconds to be updated [NodeConformance] [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/common/pods.go:439

- [x] [Fail] [k8s.io] InitContainer [NodeConformance] [It] should invoke init containers on a RestartNever pod [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/common/init_container.go:99

- [x] [Fail] [sig-storage] Secrets [It] optional updates should be reflected in volume [NodeConformance] [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/common/secrets_volume.go:367

- [x] [Fail] [sig-storage] Projected secret [It] optional updates should be reflected in volume [NodeConformance] [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/common/projected_secret.go:401

- [x] [Fail] [sig-storage] Downward API volume [It] should update annotations on modification [NodeConformance] [Conformance]
[...]

PR #325, opened 2019-12-18
[incubator][VC] back populate pod related event to tenant

zhuangqh:
related #250 

lookup virtual ns from super ns when back pop pod

fix test

```
[Fail] [k8s.io] [sig-node] Events [It] should be sent by kubelets and the scheduler about pods scheduling and running [Conformance]
/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/node/events.go:96
```

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/controllers/event/uws.go
+++ b/incubator/virtualcluster/pkg/syncer/controllers/event/uws.go
@@ -1,0 +1,130 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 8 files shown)


incubator/virtualcluster/pkg/syncer/controllers/event/uws.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	pEvent, err := c.eventLister.Events(pNamespace).Get(pName)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("could not find pEvent %s/%s in controller cache: %v", pNamespace, pName, err)
+	}
+
```
Fei-Guo: Should we check pEvent.nvolvedObject.Kind here to filter all non-Pod related events earlier?

PR #332, opened 2019-12-21
Use Cloud Build for everything and update docs

adrianludwin:
This change updates `make release` so that as little as possible happens
on the host computer. All artifacts - the manifests, kubectl plugin and
of course the image itself - are now built on Cloud Build.

The image does still have to make a round-trip onto the builder's
workstation since the Cloud Build service account doesn't have
permission to push directly to the K8s staging GCR, but you can still
verify that the image built on Cloud Build is the same one on the
official GCR.

Tested: followed these instructions to create a new build and deployed
it.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -3,38 +3,35 @@
 
 # If CONFIG is `kind`, various defaults will be optimized for deploying locally to Kind
 CONFIG ?= "default"
+
+PROJECT_ID=$(shell gcloud config get-value project)
+
+# Set default version tag for krew build and Docker image (unless version is set)
+HNC_IMG_TAG ?= latest
+
 # Image URL to use all building/pushing image targets
 ifeq ($(CONFIG),kind)
-	# The tag is `kind-local` since K8s always attempst to re-pull an image with the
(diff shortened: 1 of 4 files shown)


adrianludwin:
/assign @rjbez17 

Note that I've deleted the v0.2.0-rc1 release; once this is submitted I'll recreate it.

incubator/hnc/Makefile
```diff
@@ -180,5 +179,36 @@ kind-deploy:
 
 ###################### RELEASE ACTIONS #########################
 # Build the container image by Cloud Build and build YAMLs locally
-release: manifests
-	gcloud builds submit --config cloudbuild.yaml .
+#
+# ALL COMMANDS THAT INCLUDE THE PERSONAL ACCESS TOKEN ARE HIDDEN
+release: check-release-env test
```
rjbez17: We currently aren’t doing any logic to prevent re-pushing tags (so changing the underlying image from users). We should probably have a TODO for that as we start to move into more stable releases.

adrianludwin: Filed #333

PR #340, opened 2020-01-06
[Incubator][VC] Clean up - Refactor resource controller - Part 1

Fei-Guo:
This change mainly does the followings:

1) Move downward syncer logic to a separate dws.go for better code structure.
2) Move WaitForCacheSync to the right sync thread.
3) Add more comments.

This change touches configmap/events/namespace/event resources. There are no functional change.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/configmap/controller.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/configmap/controller.go
@@ -17,30 +17,23 @@ limitations under the License.
 package configmap
 
 import (
-	"fmt"
-
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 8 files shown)


incubator/virtualcluster/pkg/syncer/resources/configmap/controller.go
```diff
@@ -60,108 +53,18 @@ func Register(
 		return
 	}
 	c.multiClusterConfigMapController = multiClusterConfigMapController
-
 	c.configMapLister = configMapInformer.Lister()
-	c.configMapSynced = configMapInformer.Informer().HasSynced
```
zhuangqh: we still need to wait for configmapSynced, or the configMapLister cannot work.

Fei-Guo: Fixed

PR #341, opened 2020-01-07
[incubator][VC] support tenant aware dns server

zhuangqh:
add tenant dns setup script 

support tenant aware dns server
 1. get tenant name server from service kube-system/kube-dns
 2. mutate pPod's dns config to bypass kubelet's behavior

related to #250

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/cluster/cluster.go
+++ b/incubator/virtualcluster/pkg/syncer/cluster/cluster.go
@@ -30,6 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/apis/tenancy/v1alpha1"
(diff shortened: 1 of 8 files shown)


incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return ret
+}
+
+type ServiceMutateInterface interface {
+	Mutate()
+}
+
+type serviceMutator struct {
+	vc       *v1alpha1.VirtualclusterSpec
```
Fei-Guo: Does service mutator need VC at all?

zhuangqh: nope. But maybe it does later. This is a unify framework for mutator

incubator/virtualcluster/pkg/syncer/cluster/cluster.go
```diff
@@ -49,6 +50,9 @@ type Cluster struct {
 	// Config is the rest.config used to talk to the apiserver.  Required.
 	Config *rest.Config
 
+	// spec is the vc definition. Required.
+	spec *v1alpha1.VirtualclusterSpec
```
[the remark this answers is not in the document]
zhuangqh: we may use other fields later

Issue #324, opened 2019-12-18
VC: introduce master provider to vc-manager


zhuangqh:
There will be many implementations to create a tenant master according to VC cr. such as kok or AlibabaCloud openapi, etc. Introduce a provider layer to abstract different implementation.

PR #343, opened 2020-01-08
Allow user to choose different underlying platforms to provision tenant master

charleszheng44:
1. abstract Create/Delete Virtualcluster functions, and encapsulate
them into a new interface `MasterProvisioner`
2. adapt existing minikube implementation to `MasterProvisioner` interface
3. add template of `MasterProvisionerAliyun`

### DIFF
--- a/incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_aliyun.go
+++ b/incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_aliyun.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package virtualcluster
+
+import (
+	"errors"
+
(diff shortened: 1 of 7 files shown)


incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_aliyun.go
```diff
@@ -0,0 +1,30 @@
+package virtualcluster
```
zhuangqh: please add file header

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
```diff
@@ -0,0 +1,339 @@
...
+		}
+	}
+
+}
+
+// createPKISecrets creates secrets to store crt/key pairs and kubeconfigs
+// for master components of the virtual cluster
+func (mpn *MasterProvisionerNative) createPKISecrets(caGroup *vcpki.ClusterCAGroup, namespace string) error {
+	// create secret for root crt/key pair
+	rootSrt := secret.CrtKeyPairToSecret(secret.RootCASecretName,
```
zhuangqh: go style guide tell us "don't add line breaks to keep lines short when they are more readable long". Can we Keep this statement one line.
https://github.com/golang/go/wiki/CodeReviewComments#line-length

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_aliyun.go
charleszheng44: added

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
charleszheng44: line break is removed

zhuangqh:
fixes: #324

PR #349, opened 2020-01-15
Automate controller performance tests

yiqigao217:
The tests will load wide, full, skewer topologies to run seven actions
on each. It will output the completion time and the number of api
reads/writes for each action.

Tested on GKE cluster.

### DIFF
--- a/incubator/hnc/config/manager/manager.yaml
+++ b/incubator/hnc/config/manager/manager.yaml
@@ -32,8 +32,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
+            memory: 1Gi
           requests:
             cpu: 100m
-            memory: 20Mi
+            memory: 512Mi
       terminationGracePeriodSeconds: 10
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -84,11 +84,12 @@ type NamespaceSyncer interface {
 
 // Reconcile sets up some basic variables and then calls the business logic.
 func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	if !ex[req.Namespace] {
-		atomic.AddInt32(&hcTot, 1)
-		atomic.AddInt32(&hcCur, 1)
-		defer atomic.AddInt32(&hcCur, -1)
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
 	}
+	atomic.AddInt32(&hcTot, 1)
+	atomic.AddInt32(&hcCur, 1)
+	defer atomic.AddInt32(&hcCur, -1)
 
 	ctx := context.Background()
 	ns := req.NamespacedName.Namespace
@@ -150,6 +151,10 @@ func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm
 // fully written back to the apiserver yet, each namespace is reconciled in isolation (apart from
 // the in-memory forest) so this is fine.
 func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Namespace, inst *api.HierarchyConfiguration) {
+	if !ex[inst.ObjectMeta.Namespace] {
+		atomic.AddInt32(&mutexQ, 1)
+		defer atomic.AddInt32(&mutexQ, -1)
+	}
 	r.Forest.Lock()
 	defer r.Forest.Unlock()
 	ns := r.Forest.Get(inst.ObjectMeta.Namespace)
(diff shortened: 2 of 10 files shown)


incubator/hnc/config/manager/manager.yaml
```diff
@@ -32,8 +32,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 30Mi
```
adrianludwin: While 30MB was too low, 1GB is almost certainly too high. Also, this belongs in its own PR.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -84,11 +84,12 @@ type NamespaceSyncer interface {
...
 // Reconcile sets up some basic variables and then calls the business logic.
 func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	if !ex[req.Namespace] {
-		atomic.AddInt32(&hcTot, 1)
-		atomic.AddInt32(&hcCur, 1)
-		defer atomic.AddInt32(&hcCur, -1)
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
 	}
+	atomic.AddInt32(&hcTot, 1)
```
adrianludwin: Now that we're checking these into the main branch, we should probably give them more reasonable names and maybe group them together. Eg something like:

```
// in stats.go:
type counter int32
func (c *counter) Incr() {
  atomic.AddInt32(c, 1)
}
func (c *counter) Decr() {
  atomic.AddInt32(c, -1)
}
type stats struct {
  hcTotal counter
  hcCurrent counter
  ...
}
var stats stats

// here:
stats.hcCurrent.Incr()
defer stats.hcCurrent.Decr()

File: incubator/hnc/pkg/controllers/hierarchy_controller.go [hunk 3]
```diff
@@ -150,6 +151,10 @@ func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm
 // fully written back to the apiserver yet, each namespace is reconciled in isolation (apart from
 // the in-memory forest) so this is fine.
 func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Namespace, inst *api.HierarchyConfiguration) {
+	if !ex[inst.ObjectMeta.Namespace] {
+		atomic.AddInt32(&mutexQ, 1)
```
adrianludwin: I think we found that there was no mutex contention, right? If so, I'd just delete these.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -477,6 +482,11 @@ func (r *HierarchyReconciler) writeHierarchy(ctx context.Context, log logr.Logge
 		return false, nil
 	}
 
+	if !ex[inst.ObjectMeta.Namespace] {
+		atomic.AddInt32(&apiCall, 1)
+		atomic.AddInt32(&hcWriteHC, 1)
```
adrianludwin: Not sure we need _both_ `apiCall` and `hcWriteHC` (which should just be `hcWrite`). We can aggregate somewhere else, not at the call site.

Also, as a TODO, I wonder if we can replace this completely with the controller-runtime Prometheus metrics?

https://github.com/kubernetes-sigs/controller-runtime/blob/838c86fe0af6cd645301059505a61742de01ea2c/pkg/metrics/client_go_adapter.go#L35

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -150,6 +151,10 @@ func (r *HierarchyReconciler) reconcile(ctx context.Context, log logr.Logger, nm
 // fully written back to the apiserver yet, each namespace is reconciled in isolation (apart from
 // the in-memory forest) so this is fine.
 func (r *HierarchyReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Namespace, inst *api.HierarchyConfiguration) {
+	if !ex[inst.ObjectMeta.Namespace] {
```
adrianludwin: Why are we only excluding the _metrics_ if the namespace is excluded? Shouldn't this just be:

```
if ex[inst.ObjectMeta.Namespace] {
  return
}
```

And similarly elsewhere in this file?

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -14,13 +15,20 @@ import (
 
 // Counters of total/current number of hier/object reconciliation attempts.
 var (
-	hcTot int32
-	hcCur int32
-	obTot int32
-	obCur int32
+	hcTot     int32
```
adrianludwin: These all need to be documented. Also, see the other comments I made about restructuring these.

rjbez17:
@yiqigao217 any updates?

adrianludwin:
I _think_ this PR is obsolete - the golang changes went in another PR and the scripts are pending.

yiqigao217:
Yes, this one is obsolete. The reviews here will be addressed in the new PRs. The generating logs part is done in this new PR #368 .

PR #358, opened 2020-01-19
[incubator][VC] add extend info (deployment) to vPod

zhuangqh:

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
@@ -22,13 +22,14 @@ import (
 
 	"k8s.io/api/core/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	"k8s.io/kubernetes/pkg/kubelet/envvars"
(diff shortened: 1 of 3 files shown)


incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -256,12 +259,40 @@ func omitDuplicates(strs []string) []string {
 	return ret
 }
 
+// TODO(zhuangqh): make this plugable.
+func (p *podMutator) attachTenantMeta(vPod *v1.Pod) error {
+	if len(vPod.ObjectMeta.OwnerReferences) == 0 {
```
Fei-Guo: You cannot assume all Pods in tenant are deployment Pods. The Pods can be standard alone Pods and owned by any controller.

incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -256,12 +259,40 @@ func omitDuplicates(strs []string) []string {
 	return ret
 }
 
+// TODO(zhuangqh): make this plugable.
```
Fei-Guo: You can add a comment saying that for now, only Deployment Pods are mutated.

incubator/virtualcluster/pkg/syncer/conversion/mutate.go
zhuangqh: done

PR #361, opened 2020-01-23
[Incubator][VC]Refactor Cluster and MultiClusterController interface methods

Fei-Guo:
This change attempts to achieve the following best practices for building syncer framework objects.
- Expose least amount of internal state
- Avoid exposing internal pointers 

As a consequence, with this change,
- Cluster and ClusterInterface does not export GetCache() method anymore. MultiClusterController adds List method to encapsulate the access to internal cluster cache.
- Cluster does not export GetMapper and GetScheme.
- MultiClusterController does not expose GetCluster method anymore. All needs to access cluster pointer are supported by specific public methods such as GetClusterClient and GetClusterDomain.
- Many related codes are changed due to above interface changes.

Bonus fixes:
- Avoid looping the entire cluster map when calling getCluster.
- Fix a small bug in attachTenantMeta in case a standalone replicaset is deployed in tenant master.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/cluster/cluster.go
+++ b/incubator/virtualcluster/pkg/syncer/cluster/cluster.go
@@ -105,9 +105,7 @@ func (c *Cluster) GetSpec() *v1alpha1.VirtualclusterSpec {
 	return c.spec
 }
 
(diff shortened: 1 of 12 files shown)


zhuangqh:
great job.

PR #363, opened 2020-01-25
[Incubator][VC] Add storageclass object periodic checker

Fei-Guo:
This change adds storageclass periodic checker to make sure tenant and super are consistent.
We don't implement a more aggressive consistency guarantee logic such that watching tenant sc delete events to reduce the overhead of using informer.  We may add it later if necessary. Note that for storageclass, the source of truth is super master. If tenant master happens to create a sc with the same name as one of the sc in super, the checker reports a warning if their specs are different for now.

Manual Test:
Delete sc from tenant master and within 60 seconds, the sc is recreated again in tenant.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/storageclass/checker.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/storageclass/checker.go
@@ -67,7 +67,7 @@ func (c *controller) checkStorageClass() {
 	}
 
 	for _, pStorageClass := range pStorageClassList {
-		if pStorageClass.Labels[constants.PublicObjectKey] != "true" {
+		if !publicStorageClass(pStorageClass) {
 			continue
 		}
 		for _, clusterName := range clusterNames {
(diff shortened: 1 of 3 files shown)


incubator/virtualcluster/pkg/syncer/resources/storageclass/checker.go
```diff
@@ -0,0 +1,121 @@
...
+	wg.Wait()
+
+	pStorageClassList, err := c.storageclassLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("error listing storageclass from super master informer cache: %v", err)
+		return
+	}
+
+	for _, pStorageClass := range pStorageClassList {
+		if pStorageClass.Labels[constants.PublicObjectKey] != "true" {
```
zhuangqh: can we reuse `publicStorageClass` this function?

Fei-Guo: Done

PR #367, opened 2020-01-29
[incubator][VC] provide kubeconfig according to service account

zhuangqh:
This change provide a way to access to pod's own master using a kubeconfig.

- [x] create a kubeconfig according to service account when creating pod
- [x] find a mount point and mounts a configmap to pod
- [x] using a flag to control this logic
- [x] provide a sample controller yaml

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
@@ -47,7 +47,7 @@ func VC(mc *mc.MultiClusterController, clusterName string) VCMutateInterface {
 }
 
 func (m *mutator) Pod(pPod *v1.Pod) PodMutateInterface {
-	return &podMutator{mc: m.mc, clusterName: m.clusterName, pPod: pPod}
+	return &podMutateCtx{mc: m.mc, clusterName: m.clusterName, pPod: pPod}
 }
 
 func (m *mutator) Service(pService *v1.Service) ServiceMutateInterface {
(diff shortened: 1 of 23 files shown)


incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -46,67 +47,81 @@ func VC(mc *mc.MultiClusterController, clusterName string) VCMutateInterface {
...
-	// setup env var map
-	serviceEnv := getServiceEnvVarMap(p.pPod.Namespace, p.getCluster(), *p.pPod.Spec.EnableServiceLinks, services)
+	return nil
+}
 
-	for i := range p.pPod.Spec.Containers {
-		mutateContainerEnv(&p.pPod.Spec.Containers[i], vPod, serviceEnv)
-		mutateContainerSecret(&p.pPod.Spec.Containers[i], vSASecret, SASecret)
-	}
+func PodMutateDefault(vPod *v1.Pod, vSASecret, SASecret *v1.Secret, services []*v1.Service, nameServer string) PodMutator {
```
zhuangqh: PodMutateDefault is the same as the old one. In addition, I mutate the hostAlias here

incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -273,37 +291,84 @@ func omitDuplicates(strs []string) []string {
 	return ret
 }
 
-// TODO(zhuangqh): make this plugable.
 // for now, only Deployment Pods are mutated.
-func (p *podMutator) attachTenantMeta(vPod *v1.Pod) error {
-	if len(vPod.ObjectMeta.OwnerReferences) == 0 || vPod.ObjectMeta.OwnerReferences[0].Kind != "ReplicaSet" {
+func PodAddExtensionMeta(vPod *v1.Pod) PodMutator {
```
zhuangqh: The same logic as before

incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -273,37 +291,84 @@ func omitDuplicates(strs []string) []string {
...
-	}
-	labels[constants.LabelExtendDeploymentName] = replicaSetObj.ObjectMeta.OwnerReferences[0].Name
-	labels[constants.LabelExtendDeploymentUID] = string(replicaSetObj.ObjectMeta.OwnerReferences[0].UID)
-	p.pPod.SetLabels(labels)
+}
 
-	return nil
+func PodMutateAutoMountServiceAccountToken(disable bool) PodMutator {
+	return func(p *podMutateCtx) error {
+		if disable {
```
zhuangqh: if not disable, user can decide whether or not

incubator/virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -259,6 +289,60 @@ func (c *controller) reconcilePodUpdate(cluster, namespace, name string, vPod *v
...
+				ctx.AuthInfo: {
+					Token: string(token),
+				},
+			}
+			cluster := cfg.Clusters[ctx.Cluster]
+			cluster.Server = "https://kubernetes.default"
+			cfg.Clusters = map[string]*clientcmdapi.Cluster{
+				ctx.Cluster: cluster,
+			}
+
```
zhuangqh: after this transform, we got a minimize kubeconfig

incubator/virtualcluster/pkg/syncer/utils/string.go
```diff
@@ -0,0 +1,47 @@
...
+// GenAnAvailableName generate a unique valid name against the string slice passed in.
+// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+func GenAnAvailableName(names []string, prefix string, n int) string {
+	nameMap := make(map[string]struct{})
+	for _, name := range names {
+		nameMap[name] = struct{}{}
+	}
+
+	if n < len(prefix) {
+		panic("target string length cannot larger than the prefix one")
```
zhuangqh: You should got this panic during testing.  I want no error in this function

incubator/virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -174,7 +176,35 @@ func (c *controller) reconcilePodCreate(cluster, namespace, name string, vPod *v
...
+		}
+
+		kubeConfigBytes, err := createKubeConfigByServiceAccount(tenantClient.CoreV1().ServiceAccounts(vPod.Namespace), tenantClient.CoreV1().Secrets(vPod.Namespace), clientConfig, saName)
+		if err != nil {
+			return fmt.Errorf("failed to create kubeconfig from service account: %v", err)
+		}
+
+		secret := conversion.BuildKubeConfigSecret(cluster, vPod, kubeConfigBytes)
+		secret.Namespace = targetNamespace
+
```
zhuangqh: First I want to add directly `ownerReference` to the secret object. But We should first create the Pod, then we got the UID. There is a deadlock.

incubator/virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -174,7 +176,35 @@ func (c *controller) reconcilePodCreate(cluster, namespace, name string, vPod *v
...
+		_, err = c.client.Secrets(targetNamespace).Create(secret)
```
zhuangqh: Looks for help @Fei-Guo 
Actually I want to do "if exists, overwrite the old one", or we cannot handle the secret update situation.

Fei-Guo: Let us use uuid for secret name for now to workaround the problem. We should properly fix the problem by sharing the secrets for multiple Pods later.

PR #368, opened 2020-01-29
Generate controller stats logs

yiqigao217:
Create counters for controller stats, e.g. number of reconciliation
attemps, or api calls. Generate logs with all the stats if the
enable-test-log flag is set.

Tested on GKE cluster and found the generated logs in Stackdriver. The
logs stopped when the controllers were idle and once I did actions like
creating namespace or setting parents, the logs started generating
again.

### DIFF
--- a/incubator/hnc/pkg/controllers/hierarchy_controller.go
+++ b/incubator/hnc/pkg/controllers/hierarchy_controller.go
@@ -38,6 +38,7 @@ import (
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/metadata"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/stats"
 )
 
 // HierarchyReconciler is responsible for determining the forest structure from the Hierarchy CRs,
@@ -84,11 +85,12 @@ type NamespaceSyncer interface {
 
 // Reconcile sets up some basic variables and then calls the business logic.
 func (r *HierarchyReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	if !ex[req.Namespace] {
-		atomic.AddInt32(&hcTot, 1)
-		atomic.AddInt32(&hcCur, 1)
-		defer atomic.AddInt32(&hcCur, -1)
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
 	}
+
+	stats.StartHierConfigReconcile()
+	defer stats.StopHierConfigReconcile()
 
 	ctx := context.Background()
 	ns := req.NamespacedName.Namespace
@@ -477,6 +479,7 @@ func (r *HierarchyReconciler) writeHierarchy(ctx context.Context, log logr.Logge
 		return false, nil
 	}
 
+	stats.WriteHierConfig()
 	if inst.CreationTimestamp.IsZero() {
 		log.Info("Creating singleton on apiserver")
 		if err := r.Create(ctx, inst); err != nil {
@@ -499,6 +502,7 @@ func (r *HierarchyReconciler) writeNamespace(ctx context.Context, log logr.Logge
 		return false, nil
 	}
 
+	stats.WriteNamespace()
 	if inst.CreationTimestamp.IsZero() {
 		log.Info("Creating namespace on apiserver")
 		if err := r.Create(ctx, inst); err != nil {
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -18,7 +18,6 @@ package controllers
 import (
 	"context"
 	"reflect"
-	"sync/atomic"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -32,6 +31,7 @@ import (
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/metadata"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/object"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/stats"
 )
 
 // ObjectReconciler reconciles generic propagated objects. You must create one for each
@@ -50,11 +50,12 @@ type ObjectReconciler struct {
 // +kubebuilder:rbac:groups=*,resources=*,verbs=get;list;watch;create;update;patch;delete;deletecollection;use;impersonate
 
 func (r *ObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	if !ex[req.Namespace] {
-		atomic.AddInt32(&obTot, 1)
-		atomic.AddInt32(&obCur, 1)
-		defer atomic.AddInt32(&obCur, -1)
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
 	}
+
+	stats.StartObjReconcile(r.GVK)
+	defer stats.StopObjReconcile(r.GVK)
 
 	resp := ctrl.Result{}
 	ctx := context.Background()
@@ -123,6 +124,7 @@ func (r *ObjectReconciler) update(ctx context.Context, log logr.Logger, inst *un
 
 	dests, good := r.syncWithForest(ctx, log, inst, srcInst)
 	if !good {
+		stats.WriteObject(inst.GroupVersionKind())
 		return r.Delete(ctx, inst)
 	}
 
@@ -244,6 +246,7 @@ func (r *ObjectReconciler) propagate(ctx context.Context, log logr.Logger, inst 
 		metadata.SetLabel(propagated, api.LabelInheritedFrom, srcNS)
 
 		// Push to the apiserver
+		stats.WriteObject(propagated.GroupVersionKind())
 		log.Info("Propagating", "dst", dst, "origin", srcNS)
 		err := r.Update(ctx, propagated)
 		if err != nil && errors.IsNotFound(err) {
@@ -283,6 +286,7 @@ func (r *ObjectReconciler) onDelete(ctx context.Context, log logr.Logger, nnm ty
 		// TODO: double-check the label - or maybe just call deleteObsolete?
 
 		// Delete the copy
+		stats.WriteObject(propagated.GroupVersionKind())
 		log.Info("Deleting", "propagated", propagatedNnm)
 		if err := r.Delete(ctx, propagated); err != nil {
 			log.Error(err, "Coudln't delete", "propagated", propagated)
--- a/incubator/hnc/pkg/controllers/setup.go
+++ b/incubator/hnc/pkg/controllers/setup.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"fmt"
-	"time"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -12,19 +11,13 @@ import (
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 )
 
-// Counters of total/current number of hier/object reconciliation attempts.
-var (
-	hcTot int32
-	hcCur int32
-	obTot int32
-	obCur int32
-)
-
-// The ex map is only used by reconcile counters for performance testing. We explicitly
+// The ex map is used by controllers to exclude namespaces to reconcile. We explicitly
 // exclude some default namespaces with constantly changing objects.
+// TODO make the exclusion configurable - https://github.com/kubernetes-sigs/multi-tenancy/issues/374
 var ex = map[string]bool{
-	"kube-system": true,
-	"hnc-system":  true,
+	"kube-system":  true,
+	"hnc-system":   true,
+	"cert-manager": true,
 }
 
 // Create creates all reconcilers.
@@ -79,22 +72,4 @@ func createObjectReconciler(newObjectController bool, mgr ctrl.Manager, f *fores
 		GVK:    gvk,
 	}
 	return or, or.SetupWithManager(mgr)
-}
-
-// LogActivity generates logs for performance testing.
-func LogActivity() {
-	log := ctrl.Log.WithName("reconcileCounter")
-	var total, cur int32 = 0, 0
-	go func() {
-		// run forever
-		for {
-			// Log activity only when the controllers were still working in the last 0.5s.
-			time.Sleep(500 * time.Millisecond)
-			if hcTot+obTot != total || cur != 0 {
-				log.Info("Activity", "hcTot", hcTot, "hcCur", hcCur, "obTot", obTot, "obCur", obCur)
-			}
-			total = hcTot + obTot
-			cur = hcCur + obCur
-		}
-	}()
 }
--- a/incubator/hnc/pkg/stats/stats.go
+++ b/incubator/hnc/pkg/stats/stats.go
@@ -1,0 +1,192 @@
+package stats
+
+import (
+	"time"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type object struct {
+	totalReconciles counter
+	curReconciles   counter
+	apiWrites       counter
+}
+
+type objects map[schema.GroupKind]*object
+
+type stat struct {
+	// actionID is the number of controller actions devided by idles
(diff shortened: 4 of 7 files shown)


incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,90 @@
...
+	HcWrite counter
+
+	// ObRead is the total number of object reads.
+	ObRead counter
+
+	// ObWrite is the total number of object writes.
+	ObWrite counter
+}
+
+var Stats stats
```
adrianludwin: Not sure if we really need this level of grouping? They could all just be top-level variables, so you'd say `stats.ObjRead.Incr()` instead of `stats.Stats.ObjRead.Incr()` - the second `Stats` doesn't really add anything.

That's assuming you don't want to switch to a more functional style that I recommended above, like `stats.Objects.StartReconcile()` or even just `stats.StartObjectReconcile()`.

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -499,6 +503,7 @@ func (r *HierarchyReconciler) writeNamespace(ctx context.Context, log logr.Logge
 		return false, nil
 	}
 
+	stats.Stats.HcWrite.Incr()
```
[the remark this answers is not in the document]
yiqigao217: Sure, it was tracked separately in the previous commit and here I put them together just because the writes both happen in the HC reconciliation.

Separating them sounds more reasonable!

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -12,15 +11,7 @@ import (
...
-var (
-	hcTot int32
-	hcCur int32
-	obTot int32
-	obCur int32
-)
-
-// The ex map is only used by reconcile counters for performance testing. We explicitly
+// The ex map is used by controllers to exclude namespaces to reconcile. We explicitly
 // exclude some default namespaces with constantly changing objects.
```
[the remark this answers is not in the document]
yiqigao217: sure

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,90 @@
...
+	log := ctrl.Log.WithName("reconcileCounter")
+	var total, cur counter = 0, 0
+	working := false
+	actN := 1
+	go func() {
+		// run forever
+		for {
+			// Log activity only when the controllers were still working in the last 0.5s.
+			time.Sleep(500 * time.Millisecond)
+			if Stats.HcTot+Stats.ObTot != total || cur != 0 {
```
adrianludwin: I find this logic a bit hard to read. Also, why are we comparing the _current_ values for the totals, but the _last_ values for `cur`?

Maybe rename `total` to `lastTotal`, and then start this block as follows:

```
total = HcTot + ObjTot
cur = HcCur + ObjCur
if total != lastTotal || cur > 0 {
  ...
}
lastTotal = total
```

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,90 @@
...
+	ObRead counter
+
+	// ObWrite is the total number of object writes.
+	ObWrite counter
+}
+
+var Stats stats
+
+// LogActivity generates logs for performance testing.
+func LogActivity() {
```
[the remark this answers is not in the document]
yiqigao217: sg

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,90 @@
...
+			// Log activity only when the controllers were still working in the last 0.5s.
+			time.Sleep(500 * time.Millisecond)
+			if Stats.HcTot+Stats.ObTot != total || cur != 0 {
+				// If the controller was previously idle, change its status and log it's started.
+				if working == false {
+					working = true
+					msg := "Activity-" + strconv.Itoa(actN) + "-started"
+					logActivity(log, msg, Stats)
+				}
+				logActivity(log, "Activity", Stats)
```
adrianludwin: I'm wondering if "activity" should be one of the structured fields in the log, and then also have a "status" field that can be either "start," "continue" or "finish". That'll make the logs more uniform and easier to grep, I think.

incubator/hnc/pkg/controllers/hierarchy_controller.go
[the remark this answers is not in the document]
adrianludwin: I think there will be many cases where we write one and not the other, though this isn't true for reads!

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -50,12 +50,14 @@ type ObjectReconciler struct {
...
-		atomic.AddInt32(&obTot, 1)
-		atomic.AddInt32(&obCur, 1)
-		defer atomic.AddInt32(&obCur, -1)
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
 	}
 
+	stats.Stats.ObTot.Incr()
+	stats.Stats.ObCur.Incr()
+	defer stats.Stats.ObCur.Decr()
```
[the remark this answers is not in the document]
yiqigao217: It's definitely more observable to have numbers by object types, but do we just need the logs for performance testing? Since the test case will only have 1 type created in all namespaces.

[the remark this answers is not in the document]
adrianludwin: Just discussed this offline but for the record - we'll probably use these same calls to generate metrics so it makes sense to group them by GVK too. Actually, on second thought, we don't really need the version, so just group+kind is probably best.

adrianludwin:
As discussed offline: maybe let's be a bit less terse with our variable names, unless they're purely inside the stats package. So for example, `stats.HcWrite.Incr()` is a bit confusing to see in the reconciler; I'd rather see something like `stats.WriteHierarchyConfig()`. Ditto in the logs - `HierarchyConfigWrites` is more self-descriptive than `HcWrites`. But any variable that's purely internal can be short as long as you document them.

incubator/hnc/pkg/stats/stats.go
yiqigao217: In the performance testing scripts, since the logs are grep by intervals (e.g. 5s), the activity start and finish have to be numbered to be able to map to the testing actions. Currently it's grep by e.g. "Activity-$i-finished".

Do you mean generating "activity" number the same way as generating "rid"? It does look more uniform. I will make the changes in this first and figure out how to grep the log in the next PR?

incubator/hnc/pkg/controllers/hierarchy_controller.go
```diff
@@ -529,6 +534,7 @@ func (r *HierarchyReconciler) updateObjects(ctx context.Context, log logr.Logger
 
 // getSingleton returns the singleton if it exists, or creates an empty one if it doesn't.
 func (r *HierarchyReconciler) getSingleton(ctx context.Context, nm string) (*api.HierarchyConfiguration, error) {
+	stats.Stats.HcRead.Incr()
```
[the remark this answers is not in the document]
yiqigao217: Sure, just for HC Reconcile, it's supposed to do exact two reads, 1 HC read and 1 NS read.

incubator/hnc/pkg/controllers/object_controller.go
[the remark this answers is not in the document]
yiqigao217: After discussion offline, I will break it by object type for not only the performance testing but also later metrics.

incubator/hnc/pkg/stats/stats.go
adrianludwin: Yeah, pretty much. You could probably do a grep like:

```
cat log | grep "\"activity\": \"5\"" | grep "\"status\": \"finished\"
```

which isn't really as pretty but is definitely more in keeping with how structured logs usually work. wdyt?

incubator/hnc/pkg/stats/stats.go
yiqigao217: I did more abstraction to have `stats.StartObjReconcile(gk)`, `stats.StartHierConfigReconcile()`, `stats.WriteHierConfig()`, etc. in the new change and kept `stats` to have `objects` field with `map[gk]*object` type.

incubator/hnc/pkg/stats/stats.go
yiqigao217: I used the `lastCur != 0` to still output the log for what happened in the past 0.5s. Otherwise, the last active log will be missing.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -50,12 +50,13 @@ type ObjectReconciler struct {
...
 func (r *ObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	if !ex[req.Namespace] {
-		atomic.AddInt32(&obTot, 1)
-		atomic.AddInt32(&obCur, 1)
-		defer atomic.AddInt32(&obCur, -1)
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
 	}
 
+	stats.StartObjReconcile(r.GVK.GroupKind())
```
adrianludwin: Nit: it's sometimes better to make the _caller_ look simpler. So you could say `stats.StartObjReconcile(r.GVK)` and just call `.GroupKind()` inside the function.

There's another school of thought that says "it's better to pass in the minimum amount of information so you don't accidentally come to depend on things (like Version) that you shouldn't care about." In this situation, I'd say that's unlikely to happen so I'd prioritize making the call sites simpler - but it's up to you, this is just a nit.

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -12,19 +11,12 @@ import (
...
 var ex = map[string]bool{
```
adrianludwin: Add a TODO comment to make this configurable

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+	i := int32(*c)
+	atomic.AddInt32(&i, 1)
+	*c = counter(i)
+}
+
+func (c *counter) decr() {
+	i := int32(*c)
+	atomic.AddInt32(&i, -1)
+	*c = counter(i)
+}
```
adrianludwin: Nit: I like to keep the methods of a type close to the definition of the type. I noticed you used to do that and then moved them here? You could also move these to counter.go if you don't want to clutter up the main file.

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+func StopHierConfigReconcile() {
+	stats.curHierConfigReconciles.decr()
+}
+
+// StartObjReconcile updates the stats for objects with common GK
+// when an object reconciliation starts.
+func StartObjReconcile(gk schema.GroupKind) {
+	if stats.objects == nil {
+		stats.objects = make(map[schema.GroupKind]*object)
+	}
```
adrianludwin: Why is this needed? Don't we always call `initStats` before starting the controllers?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+	"github.com/go-logr/logr"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type counter int32
+
+type object struct {
+	gk              schema.GroupKind
```
adrianludwin: Is this ever used?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+// when an object reconciliation finishes.
+func StopObjReconcile(gk schema.GroupKind) {
+	if stats.objects == nil {
+		stats.objects = make(map[schema.GroupKind]*object)
+	}
+	if _, ok := stats.objects[gk]; !ok {
+		stats.objects[gk] = &object{
+			gk: gk,
+		}
+	}
```
adrianludwin: This shouldn't be needed because there's no way to call `StopXXX` without first calling `StartXXX`, right?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+	stats.objects[gk].apiWrites.incr()
+}
+
+func initStats() {
+	objects := make(map[schema.GroupKind]*object)
+	for _, gvk := range config.GVKs {
+		objects[gvk.GroupKind()] = &object{
+			gk: gvk.GroupKind(),
+		}
+	}
```
adrianludwin: Since GVKs will become configurable and changeable in the future, and since you already have code to do this initialization in `StartXXX`, can we remove this?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+
+// StartLoggingActivity generates logs for performance testing.
+func StartLoggingActivity() {
+	initStats()
+	log := ctrl.Log.WithName("reconcileCounter")
+	var total, lastTotal, lastCur counter = 0, 0, 0
+	working := false
+	go func() {
+		// run forever
+		for {
```
adrianludwin: Suggestion: factor this out into its own function, then you can get rid of two extra levels of indentation.

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+			time.Sleep(500 * time.Millisecond)
+			total = stats.totalHierConfigReconciles + getTotalObjReconciles()
+			// If lastCur is not 0 yet, still generate a log for the past 0.5s.
+			if total != lastTotal || lastCur != 0 {
+				// If the controller was previously idle, change its status and log it's started.
+				if working == false {
+					working = true
+					logActivity(log, "start")
+				}
+				logActivity(log, "continue")
```
adrianludwin: Should this be an `else`?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+			} else {
+				// If the controller was previously working, change its status and log it's finished.
+				if working == true {
+					working = false
+					logActivity(log, "finish")
+					stats.actionID.incr()
```
adrianludwin: `actionID` doesn't _need_ to be a `counter` because it's only ever accessed from one thread. So you could just say `stats.actionID++` here if you wanted. Up to you.

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+				}
+			}
+			lastTotal = stats.totalHierConfigReconciles + getTotalObjReconciles()
```
adrianludwin: Maybe `lastTotal = total`?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -0,0 +1,187 @@
...
+			lastCur = stats.curHierConfigReconciles + getCurObjReconciles()
+		}
+	}()
+}
+
+func logActivity(log logr.Logger, status string) {
+	log.Info("Activity", "Action", stats.actionID, "Status", status, "HierConfigWrites", stats.hierConfigWrites, "NamespaceWrites", stats.namespaceWrites, "ObjectWrites", getObjWrites(), "TotalHierConfigReconciles", stats.totalHierConfigReconciles, "CurHierConfigReconciles", stats.curHierConfigReconciles, "TotalObjReconciles", getTotalObjReconciles(), "CurObjReconciles", getCurObjReconciles())
```
adrianludwin: Nit: maybe format this to make it easier to read?

```
log.Info("Activity",
  "Action", stats.actionID,
  "Status", status,
  ...
)
```

rjbez17:
This lgtm after comments from @adrianludwin  are addressed

incubator/hnc/pkg/stats/stats.go
yiqigao217: yes removed.

incubator/hnc/pkg/stats/stats.go
yiqigao217: agree, removed.

incubator/hnc/pkg/stats/stats.go
yiqigao217: I got error if I remove this part:
```
panic: assignment to entry in nil map [recovered]
        panic: assignment to entry in nil map
```
It's true that `stats.StartLoggingActivity()` is called before creating controllers. Maybe compiler cannot tell?

adrianludwin: That's a runtime error, not a compiler one. Which goroutine is generating that panic?

adrianludwin: Ah - if `testLog` is false, `stats.StartLoggingActivity()` is never called - maybe that's what you're seeing? If so, we can probably just say `if stats.objects == nil {return}` instead of creating it.

yiqigao217: ah right, exactly! ok will do that.

yiqigao217: This is done. I added a `testLogEnabled` flag and it's set to true in `stats.StartLoggingActivity()` func.

In each stats update func, it will check the flag first by:
```
if !testLogEnabled {
    return
}
```

PR #370, opened 2020-01-30
[Incubator][VC]Implement vNode Garbage Collection

Fei-Guo:
This change implements removing orphan vNode in tenant master. Orphan vNode means no Pods in the tenant master are bound to it, which can happen when vPod is deleted or recreated and bound to another vNode.  vNode is created by Syncer only. 

Why do we need to do this?
- Less pressure on sending vNode heartbeats to tenant master. If we don't do vNode GC, there could accumulate a lot of orphan vNodes in tenant master overtime.
- Better user serverless experience - you see what you use.

Technical challenges 
- K8s does not provide multi-objects consistency semantic. Even though Pod and the bound node are associated, APIServer does not provide an atomic way to perform "delete a node if no pods are bound to it". 
- From syncer's perspective, the biggest race is that uws tries to bind a vPod to a vNode and the GC thread tries to delete the vNode from tenant master. Without APISever support, it is impossible to completely resolve the race. In the worst case, a vPod is bound to a vNode and the vNode is removed by GC thread right away.  

Solutions
The problem can be divided to two parts:
- Find orphan vNodes: This is done by tracking vPod and vNode association in Pod mccontroller dws thread using `clusterVNodePodMap`. Upon vPod deletion event, if we found no more Pods are available in the vNode, the vNode is moved to `clusterVNodeGCMap` as a GC candidate. 
- GC vNode: To avoid accidentally deleting a vNode which is just bound to a vPod by uws, we introduce a grace period for removing an orphan vNode. A vNode GC candidate has two deletion phases, `VNodeQuiescing` and `VNodeDeleting`. 
    - In `VNodeQuiescing` phase, if any vPod is found associated with this vNode or uws attempts to bind a vPod on it, it is removed from `clusterVNodeGCMap`. 
    - If a vNode has been in `VNodeQuiescing` phase for more than `DefaultvNodeGCGracePeriod`, the GC thread will change the vNode to `VNodeDeleting` phase and issue the delete API to remove the vNode from tenant master. Once we confirmed the vNode is deleted, the vNode is removed from `clusterVNodeGCMap`.
    - If uws attempts to bind a vPod to a vNode and the vNode is in `VNodeDeleting` phase, uws will retry until the vNode is removed from APIServer.

The above solution still cannot handle a case - a uws bind API call for some reason takes more than `DefaultvNodeGCGracePeriod`, the race may still happen. However, if we set `DefaultvNodeGCGracePeriod` long enough, the race practically should be extremely rare. 

Finally, syncer restart may cause some orphan vNodes left in tenant master, and they will not be picked up by GC thread after syncer is restarted. To resolve the problem, we still periodically check tenant master to find missing orphan vNodes leveraging the Pod checker thread. This check can be somehow expensive and may be invoked with a much slower rate if necessary.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/syncer/constants"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/syncer/conversion"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/syncer/reconciler"
 )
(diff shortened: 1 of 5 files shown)


incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
```diff
@@ -48,12 +49,88 @@ func (c *controller) StartPeriodChecker(stopCh <-chan struct{}) error {
...
+	wait.Until(c.vNodeGCDo, c.periodCheckerPeriod, stopCh)
+}
+
+type Candidate struct {
+	cluster  string
+	nodeName string
+}
+
+func (c *controller) vNodeGCDo() {
+
```
adohe: should remove this blank line.

incubator/virtualcluster/pkg/syncer/resources/pod/controller.go
```diff
@@ -143,6 +164,73 @@ func (c *controller) enqueuePod(obj interface{}) {
...
+// c.Mutex needs to be Locked before calling addToClusterVNodeGCMap
+func (c *controller) addToClusterVNodeGCMap(cluster string, nodeName string) {
+	if _, exist := c.clusterVNodeGCMap[cluster]; !exist {
+		c.clusterVNodeGCMap[cluster] = make(map[string]VNodeGCStatus)
+	}
+	c.clusterVNodeGCMap[cluster][nodeName] = VNodeGCStatus{
+		QuiesceStartTime: metav1.Now(),
+		Phase:            VNodeQuiescing,
+	}
+
```
adohe: should remove this blank line.

incubator/virtualcluster/pkg/syncer/resources/pod/controller.go
```diff
@@ -143,6 +164,73 @@ func (c *controller) enqueuePod(obj interface{}) {
...
+}
+
+// c.Mutex needs to be Locked before calling removeFromClusterVNodeGCMap
```
adohe: removeFromClusterVNodeGCMap -> removeQuiescingNodeFromClusterVNodeGCMap

adohe:
overall lgtm except small nits.

incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
Fei-Guo: done

incubator/virtualcluster/pkg/syncer/resources/pod/controller.go
Fei-Guo: done

incubator/virtualcluster/pkg/syncer/resources/pod/controller.go
Fei-Guo: done

incubator/virtualcluster/pkg/syncer/resources/pod/uws.go
```diff
@@ -122,16 +122,27 @@ func (c *controller) backPopulate(key string) error {
...
-	// first check whether tenant pod has assigned.
-	if vPod.Spec.NodeName != pPod.Spec.NodeName {
+	// If tenant Pod has not been assigned, bind to virtual Node.
+	if vPod.Spec.NodeName == "" {
 		n, err := c.client.Nodes().Get(pPod.Spec.NodeName, metav1.GetOptions{})
 		if err != nil {
 			return fmt.Errorf("failed to get node %s from super master: %v", pPod.Spec.NodeName, err)
 		}
+		// We need to handle the race with vNodeGC thread here.
+		if err = func() error {
```
zhuangqh: why do you add a closure function here?

incubator/virtualcluster/pkg/syncer/resources/pod/controller.go
```diff
@@ -143,6 +164,72 @@ func (c *controller) enqueuePod(obj interface{}) {
...
+		c.Lock()
+		defer c.Unlock()
+		if event == reconciler.AddEvent || event == reconciler.UpdateEvent {
+			if _, exist := c.clusterVNodePodMap[cluster]; !exist {
+				c.clusterVNodePodMap[cluster] = make(map[string]map[string]struct{})
+			}
+			if _, exist := c.clusterVNodePodMap[cluster][nodeName]; !exist {
+				c.clusterVNodePodMap[cluster][nodeName] = make(map[string]struct{})
+			}
+			c.clusterVNodePodMap[cluster][nodeName][vPod.Name] = struct{}{}
```
zhuangqh: pod name is unique among each namespace?

incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
```diff
@@ -165,3 +242,46 @@ func (c *controller) checkPodsOfTenantCluster(clusterName string) {
...
+	if err != nil {
+		klog.Errorf("failed to list vNode from cluster %s config: %v", clusterName, err)
+		return
+	}
+
+	for _, vNode := range nodeList.Items {
+		if vNode.Labels[constants.LabelVirtualNode] != "true" {
+			continue
+		}
+		func() {
```
zhuangqh: why do you create a anonymous function here? if you want to create a new context, just put the code inside a new `{}`.

incubator/virtualcluster/pkg/syncer/resources/pod/controller.go
Fei-Guo: Nice catch. Fixed.

PR #372, opened 2020-01-31
[Incubator][VC] Update the apiserver's certificate of the Virtualcluster during the runtime

charleszheng44:
Update the apiserver's certificate of the Virtualcluster to allow the pod

to visit the apiserver from inside the cluster. New creation steps including,

  * wait for the default/kubernetes svc being synced to super master,
  * regenerate a certificate for the apiserver of the vc by adding the clusterIP of 
     the svc that is corresponding to the default/kubernetes on the super master
  * change the vc status to Updating,
  * delete the apiserver pod,
  * wait for apiserver pod being recreated by the statefulset controller,
  * change the vc status to Ready

Other minor changes,

  * change Makefile to allow building image for specific binaries
  * add auxiliary fields (e.g. shortNames) to vc and cv yamls
  * encapsulate common functionality (e.g. polling Statefulset, update VC status)

### DIFF
--- a/incubator/virtualcluster/Makefile
+++ b/incubator/virtualcluster/Makefile
@@ -78,7 +78,7 @@ release-images: test build-images
 # 1. build all binaries.
 # 2. copy binaries to the corresponding docker image.
 build-images:
-	hack/make-rules/release-images.sh
+	hack/make-rules/release-images.sh $(WHAT)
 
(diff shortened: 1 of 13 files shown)


charleszheng44:
instead of adding the clusterip during runtime, we will add the clusterip preflight #375

Issue #369, opened 2020-01-30
HNC: make client QPS limit configurable and set default limit at 50QPS


adrianludwin:
We discovered that HNC is very slow because the default client-go library limits you to 5 QPS by default with a burst rate of 10. This is very, very low if you have to sync a large number of namespaces, as will typically be the case for HNC.

We should make this configurable and set a good default limit. On GKE, apiservers for small clusters have a limit of 100QPS, and I observed unfortunate things happening when I drove HNC to this level (eg leader election started failing elsewhere on the cluster). So I'm going to propose a default of 50QPS, with the burst set to 50% higher than that (eg 75QPS).

adrianludwin:
@sophieliu15 - see this line: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/cmd/manager/main.go#L73

```
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
```

The config returned by `GetConfigOrDie` is what we need to modify. Something like this:

```
cfg := ctrl.GetConfigOrDie()
cfg.QPS = 50
cfg.Burst = cfg.QPS * 1.5
```

except instead of hardcoding these values, we should read them from a flag, and then set the default here: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/config/default/manager_auth_proxy_patch.yaml#L23

adrianludwin:
@rjbez17 Do you think we should hardcode an upper limit in HNC itself? Or just let people shoot themselves in the foot if they really want to (plus other K8s clusters might have a higher limit than small GKE clusters)?

rjbez17:
I generally think shooting yourself in the foot is fine if it’s clear what I used to do it with.

Basically, if we can identify that we are having issues because too high QPS and we log it, that’s totally enough in my opinion.

adrianludwin:
sgtm, thanks.

adrianludwin:
This is now done

/close

PR #377, opened 2020-02-03
HNC: Make API server QPS throttle configurable with a default value. 

sophieliu15:
Add command line flags for qps and burst limit for apiservers. Set the default qps to 50 and default burst to 75 respectively.

This is a fix for #369 which reports low HNC performance due to default limit of 5 QPS and 10 burst rate set by client-go library.

Test: Tested via debug logs to ensure the default values are set and the config values changes correctly based on the input from 
manager_autho_proxy_pathc.yaml 

/assign @adrianludwin

### DIFF
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -58,6 +58,7 @@ func main() {
 		debugLogs            bool
 		newObjectController  bool
 		testLog              bool
+		qps                  int
 	)
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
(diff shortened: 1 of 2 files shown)


adrianludwin:
Can you please squash your commits (https://blog.carbonfive.com/2017/08/28/always-squash-and-rebase-your-git-commits/)? It's usually best if there's only one commit per PR.

adrianludwin:
You'll need to force-push your changes, but that's fine

sophieliu15:
Thanks for the suggestions. Squashed commits into one.

yiqigao217:
The code changes lgtm! The only thing is that please copy the PR message also into the commit message.

sophieliu15:
@yiqigao217  Updated commit messages. Thanks for the review and suggestions.

incubator/hnc/cmd/manager/main.go
```diff
@@ -67,11 +69,17 @@ func main() {
 	flag.BoolVar(&newObjectController, "enable-new-object-controller", false, "Enables new object controller.")
 	flag.BoolVar(&testLog, "enable-test-log", false, "Enables test log.")
 	flag.IntVar(&maxReconciles, "max-reconciles", 1, "Number of concurrent reconciles to perform.")
+	flag.IntVar(&qps, "qps", 50, "The maximum QPS to the API server.")
+	flag.IntVar(&burst, "burst", 75, "Maximum burst for throttle.")
```
adrianludwin: I think my preference would be just have a single number (qps) and always set "burst" to maybe qps*1.5.

Also, "--qps" doesn't really mean a lot on the command line - it's not self-descriptive. How about `--apiserver-qps-throttle`? That makes it clear what we're protecting with this flag. Internally we can still call it the same thing.

sophieliu15: Thanks for the suggestion. Please see the updated code.

incubator/hnc/cmd/manager/main.go
```diff
@@ -67,11 +68,16 @@ func main() {
...
 	flag.IntVar(&maxReconciles, "max-reconciles", 1, "Number of concurrent reconciles to perform.")
+	flag.IntVar(&qps, "apiserver-qps-throttle", 50, "The maximum QPS to the API server.")
 	flag.Parse()
 
 	ctrl.SetLogger(zap.Logger(debugLogs))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	cfg := ctrl.GetConfigOrDie()
+	cfg.QPS = float32(qps)
+	cfg.Burst = int(cfg.QPS * 1.5)
```
adrianludwin: Maybe add a comment here, something like:

```
// By default, Burst is about 2x QPS, but since HNC's "bursts" can last for ~minutes
// we need to raise the QPS param to be much higher than we ordinarily would. As a
// result, doubling this higher threshold is probably much too high, so lower it to a more
// reasonable number.
//
// TODO: better understand the behaviour of Burst, and consider making it equal to QPS if
// it turns out to be harmful.
```

sophieliu15: Thanks for the comments. Does "~minutes" mean several minutes or was there a number missing (e.g., ~2 minutes)? Do we have a rough metric about how long it will last?

adrianludwin:
I just meant several minutes. Maybe O(minutes) would have been better but
(despite having been at Google for three years) I still feel like a
disappointment to Computer Science whenever I write that.

incubator/hnc/cmd/manager/main.go
sophieliu15: Cool! I will leave it like that.

PR #380, opened 2020-02-04
Remove old object controller and the flag

yiqigao217:
The old object controller is removed. The flags to toggle the old/new
object controllers are also cleaned.

Tested by 'make test' and manual testing on GKE cluster.


incubator/hnc/config/rbac/role.yaml
```diff
@@ -13,13 +13,10 @@ rules:
   verbs:
   - create
   - delete
-  - deletecollection
   - get
-  - impersonate
   - list
   - patch
   - update
-  - use
```
yiqigao217: Not sure if this needs to be checked in.

adrianludwin: Weird. I think I removed those permissions a while ago and just forgot to regenerate the files? Maybe remove this from the commit if you can and resubmit separately?

yiqigao217: Done.

PR #387, opened 2020-02-05
Add performance tests on controller start/restart

yiqigao217:
Add scripts to load different topologies of namespaces into the cluster.
Start/restart controllers and output completion time and number of api
reads/writes to see the performance.

Tested on GKE cluster.


adrianludwin:
This lgtm but it would be nice if a non-Googler could have a look. Ryan, will you have a chance this week? If not I'll just approve it myself.

/assign @rjbez17

Issue #330, opened 2019-12-20
VC: ignore object labels which were set by super master controller


zhuangqh:
tenant's Pods in super master may be attached some label by super master controller.
We should ignore these label when computing the diff of pPod and vPod.

zhuangqh:
Implement tips
1. create a flag to receive a slice of string prefix, which would be used to filter object label or annotations. may named 'standalone-object-annotation-prefixes'
2. change the kvequality checker here https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/virtualcluster/pkg/syncer/conversion/equality.go#L108
3. find a way this pass the prefix parameters to the EqualityChecker.
I prefer firstly create a context object, then do equality check. The user interface like
`Syncer(prefix parameter).CheckEquality(pPod, vPod)`  because there are many place use this func.

`CheckEquality(config, pPod, vPod)` this design worse


if the key match the prefix, just ignore this key. Let this key control by their master independently.

zhuangqh:
@charleszheng44

charleszheng44:
PR #389

PR #389, opened 2020-02-06
[Incubator][VC] ignore annotations that have specific prefixes 

charleszheng44:
Try to resolve issue #330 

when comparing virtual object and the real object, syncer will ignore annotation
prefixes specified in env `STDLN_OBJ_ANNO_PREFXS`

### DIFF
--- a/incubator/virtualcluster/config/setup/all_in_one.yaml
+++ b/incubator/virtualcluster/config/setup/all_in_one.yaml
@@ -239,11 +239,11 @@ spec:
           imagePullPolicy: Always
           name: vc-syncer
           env:
-          - name: STDLN_OBJ_ANNO_PREFXS
+          - name: STANDALONE_OBJ_ANNO_PREFIXES
             valueFrom: 
               configMapKeyRef:
                 name: syncer-envs
-                key: STDLN_OBJ_ANNO_PREFXS
+                key: STANDALONE_OBJ_ANNO_PREFIXES
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -252,7 +252,7 @@ metadata:
   namespace: vc-manager
 data:
   # use space as delimiter. e.g. STDLN_OBJ_ANNO_PREFXS: "a b c"
-  STDLN_OBJ_ANNO_PREFXS: "" 
+  STANDALONE_OBJ_ANNO_PREFIXES: "" 
 ---
 apiVersion: apps/v1
 kind: DaemonSet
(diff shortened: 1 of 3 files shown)


incubator/virtualcluster/config/setup/all_in_one.yaml
```diff
@@ -238,6 +238,21 @@ spec:
           image: virtualcluster/syncer-amd64
           imagePullPolicy: Always
           name: vc-syncer
+          env:
+          - name: STDLN_OBJ_ANNO_PREFXS
```
zhuangqh: how about `STDIN_OBJ_ANNO_PREFIXES` ?

PREFXS  this abbreviation is confusing

incubator/virtualcluster/pkg/syncer/conversion/equality.go
```diff
@@ -25,8 +26,25 @@ import (
...
+	// load the StandaloneObjAnnotationPrefixex from envs. When the syncer
+	// run in a pod, the env will be set through a configmap
+	// TODO should we add a feature to restart the syncer if the configmap
+	// is updated
+	soap := os.Getenv("STDLN_OBJ_ANNO_PREFXS")
+	if soap == "" {
+		return
+	}
+	// the delimiter of the prefix list is space
+	StandaloneObjAnnotationPrefixes = append(StandaloneObjAnnotationPrefixes, strings.Split(soap, " ")...)
```
zhuangqh: dedup and remove space-only prefix

may there is a continuous space.

incubator/virtualcluster/pkg/syncer/utils/string.go
```diff
@@ -45,3 +48,13 @@ func GenAnAvailableName(names []string, prefix string, n int) string {
 		}
 	}
 }
+
+// HasPrefixs checks whether the string 's' begins with any prefix in 'prefixs'.
+func HasPrefixs(s string, prefixs []string) bool {
```
zhuangqh: HasPrefixes

incubator/virtualcluster/pkg/syncer/conversion/equality.go
charleszheng44: use `strings.Field` instead

incubator/virtualcluster/pkg/syncer/utils/string.go
charleszheng44: typo fixed

incubator/virtualcluster/config/setup/all_in_one.yaml
charleszheng44: change to `STDLN_OBJ_ANNO_PREFIXES`

PR #396, opened 2020-02-10
Set stats.Init() as default

yiqigao217:
Previously, the stats were only initialized to serve data for logging
activities when the "enable-test-log" flag is set. Since we are going to
use the stats data to export metrics, the stats are now set as default
without flags.

Tested on GKE cluster.

### DIFF
--- a/incubator/hnc/pkg/stats/stats.go
+++ b/incubator/hnc/pkg/stats/stats.go
@@ -35,17 +35,11 @@ type stat struct {
 	objects objects
 }
 
-var (
-	testLogEnabled bool
-	stats          stat
-)
+var stats stat
 
 // StartHierConfigReconcile updates stats when hierarchyConfig
 // reconciliation starts.
 func StartHierConfigReconcile() {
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/stats/stats.go
```diff
@@ -76,39 +64,28 @@ func StartObjReconcile(gvk schema.GroupVersionKind) {
...
-	if !testLogEnabled {
-		return
-	}
 	gk := gvk.GroupKind()
 	stats.objects[gk].apiWrites.incr()
 }
 
-func initStats() {
+// Init initializes the stat counters.
+func Init() {
```
adrianludwin: If you just call this `init()`, it will be called when HNC starts and you don't need to call it manually (e.g. see https://medium.com/golangspec/init-functions-in-go-eac191b3860a)

yiqigao217: Got you, thanks! Fixed.

PR #410, opened 2020-02-11
Add OpenCencus metrics and export to Stackdriver

yiqigao217:
Export hierconfig_reconcile_total and hierconfig_writes_total metrics to
Stackdriver using OpenCensus. The metrics come from stats counters.

Tested on GKE cluster. I can see the hnc metrics in Stackdriver and the
data changed when new namespaces were created or set parent.

### DIFF
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -19,6 +19,7 @@ import (
 	"flag"
 	"os"
 
+	"contrib.go.opencensus.io/exporter/stackdriver"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
@@ -68,6 +69,25 @@ func main() {
 	flag.IntVar(&maxReconciles, "max-reconciles", 1, "Number of concurrent reconciles to perform.")
 	flag.IntVar(&qps, "apiserver-qps-throttle", 50, "The maximum QPS to the API server.")
 	flag.Parse()
+
+	// Enable OpenCensus exporters to export metrics
+	// to Stackdriver Monitoring.
+	// Exporters use Application Default Credentials to authenticate.
+	// See https://developers.google.com/identity/protocols/application-default-credentials
+	// for more details.
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{})
+	if err != nil {
+		setupLog.Error(err, "cannot create Stackdriver exporter")
+		os.Exit(1)
+	}
+	// Flush must be called before main() exits to ensure metrics are recorded.
+	defer exporter.Flush()
+
+	if err := exporter.StartMetricsExporter(); err != nil {
+		setupLog.Error(err, "cannot start metric exporter")
+		os.Exit(1)
+	}
+	defer exporter.StopMetricsExporter()
 
 	ctrl.SetLogger(zap.Logger(debugLogs))
 
--- a/incubator/hnc/pkg/stats/metrics.go
+++ b/incubator/hnc/pkg/stats/metrics.go
@@ -1,0 +1,44 @@
+package stats
+
+import (
+	"context"
+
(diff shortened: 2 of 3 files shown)


incubator/hnc/cmd/manager/main.go
```diff
@@ -69,6 +70,25 @@ func main() {
...
+	exporter, err := stackdriver.NewExporter(stackdriver.Options{})
+	if err != nil {
+		setupLog.Error(err, "cannot create Stackdriver exporter")
+		os.Exit(1)
+	}
+	// Flush must be called before main() exits to ensure metrics are recorded.
+	defer exporter.Flush()
+
+	if err := exporter.StartMetricsExporter(); err != nil {
+		setupLog.Error(err, "cannot start metric exportor")
```
adrianludwin: s/exportor/exporter/

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+var (
+	hierConfigReconcileTotal = ocstats.Int64("hierconfig_reconcile_total", "The number of HierConfig reconciliations happened", "times")
+	hierConfigWritesTotal    = ocstats.Int64("hierconfig_writes_total", "The number of HierConfig writes happened during HierConfig reconciliations", "times")
+)
+
+var (
+	hierReconcileView = &ocview.View{
+		Name:        "hnc/hierconfig_reconciler/reconcile_total",
```
adrianludwin: Nit suggestion: about about `hnc/reconcilers/hierconfig/total`? Why does "reconcile" need to be repeated?

The alternative to `reconcilers` will eventually be things like `validators` and (maybe) `mutators`, while the alternatives to `hierconfig` will be `objects` and `hncconfig`.

yiqigao217: Sg, done. The other one is renamed as `hnc/reconcilers/hierconfig/hierconfig_writes_total` since there is also a `hnc/reconcilers/hierconfig/namespace_writes_total` in addition.

adrianludwin: sgtm

rjbez17:
Any reason we are only supporting stackdriver? Also, I thought we were going to use OpenTelemetry (Merging of OpenCensus and OpenTracing)

adrianludwin:
> Any reason we are only supporting stackdriver? Also, I thought we were going to use OpenTelemetry (Merging of OpenCensus and OpenTracing)

Ah sorry for the lack of update:

- Stackdriver: it's a priority for us (you know... GKE and all...).
- Why not exporting to Prometheus? We just haven't figured out a) how to install and configure it properly and b) how to get it to play nicely with the Prometheus endpoint exposed by `controller-runtime`. Either we'll get to it or we'll welcome anyone who wants to figure out how Prometheus export should work.
- Why not OpenTelemetry? Simply because it's not mature enough yet, apparently - no Stackdriver exporter, and the OT Collector isn't stable yet either (plus it would require a second deployment, which we wanted to avoid).

Does that all make sense?

rjbez17:
> > Any reason we are only supporting stackdriver? Also, I thought we were going to use OpenTelemetry (Merging of OpenCensus and OpenTracing)
> 
> Ah sorry for the lack of update:
> 
>     * Stackdriver: it's a priority for us (you know... GKE and all...).
> 
>     * Why not exporting to Prometheus? We just haven't figured out a) how to install and configure it properly and b) how to get it to play nicely with the Prometheus endpoint exposed by `controller-runtime`. Either we'll get to it or we'll welcome anyone who wants to figure out how Prometheus export should work.
> 
>     * Why not OpenTelemetry? Simply because it's not mature enough yet, apparently - no Stackdriver exporter, and the OT Collector isn't stable yet either (plus it would require a second deployment, which we wanted to avoid).
> 
> 
> Does that all make sense?

Makes sense, but I do think Prometheus is a requirement for this to be GA.

adrianludwin:
> Makes sense, but I do think Prometheus is a requirement for this to be GA.

sgtm, filed #416

Issue #411, opened 2020-02-11
HNC: Implement HNC type configuration


sophieliu15:
HNC type configuration makes it possible for cluster administrators to configure the propagation behavior for each KIND by HNC. This is a cluster-wide configuration.

Design doc: http://bit.ly/hnc-type-configuration

sophieliu15:
/close

This issue is fixed.

PR #412, opened 2020-02-11
Define HNCConfiguration CR and set up a reconciler for the CR

sophieliu15:
This PR implements HNC type configuration by creating HNCConfiguration CR and setting up a corresponding reconciler for it. Specifically, it does the following:

 - Implements the Go structs for HNCConfiguration and generates corresponding yaml files by running "make manifests".
 - Creates a basic reconciler for HNCConfiguration and configures it in setup.go. The reconciler does the following:
    - Creates a default singleton if it does not exist. 
    - Writes or updates the singleton on the apiserver.
    - Logs the Spec of the HNCConfiguration when a reconciliation happens.

Tested:
  - Verified that the CRD were created correctly.
  - Created a sample HNCConfiguration and applied to KIND.
  - Verified the logs are expected on a GKE cluster during a reconciliation.

Issue: #411


incubator/hnc/pkg/config/default_config.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+// GetDefaultConfig creates the default configuration for HNCConfiguration Spec.
+// By default, HNC configuration should always propagate Roles and RoleBindings.
+// See details in http://bit.ly/hnc-type-configuration
+func GetDefaultConfig() *api.HNCConfigurationSpec {
+	return &api.HNCConfigurationSpec{
```
adrianludwin: Can you please switch to two-space tabs to align with the rest of the code? Thanks!

incubator/hnc/pkg/config/default_config.go
```diff
@@ -0,0 +1,16 @@
+package config
+
+import (
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+// GetDefaultConfig creates the default configuration for HNCConfiguration Spec.
+// By default, HNC configuration should always propagate Roles and RoleBindings.
+// See details in http://bit.ly/hnc-type-configuration
+func GetDefaultConfig() *api.HNCConfigurationSpec {
```
adrianludwin: This is only returning the spec, so it should probably be called `GetDefaultConfigSpec` or something like that. Also, no need to return a pointer if the caller doesn't need it to be one (and I don't _think_ it does).

incubator/hnc/pkg/controllers/config_controller.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	ctx := context.Background()
+	inst, err := r.getSingleton(ctx)
+	if err != nil {
+		r.Log.Error(err, "Couldn't read singleton.")
+		return ctrl.Result{}, nil
+	}
+
+	r.Log.Info("Reconciling cluster-wide HNC configuration.")
```
adrianludwin: Add "TODO" here to actually modify the other reconcilers

incubator/hnc/pkg/config/default_config.go
adrianludwin: I have no idea what I was thinking. Ignore this.

rjbez17:
lgtm after that last few comments resolved

incubator/hnc/pkg/controllers/setup.go
```diff
@@ -55,5 +55,14 @@ func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int) error {
 		return fmt.Errorf("cannot create Hierarchy controller: %s", err.Error())
 	}
 
+	// Create the ConfigReconciler, passing it the object reconcillers so it can call them.
```
yiqigao217: Remove ", passing it the object reconcillers so it can call them" in the comment?

incubator/hnc/config/rbac/role.yaml
```diff
@@ -13,13 +13,10 @@ rules:
   verbs:
   - create
   - delete
-  - deletecollection
   - get
-  - impersonate
   - list
   - patch
   - update
-  - use
```
yiqigao217: Do we need to exclude role.yaml to a single PR by itself? @adrianludwin

adrianludwin: That would be nice - Yiqi, do you mind submitting that PR? Sophie, would you please exclude this file from your commit?

yiqigao217: sure will do shortly.

sophieliu15: Reverted this file. Thanks Yiqi for the clean up!

incubator/hnc/pkg/controllers/setup.go
sophieliu15: Good catch! Removed.

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: Done. Feel free to take a look.

incubator/hnc/pkg/config/default_config.go
sophieliu15: Thanks for the suggestion. Changed the name and returned the object instead of pointer. You are right, it does not need to be a pointer.

sophieliu15:
Thanks everyone for the review.

@adrianludwin  and @yiqigao217 I just updated code based on your comments. Feel free to take a look!

incubator/hnc/config/samples/hnc_v1_config.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: config
+spec:
+  types:
+    - apiVersion: v1
+      kind: Secret
+      mode: propagate
+    - apiVersion: rbac.authorization.k8s.io/v1
+      kind: Role
+      mode: ignore
```
adrianludwin: Oh, can you reverse these two? Role must always be `propagate`, but Secret can be `ignore`.

incubator/hnc/pkg/controllers/config_controller.go
adrianludwin: lgtm

incubator/hnc/config/samples/hnc_v1_config.yaml
sophieliu15: Sure! Done.

PR #414, opened 2020-02-12
Update the generated role.yaml

yiqigao217:
This update is generated from a previous change.

### DIFF
--- a/incubator/hnc/config/rbac/role.yaml
+++ b/incubator/hnc/config/rbac/role.yaml
@@ -13,13 +13,10 @@ rules:
   verbs:
   - create
   - delete
-  - deletecollection
   - get
-  - impersonate
   - list
   - patch
   - update
-  - use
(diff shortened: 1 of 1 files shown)


adrianludwin:
Looks like this was a consequence of #380  - the old object controller (incorrectly) had these permissions, so when it was deleted, they should have been removed from the generated files as well. Not sure why they weren't 🤷‍♂️

yiqigao217:
> Looks like this was a consequence of #380 - the old object controller (incorrectly) had these permissions, so when it was deleted, they should have been removed from the generated files as well. Not sure why they weren't

Ah now it makes sense. The file was generated then in #380 . See history here: https://github.com/kubernetes-sigs/multi-tenancy/pull/380#pullrequestreview-353199173. We thought it was supposed to be generated from even previous changes and thus excluded from #380

Issue #416, opened 2020-02-12
HNC: export from OpenCensus to Prometheus


adrianludwin:
cc @rjbez17 @yiqigao217 

As of #410, HNC is now able to exporting metrics to Stackdriver. It doesn't yet support Prometheus because:

1. None of us HNC developers were able to install and configure Prometheus in a reasonable amount of time; 
2. We couldn't figure out how to get OpenCensus' Prometheus endpoint to play nicely with controller-runtime's Prometheus endpoint. Maybe we can expose them on different ports?

For anyone with Prometheus experience, it's probably fairly easy to fix this problem but we don't currently have the expertise to do it ourself.

/good-first-issue

rjbez17:
Sounds like I can actually be useful! I'll try to take a wack at this next week.

yeya24:
@rjbez17  Hi, I think I can work on this issue this week. Is it okay?

rjbez17:
I’ll never complain about someone doing work for me faster than me! Go for it

yeya24:
I am checking this issue now. But I am wondering what metrics do we have currently? If we don't have any metrics except the metrics from controller-runtime, why we need to add Prometheus?

adrianludwin:
Hi Ben, have you worked with metrics and Prometheus before?

We have metrics exported from controller-runtime on a Prometheus endpoint,
and also from OpenCensus via Stackdriver (see
https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/pkg/stats/metrics.go).
We'd like to also export from OpenCensus to Prometheus, but without
disrupting the endpoint from controller-runtime. Does that all make sense?
If not, it's probably not a good starter project on HNC since this area is
quite complicated if you don't already know all these monitoring systems.
Thanks!

yeya24:
@adrianludwin  Hi, so you want to add the metrics from Stackdriver to the /metrics endpoint exposed by controller-runtime?

adrianludwin:
If possible, yes. Otherwise, we'd like to expose a *second* /metrics
endpoint (e.g. on another port) and have instructions for how to configure
Prometheus to use both.

yeya24:
Yes, I have implemented this to use the same /metrics. But I am not familiar with StackDriver so I am not sure whether it works or not. I don't have an env for testing

adrianludwin:
Cool - maybe throw up a PR and we can see if StackDriver is still working?

[...]

PR #418, opened 2020-02-12
Add all controller metrics

yiqigao217:
This commit adds metrics of HierConfig and all other object controllers.
It includes total reconciliations and peak concurrent reconciliations
and related api writes.

Tested on GKE cluster and the metrics can all be viewed in Stackdriver.


incubator/hnc/pkg/stats/metrics.go
```diff
@@ -2,22 +2,45 @@ package stats
...
 		Measure:     hierConfigReconcileTotal,
-		Description: "The number of HierConfig reconciliations happened",
+		Description: "The total number of HierConfig reconciliations happened",
+		Aggregation: ocview.LastValue(),
+	}
+
+	hierReconcileConcurrentView = &ocview.View{
+		Name:        "hnc/reconcilers/hierconfig/concurrent_peak",
+		Measure:     hierConfigReconcileConcurrent,
+		Description: "The peak concurrent HierConfig reconciliations happened in the last reporting period",
```
adrianludwin: Maybe define what the "reporting period" is? Is it configurable?

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -2,22 +2,45 @@ package stats
...
+	objectWritesTotal             = ocstats.Int64("object_writes_total", "The number of object writes happened during object reconciliations", "writes")
 )
 
 var (
-	hierReconcileView = &ocview.View{
+	maxConcurrentHierConfigReconcile int64
+	maxConcurrentObjectReconcile     map[schema.GroupKind]int64
+)
+
+var KeyGK, _ = tag.NewKey("GK")
```
adrianludwin: Doc comment?

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -2,22 +2,45 @@ package stats
...
 	ocview "go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 var (
-	hierConfigReconcileTotal = ocstats.Int64("hierconfig_reconcile_total", "The number of HierConfig reconciliations happened", "times")
-	hierConfigWritesTotal    = ocstats.Int64("hierconfig_writes_total", "The number of HierConfig writes happened during HierConfig reconciliations", "times")
+	hierConfigReconcileTotal      = ocstats.Int64("hierconfig_reconcile_total", "The total number of HierConfig reconciliations happened", "reconciliations")
```
adrianludwin: Maybe worth some higher-level comment that each of these stats has exactly one view (if that's correct)

incubator/hnc/pkg/stats/metrics.go
adrianludwin: Also maybe `KeyGroupKind` (if that's what is is) would be nice? I don't feel strongly though, up to you.

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -27,18 +50,84 @@ var (
...
+		objectReconcileConcurrentView,
+		objectWritesView,
+	); err != nil {
 		log.Error(err, "Failed to register the views")
 	}
+
+	go resetPeriodicMax()
+}
+
+func recordMetrics(m int64, ms *ocstats.Int64Measure) {
```
adrianludwin: Why is this `recordMetrics`, plural? Why not `recordMetric`?

Some comments about what each of these functions are for might be nice.

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -27,18 +50,84 @@ var (
...
 }
 
-func recordMetricsInt64(c counter, ms *ocstats.Int64Measure) {
-	ocstats.Record(context.Background(), ms.M(int64(c)))
+func recordTagMetrics(m int64, ms *ocstats.Int64Measure, k tag.Key, v string) {
+	ctx, _ := tag.New(context.Background(), tag.Insert(k, v))
+	ocstats.Record(ctx, ms.M(m))
+}
+
+func max(tmpMax int64, cur counter) int64 {
```
adrianludwin: This isn't used in this file - maybe move it to the other file?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -42,7 +42,12 @@ var stats stat
 func StartHierConfigReconcile() {
 	stats.totalHierConfigReconciles.incr()
 	stats.curHierConfigReconciles.incr()
-	recordMetricsInt64(stats.totalHierConfigReconciles, hierConfigReconcileTotal)
+
+	m := int64(stats.totalHierConfigReconciles)
+	recordMetrics(m, hierConfigReconcileTotal)
+	// Only update the maxConcurrent value in StartReconcile() since it's impossible to
+	// get it from StopReconcile().
```
adrianludwin: Why's it impossible?

incubator/hnc/pkg/stats/stats.go
```diff
@@ -72,22 +84,32 @@ func StopObjReconcile(gvk schema.GroupVersionKind) {
...
+	recordMetrics(m, namespaceWritesTotal)
 }
 
 // WriteHierConfig updates stats when writing hierarchyConfig instance.
 func WriteHierConfig() {
 	stats.hierConfigWrites.incr()
-	recordMetricsInt64(stats.hierConfigWrites, hierConfigWritesTotal)
+
+	m := int64(stats.hierConfigWrites)
+	recordMetrics(m, hierConfigWritesTotal)
```
adrianludwin: Are you always starting with an `int`? If so, why not have `recordMetrics` accept `int` and cast it inside the function?

yiqigao217: The temp max values are using int64, the same as the metrics type. Since `recordMetrics()` also records the int64 max value, I moved the cast outside of the function.

I do feel it's verbose too. I will change the max types from int64 to counter, so they can all be cast inside the function.

incubator/hnc/pkg/stats/stats.go
yiqigao217: Updated the comment with "since it's impossible to get the max from StopReconcile() when the reconcile number decrements"

incubator/hnc/pkg/stats/metrics.go
yiqigao217: It's configurable in the exporter. I added the 1 minute `ReportingInterval` and commented that the minimum Stackdriver reporting period must be above 60 seconds.

incubator/hnc/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/kubernetes-sigs/multi-tenancy/incubator/hnc
 go 1.12
 
 require (
+	contrib.go.opencensus.io/exporter/stackdriver v0.13.0
```
adrianludwin: Why wasn't this part of the last PR?

incubator/hnc/pkg/stats/metrics.go
adrianludwin: Since we're hardcoding this as 60s, maybe just say that in the comment? Eg, "this is the peak concurrent  in the past 60s, which is also the minimum Stackdriver reporting period and the one we're using"?

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -27,18 +49,90 @@ var (
...
+// The tag is used to group and filter collected metrics later on.
+func recordTagMetric(m counter, ms *ocstats.Int64Measure, k tag.Key, v string) {
+	ctx, _ := tag.New(context.Background(), tag.Insert(k, v))
+	ocstats.Record(ctx, ms.M(int64(m)))
+}
+
+func resetPeriodicPeak() {
+	for {
+		// This runs forever. It records and resets the peakConcurrent_ values every 1 minute,
+		// which is the same as the Stackdriver exporter's reporting interval.
```
adrianludwin: Any way we can create a constant and use it in all the places it's needed?

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -27,18 +49,90 @@ var (
...
+}
+
+// recordTagMetric inserts a tag to the context before recording a metric.
+// The tag is used to group and filter collected metrics later on.
+func recordTagMetric(m counter, ms *ocstats.Int64Measure, k tag.Key, v string) {
+	ctx, _ := tag.New(context.Background(), tag.Insert(k, v))
+	ocstats.Record(ctx, ms.M(int64(m)))
+}
+
+func resetPeriodicPeak() {
```
adrianludwin: I think this might make more sense as "recordPeakConcurrentReconciles" or something like that. This function doesn't _just_ reset - it actually records as well.

incubator/hnc/pkg/stats/stats.go
```diff
@@ -57,9 +62,15 @@ func StartObjReconcile(gvk schema.GroupVersionKind) {
 	gk := gvk.GroupKind()
 	if _, ok := stats.objects[gk]; !ok {
 		stats.objects[gk] = &object{}
+		peakConcurrentObjectReconcile[gk] = 0
```
adrianludwin: Is this needed? I _think_ `m[k] = max(m[k], foo)` will work even if the key `k` doesn't exist yet, because `m[k]` will evaluate to the default value, which for an integer is zero. But I'm not certain.

incubator/hnc/go.mod
yiqigao217: It should be included in the last PR, but I excluded them.

incubator/hnc/pkg/stats/metrics.go
yiqigao217: Yes, correct. Will update the comment.

incubator/hnc/pkg/stats/metrics.go
yiqigao217: sure, sg.

incubator/hnc/go.mod
adrianludwin: Thanks

incubator/hnc/pkg/stats/metrics.go
```diff
@@ -27,18 +53,96 @@ var (
...
+	ocstats.Record(ctx, ms.M(int64(m)))
+}
+
+func recordPeakConcurrentReconciles() {
+	for {
+		// This runs forever. It records and resets the peakConcurrent_ values every 1 minute,
+		// which is the same as the Stackdriver exporter's reporting interval.
+		time.Sleep(ReportingInterval)
+		peak.lock.Lock()
+		defer peak.lock.Unlock()
```
adrianludwin: `defer` is only invoked when the _function_ exits, not the _scope_. So I think you need an explicit call to Unlock here.

adrianludwin:
lgtm!
/assign @rjbez17 
/unassign @adrianludwin

yiqigao217:
Hi @rjbez17, these are the rest of the controller metrics exported to Stackdriver. The prometheus metrics will be added through #416 .

rjbez17:
🎉 

/lgtm
/approve

PR #424, opened 2020-02-14
Dynamically create object reconcilers for new types.

sophieliu15:
This PR creates object reconciler(s) when new type(s) are added to the
HNCConfiguration and informs the hierarchy reconciler about the new
object reconciler(s).

Specifically, it does the following tasks:
  - Dynamically creates an object reconciler when a new type is added
    in HNCConfiguration.
  - Informs the hierarchy reconciler about the newly added object
    reconciler so the corresponding objects can be propagated correctly
    when the hierarchy structure changes.

Test: Verified on a GKE cluster that
   - When adding a new type in HNCConfiguration, corresponding object
     reconciler is created successfully.
   - When adding a new object of the same type to a node in the namespace
     tree, the object is propagated correctly to the descendants of the
     node.
   - When the tree structure changes (e.g., add another descendant to the
     node above), the new object of the same type is propagated
     correctly.

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/controllers/config_controller.go
+++ b/incubator/hnc/pkg/controllers/config_controller.go
@@ -2,15 +2,20 @@ package controllers
 
 import (
 	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 )
 
 // ConfigReconciler is responsible for determining the HNC configuration from the HNCConfiguration CR,
@@ -18,7 +23,16 @@ import (
 // It can also set the status of the HNCConfiguration CR.
 type ConfigReconciler struct {
 	client.Client
-	Log logr.Logger
+	Log     logr.Logger
+	Manager ctrl.Manager
+
+	// Forest is the in-memory data structure that is shared with all other reconcilers.
+	Forest *forest.Forest
+
+	// HierarchyConfigUpdates is a channel of events used to update hierarchy configuration changes performed by
+	// ObjectReconcilers. It is passed on to ObjectReconcilers for the updates. The ConfigReconciler itself does
+	// not use it.
+	HierarchyConfigUpdates chan event.GenericEvent
 }
 
 // Reconcile sets up some basic variable and logs the Spec.
@@ -42,13 +56,23 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// achieve the reconciliation.
 	r.Log.Info("Reconciling cluster-wide HNC configuration.")
 
+	// Create cooresponding ObjectReconcilers for newly added types, if needed.
+	// TODO: Returning an error here will make controller-runtime to requeue the object to be reconciled again.
+	// A better way to handle this is to display the error as part of the HNCConfig.status field to suggest the error is a
+	// permanent problem and to stop controller-runtime from retrying.
+	// TODO: Rename the method syncObjectReconcilers because we might need more than creating ObjectReconcilers in future.
+	// For example, we might need to delete an ObjectReconciler if its corresponding type is deleted in the HNCConfiguration.
+	if err := r.createObjectReconcilers(inst); err != nil {
+		r.Log.Error(err, "Couldn't create the new object reconciler.")
+		return ctrl.Result{}, nil
+	}
+
 	// Write back to the apiserver.
+	// TODO: Update HNCConfiguration.Status before writing the singleton back to the apiserver.
 	if err := r.writeSingleton(ctx, inst); err != nil {
 		r.Log.Error(err, "Couldn't write singleton.")
 		return ctrl.Result{}, err
 	}
-
-	r.logSpec(inst)
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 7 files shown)


adrianludwin:
/assign @rjbez17 

Note that this change will also stop syncing everything except Roles and RoleBindings by default.

incubator/hnc/pkg/controllers/config_controller.go
```diff
@@ -42,13 +56,24 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// achieve the reconciliation.
 	r.Log.Info("Reconciling cluster-wide HNC configuration.")
 
+	// TODO: This is only for debuging and testing in the early development stage. Remove this when the
+	// implementation is completed.
+	r.logSpec(inst)
+
```
adrianludwin: Maybe remove this now? And just put out logs whenever something interesting happens (which I think you're already doing).

incubator/hnc/pkg/controllers/config_controller.go
```diff
@@ -42,13 +56,24 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	// Create cooresponding ObjectReconcilers for newly added types, if needed.
+	// TODO: Returning an error here will make controller-runtime to requeue the object to be reconciled again.
```
adrianludwin: I think we shouldn't return the error in this PR - just log it, and return nil. Just because you have an error doesn't mean you _have_ to return it (though lots of golang coders behave that way!)

incubator/hnc/pkg/controllers/config_controller.go
```diff
@@ -91,6 +116,63 @@ func (r *ConfigReconciler) writeSingleton(ctx context.Context, inst *api.HNCConf
...
+	r.Forest.Lock()
+	defer r.Forest.Unlock()
+	for _, t := range inst.Spec.Types {
+		gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+		if r.Forest.HasTypeSyncer(gvk) {
+			continue
+		}
+		if err := r.createObjectReconciler(gvk); err != nil {
+			r.Log.Error(err, "while creating an object controller")
+			return err
```
adrianludwin: Generally, you should _either_ log _or_ return an error. If an upstream caller needs to know there was an error, you add information to the error (eg `return fmt.Errorf("while trying to foo: %s", err)`); if the caller doesn't need to know, log it and don't return anything.

Otherwise, one actual error ends up causing N errors in the log, where N is the call stack depth of the code in question.

incubator/hnc/pkg/forest/forest.go
```diff
@@ -45,6 +67,30 @@ func (f *Forest) Unlock() {
...
+	}
+	return false
+}
+
+// GetTypeSyncers returns the types list.
+func (f *Forest) GetTypeSyncers() []TypeSyncer {
+	// We can make a shallow copy here because the TypeSyncer is currently implemented by
+	// *ObjectReconciler (the pointer type) which means we are copying addresses here. We might need
+	// to change this to a deep copy if the TypeSyncer is implemented by a value type in future.
+	types := f.types
```
adrianludwin: I'm not sure this is true - copying a slice is a shallow operation, see https://play.golang.org/p/huiy48tguSM.

Now, the only think you'll ever do to modify this slice is to append to it, so maybe that's safe? But better safe than sorry, maybe just use `copy` instead: https://stackoverflow.com/questions/30182538/why-cant-i-duplicate-a-slice-with-copy

incubator/hnc/pkg/controllers/config_controller.go
```diff
@@ -42,13 +56,24 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	// TODO: Returning an error here will make controller-runtime to requeue the object to be reconciled again.
+	// A better way to handle this is to display the error as part of the HNCConfig.status field to suggest the error is a
+	// permanent problem and to stop controller-runtime from retrying.
+	if err := r.createObjectReconcilers(inst); err != nil {
+		r.Log.Error(err, "Couldn't create the new object reconciler.")
+		return ctrl.Result{}, err
+	}
+
 	// Write back to the apiserver.
 	if err := r.writeSingleton(ctx, inst); err != nil {
```
yiqigao217: Just curious: what is the scenario that we need to write the singleton during `Reconcile()`?

incubator/hnc/pkg/controllers/config_controller.go
```diff
@@ -42,13 +56,24 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
+	// TODO: This is only for debuging and testing in the early development stage. Remove this when the
+	// implementation is completed.
+	r.logSpec(inst)
+
+	// Create cooresponding ObjectReconcilers for newly added types, if needed.
+	// TODO: Returning an error here will make controller-runtime to requeue the object to be reconciled again.
+	// A better way to handle this is to display the error as part of the HNCConfig.status field to suggest the error is a
+	// permanent problem and to stop controller-runtime from retrying.
+	if err := r.createObjectReconcilers(inst); err != nil {
```
yiqigao217: Maybe add a `TODO` for something like `syncObjectReconcilers()`? It will also include removing obsolete object reconcilers besides `createObjectReconcilers()`.

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: Sure! Done!

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: Done.

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: Thanks so much for the explanation. In this case, I changed to return the error without logging. The Reconcile() function will log the error.

incubator/hnc/pkg/forest/forest.go
sophieliu15: Thanks for the code pointer. Updated the code by following the example.

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: IIUC, if we update HNCConfiguration.Status during reconciliation, we need to write the updated HNCConfiguration object back to the apiserver. Feel free to correct me if this is not the case.

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: Good point. Added the TODO for renaming in future. Feel free to take a look.

sophieliu15:
The test should be fixed by now. Feel free to take a look.

incubator/hnc/pkg/controllers/config_controller.go
adrianludwin: This is correct - we're not doing it yet, but you should probably add that as another step in your plan doc.

incubator/hnc/pkg/controllers/object_controller_test.go
```diff
@@ -34,6 +38,9 @@ var _ = Describe("Secret", func() {
 	})
 
 	It("should be copied to descendents", func() {
+		// Add secret to HNCConfiguration so that an ObjectReconciler will be created for secret.
+		addSecretToHNCConfig(ctx, hncConfig)
+
```
adrianludwin: These should all be in the `BeforeEach` function at the top of this file.

incubator/hnc/pkg/forest/forest.go
```diff
@@ -45,6 +67,28 @@ func (f *Forest) Unlock() {
...
+func (f *Forest) HasTypeSyncer(gvk schema.GroupVersionKind) bool {
+	for _, t := range f.types {
+		if t.GetGVK() == gvk {
+			return true
+		}
+	}
+	return false
+}
+
+// GetTypeSyncers returns the types list.
```
adrianludwin: Maybe worth explaining why we're returning a copy instead of the list itself. Eg, this function requires that the mutex is held, but the returned list is safe to use without the mutex being held; also, the caller can modify the returned array without fear of corrupting data.

sophieliu15: Thanks for the suggestion! Updated the comments. Feel free to take a look.

incubator/hnc/pkg/controllers/object_controller_test.go
sophieliu15: Good point. Done.

incubator/hnc/pkg/controllers/config_controller.go
sophieliu15: Thanks for the confirmation. Doen.

incubator/hnc/pkg/controllers/object_controller_test.go
```diff
@@ -11,18 +11,25 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
 )
 
-var _ = Describe("Secret", func() {
+var _ = Describe("Role", func() {
```
adrianludwin: Didn't we switch this back to "Secret" now?

rjbez17:
This lgtm, @adrianludwin  and @yiqigao217  are you satisfied with the updates from you last few comments? Feel free to remove hold when happy.

/assign @adrianludwin 
/assign @yiqigao217 

/lgtm
/approve
/hold

adrianludwin:
Thanks Ryan! Just one minor nit left then I'll remove the hold.

incubator/hnc/pkg/controllers/object_controller_test.go
sophieliu15: Good catch! Switched back to "Secret".

PR #438, opened 2020-02-19
Add a unit test for config_controller.go

sophieliu15:
Add a unit test for config_controller.go to test the case that when a new type is added to HNCConfiguration singleton, the corresponding object reconciler can be created correctly.

This PR also does the following refactoring: 
   -  Moves shared helper functions from each controller test files to a common file: test_helpers_test.go
   -  Switches the object type in object_controller_test.go from Secret to Role because:
       -  The behaviors that are tested in the object_controller_test.go are irrelevant to the object type or the creation of a new object reconciler.
       -  Since the test suite creates the object reconciler for Role by default, object_controller_test.go can use the existing Role object reconciler for the tests 
       instead of creating a new object reconciler.
       -  The tests inside controllers_test package share the same test environment, which means if the test in one Describe container adds a type to 
       HNCConfiguration singleton, the corresponding object reconciler for that type will be created and will still be there when running subsequent Describe 
       containers in the package. Until we find a way to "cleanup" object reconcilers, it is better to only create new object reconcilers when necessary.

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/controllers/config_controller_test.go
+++ b/incubator/hnc/pkg/controllers/config_controller_test.go
@@ -1,0 +1,111 @@
+package controllers_test
+
+import (
+	"context"
+	"time"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+var _ = Describe("HNCConfiguration", func() {
+	ctx := context.Background()
+
+	var (
+		fooName string
+		barName string
+	)
+
+	BeforeEach(func() {
+		fooName = createNS(ctx, "foo")
+		barName = createNS(ctx, "bar")
+	})
+
+	AfterEach(func() {
+		// Change current config back to the default value.
+		Eventually(func() error {
+			return resetHNCConfigToDefault(ctx)
+		}).Should(Succeed())
+	})
+
+	It("should propagate objects whose types have been added to HNCConfiguration", func() {
+		setParent(ctx, barName, fooName)
+		makeSecret(ctx, fooName, "foo-sec")
+
+		// Wait 1 second to give "foo-sec" a chance to be propagated to bar, if it can be propagated.
+		time.Sleep(1 * time.Second)
+		// Foo should have "foo-sec" since we created it there.
+		Eventually(hasSecret(ctx, fooName, "foo-sec")).Should(BeTrue())
+		// "foo-sec" is not propagated to bar because Secret hasn't been configured in HNCConfiguration.
+		Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeFalse())
+
+		Eventually(func() error {
+			c := getHNCConfig(ctx)
+			return addSecretToHNCConfig(ctx, c)
+		}).Should(Succeed())
+
+		// "foo-sec" should now be propagated from foo to bar.
+		Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeTrue())
+		Expect(secretInheritedFrom(ctx, barName, "foo-sec")).Should(Equal(fooName))
+	})
+})
+
+func resetHNCConfigToDefault(ctx context.Context) error {
+	c := getHNCConfig(ctx)
+	c.Spec = config.GetDefaultConfigSpec()
+	return k8sClient.Update(ctx, c)
+}
+
+func getHNCConfig(ctx context.Context) *api.HNCConfiguration {
+	return getHNCConfigWithOffset(1, ctx)
+}
+
+func getHNCConfigWithOffset(offset int, ctx context.Context) *api.HNCConfiguration {
+	snm := types.NamespacedName{Name: api.HNCConfigSingleton}
+	config := &api.HNCConfiguration{}
+	EventuallyWithOffset(offset+1, func() error {
+		return k8sClient.Get(ctx, snm, config)
+	}).Should(Succeed())
+	return config
+}
+
+func addSecretToHNCConfig(ctx context.Context, c *api.HNCConfiguration) error {
+	secSpec := api.TypeSynchronizationSpec{APIVersion: "v1", Kind: "Secret", Mode: api.Propagate}
+	c.Spec.Types = append(c.Spec.Types, secSpec)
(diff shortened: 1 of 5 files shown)


incubator/hnc/pkg/controllers/config_controller_test.go
```diff
@@ -0,0 +1,37 @@
...
+		fooName string
+		barName string
+	)
+
+	BeforeEach(func() {
+		fooName = createNS(ctx, "foo")
+		barName = createNS(ctx, "bar")
+	})
+
+	It("should create a new ObjectReconciler when adding a new type to the HNCConfiguration", func() {
```
adrianludwin: You're not testing that an ObjectReconciler exists - you're testing that the objects get propagated. So maybe:

`It("should propagate objects whose types have been added to HNCConfiguration")`

incubator/hnc/pkg/controllers/config_controller_test.go
```diff
@@ -0,0 +1,37 @@
...
+		setParent(ctx, barName, fooName)
+		makeSecret(ctx, fooName, "foo-sec")
+
+		// "foo-sec" is not propagated to bar because Secret hasn't been configured in HNCConfiguration.
+		Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeFalse())
```
yiqigao217: Probably better to add more confirmation here by either:
1) add a sleep time before this to give it a "chance" to propagate object; OR
2) add another case for comparison: when creating an object in the `foo` NS with a type configured in HNCConfiguration, e.g. `configmap`,  the parent `configmap` object is propagated to the `bar` NS while the `secret` object is not. It will look like:
```
config := getHNCConfig(ctx)
// Omit this, if the configmap is already configured in the original singleton.
addConfigmapToHNCConfig(ctx, config)

makeSecret(ctx, fooName, "foo-sec")
makeConfigmap(ctx, fooName, "foo-cfg")

Eventually(hasConfigmap(ctx, barName, "foo-cfg")).Should(BeTrue())
Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeFalse())
```

sophieliu15: Thanks so much for the suggestion! I used option 1 for now to let it sleep for 1 second to have a chance to propagate.

I will add another test case later to compared the propagation of an object with type that is configured in the HNCConfiguration with another object with type that is not configured. We can do that by compare any type with RBAC types on start up because we should have configured RBAC types by default. Right now this part is tricky since the code does not configure RBAC types by default and we set it to default in suite_test.go. After I correct the code and remove the default setup in suite_test.go, I will add that test case.

incubator/hnc/pkg/controllers/config_controller_test.go
sophieliu15: Thanks for the suggestion. Changed the description. 

I also reset the singleton to default in AfterEach in object_controller_test.go and config_controller_test.go as we discussed. Feel free to take a look!

Thanks so much for the discussion!

incubator/hnc/pkg/controllers/config_controller_test.go
yiqigao217: sgtm, yes option 2 will be added much easier once default singleton is configured or you have more cases to test later.

adrianludwin: I agree, option #1 seems simpler and also sufficient in this case.

incubator/hnc/pkg/controllers/config_controller_test.go
```diff
@@ -0,0 +1,52 @@
...
+	BeforeEach(func() {
+		fooName = createNS(ctx, "foo")
+		barName = createNS(ctx, "bar")
+	})
+
+	AfterEach(func() {
+		// Change current singleton back to the default value.
+		Eventually(func() error {
+			c := getHNCConfig(ctx)
+			return resetHNCConfigToDefault(ctx, c)
```
adrianludwin: Maybe resetHNCConfigToDefault should call getHNCConfig itself? Otherwise, that suggests that you could call it with some special config, which would kind of defeat the purpose.

Also, I'd keep this helper function (and others) in this file until it needs to be called from other files too, but that's just a nit.

adrianludwin:
/ok-to-test

lgtm after a few nits. @rjbez17 , wdyt?

/assign @rjbez17

incubator/hnc/pkg/controllers/config_controller_test.go
rjbez17: Is 1 second long enough? Wont normal flakiness of a running test system potentially take longer complete a reconcile? If I'm reading this correctly we wait 1 second to ensure HNC didn't accidentally propagate something it shouldn't.

adrianludwin: In practice, if things propagate in the test, they usually propagate in well under 1s. And yes, we're trying to ensure that HNC _hasn't_ done something that it _shouldn't_, so even if this only "works" 80% of the time, that should be more than enough to inform the developer if they accidentally break this.

rjbez17: My concern is more around the point (months from now) when it starts taking longer than 1s and it only starts to work <10% of the time. Realistically now we don't have any indication if the test is actually successfully testing what it should. I think option 2 is the overall better test, and I assume we'll implement it eventually? I'm not concerned enough to hold up the merge.

incubator/hnc/pkg/controllers/config_controller_test.go
sophieliu15: Thanks for the suggestion! Moved getHNCConfig into resetHNCConfigToDefault. It makes sense to put only shared functions in the helper file. There are not as many shared functions as before once we switch to test on Role in object_controller_test.go. I moved the ones that are not shared back to the individual files. 

I keep createNSWithLabel in the helper file even though it is only used by hierarchy_controller_test.go for now because createNSWithLabel is similar to createNS and the latter is used in many places.

incubator/hnc/pkg/controllers/config_controller_test.go
sophieliu15: Thanks Adrian and Ryan for the suggestions. In this case, I will keep the option 1 in this PR and switch to Option 2 in my next PR when I fix the default config setup as I discussed in the previous comment.

adrianludwin: Ok, sounds reasonable, since Sophie is going to be adding a special-purpose test for the default configuration.

incubator/hnc/pkg/controllers/config_controller_test.go
adrianludwin: sgtm!

PR #442, opened 2020-02-20
Propagate empty source object to descendants if the descendants do not have not copy of the object.

### LINKED ISSUE #441: HNC: Empty source object cannot be propagated by object reconciler
Current object reconciler will only propagate an object when a descendant's copy of the object is different from the source object. This is currently handled in [syncPropagated](https://github.com/kubernetes-sigs/multi-tenancy/blob/95218a0fb87fee9d024202ad9a4e1755d9964b56/incubator/hnc/pkg/controllers/object_controller.go#L222) method.  

However, when source object is empty (but not nil) and when a descendant does not have any copy of the object, syncPropagated will return ["ignore"](https://github.com/kubernetes-sigs/multi-tenancy/blob/95218a0fb87fee9d024202ad9a4e1755d9964b56/incubator/hnc/pkg/controllers/object_controller.go#L227) action and will not propagate the empty object to the descendant. When source object is not nil and when a descendant does not have any copy of the object, we would expect the object to be propagated to the descendant, even when the object is empty.

To reproduce this issue:
* Create acme-org, team-a namespaces. Set team-a as child of acme-org
     * kubectl create namespace acme-org
     * kubectl create namespace team-a
     * kubectl hns set team-a --parent  acme-org
* Create an empty and an non-empty configmaps in acme-org
     * kubectl -n acme-org create cm emptycf
     * kubectl -n acme-org create cm nonemptycf --from-literal key=value
* Check that acme-org has two configmaps and team-a does not have any configmap
     * kubectl get cm -n acme-org
     * kubectl get cm -n team-a
* Add ConfigMap in HNCConfiguration
     * kubectl get hncconfiguration config -o yaml > hncconfig.yaml
     * Add following in hncconfig.yaml
          - apiVersion: v1
             kind: ConfigMap
             mode: propagate
     * kubectl apply -f hncconfig.yaml
* Now you will see team-a has nonemptycf but does not have emptycf
     * kubectl get cm -n team-a

sophieliu15:
Currently an empty source object will not be propagated to descendants if the descendants do not have any copy of the object as reported in #441. The PR fixes this issue so that empty source object will be propagated to descendants if the descendants do not have the copy of the object.

Fixes #441

### DIFF
--- a/incubator/hnc/pkg/controllers/config_controller_test.go
+++ b/incubator/hnc/pkg/controllers/config_controller_test.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
-	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
@@ -53,25 +52,6 @@ var _ = Describe("HNCConfiguration", func() {
 		Expect(secretInheritedFrom(ctx, barName, "foo-sec")).Should(Equal(fooName))
 	})
 })
-
-func resetHNCConfigToDefault(ctx context.Context) error {
-	c := getHNCConfig(ctx)
-	c.Spec = config.GetDefaultConfigSpec()
-	return k8sClient.Update(ctx, c)
-}
-
-func getHNCConfig(ctx context.Context) *api.HNCConfiguration {
-	return getHNCConfigWithOffset(1, ctx)
-}
-
-func getHNCConfigWithOffset(offset int, ctx context.Context) *api.HNCConfiguration {
-	snm := types.NamespacedName{Name: api.HNCConfigSingleton}
-	config := &api.HNCConfiguration{}
-	EventuallyWithOffset(offset+1, func() error {
-		return k8sClient.Get(ctx, snm, config)
-	}).Should(Succeed())
-	return config
-}
 
 func addSecretToHNCConfig(ctx context.Context, c *api.HNCConfiguration) error {
 	secSpec := api.TypeSynchronizationSpec{APIVersion: "v1", Kind: "Secret", Mode: api.Propagate}
--- a/incubator/hnc/pkg/controllers/object_controller.go
+++ b/incubator/hnc/pkg/controllers/object_controller.go
@@ -218,8 +218,11 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger, 
 		return remove, nil
 	}
 
-	// If the copy is different from the source, return the write action and the source instance.
-	if !reflect.DeepEqual(object.Canonical(inst), object.Canonical(srcInst)) {
+	// If an object doesn't exist, assume it's been deleted or not yet created.
+	exist := inst.GetCreationTimestamp() != v1.Time{}
+
+	// If the copy does not exist or is different from the source, return the write action and the source instance.
+	if !exist || !reflect.DeepEqual(object.Canonical(inst), object.Canonical(srcInst)) {
 		metadata.SetLabel(inst, api.LabelInheritedFrom, srcInst.GetNamespace())
 		return write, srcInst
 	}
--- a/incubator/hnc/pkg/controllers/object_controller_test.go
+++ b/incubator/hnc/pkg/controllers/object_controller_test.go
@@ -34,6 +34,13 @@ var _ = Describe("Secret", func() {
 		makeRole(ctx, bazName, "baz-role")
 	})
 
+	AfterEach(func() {
+		// Change current config back to the default value.
+		Eventually(func() error {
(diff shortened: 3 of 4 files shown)


incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -218,6 +218,14 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 		return remove, nil
 	}
 
+	// If an object doesn't exist, assume it's been deleted or not yet created. In this case, we will
+	// return write action and the source instance.
+	exist := inst.GetCreationTimestamp() != v1.Time{}
```
adrianludwin: See if `inst.GetCreationTimestamp().IsZero()` works here. I don't think we ever figured out why it sometimes works but sometimes wouldn't compile.

incubator/hnc/pkg/controllers/object_controller.go
```diff
@@ -218,6 +218,14 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+	if !exist {
+		metadata.SetLabel(inst, api.LabelInheritedFrom, srcInst.GetNamespace())
```
adrianludwin: In order to avoid calling SetLabel twice, maybe merge the two `if` statements?

incubator/hnc/pkg/controllers/config_controller_test.go
```diff
@@ -52,6 +54,29 @@ var _ = Describe("HNCConfiguration", func() {
 		Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeTrue())
 		Expect(secretInheritedFrom(ctx, barName, "foo-sec")).Should(Equal(fooName))
 	})
+
+	It("should propagate an empty object whose types have been added to HNCConfiguration", func() {
```
adrianludwin: Ideally this would be in object-controller-test, since it's only testing a pure feature of the object controller? Would it be possible to move this there?

incubator/hnc/pkg/controllers/config_controller_test.go
```diff
@@ -52,6 +54,29 @@ var _ = Describe("HNCConfiguration", func() {
...
+		Eventually(hasConfigMap(ctx, barName, "foo-config")).Should(BeFalse())
+
+		Eventually(func() error {
+			c := getHNCConfig(ctx)
+			return addConfigMapToHNCConfig(ctx, c)
+		}).Should(Succeed())
+
+		// "foo-config" should now be propagated from foo to bar.
+		Eventually(hasConfigMap(ctx, barName, "foo-config")).Should(BeTrue())
+		Expect(configMapInheritedFrom(ctx, barName, "foo-config")).Should(Equal(fooName))
```
adrianludwin: This is the only "real" part of this test - everything else is tested elsewhere. Maybe create a helper to add ConfigMaps as a supported type in the object controller (e.g. change `addConfigMapToHNCConfig` to include the whole `Eventually` and `getHNCCConfig` stuff) so you can call it from object-controller-test, and skip the stuff where you try to propagate _before_ creating the object controller, since you already test that.

sophieliu15: Thanks for the suggestion! PTAL

incubator/hnc/pkg/controllers/config_controller_test.go
sophieliu15: Good idea. Moved to the object_controller_test.go. Also moved some more shared functions to the test_helpers_test.go

incubator/hnc/pkg/controllers/object_controller.go
sophieliu15: Thanks for the suggestion! Done.

incubator/hnc/pkg/controllers/object_controller.go
sophieliu15: Looks like it still doesn't work. It has compile errors:
pkg/controllers/object_controller.go:224:38: cannot call pointer method on inst.GetCreationTimestamp()
pkg/controllers/object_controller.go:224:38: cannot take the address of inst.GetCreationTimestamp()

adrianludwin: Ah well, worth a shot.

incubator/hnc/pkg/controllers/object_controller_test.go
```diff
@@ -48,6 +55,18 @@ var _ = Describe("Secret", func() {
 		Expect(roleInheritedFrom(ctx, bazName, "bar-role")).Should(Equal(barName))
 	})
 
+	It("should be copied to descendents when source object is empty", func() {
+		setParent(ctx, barName, fooName)
+		// Creates an empty ConfigMap. We use ConfigMap for this test because the apiserver will not
+		// add additional fields to an empty ConfigMap object to make it non-empty.
+		makeConfigMap(ctx, fooName, "foo-config")
+		addConfigMapToHNCConfig(ctx)
```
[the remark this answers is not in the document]
sophieliu15: Thanks for the suggestion! Will do!

PR #447, opened 2020-02-24
Replace "controller" with "reconciler" in hnc/pkg/controllers directory

sophieliu15:
Replace "controller" with "reconciler" in hnc/pkg/controllers directory. "Controllers" can refer to both reconcilers and admission controllers. Since the controllers we implemented under hnc/pkg/controllers are actually reconcilers, using "reconciler" instead of "controller" would be more precise.

### DIFF
--- a/incubator/hnc/README.md
+++ b/incubator/hnc/README.md
@@ -177,16 +177,18 @@ interesting directories are probably:
 
 * `/api`: the API definition.
 * `/cmd`: top-level executables. Currently the manager and the kubectl plugin.
-* `/pkg/controllers`: the reconcilers and their tests
+* `/pkg/reconcilers`: the reconcilers and their tests
 * `/pkg/validators`: validating admission controllers
-* `/pkg/forest`: the in-memory data structure, shared between the controllers
+* `/pkg/forest`: the in-memory data structure, shared between the reconcilers
   and validators.
 
-Within the `controllers` directory, there are two controller:
+Within the `reconcilers` directory, there are three reconcilers:
 
-* **Hierarchy controller:** manages the hierarchy via the `Hierarchy` singleton
+* **HNCConfiguration reconciler:** manages the HNCConfiguration via the
+  cluster-wide `config` singleton.
+* **Hierarchy reconciler:** manages the hierarchy via the `Hierarchy` singleton
   as well as the namespace in which it lives.
-* **Object controller:** Propagates (copies and deletes) the relevant objects
+* **Object reconciler:** Propagates (copies and deletes) the relevant objects
   from parents to children. Instantiated once for every supported object GVK
   (group/version/kind) - e.g., `Role`, `Secret`, etc.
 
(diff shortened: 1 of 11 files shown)


incubator/hnc/README.md
```diff
@@ -177,13 +177,15 @@ interesting directories are probably:
...
+* `/pkg/reconcilers`: the reconcilers and their tests
 * `/pkg/validators`: validating admission controllers
-* `/pkg/forest`: the in-memory data structure, shared between the controllers
+* `/pkg/forest`: the in-memory data structure, shared between the reconcilers
   and validators.
 
-Within the `controllers` directory, there are two controller:
+Within the `reconcilers` directory, there are three reconcilers:
 
+* **HNCConfiguration controller:** manages the HNCConfiguration via the
```
yiqigao217: s/controller/reconciler?

sophieliu15: Good catch! Updated.

incubator/hnc/go.sum
```diff
@@ -142,6 +142,7 @@ github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/pty v1.1.5/go.mod h1:9r2w37qlBe7rQ6e1fg1S/9xpWHSnaqNdHD3WcMdbPDA=
 github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
+github.com/kubernetes-sigs/multi-tenancy v0.0.0-20200221171633-2b5c81967af0 h1:JzZCwDzWZpZOMNgir26AiVSDuCWsJSGB2CX5pzT2/fM=
```
adrianludwin: This doesn't look right - this means it tried to download a package. Try reverting go.sum and re-running `make` to ensure it's not changed again.

incubator/hnc/pkg/reconcilers/config_reconciler.go
```diff
@@ -1,4 +1,4 @@
-package controllers
+package reconcilers
```
adrianludwin: Maybe change the filename too? `reconcilers/config_reconciler.go` seems both redundant and a bit ambiguous because of HierarchyConfiguration. How about `reconcilers/hncconfig.go` (and also `object.go` and `hierarchyconfig.go`)?

sophieliu15: Changed to hnc_config.go and hierarchy_config.go since there are two words. Changed object_reconciler.go to object.go. Feel free to take a look.

incubator/hnc/go.sum
sophieliu15: Reverted.

incubator/hnc/pkg/reconcilers/config_reconciler.go
adrianludwin: Hmm - seems like the convention is _not_ to use underscores? See https://stackoverflow.com/questions/25161774/what-are-conventions-for-filenames-in-go. Have we not been following this?

adrianludwin: Hmm, looks like we don't follow that. lgtm :)

PR #448, opened 2020-02-24
Create HierarchicalNamespace CRD

yiqigao217:
This CRD is created as the UX of the HNC self-service namespace. See
design at http://bit.ly/hnc-self-serve-ux.

Tested on a GKE cluster by "kubectl apply -f" the generated CRD yaml
file. The CRD was created with success. I also added a sample hns.yaml
and ran "kubectl create -f", the hns object was created and showed up
when I ran "kubectl get hns".

The possibility of "kubectl create hns" is still under investigation,
since the "kubectl create" only supports a selected list of built-in
resource types (even "kubectl create pod" is not supported). We will see
if "kubectl create" a CR object is possible.


yiqigao217:
I noticed the changes in hierarchyconfigurations.yaml and hncconfigurations.yaml after running `make manifests`. I am not sure why it happened since there's no change in the hierarchy_types.go or hnc_config.go files in api/v1alpha1/. WDYT?

/assign @adrianludwin

incubator/hnc/api/v1alpha1/hierarchical_namespace.go
```diff
@@ -0,0 +1,67 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// HNSStatus describes the status of a hierarchical namespace. The status could be
+// "missing", "ok" or "conflict". The definitions will be described below.
+type HNSStatus string
```
adrianludwin: Right now, this would give us `status.status`, which is a bit odd. Pods and containers use `state`, eg:

```
status:
  state: ok | missing | conflict
```

I'd recommend following that, sgty?

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchicalnamespaces.yaml
```diff
@@ -0,0 +1,52 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: hierarchicalnamespaces.hnc.x-k8s.io
+spec:
+  group: hnc.x-k8s.io
+  names:
```
adrianludwin: Add `shortName: hns`? You'll need to put this in the magic kubebuilder comment.

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -14,7 +12,7 @@ spec:
     listKind: HierarchyConfigurationList
     plural: hierarchyconfigurations
     singular: hierarchyconfiguration
-  scope: Namespaced
+  scope: ""
```
adrianludwin: I think you need to update your version of kubebuilder? See what version Sophie has - I think one of her PRs added the scope.

yiqigao217:
> I noticed the changes in hierarchyconfigurations.yaml and hncconfigurations.yaml after running `make manifests`. I am not sure why it happened since there's no change in the hierarchy_types.go or hnc_config.go files in api/v1alpha1/. WDYT?
> 
> /assign @adrianludwin

The reason is that different versions of `controller-gen` is used. @adrianludwin and I used an older version, while @sophieliu15 just downloaded `kubebuilder` with built-in v0.2.4 `controller-gen`. That's why our generated yaml files are different - see example https://github.com/kubernetes-sigs/multi-tenancy/pull/412/commits/5d434459224f3d62aa10e1c1345060e077cef2c9#diff-cc3fafb48bba9b066ffb7c74baeebf6d

I submitted #450 , but there are some issues with it. I will dig more.

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
yiqigao217: Ah sg. Didn't saw your comment here. I will just update my kubebuilder to solve this issue!

incubator/hnc/api/v1alpha1/hierarchical_namespace.go
```diff
@@ -0,0 +1,67 @@
...
+	// Ok means the child namespace exists.
+	Ok HNSState = "ok"
+
+	// Conflict means a namespace of the same name already exists.
+	Conflict HNSState = "conflict"
+)
+
+// HierarchicalNamespaceStatus defines the observed state of HierarchicalNamespace.
+type HierarchicalNamespaceStatus struct {
+	// State describes the state of a hierarchical namespace.
```
adrianludwin: Can you please document the values here, and not as part of the constants, since this documentation becomes part of the CRD definition while the constants do not. See, eg: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/api/v1alpha1/hierarchy_types.go#L126 (note the blank lines between list elements).

The constants themselves can just get an overall comment like this: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/api/v1alpha1/hierarchy_types.go#L34

yiqigao217: sg! done.

adrianludwin:
lgtm. Can you confirm that if you just say `kubectl create hns bar -n foo` (or `kubectl create hierarchicalnamespace bar -n foo`, then an empty HNS object called "bar" gets created in namespace "foo"?

adrianludwin:
/hold
/lgtm
/approve

For Ryan to review.

yiqigao217:
> lgtm. Can you confirm that if you just say `kubectl create hns bar -n foo` (or `kubectl create hierarchicalnamespace bar -n foo`, then an empty HNS object called "bar" gets created in namespace "foo"?

No. I got
`No resources found in default namespace.` for `k get hierarchicalnamespace.hnc.x-k8s.io`

`error: unknown command "hierarchicalnamespace.hnc.x-k8s.io"` for `k create hierarchicalnamespace.hnc.x-k8s.io`.

However, I got
`customresourcedefinition.apiextensions.k8s.io/hierarchicalnamespaces.hnc.x-k8s.io configured`
for `k apply -f config/crd/bases/hnc.x-k8s.io_hierarchicalnamespaces.yaml` as stated in the commit.

Seems the CR is created, but I cannot create an object using `kubectl create`.

I created a sample hns.yaml with `HierarchicalNamespace` as kind and applied it. I can get the sample object using `k get hns`.

yiqigao217:
I added a sample hns.yaml for `kubectl` to apply to create an hns object and updated the commit body with the explanation that all commands `kubectl apply -f crd.yaml`, `kubectl create -f hns.yaml` and `kubectl get hns` worked while `kubectl create hns` not.

It's a `kubectl` related issue that CR objects can be created by `kubectl create -f` but not directly by `kubectl create`, which is out of the scope of this PR.

adrianludwin:
Ok, pity that we can't do `k create hns`, we'll just have to go with `k hns create` for this one case.

/shrug

I'm ok with this one @rjbez17 is. I think he's OOO today so perhaps he'll reply tomorrow, if not lmk and I'll approve this myself.

PR #453, opened 2020-02-25
Create a default HNCConfiguration singleton when starting HNC if the singleton does not exist

sophieliu15:
This PR creates a default HNCConfiguration singleton when starting HNC if the singleton does not exist. The feature is achieved by adding a channel in reconciler watch to trigger the reconciliation at startup. The reconciliation will ensure the default singleton will be created if it does not exist.

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hnc_config.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -28,6 +30,11 @@ type ConfigReconciler struct {
 
 	// Forest is the in-memory data structure that is shared with all other reconcilers.
 	Forest *forest.Forest
+
+	// Igniter is a channel of event.GenericEvent (see "Watching Channels" in
+	// https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches.html)
+	// that is used to enqueue the singleton for initial reconciliation.
+	Igniter chan event.GenericEvent
 
 	// HierarchyConfigUpdates is a channel of events used to update hierarchy configuration changes performed by
 	// ObjectReconcilers. It is passed on to ObjectReconcilers for the updates. The ConfigReconciler itself does
@@ -56,7 +63,7 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// achieve the reconciliation.
 	r.Log.Info("Reconciling cluster-wide HNC configuration.")
 
-	// Create cooresponding ObjectReconcilers for newly added types, if needed.
+	// Create corresponding ObjectReconcilers for newly added types, if needed.
 	// TODO: Returning an error here will make controller-runtime to requeue the object to be reconciled again.
 	// A better way to handle this is to display the error as part of the HNCConfig.status field to suggest the error is a
 	// permanent problem and to stop controller-runtime from retrying.
@@ -171,9 +178,38 @@ func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind) e
 	return nil
 }
 
+// forceInitialReconcile forces reconciliation to start after setting up the
+// controller with the manager. This is used to create a default singleton if
+// there is no singleton in the cluster. This occurs in a goroutine so the
+// caller doesn't block; since the reconciler is never garbage-collected,
+// this is safe.
+func (r *ConfigReconciler) forceInitialReconcile(log logr.Logger, reason string) {
+	go func() {
+		log.Info("Enqueuing for reconciliation", "reason", reason)
+		// The watch handler doesn't care about anything except the metadata.
+		inst := &api.HNCConfiguration{}
+		inst.ObjectMeta.Name = api.HNCConfigSingleton
+		r.Igniter <- event.GenericEvent{Meta: inst}
+	}()
+}
+
 // SetupWithManager builds a controller with the reconciler.
 func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
+	if err := ctrl.NewControllerManagedBy(mgr).
 		For(&api.HNCConfiguration{}).
-		Complete(r)
+		Watches(&source.Channel{Source: r.Igniter}, &handler.EnqueueRequestForObject{}).
+		Complete(r); err != nil {
+		return err
+	}
+	// Create a default singleton if there is no singleton in the cluster.
+	//
+	// The cache used by the client to retrieve objects might not be populated
+	// at this point. As a result, we cannot use r.Get() to determine the existence
+	// of the singleton and then use r.Create() to create the singleton if
+	// it does not exist. As a workaround, we decide to enforce reconciliation. The
+	// cache is populated at the reconciliation stage. A default singleton will be
+	// created during the reconciliation if there is no singleton in the cluster.
+	r.forceInitialReconcile(r.Log, "Enforce reconciliation to create a default"+
+		"HNCConfiguration singleton if it does not exist.")
+	return nil
 }
--- a/incubator/hnc/pkg/reconcilers/hnc_config_test.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config_test.go
@@ -2,12 +2,13 @@ package reconcilers_test
 
 import (
 	"context"
-	"time"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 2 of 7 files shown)


incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -29,6 +31,11 @@ type ConfigReconciler struct {
 	// Forest is the in-memory data structure that is shared with all other reconcilers.
 	Forest *forest.Forest
 
+	// Affected is a channel of event.GenericEvent (see "Watching Channels" in
+	// https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches.html)
+	// that is used to  enqueue the singleton that needs reconciling.
```
yiqigao217: Remove the extra space before "enqueue"

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -31,7 +38,28 @@ var _ = Describe("HNCConfiguration", func() {
 		}).Should(Succeed())
 	})
 
-	It("should propagate objects whose types have been added to HNCConfiguration", func() {
+	It("should set mode of Roles and RoleBindings as propagate by default", func() {
+		config := getHNCConfig(ctx)
+
+		Eventually(hasTypeWithMode("rbac.authorization.k8s.io/v1", "Role", api.Propagate, config)).Should(BeTrue())
+		Eventually(hasTypeWithMode("rbac.authorization.k8s.io/v1", "RoleBinding", api.Propagate, config)).Should(BeTrue())
```
yiqigao217: Just curious, where are Role and RoleBinding added to the singleton?

sophieliu15: When there is no singleton, the reconciler will create a default one and write it to api server in getSingleton() and writeSingleton(). The default singleton has Role and Rolebindings (http://shortn/_LxQ9fWO7i6). The two methods are called in the Reconcile() method. We force the reconciliation in SetupWithManager() so if there is no singleton in the cluster, the Reconcile() method will create a default one.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Good catch! Removed.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
yiqigao217: Got you thanks!

yiqigao217:
lgtm except one last comment: maybe change the PR/commit title and body to say something like "add a channel in reconciler watch to trigger reconciliation at startup to ensure the default singleton is created..." since the creation of singleton is done and not included in this PR.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -29,6 +31,11 @@ type ConfigReconciler struct {
 	// Forest is the in-memory data structure that is shared with all other reconcilers.
 	Forest *forest.Forest
 
+	// Affected is a channel of event.GenericEvent (see "Watching Channels" in
```
adrianludwin: Since we only use this for initialization, maybe call this something more limited? Eg, `StartupQueue` or `Kickstarter` or `Igniter` or something like that :)

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -173,7 +180,23 @@ func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind) e
 
 // SetupWithManager builds a controller with the reconciler.
 func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.enqueueAffected(r.Log, "Enforce reconciliation to create a default"+
+		"HNCConfiguration singleton if it does not exist.")
```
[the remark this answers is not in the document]
adrianludwin: Also, insead of `enqueueAffected`, maybe `forceInitialReconcile`?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -41,22 +69,19 @@ var _ = Describe("HNCConfiguration", func() {
 		Eventually(hasSecret(ctx, fooName, "foo-sec")).Should(BeTrue())
 		// "foo-sec" is not propagated to bar because Secret hasn't been configured in HNCConfiguration.
 		Eventually(hasSecret(ctx, barName, "foo-sec")).Should(BeFalse())
```
adrianludwin: You could take out the "Eventually" here because we don't expect this to _become_ false - we expect it to _stay_ false.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -41,22 +69,19 @@ var _ = Describe("HNCConfiguration", func() {
...
-
-		Eventually(func() error {
-			c := getHNCConfig(ctx)
-			return addSecretToHNCConfig(ctx, c)
```
adrianludwin: Do we not need a test for this anymore? E.g. that a type that _is_ explicitly added to the config _should_ be propagated? It can be its own test now (eg we don't need a "before-and-after" test, we can have two tests, one that adds Secrets to the config and one that doesn't).

sophieliu15: I changed the test to option 2 that we discussed in the previous PR (https://github.com/kubernetes-sigs/multi-tenancy/pull/438#discussion_r381444776). I create objects of two types in the parent namespace and only add one type to config. The test shows that only objects of the type that is added to the config are populated to the child namespace. Feel free to take a look.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Thanks for the suggestion. I changed the test to address your comment below.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Sure! Changed to Igniter.

incubator/hnc/pkg/reconcilers/hnc_config.go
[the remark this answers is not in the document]
sophieliu15: Thanks for the suggestion! Changed to forceInitialReconcile() and called it after SetupWithManager() in setup.go. Also added the comments in setup.go to discuss the issue with r.Get() outside of Reconcile() that we discovered.

incubator/hnc/pkg/reconcilers/setup.go
```diff
@@ -41,11 +41,22 @@ func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int) error {
 		Log:                    ctrl.Log.WithName("reconcilers").WithName("HNCConfiguration"),
 		Manager:                mgr,
 		Forest:                 f,
+		Igniter:                make(chan event.GenericEvent),
 		HierarchyConfigUpdates: hcChan,
 	}
 	if err := cr.SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("cannot create Config reconciler: %s", err.Error())
 	}
+	// Create a default singleton if there is no singleton in the cluster.
```
adrianludwin: Hmm, why not move all this into `cr.SetupWithManager`? This seems like we're leaking a bunch of internal logic about the HNCConfig reconcilers into setup.go.

incubator/hnc/pkg/reconcilers/hnc_config.go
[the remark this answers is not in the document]
sophieliu15: > I have no _reason_ to be scared of this, but this makes me a little scared. I feel like it might be safer to enqueue something _after_ the reconciler's been started?

I might misunderstand this. Did you mean doing the enqueue after SetupWithManager() is called? That's why I moved it to setup.go and called it after SetupWithManaget().

incubator/hnc/pkg/reconcilers/setup.go
sophieliu15: I might misunderstand your previous comment... In my previous version, I enqueued the object in SetupWithManager before the return statement. You commented "I have no reason to be scared of this, but this makes me a little scared. I feel like it might be safer to enqueue something after the reconciler's been started?" (see my comment reply above). That's why I moved it out of SetupWithManager. Did I miss anything here?

sophieliu15: I might misunderstand your previous comment... In my previous version, I enqueued the object in SetupWithManager before the return statement. You commented "I have no reason to be scared of this, but this makes me a little scared. I feel like it might be safer to enqueue something after the reconciler's been started?" (see my comment reply above). That's why I moved it out of SetupWithManager. Did I miss anything here?

incubator/hnc/pkg/reconcilers/hnc_config.go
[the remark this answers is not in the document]
adrianludwin: +1

incubator/hnc/pkg/reconcilers/setup.go
adrianludwin: The return statement is what does all the real work. I just meant that instead of saying:

```
enqueue()
return ctrl.NewControllerManagedBy(mgr).A().B().Complete(r)
```

It should be something like:
```
err := ctrl.NewControllerManagedBy(mgr).A().B().Complete(r)
if err != nil {
  return err
}
enqueue()
return nil
```

A little more wordy but it keeps everything in one place.

sophieliu15: Thanks so much for the explanation! Feel free to take a look at the updated code.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -171,9 +178,38 @@ func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind) e
...
+		inst := &api.HNCConfiguration{}
+		inst.ObjectMeta.Name = api.HNCConfigSingleton
+		r.Igniter <- event.GenericEvent{Meta: inst}
+	}()
+}
+
 // SetupWithManager builds a controller with the reconciler.
 func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewControllerManagedBy(mgr).
+	if err := ctrl.NewControllerManagedBy(mgr).
```
adrianludwin: Usually we'd put these in separate statements; the `if err := foo(); err != nil` pattern really only works as a one-liner. But I'll approve anyway :)

PR #455, opened 2020-02-26
Create the empty HierarchicalNamespaceReconciler

yiqigao217:
Create the empty HierarchicalNamespaceReconciler to only generate a log
with the information of the object being reconciled and setup with
controller manager.

Add "forbidden" state to CRD status.
Rename HierarchyReconciler to HierarchyConfigReconciler.

Tested on GKE: I created a sample hns object and found the log of
reconciling that hns object in Stackdriver.

Part of #457

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchical_namespace.go
+++ b/incubator/hnc/api/v1alpha1/hierarchical_namespace.go
@@ -20,14 +20,15 @@ import (
 )
 
 // HNSState describes the state of a hierarchical namespace. The state could be
-// "missing", "ok" or "conflict". The definitions will be described below.
+// "missing", "ok", "conflict" or "forbidden". The definitions will be described below.
 type HNSState string
 
 // HNSStates, which are documented in the comment to HierarchicalNamespaceStatus.State.
 const (
-	Missing  HNSState = "missing"
-	Ok       HNSState = "ok"
-	Conflict HNSState = "conflict"
+	Missing   HNSState = "missing"
+	Ok        HNSState = "ok"
+	Conflict  HNSState = "conflict"
+	Forbidden HNSState = "forbidden"
 )
 
 // HierarchicalNamespaceStatus defines the observed state of HierarchicalNamespace.
@@ -41,8 +42,11 @@ type HierarchicalNamespaceStatus struct {
 	//
 	// - "ok": the child namespace exists.
 	//
-	// - "conflict": a namespace of the same name already exists. If the validation webhook
-	// works properly, this should never happen since the HNS won't be created at the first place.
+	// - "conflict": a namespace of the same name already exists. The admission controller
+	// will attempt to prevent this.
+	//
+	// - "forbidden": the HNS was created in a namespace that doesn't allow children, such
+	// as kube-system or hnc-system. The admission controller will attempt to prevent this.
 	State HNSState `json:"status,omitempty"`
 }
 
--- a/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
@@ -1,0 +1,51 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconcilers
+
+import (
+	"github.com/go-logr/logr"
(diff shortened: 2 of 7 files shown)


incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"github.com/go-logr/logr"
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// HierarchicalNamespaceReconciler reconciles HierarchicalNamespace CRs to make sure
+// all the hierarchical namespaces are properly maintained.
+type HierarchicalNamespaceReconciler struct {
```
adrianludwin: I'm thinking it's a bit confusing to have a `HierarchicalNamespace` and `Hierarchy` reconciler. Maybe the other one should switch to `HierarchyConfig`?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -0,0 +1,49 @@
...
+	Log logr.Logger
+}
+
+// Reconcile sets up some basic variables and then calls the business logic.
+// It currently only generates some logs for testing.
+func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	if ex[req.Namespace] {
+		return ctrl.Result{}, nil
+	}
+
```
adrianludwin: Hmm. What does it mean if a user creates an HNS object in a namespace that should be excluded by HNC? Should we just ignore it?

yiqigao217: I think we can prevent it by webhook when users create it. We can still keep the exclusion here. I will add a comment here and a todo for webhook?

Or do you think we should add another status to HNS and report it here?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: sg

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
adrianludwin: I kind of like the idea of a status. I never like relying solely on webhooks :)

adrianludwin: smythe@ actually pointed out a downside of webhooks to me yesterday which I hadn't thought of (but fits my overall philosophy): imagine we make the list of excluded namespaces configurable one day. Then someone could exclude a namespace just as someone else creates an HNS. The webhook might pass, but by the time the object's _actually_ submitted, the NS is now excluded. Granted, this seems unlikely but it's an example of why we shouldn't assume that all our state is actually good.

yiqigao217: I added an "error" state to the CRD status and a todo comment here.

incubator/hnc/api/v1alpha1/hierarchical_namespace.go
```diff
@@ -43,6 +44,10 @@ type HierarchicalNamespaceStatus struct {
 	//
 	// - "conflict": a namespace of the same name already exists. If the validation webhook
 	// works properly, this should never happen since the HNS won't be created at the first place.
+	//
+	// - "error": there's something wrong with the HNS. Currently, it would be that
```
adrianludwin: Maybe make this more specific - eg `"forbidden": the HNS was created in a namespace that doesn't allow children, such as kube-system or hnc-system`.

We don't have a human-readable field so we should be as specific as possible IMO.

incubator/hnc/api/v1alpha1/hierarchical_namespace.go
```diff
@@ -43,6 +44,10 @@ type HierarchicalNamespaceStatus struct {
...
+	// the parent namespace is not allowed to have subnamespaces. If the validation webhook
+	// works properly, this should not happen since the HNS won't be created at the first place.
```
adrianludwin: Maybe instead of "If the validation webhook works properly...", replace this with "The admission controller will attempt to prevent this."?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -0,0 +1,50 @@
...
+// all the hierarchical namespaces are properly maintained.
+type HierarchicalNamespaceReconciler struct {
+	client.Client
+	Log logr.Logger
+}
+
+// Reconcile sets up some basic variables and then calls the business logic.
+// It currently only generates some logs for testing.
+func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	// Todo change hns.status.state to "error" if the namespace is excluded.
```
adrianludwin: File issue and reference the issue here?

adrianludwin:
/lgtm
/approve
/hold
/assign @rjbez17 

lgtm after the last few comments are resolved

yiqigao217:
All the comments are resolved.

Issue #457, opened 2020-02-26
HNC: Implement self-service namespace


yiqigao217:
Allow tenant admins who do not have cluster-wide namespace creation privileges to still create subnamespaces by creating custom resource instances locally. 

The implementation includes:
* Define `HierarchicalNamespace` (`hns`) CRD
* Add `allowCascadingDelete` field to `HierarchyConfig.spec`
* Add `kubectl` plugin to toggle `HierarchyConfig.spec.allowCascadingDelete`
* Create an HNS reconciler to create/delete subnamespaces, interact with HC reconciler, etc. 
* Add webhook on the creation of HNS
* Add `kubectl` plugin to create an HNS

See design: [HNC self-service namespaces UX](http://bit.ly/hnc-self-serve-ux)

adrianludwin:
I think we're done with this? Congrats!

yiqigao217:
> I think we're done with this? Congrats!

Thanks! The only thing left is the flaky test #560 . As for `EX` map, I unlinked the issue #495  to this main issue.

adrianludwin:
I think you can close this one, the hnc-v0.3 milestone is sufficiently
tracking IMO.

yiqigao217:
I see, sure!

PR #460, opened 2020-02-26
Add `kubectl hns create2` to only create hns in parent namespace

yiqigao217:
Add a new `kubectl hns create` command `create2` to only create hns
instances in the parent namespace when creating subnamespaces.
Currently, `create` creates the hc singleton in the parent namespace.
`create2` will replace the current `create` when the hns reconciler is
implemented. The hc singletons will then be created by the hc or the hns
reconciler.

Tested on GKE cluster. I created a subnamespace using "kubectl hns
create2 -n parent child". The "child" hns instance was created in the
"parent" namespace and no "hierarchy" singleton hc instance was created
by this command. Then I saw the hns reconciler log on the hns instance
in Stackdriver.

Part of #457

### DIFF
--- a/incubator/hnc/pkg/kubectl/create2.go
+++ b/incubator/hnc/pkg/kubectl/create2.go
@@ -1,0 +1,46 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubectl
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var create2Cmd = &cobra.Command{
+	Use:   "create2 -n parent child",
+	Short: "Creates a subnamespace under the given parent.",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		parent, _ := cmd.Flags().GetString("namespace")
+		if parent == "" {
+			fmt.Println("Error: parent must be set via --namespace or -n")
+			os.Exit(1)
+		}
+		// Create the hns instance, the custom resource representing the self-serve subnamespace.
+		// TODO: ensure the specified child doesn't already exist and the parent is allowed
+		//  to create self-serve subnamespaces through the admission controller. See issue
+		//  https://github.com/kubernetes-sigs/multi-tenancy/issues/458
+		client.createHierarchicalNamespace(parent, args[0])
+	},
+}
+
+func newCreate2Cmd() *cobra.Command {
+	create2Cmd.Flags().StringP("namespace", "n", "", "The parent namespace for the new self-serve subnamespace")
+	return create2Cmd
+}
--- a/incubator/hnc/pkg/kubectl/root.go
+++ b/incubator/hnc/pkg/kubectl/root.go
@@ -43,6 +43,7 @@ type realClient struct{}
 type Client interface {
 	getHierarchy(nnm string) *api.HierarchyConfiguration
 	updateHierarchy(hier *api.HierarchyConfiguration, reason string)
+	createHierarchicalNamespace(nnm string, hnnm string)
 }
 
 func init() {
@@ -87,6 +88,7 @@ func init() {
 	rootCmd.AddCommand(newDescribeCmd())
 	rootCmd.AddCommand(newTreeCmd())
 	rootCmd.AddCommand(newCreateCmd())
+	rootCmd.AddCommand(newCreate2Cmd())
 }
 
 func Execute() {
@@ -111,7 +113,7 @@ func (cl *realClient) getHierarchy(nnm string) *api.HierarchyConfiguration {
 	hier := &api.HierarchyConfiguration{}
 	hier.Name = api.Singleton
 	hier.Namespace = nnm
-	err := hncClient.Get().Resource(api.Resource).Namespace(nnm).Name(api.Singleton).Do().Into(hier)
+	err := hncClient.Get().Resource(api.HierarchyConfigurations).Namespace(nnm).Name(api.Singleton).Do().Into(hier)
 	if err != nil && !errors.IsNotFound(err) {
 		fmt.Printf("Error reading hierarchy for %s: %s\n", nnm, err)
 		os.Exit(1)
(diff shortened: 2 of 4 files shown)


incubator/hnc/pkg/kubectl/create.go
```diff
@@ -36,9 +38,20 @@ var createCmd = &cobra.Command{
 		// another namespace, this will be caught by the admission controller.
 		updates := hcUpdates{requiredChildren: args}
 		updateHC(client, updates, parent)
+		// Create hns instances.
+		createHNS(client, parent, args)
```
adrianludwin: I'm not sure if it's a good idea to do these both in the same command (`k hns create`). Shouldn't we be able to test these independently? I'd prefer `k hns create` to stay as it is for now, and maybe add `k hns create2` to test the new version. Or is there a good reason to keep them together?

incubator/hnc/pkg/kubectl/create.go
```diff
@@ -36,9 +38,20 @@ var createCmd = &cobra.Command{
...
 	},
 }
 
+func createHNS(cl Client, ns string, nhnss []string) {
+	for _, nhns := range nhnss {
+		hns := &api.HierarchicalNamespace{}
+		hns.Name = nhns
+		hns.Namespace = ns
+		cl.createHierarchicalNamespace(hns)
```
adrianludwin: Since we'll never modify any properties other than the name and ns, maybe the client should just take those explicitly and do everything internally? Eg:

`cl.createHNS(parentNS, childNS)`

incubator/hnc/pkg/kubectl/create.go
yiqigao217: SG. I created a new `create2` and updated the commit title and body with more details.

incubator/hnc/pkg/kubectl/root.go
```diff
@@ -123,16 +125,28 @@ func (cl *realClient) updateHierarchy(hier *api.HierarchyConfiguration, reason s
...
 	}
 }
 
+func (cl *realClient) createHierarchicalNamespace(nnm string, hnnm string) {
+	hns := &api.HierarchicalNamespace{}
+	hns.Name = hnnm
+	hns.Namespace = nnm
+	err := hncClient.Post().Resource(api.HierarchicalNamespaces).Namespace(nnm).Name(hnnm).Body(hns).Do().Error()
+	if err != nil {
+		fmt.Printf("\nCould not create hierarchicalnamespace instance.\nReason: %s\n", err)
```
sophieliu15: nit: Do we need space between "hierarchical" and "namespace"? Or change the message below to remove the space for consistency.

incubator/hnc/pkg/kubectl/create2.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func createHNSs(cl Client, nnm string, hnnms []string) {
+	for _, hnnm := range hnnms {
+		cl.createHierarchicalNamespace(nnm, hnnm)
+	}
+}
+
+func newCreate2Cmd() *cobra.Command {
+	create2Cmd.Flags().StringP("namespace", "n", "", "The parent namespace for the new child namespace(s)")
```
sophieliu15: nit: Maybe we need to unify the name for subnamespaces? In line 27, it is "subnamespaces", here it is "child namesoace(s)", and line 35 it is "self-serve subnamespaces" (may not be a concern since it is a comment). Maybe we change "child namespace(s)" here to "subnamespaces" too? Also I assume subnamespaces == self-serve subnamespaces?

sophieliu15: nit: Maybe we need to unify the name for subnamespaces? In line 27, it is "subnamespaces", here it is "child namesoace(s)", and line 35 it is "self-serve subnamespaces" (may not be a concern since it is a comment). Maybe we change "child namespace(s)" here to "subnamespaces" too? Also I assume subnamespaces == self-serve subnamespaces?

incubator/hnc/pkg/kubectl/root.go
sophieliu15: nit: Do we need space between "hierarchical" and "namespace"? Or change the message below to remove the space for consistency.

incubator/hnc/pkg/kubectl/create2.go
yiqigao217: SG! I will change all to "self-serve subnamespaces" since it's the most accurate.

incubator/hnc/pkg/kubectl/root.go
yiqigao217: sg. I will change the below to "hierarchicalnamespace" since is the resource type.

incubator/hnc/pkg/kubectl/create2.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var create2Cmd = &cobra.Command{
+	Use:   "create2 -n parent child1 [child2 ...]",
+	Short: "Creates subnamespaces under the given parent.",
+	Args:  cobra.RangeArgs(1, 10),
```
adrianludwin: Does `k create ns` allow multiple args? If not, we shouldn't either, IMO?

incubator/hnc/pkg/kubectl/create2.go
```diff
@@ -0,0 +1,52 @@
...
+		parent, _ := cmd.Flags().GetString("namespace")
+		if parent == "" {
+			fmt.Println("Error: parent must be set via --namespace or -n")
+			os.Exit(1)
+		}
+		// Create hns instances, the custom resources representing the self-serve subnamespaces.
+		// TODO: ensure the specified children don't already exist and the parent is allowed
+		//  to create self-serve subnamespaces through the admission controller. See issue
+		//  https://github.com/kubernetes-sigs/multi-tenancy/issues/458
+		createHNSs(client, parent, args)
```
adrianludwin: `createHNSes` (if this is still plural). Although this is sufficiently short that you could also just inline it.

incubator/hnc/pkg/kubectl/create2.go
yiqigao217: No. sg.

incubator/hnc/pkg/kubectl/create2.go
yiqigao217: Inlined it.

incubator/hnc/pkg/kubectl/create2.go
```diff
@@ -0,0 +1,48 @@
...
+			fmt.Println("Error: parent must be set via --namespace or -n")
+			os.Exit(1)
+		}
+		// Create the hns instance, the custom resource representing the self-serve subnamespace.
+		// TODO: ensure the specified child doesn't already exist and the parent is allowed
+		//  to create self-serve subnamespaces through the admission controller. See issue
+		//  https://github.com/kubernetes-sigs/multi-tenancy/issues/458
+		for _, child := range args {
+			client.createHierarchicalNamespace(parent, child)
+		}
```
adrianludwin: There's only ever one arg now :) So we don't need the loop.

PR #462, opened 2020-02-27
Add Conditions in HNCConfiguration.Status to indicate error scenarios.

sophieliu15:
This PR adds Conditions in HNCConfiguration.Status to indicate error scenarios. Condition specifies error code and message for a specific error. Currently we support following error code:

    - critSingletonNameInvalid: the specified singleton name is invalid.
    - objectReconcilerCreationFailed: an error exists when creating the object reconciler for a specific type.

This PR includes changes introduced in #471

Issue: #411


sophieliu15:
I cannot modify reviewers as described [here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/requesting-a-pull-request-review) because the gear icon in the "Reviewer" section is gone. Maybe there are some issues with the github UI.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -38,33 +45,50 @@ type ConfigReconciler struct {
...
 func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	// TODO: Surface the error more prominently and add a validating admission controller to prevent
-	// the problem in the first place.
+	ctx := context.Background()
+
+	// Validate the singleton name.
+	// TODO: Add a validating admission controller to prevent the problem in the first place.
 	if req.NamespacedName.Name != api.HNCConfigSingleton {
-		r.Log.Error(nil, "Singleton name is wrong. It should be 'config'.")
+		nnm := types.NamespacedName{Name: req.NamespacedName.Name}
```
adrianludwin: Rather than reading/writing the "bad" config, maybe let's just add a condition to the "real" config and leave the bad one alone? That seems like a lot less code.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -38,33 +45,50 @@ type ConfigReconciler struct {
 // Reconcile sets up some basic variable and logs the Spec.
 // TODO: Updates the comment above when adding more logic to the Reconcile method.
 func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
-	// TODO: Surface the error more prominently and add a validating admission controller to prevent
-	// the problem in the first place.
+	ctx := context.Background()
```
adrianludwin: We've tried to keep most functions fairly short by aggressively moving code to subfunctions.  Maybe try to do the same here?

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -38,33 +45,50 @@ type ConfigReconciler struct {
...
+	inst.Status.Conditions = nil
+
+	// Create corresponding ObjectReconcilers for newly added types, if needed.
 	// TODO: Rename the method syncObjectReconcilers because we might need more than creating ObjectReconcilers in future.
 	// For example, we might need to delete an ObjectReconciler if its corresponding type is deleted in the HNCConfiguration.
-	if err := r.createObjectReconcilers(inst); err != nil {
-		r.Log.Error(err, "Couldn't create the new object reconciler.")
-		return ctrl.Result{}, nil
+	if hasError := r.createObjectReconcilers(inst); hasError {
+		r.Log.Info("Object reconciler creation has an error.")
```
adrianludwin: Most of the error handling is already handled in `createObjectReconciler` so I'd log the problem there as well (you also have far more information there). Also, there's no need to name a temporary variable like this unless you want to refer to it again - though in this case, you can just change `createObjectReconcilers` so it doesn't return anything.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -133,17 +157,25 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) e
...
+
+	hasError := false
 	for _, t := range inst.Spec.Types {
 		gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
 		if r.Forest.HasTypeSyncer(gvk) {
 			continue
 		}
 		if err := r.createObjectReconciler(gvk); err != nil {
-			return fmt.Errorf("Error while trying to create ObjectReconciler: %s", err)
+			msg := fmt.Sprintf("Error while trying to create ObjectReconciler for type %s: %s", gvk, err)
```
adrianludwin: Maybe:

```
err := r.createObjectReconciler(gvk)
if err == nil {
  continue
}
... rest of error handling code ...
```

Alternatively, pass `inst` into `createObjectReconciler` and handle the error there?

adrianludwin: "Error while" makes sense for a log message, but not really for a status. The status should be more something like "Couldn't create reconciler for type %s: %s".

Also, if you want to log this here, log it as a real error: `r.log.Error(err, "Couldn't create reconciler", "type", gvk)`

sophieliu15:
Looks like I can only use the tag ("/assign") for assignee not for reviewer. @yiqigao217 feel free to take a look at the PR and leave any comments.

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -42,6 +42,13 @@ const (
 	Remove SynchronizationMode = "remove"
 )
 
+// HNCConfigurationCondition codes. *All* codes must also be documented in the
+// comment to HNCConfigurationCondition.Code.
+const (
+	CritSingletonNameInvalid       HNCConfigurationCode = "CRIT_SINGLETON_NAME_INVALID"
```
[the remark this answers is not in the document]
sophieliu15: I assume we are talking about the string literal here (e.g, "CRIT_SINGLETON_NAME_INVALID")? Changed to the camelCase for that. :)

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -88,6 +95,9 @@ type HNCConfigurationSpec struct {
 type HNCConfigurationStatus struct {
 	// Types indicates the observed synchronization states of kinds, if any.
 	Types []TypeSynchronizationStatus `json:"types,omitempty"`
+
+	// Conditions describes the errors, if any.
+	Conditions []HNCConfigurationCondition `json:"conditions,omitempty"`
```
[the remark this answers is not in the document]
sophieliu15: I was considering adding another field in `HNCConfigurationCondition` struct to indicate types when I did the implementation, e.g.,

```go
  type HNCConfigurationCondition struct {
    Code HNCConfigurationCode `json:"code"`
    Msg string `json:"msg,omitempty"`
    Affects []AffectedType `json:"affects, omitempty"`
 }

type AffectedType struct {
	APIVersion string `json:"apiVersion,omitempty"`
	Kind string `json:"kind,omitempty"`
}
```

But I didn't end up doing that because:
- Right now we only have two error conditions. The one that are associated with types (`objectReconcilerCreationFailed`) already has the type information in the Msg field. Adding the Affects field does not seem to provide any additional information.
- Also the other condition (`critSingletonNameInvalid`) is not related to any type and will have an empty `Affects` field. This might cause some confusion because when the field is empty we might not be able to tell if it impacts all types or if we miss the type information.

I am thinking about adding the field later if we end up having many error codes associated with specific types or if we find it is not enough to provide type information in the error message. Feel free to let me know if we want to include it in this PR or if there are any other approaches to associate a condition with types (I agreed with what you said that putting `Condition` in `Type` might be troublesome because some errors are not tied to any types).

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: I feel putting the error in the real config might be confusing: the singleton has the correct name "config" but has the error that ""Singleton name is wrong. It should be 'config'". 

Furthermore, when people create the object with a wrong name, they probably don't know the "config" object exists. If something is wrong, they are more likely to lookup the status of the object that they created to see what is going on instead of the status of the real "config" object.

Feel free to let me know your thoughts. :)

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Good idea, thanks for the suggestion. PTAL.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion. I moved the error log and condition change to `createObjectReconciler`. Changed both `createObjectReconcilers` and `creatObjectReconciler` to return nothing. Feel free to take a look.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Passed `inst` into `createObjectReconciler` to handle the error there.

Just out of curiosity... Is there any reason we prefer to handle the `nil` case? Seems to me that the opposite way (handle `err != nil` case) is quite common in our code ([example](https://github.com/kubernetes-sigs/multi-tenancy/blob/bb7eaf93524e68dc2ea42885b676b5a3d9ee728f/incubator/hnc/pkg/reconcilers/hierarchy_config.go#L242)) and in other places ([example](https://tour.golang.org/methods/19)).

Thanks so much for the suggestion!

sophieliu15: Thanks for the suggestion! Moved the error handling in `createObjectReconciler`. Logged it as real error there and also change the status message as you suggested. PTAL.

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -99,6 +109,35 @@ type HNCConfigurationList struct {
...
+	//
+	// All codes that begin with the prefix `crit` indicate that reconciliation has
+	// been paused for this configuration. Future changes of the configuration will be
+	// ignored by HNC until the condition has been resolved. Non-critical conditions
+	// typically indicate some kinds of error that HNC itself can ignore. However,
+	// the behaviors of some types might be out-of-sync with the users' expectations.
+	//
+	// Currently, the supported values are:
+	//
+	// - "critSingletonNameInvalid": the specified singleton name is invalid.
```
yiqigao217: Maybe add some additional guide information like users should either 1) rename it to "config" if this is the right and only existing `HNCConfiguration` resource they want to use or 2) delete it if the right "config" already exists?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -31,32 +37,85 @@ var _ = Describe("HNCConfiguration", func() {
...
+	It("should unset ObjectReconcilerCreationFailed condition if an object reconciler creation later succeeds", func() {
+		// API version of LimitRange should be "v1"
+		addToHNCConfig(ctx, "v2", "LimitRange", api.Propagate)
+
+		Eventually(hasHNCConfigurationCondition(ctx, api.ObjectReconcilerCreationFailed)).Should(BeTrue())
+
+		updateHNCConfigSpec(ctx, "v2", "v1", "LimitRange", "LimitRange", api.Propagate, api.Propagate)
+
+		Eventually(hasHNCConfigurationCondition(ctx, api.ObjectReconcilerCreationFailed)).Should(BeFalse())
+	})
```
yiqigao217: Maybe add that the type matches in addition to the error code? I saw your discussions on if to add the type to the struct. Once you decide, maybe add the type match here either by matching string or type subfield.

sophieliu15: I will add the match for the message for now. :) Thanks for the suggestion!

incubator/hnc/api/v1alpha1/hnc_config.go
sophieliu15: Good idea! I added another sentence to indicate what the correct name should be. It depends on the users if they want to delete the object that has the wrong name or not. When they rename it to "config" and do k apply -f  config.yaml  the "config" object will be updated with the contents in config.yaml.

incubator/hnc/api/v1alpha1/hnc_config.go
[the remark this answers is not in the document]
adrianludwin: Roger

incubator/hnc/api/v1alpha1/hnc_config.go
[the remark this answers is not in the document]
adrianludwin: No, I think you're right. I've been focused on trying to make everything machine-readable, but in this case I think it's fine to just mention the bad Kind in the human-readable error message. Maybe just make sure the documentation to the codes clearly state that `msg` will specify whatever information is needed.

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -99,6 +109,36 @@ type HNCConfigurationList struct {
...
+	//
+	// All codes that begin with the prefix `crit` indicate that reconciliation has
+	// been paused for this configuration. Future changes of the configuration will be
+	// ignored by HNC until the condition has been resolved. Non-critical conditions
+	// typically indicate some kinds of error that HNC itself can ignore. However,
+	// the behaviors of some types might be out-of-sync with the users' expectations.
+	//
+	// Currently, the supported values are:
+	//
+	// - "critSingletonNameInvalid": the specified singleton name is invalid. The name should be the
```
adrianludwin: Maybe: "An HNCConfiguration object exists with a name other than `config`. The object will be identified in the message."

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -99,6 +109,36 @@ type HNCConfigurationList struct {
...
+	// same as HNCConfigSingleton.
+	//
+	// - "objectReconcilerCreationFailed": an error exists when creating the object
+	// reconciler for a specific type.
```
adrianludwin: ... "for the type specified in `msg`."

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -38,34 +45,32 @@ type ConfigReconciler struct {
...
-	if req.NamespacedName.Name != api.HNCConfigSingleton {
-		r.Log.Error(nil, "Singleton name is wrong. It should be 'config'.")
+	ctx := context.Background()
+
+	// Validate the singleton name.
+	// TODO: Add a validating admission controller to prevent the problem in the first place.
+	if err := r.validateSingletonName(ctx, req.NamespacedName.Name); err != nil {
+		r.Log.Error(err, "Singleton name validation failed.")
 		return ctrl.Result{}, nil
 	}
```
adrianludwin: I'd put the error message in the "real" singleton - after all, that's where users will go to look for their problems. So this should really be something like:

`r.validateSingletonName(ctx, inst, req.NamespacedName);`

Note that this doesn't return an error - it just adds the condition to the _real_ singleton. You could _also_ add a condition to the "bad" config as well, if you like, but note that if you can't write back to it for any reason, you should probably just log that error and move on to try to write to the "real" one.

incubator/hnc/api/v1alpha1/hnc_config.go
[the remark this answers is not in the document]
sophieliu15: Cool! Updated the documentation of Msg to say that it will include type information if the condition is for specific types.

incubator/hnc/api/v1alpha1/hnc_config.go
sophieliu15: I didn't change this documentation because we probably cannot write the error message in the correct singleton. See my comment reply below.

incubator/hnc/api/v1alpha1/hnc_config.go
sophieliu15: Updated. Thanks for the suggestion.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: It doesn't look like this will work (also tested on the real cluster). When we do reconcile, we always update Status (i.e., Condition in the Status) to the latest one (feel free to correct me if this is not the case). If we write the error message to the correct singleton (one with name 'config'), the workflow will be like:
 
 1. Write the error condition to the correct singleton
 2.  Singleton change above will trigger a new reconciliation for the correct singleton. The request for the reconciliation will have the correct name so the Status of the correct singleton will be updated again and the error message will be removed.
 3.  Singleton change above will trigger a new reconciliation for the correct singleton. Nothing changed at this point, no more subsequent reconciliations triggered by invalid name.  

When user check the status of the correct singleton, it will be after step 3 when all the reconciliation chain finishes. At that point, the error message is already removed.  

Feel free to let me know if I miss anything. 

------------------------------------------------------------

Regarding to `validateSingletonName` not returning an error... I don't think it has to return an error, but I feel it should return some signal to indicate if the name is wrong or not. If the name is wrong, we don't need to execute rest of steps in the `Reconcile` (e.g., creating/updating object reconciliers). IIUC, we only consider the `Spec` in the correct singleton, not the wrong one.

incubator/hnc/pkg/reconcilers/hnc_config.go
adrianludwin: Yeah, maybe both are needed. Though I realize that if someone adds the "bad" config, we'll have to manually re-enqueue the good one too - the "ignite" channel can be used for that though.

On the good one, you can use a more descriptive error like `The HNC Config must be called "config", but an HNC Config object with the name "foo" exists as well.`

incubator/hnc/pkg/reconcilers/hnc_config.go
adrianludwin: I prefer whichever leads to the least indented code - nothing more interesting than that. Up to you.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Hi @adrianludwin 

Friendly ping in case you miss this reply. There are some issues about writing the error message to the correct object as I described above. Feel free to take a look. Thank you!

adrianludwin: Ok I think this is fine for now. I'll approve.

sophieliu15:
Hi @rjbez17 

Feel free to take a look at this PR when you have time. Adrian and Yiqi did the first round of the code review. Since I have a chain of PRs for review, it would be great if we can merge this in to reduce the code review burden for subsequent PRs. 

Thanks so much for your help!

Sophie

PR #464, opened 2020-02-28
[Incubator][VC]Make DWS request only contain immutable fields

Fei-Guo:
This change makes sure the request in DWS queue does not contain any mutable fields which can potentially cause parallel reconcile for the same object. 

Since event_type is removed, the DWS reconcile logic of figuring out whether it is Add or Update or Delete is changed by checking current tenant/super master informer cache.

We also add UID consistency check for Pod/Service/Namespace objects (other resources will be handled in future change) in order to avoid confusing update/create sync events when the object with the same name is frequently created and deleted.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/namespace/dws.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/namespace/dws.go
@@ -18,7 +18,6 @@ package namespace
 
 import (
 	"fmt"
-	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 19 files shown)


incubator/virtualcluster/pkg/syncer/resources/namespace/dws.go
```diff
@@ -40,64 +39,85 @@ func (c *controller) StartDWS(stopCh <-chan struct{}) error {
 
 // The reconcile logic for tenant master namespace informer
 func (c *controller) Reconcile(request reconciler.Request) (reconciler.Result, error) {
-	klog.V(4).Infof("reconcile namespace %s %s event for cluster %s", request.Name, request.Event, request.Cluster.Name)
+	klog.V(4).Infof("reconcile namespace %s for cluster %s", request.Name, request.ClusterName)
+	targetNamespace := conversion.ToSuperMasterNamespace(request.ClusterName, request.Name)
```
zhuangqh: request.Namespace

incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
```diff
@@ -174,34 +174,40 @@ func (c *controller) checkPods() {
...
-				if pPod.Spec.TerminationGracePeriodSeconds != nil {
-					gracePeriod = *pPod.Spec.TerminationGracePeriodSeconds
-				}
-				deleteOptions := metav1.NewDeleteOptions(gracePeriod)
-				deleteOptions.Preconditions = metav1.NewUIDPreconditions(string(pPod.UID))
-				if err = c.client.Pods(pPod.Namespace).Delete(pPod.Name, deleteOptions); err != nil {
-					klog.Errorf("error deleting pPod %v/%v in super master: %v", pPod.Namespace, pPod.Name, err)
-				} else {
-					metrics.CheckerRemedyStats.WithLabelValues("numDeletedOrphanSuperMasterPods").Inc()
+		if errors.IsNotFound(err) && pPod.DeletionTimestamp == nil {
```
zhuangqh: if err != nil && errors.IsNotFound(err) ...

PR #476, opened 2020-03-02
Support 'ignore' mode in ConfigReconciler.

sophieliu15:
This PR adds the support for 'ignore' mode in ConfigReconciler. Specifically, it adds following features:

- If the mode of a type is set to 'ignore', the corresponding object reconciler will ignore all subsequent reconciliations (e.g., new or changed objects will not be propagated).
- If the mode of a type is changed from 'ignore' to 'propagate', reconciliations for all existing objects of the type in the cluster will be triggered.
- If a mode of a type is unrecognized, it will be treated as 'ignore'.
- If a mode of a type is unset, it will be treated as 'propagate'.

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/forest/forest.go
+++ b/incubator/hnc/pkg/forest/forest.go
@@ -29,6 +29,9 @@ type TypeSyncer interface {
 	SyncNamespace(context.Context, logr.Logger, string) error
 	// Provides the GVK that is handled by the reconciler who implements the interface.
 	GetGVK() schema.GroupVersionKind
+	// SetMode sets the propagation mode of objects that are handled by the reconciler who implements the interface.
+	// The method also syncs objects in the cluster for the type handled by the reconciler if necessary.
+	SetMode(context.Context, api.SynchronizationMode, logr.Logger) error
 }
 
 // Forest defines a forest of namespaces - that is, a set of trees. It includes methods to mutate
@@ -40,7 +43,7 @@ type Forest struct {
 	lock       sync.Mutex
 	namespaces namedNamespaces
 
-	// types is a list of other reconcillers that HierarchyReconciler can call if the hierarchy
+	// types is a list of other reconcilers that HierarchyReconciler can call if the hierarchy
 	// changes. This will force all objects to be re-propagated.
 	//
 	// This is probably wildly inefficient, and we can probably make better use of things like
@@ -72,14 +75,15 @@ func (f *Forest) AddTypeSyncer(nss TypeSyncer) {
 	f.types = append(f.types, nss)
 }
 
-// HasTypeSyncer returns true if there is already a reconciler with the given GVK.
-func (f *Forest) HasTypeSyncer(gvk schema.GroupVersionKind) bool {
+// GetTypeSyncer returns the reconciler for the given GVK or nil if the reconciler
+// does not exist.
+func (f *Forest) GetTypeSyncer(gvk schema.GroupVersionKind) TypeSyncer {
 	for _, t := range f.types {
 		if t.GetGVK() == gvk {
-			return true
+			return t
 		}
 	}
-	return false
+	return nil
 }
 
 // GetTypeSyncers returns the types list.
@@ -113,6 +117,15 @@ func (f *Forest) Get(nm string) *Namespace {
 	}
 	f.namespaces[nm] = ns
 	return ns
+}
+
+// GetNamespaceNames returns names of all namespaces in the cluster.
+func (f *Forest) GetNamespaceNames() []string {
+	var keys []string
+	for k := range f.namespaces {
+		keys = append(keys, k)
+	}
+	return keys
 }
 
 type namedNamespaces map[string]*Namespace
--- a/incubator/hnc/pkg/reconcilers/hnc_config.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config.go
@@ -50,35 +50,35 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// Validate the singleton name.
 	// TODO: Add a validating admission controller to prevent the problem in the first place.
 	if err := r.validateSingletonName(ctx, req.NamespacedName.Name); err != nil {
-		r.Log.Error(err, "Singleton name validation failed.")
+		r.Log.Error(err, "Singleton name validation failed")
 		return ctrl.Result{}, nil
 	}
 
 	inst, err := r.getSingleton(ctx)
 	if err != nil {
-		r.Log.Error(err, "Couldn't read singleton.")
+		r.Log.Error(err, "Couldn't read singleton")
 		return ctrl.Result{}, err
 	}
 
 	// TODO: Modify this and other reconcilers (e.g., hierarchy and object reconcilers) to
 	// achieve the reconciliation.
-	r.Log.Info("Reconciling cluster-wide HNC configuration.")
+	r.Log.Info("Reconciling cluster-wide HNC configuration")
 
 	// Clear the existing conditions because we will reconstruct the latest conditions.
 	inst.Status.Conditions = nil
 
-	// Create corresponding ObjectReconcilers for newly added types, if needed.
-	// TODO: Rename the method syncObjectReconcilers because we might need more than creating ObjectReconcilers in future.
-	// For example, we might need to delete an ObjectReconciler if its corresponding type is deleted in the HNCConfiguration.
-	r.createObjectReconcilers(inst)
+	// Create or sync corresponding ObjectReconcilers, if needed.
+	syncErr := r.syncObjectReconcilers(ctx, inst)
 
 	// Write back to the apiserver.
 	// TODO: Update HNCConfiguration.Status before writing the singleton back to the apiserver.
 	if err := r.writeSingleton(ctx, inst); err != nil {
-		r.Log.Error(err, "Couldn't write singleton.")
+		r.Log.Error(err, "Couldn't write singleton")
 		return ctrl.Result{}, err
 	}
-	return ctrl.Result{}, nil
+
+	// Retry reconciliation if there is a sync error.
+	return ctrl.Result{}, syncErr
 }
 
 // getSingleton returns the singleton if it exists, or creates a default one if it doesn't.
@@ -120,11 +120,20 @@ func (r *ConfigReconciler) writeSingleton(ctx context.Context, inst *api.HNCConf
 	return nil
 }
 
-// createObjectReconcilers creates corresponding ObjectReconcilers for the newly added types in the
-// HNC configuration, if there is any. It also informs the in-memory forest about the newly created
-// ObjectReconcilers. If a type is misconfigured, the corresponding object reconciler will not be created.
-func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
-	// This method is guarded by the forest mutex. The mutex is guarding two actions: creating ObjectReconcilers for
+// syncObjectReconcilers creates or syncs ObjectReconcilers.
+//
+// For newly added types in the HNC configuration, the method will create corresponding ObjectReconcilers and
+// informs the in-memory forest about the newly created ObjectReconcilers. If a newly added type is misconfigured,
+// the corresponding object reconciler will not be created. The method will not return error while creating
+// ObjectReconcilers. Instead, any error will be written into the Status field of the singleton. This is
+// intended to avoid infinite reconciliation when the type is misconfigured.
+//
+// If a type exists, the method will sync the mode of the existing object reconciler
+// and update corresponding objects if needed. An error will be return to trigger reconciliation if sync fails.
+func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.HNCConfiguration) error {
+	// This method is guarded by the forest mutex.
+	//
+	// For creating an object reconciler, the mutex is guarding two actions: creating ObjectReconcilers for
 	// the newly added types and adding the created ObjectReconcilers to the types list in the Forest (r.Forest.types).
 	//
 	// We use mutex to guard write access to the types list because the list is a shared resource between various
@@ -136,28 +145,36 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
 	// from the forest for the object reconciliation, after we create ObjectReconcilers but before we write the
 	// ObjectReconcilers to the types list. As a result, objects of the newly added types might not be propagated correctly
 	// using the latest forest structure.
+	//
+	// For syncing an object reconciler, the mutex is guarding the read access to the `namespaces` field in the forest. The
+	// `namespaces` field is a shared resource between various reconcilers.
 	r.Forest.Lock()
 	defer r.Forest.Unlock()
 
 	for _, t := range inst.Spec.Types {
 		gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
-		if r.Forest.HasTypeSyncer(gvk) {
-			continue
+		if ts := r.Forest.GetTypeSyncer(gvk); ts != nil {
+			if err := ts.SetMode(ctx, t.Mode, r.Log); err != nil {
+				return err // retry the reconciliation
+			}
+		} else {
+			r.createObjectReconciler(gvk, t.Mode, inst)
 		}
-		r.createObjectReconciler(gvk, inst)
 	}
+
+	return nil
 }
 
 // createObjectReconciler creates an ObjectReconciler for the given GVK and informs forest about the reconciler.
-// TODO: May need to pass in spec instead to provide information about Mode to the ObjectReconciler in future.
-func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind, inst *api.HNCConfiguration) {
-	r.Log.Info("Creating an object reconciler.", "GVK", gvk)
+func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind, mode api.SynchronizationMode, inst *api.HNCConfiguration) {
+	r.Log.Info("Creating an object reconciler", "gvk", gvk, "mode", mode)
 
 	or := &ObjectReconciler{
 		Client:            r.Client,
 		Log:               ctrl.Log.WithName("reconcilers").WithName(gvk.Kind),
 		Forest:            r.Forest,
 		GVK:               gvk,
+		Mode:              mode,
 		Affected:          make(chan event.GenericEvent),
 		AffectedNamespace: r.HierarchyConfigUpdates,
 	}
@@ -188,7 +205,7 @@ func (r *ConfigReconciler) validateSingletonName(ctx context.Context, nm string)
 		return err
 	}
 
-	msg := "Singleton name is wrong. It should be 'config'."
+	msg := "Singleton name is wrong. It should be 'config'"
 	condition := api.HNCConfigurationCondition{
 		Code: api.CritSingletonNameInvalid,
 		Msg:  msg,
@@ -220,10 +237,11 @@ func (r *ConfigReconciler) forceInitialReconcile(log logr.Logger, reason string)
 
 // SetupWithManager builds a controller with the reconciler.
 func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	if err := ctrl.NewControllerManagedBy(mgr).
+	err := ctrl.NewControllerManagedBy(mgr).
 		For(&api.HNCConfiguration{}).
 		Watches(&source.Channel{Source: r.Igniter}, &handler.EnqueueRequestForObject{}).
-		Complete(r); err != nil {
+		Complete(r)
+	if err != nil {
 		return err
 	}
 	// Create a default singleton if there is no singleton in the cluster.
@@ -235,6 +253,6 @@ func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// cache is populated at the reconciliation stage. A default singleton will be
 	// created during the reconciliation if there is no singleton in the cluster.
 	r.forceInitialReconcile(r.Log, "Enforce reconciliation to create a default"+
-		"HNCConfiguration singleton if it does not exist.")
+		"HNCConfiguration singleton if it does not exist")
 	return nil
 }
--- a/incubator/hnc/pkg/reconcilers/hnc_config_test.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config_test.go
@@ -3,6 +3,7 @@ package reconcilers_test
 import (
 	"context"
 	"strings"
+	"time"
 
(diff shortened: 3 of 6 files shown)


sophieliu15:
@adrianludwin and @yiqigao217 PTAL. As usual, I cannot change existing reviewers.

adrianludwin:
I'll review this when the other change goes in but for now:

/ok-to-test

incubator/hnc/pkg/forest/forest.go
```diff
@@ -72,14 +78,14 @@ func (f *Forest) AddTypeSyncer(nss TypeSyncer) {
 	f.types = append(f.types, nss)
 }
 
-// HasTypeSyncer returns true if there is already a reconciler with the given GVK.
-func (f *Forest) HasTypeSyncer(gvk schema.GroupVersionKind) bool {
+// GetTypeSyncer returns the reconciler for the given GVK.
```
adrianludwin: ... or nil if it doesn't exist.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -136,28 +138,52 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
...
+			r.syncObjectReconciler(ctx, mode, ts)
+		} else {
+			r.createObjectReconciler(gvk, mode, inst)
+		}
+	}
+}
+
+// syncObjectReconciler syncs the mode of an ObjectReconciler. It also updates corresponding objects, if needed.
+func (r *ConfigReconciler) syncObjectReconciler(ctx context.Context, mode api.SynchronizationMode, ts forest.TypeSyncer) {
+	r.Log.Info("Syncing an object reconciler.", "gvk", ts.GetGVK())
```
adrianludwin: Nit: we generally don't have periods at the end of the messages, I don't think.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -136,28 +138,52 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
...
+
+	oldMode := ts.GetMode()
+	if mode != oldMode {
```
adrianludwin: Nit:
```
if mode == oldMode {
  return
}
```

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -136,28 +138,52 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
...
+		r.Log.Info("Changing Mode of the object reconciler", "old Mode", oldMode, "new Mode", mode)
```
adrianludwin: The keys should generally not have spaces in them so that they're easier to query. Also, you should include the GVK otherwise there's no way to know _which_ reconciler this line applies to.

```
r.Log.Info("Changing object reconciler mode", "gvk", ts.GetGVK(), "old", oldMode, "new", mode)
```

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -136,28 +138,52 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
...
+		ts.SetMode(mode)
+		// If we change from "ignore" mode to other modes (e.g., "propagate"), we need to update objects in the cluster
+		// (e.g., propagate existing objects).
+		if oldMode == api.Ignore {
+			err := ts.SyncCluster(ctx, r.Log)
+			if err != nil {
+				r.Log.Error(err, "Error while trying to sync ObjectReconciler", "gvk", ts.GetGVK())
+			}
```
adrianludwin: This is one of those cases where we probably _do_ want to return the error so that we can retry if something went wrong. It's probably best to log at the deepest level (i.e. inside `ts.SyncCluster`), but return the error as well.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -136,28 +138,52 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
...
-		r.createObjectReconciler(gvk, inst)
 	}
 }
 
 // createObjectReconciler creates an ObjectReconciler for the given GVK and informs forest about the reconciler.
-// TODO: May need to pass in spec instead to provide information about Mode to the ObjectReconciler in future.
-func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind, inst *api.HNCConfiguration) {
-	r.Log.Info("Creating an object reconciler.", "GVK", gvk)
+func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind, mode api.SynchronizationMode, inst *api.HNCConfiguration) {
+	r.Log.Info("Creating an object reconciler.", "gvk", gvk, "Mode", mode)
```
adrianludwin: Nit: no period, also lowercase `"Mode"`

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -218,12 +244,28 @@ func (r *ConfigReconciler) forceInitialReconcile(log logr.Logger, reason string)
 	}()
 }
 
+// getMode returns Mode field from a TypeSynchronizationSpec object. Please
+// see the comments of api.SynchronizationMode for currently supported modes.
+// If Mode is not set, it will be api.Propagate by default. Any unknown mode is
+// treated as api.Ignore.
+func (r *ConfigReconciler) getMode(t api.TypeSynchronizationSpec) api.SynchronizationMode {
+	switch mode := t.Mode; mode {
```
adrianludwin: Why not just `switch t.Mode`? I don't know this part of golang that well, is there a difference?

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -91,22 +95,53 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
+}
+
 func (r *ObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	if ex[req.Namespace] {
 		return ctrl.Result{}, nil
 	}
 
+	log := r.Log.WithValues("trigger", req.NamespacedName)
+	if r.Mode == api.Ignore {
+		log.Info("Ignore reconciliation.", "gvk", r.GVK)
```
adrianludwin: This should be log.V(1).Info at best, but you can probably just remove it too.

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -91,22 +95,53 @@ func (r *ObjectReconciler) SyncNamespace(ctx context.Context, log logr.Logger, n
...
 }
 
+// GetMode provides mode of objects that are handled by the reconciler.
+func (r *ObjectReconciler) GetMode() api.SynchronizationMode {
+	return r.Mode
+}
+
+// SetMode provides mode of objects that are handled by the reconciler.
+func (r *ObjectReconciler) SetMode(mode api.SynchronizationMode) {
+	r.Mode = mode
```
adrianludwin: Shouldn't we validate the value here? Below, we do a direct comparison to `if r.Mode == api.Ignore`, which means that if I accidentally say `mode: "ingore"`, the objects will _still_ be propagated.

sophieliu15: Right now we validate the mode when we get it from the spec (see `getMode` in `hnc_config.go`). Then we call `SetMode` and trigger the `Reconcile`. The mode should already be validated when we do the `if r.Mode == api.Ignore` comparison in `Reconcile` of `object.go`.

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: Makes sense. Removed the log.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: You are right. We can just use `switch t.Mode`.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Removed period and changed to lowercase.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion. Logged the error inside `ts.SyncCluster` and surfaced the error to trigger reconciling again.

We hope to retry in this case because we expect sync to succeed?

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion. Added "gvk" and removed spaces.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Done. Thanks for the suggestion.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion. Removed all periods in all log messages in this file.

incubator/hnc/pkg/forest/forest.go
sophieliu15: Done.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -105,6 +106,52 @@ var _ = Describe("HNCConfiguration", func() {
 		// is not added to HNCConfiguration.
 		Expect(hasResourceQuota(ctx, barName, "foo-resource-quota")).Should(BeFalse())
 	})
+
+	It("should not propagate objects if the mode of a type is ignore", func() {
+		setParent(ctx, barName, fooName)
+		makeSecret(ctx, fooName, "foo-sec")
+		makeResourceQuota(ctx, fooName, "foo-resource-quota")
```
yiqigao217: What is the mode here when the secret and resourceQuota are created in parent?

sophieliu15: Secret is "propagate" because it is set as that in previous test case. ResourceQuota hasn't been configured yet.

sophieliu15: Discussed offline. Moved `AddHNCConfig` at the beginning to make it clear what the mode will be for each type in the test case.

I will have a followup PR to change all mode of object reconcilers to 'ignore' if the corresponding type is no longer in the `config` singleton.

incubator/hnc/pkg/reconcilers/hnc_config.go
adrianludwin: Exactly, yes. We _expect_ all our calls to the K8s apiserver to succeed, which means that we _expect_ any failures to be transient and therefore retriable.

incubator/hnc/pkg/reconcilers/object.go
adrianludwin: Yup, I agree that this works today. I'm just worried about accidentally adding a new codepath in the future that bypasses the check.

rjbez17:
Bear with my dumb question here, but can you confirm the workflow in the following scenario?

1) I deploy HNC, I set the Config to `Remove` objects of type `Foo.Bar`
2) `Foo.Bar` Object types DNE on the cluster so it is set to `Ignore`
3) I deploy a CRD of type `Foo.Bar` into the cluster
4) ?

Looking at the code, it seems that `Foo.Bar` will be locked into `Ignore` until you essentially 'touch' the HNC Config. Is that correct? Is that desired? (I could have missed something in the code, I haven't had coffee yet)

I'm less concerned about `Remove` here as it wouldn't really matter in this usecase. But if we add more mode types in the future, I'd like to understand the flow.

Similarly, it'd seem that in step 1, if I explicitly set `Foo.Bar` to `Propagate` it would end up being `Ignore` while if I did not explicitly set it, it would be set to `Propagate`. 

If we self correct without a manual 'touch' here then just ignore me :)

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -136,28 +145,45 @@ func (r *ConfigReconciler) createObjectReconcilers(inst *api.HNCConfiguration) {
...
 	for _, t := range inst.Spec.Types {
 		gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
-		if r.Forest.HasTypeSyncer(gvk) {
-			continue
+		if ts := r.Forest.GetTypeSyncer(gvk); ts != nil {
+			err := r.syncObjectReconciler(ctx, t.Mode, ts)
+			// Return sync error here so that we can retry the reconciliation.
+			if err != nil {
+				return err
+			}
```
adrianludwin: For future reference, this is a good time to combine the lines:

```
if err := r.syncObjectReconciler(ctx, t.Mode, ts); err != nil {
  return err // retry the reconciliation
}
```

But I won't hold up the merge for this :)

adrianludwin: Also, why not fold the two-line function below into this one?

```
if err := ts.SetMode(ctx, r.Log, t.Mode); err != nil {
  return err // retry the reconciliation
}
```

You can move the logging into the method.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -14,6 +15,13 @@ import (
 )
 
 var _ = Describe("HNCConfiguration", func() {
+	// sleepSec is the time (in seconds) to sleep for objects propagation to take effect.
+	// time.Sleep only accepts integer values as input and from experiment 1
+	// second is the minimal integer seconds HNC needs to propagate an object.
+	// Therefore, we set sleep time as double of the minimal sleep time.
+	//
+	// We may need to increase the sleep time in future if HNC takes longer to propagate objects.
+	const sleepSec = 2
```
adrianludwin: Ah - this isn't how `time` works in golang :) See: https://golang.org/pkg/time/#Sleep:

```
time.Sleep(100 * time.Millisecond)
```

You've created an untyped constant here, which can't be passed directly to `time.Sleep()`. If you said:

```
const sleepSec = 2 * time.Second
```

then you wouldn't need to pass the units in the call to `time.Sleep`. Better yet, you could experiment with, say, `100 * time.Millisecond` to see if shorter times are still applicable.

Do you want to try out shorter times in this PR or another one?

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -96,17 +100,79 @@ func (r *ObjectReconciler) GetGVK() schema.GroupVersionKind {
...
+// SetMode sets the Mode field of an object reconciler and syncs objects in the cluster if needed.
+// The method will return an error if syncs fail.
+func (r *ObjectReconciler) SetMode(ctx context.Context, mode api.SynchronizationMode, log logr.Logger) error {
+	log = log.WithValues("gvk", r.GVK)
+	newMode := r.getValidateMode(mode, log)
+	oldMode := r.Mode
+	if newMode == oldMode {
+		return nil
+	}
+	r.Log.V(1).Info("Changing Mode of the object reconciler", "old", oldMode, "new", newMode)
```
adrianludwin: This doesn't need to be V(1) since it will only occur rarely. In fact, this is probably a pretty important thing to log, especially if you get rid of the log message in the caller like I suggested.

Also, no need to capitalize "Mode."

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -96,17 +100,79 @@ func (r *ObjectReconciler) GetGVK() schema.GroupVersionKind {
...
+	// 'remove' mode will be treated as 'ignore'. We will add api.Remove to the first
+	// case when we start to support the mode.
+	case api.Propagate, api.Ignore:
+		return mode
+	case "":
+		log.V(1).Info("Add mode because mode is not set", "mode", api.Propagate)
+		return api.Propagate
+	default:
+		log.V(1).Info("Update mode because mode is unrecognized", "mode", api.Ignore)
+		return api.Ignore
```
adrianludwin: These are also probably better as normal logs, not V(1).

They could probably also be made a bit more clear, IMO - for the first one:

`"Unset mode; using "propagate"`

For the second one:

`"Unrecognized mode; using "ignore", "mode", mode`

This ensures that the log includes the _bad_ value that was passed in.

sophieliu15:
Hi @rjbez17 

IIUC, is your concern about configuring a custom resource in the HNCConfiguration object instead of build-in resource (e.g., roles, rolebindings, secrets)? That is a good point. I haven't considered supporting customized resource in 
HNCConfiguration yet. Right now I am assuming the CRD exists before adding the type in the config (i.e., mostly I only consider types we previously supported in [gvk.go](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/pkg/config/gvk.go)). I experimented the workflow with the `CronTab` custom resource as defined [here](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition). Please see my comments below about current behaviors and desired behaviors.

> Bear with my dumb question here, but can you confirm the workflow in the following scenario?
> 
> 1. I deploy HNC, I set the Config to `Remove` objects of type `Foo.Bar`

At this point, if `Foo.Bar` is a random type that hasn't been defined (i.e., CRD does not exist), we will not create an object reconciler for it.
For the `config` object, it will have an entry in `Status` indicating the object reconciler creation failed. Here is an example, `CronTab` is a random type that I haven't defined yet.

```
Spec:
  Types:
    API Version:  v1
    Kind:         CronTab
    Mode:         remove
    API Version:  rbac.authorization.k8s.io/v1
    Kind:         Role
    Mode:         propagate
    API Version:  rbac.authorization.k8s.io/v1
    Kind:         RoleBinding
    Mode:         propagate
Status:
  Conditions:
    Code:  objectReconcilerCreationFailed
    Msg:   Couldn't create ObjectReconciler for type /v1, Kind=CronTab: no matches for kind "CronTab" in version "v1"
```
 

> 2. `Foo.Bar` Object types DNE on the cluster so it is set to `Ignore`

We do not change the spec of `config` object, so if you look at the content of `config` object (above), you will see the `Mode` for `Foo.Bar` is what you set before (i.e., `remove`). In terms of the "mode" of object reconciler (you can think of that as the behavior of the object reconciler or the `Mode` field of the `ObjectReconciler` struct), it is not set to anything because the object reconciler does not even exists.

> 3. I deploy a CRD of type `Foo.Bar` into the cluster

The object reconciler for `Foo.Bar` still does not exist; therefore, if you start to create `Foo.Bar` objects, it will not be propagated. This is the same behavior as `ignore` mode if the reconciler exists.

From my experiment, even you retouch the `config` object to trigger the reconciliation of `ConfigReconciler`, the object reconciler for `Foo.Bar` still cannot be created with the same error. Only when I reinstall HNC (`make deploy`), the object reconciler will be created.

Once the object reconciler is created, the mode of that object reconciler will be `ignore` because we treat `remove` as an unrecognized mode in this PR (see `getValidateMode` in `object.go`), but it will be `remove` mode once we support `remove`.

> 4. ?
> 
> Looking at the code, it seems that `Foo.Bar` will be locked into `Ignore` until you essentially 'touch' the HNC Config. Is that correct? Is that desired? (I could have missed something in the code, I haven't had coffee yet)

The `Foo.Bar` objects will not be reconciled because the object reconciler for `Foo.Bar` does not exist. We will create object reconciler for it not after we retouch `config` object, but after we reinstall HNC.

This may not be ideal but I feel the solution might more on the documentation side than technical side. For example, we can add in the user guide that:
 * Only add type in `config` object if the type exists or
 * Reinstall HNC if they add a new CRD while the HNC is running.

At the same time, we can also add a feature when we implement the admission controller to automatically remove undefined type from the spec (not sure if this is feasible), but this does not solve the issue that we need to reinstall HNC to recognize custom resource defined while running HNC.

> 
> I'm less concerned about `Remove` here as it wouldn't really matter in this usecase. But if we add more mode types in the future, I'd like to understand the flow.
> 
> Similarly, it'd seem that in step 1, if I explicitly set `Foo.Bar` to `Propagate` it would end up being `Ignore` while if I did not explicitly set it, it would be set to `Propagate`.

For step 1, if the `Foo.Bar` type does not exist, the mode in Spec will be whatever you set it to be. However, this does not matter since the object reconciler for `Foo.Bar` does not exist.

If the type exists, `Propagate` and `Ignore` will still be `Propagate` and `Ignore`. If it is not set, it will be `Propagate`, anything else will be `Ignore` (including `Remove` until we start to support that mode). See details in `getValidateMode` in `object.go`

> 
> If we self correct without a manual 'touch' here then just ignore me :)

Feel free to let me know if there are any questions or your thoughts about this issue.

(Also add @adrianludwin in the discussion for his thoughts)

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: Thanks for the explanation. Done.

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: Thanks for the explanation. Done.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Thanks so much for the explanation. I tried the millisecond. Looks like the minimal time it needs is around 15 millisecond (0.015s). As we discussed, I changed sleep time to 1 second.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion! Updated to use the second example.

adrianludwin:
> From my experiment, even you retouch the `config` object to trigger the reconciliation of `ConfigReconciler`, the object reconciler for `Foo.Bar` still cannot be created with the same error. Only when I reinstall HNC (`make deploy`), the object reconciler will be created.

This is a bug, and I'm pretty sure we can find a way to fix it - let's just file an issue for now. For example, we can add a watch onto CRD types - or maybe just use the webhook to stop it from happening. You can also always just kill the manager pod to restart HNC, you don't need to fully reinstall it.

adrianludwin:
@rjbez17 , did Sophie answer your questions?

sophieliu15:
> > From my experiment, even you retouch the `config` object to trigger the reconciliation of `ConfigReconciler`, the object reconciler for `Foo.Bar` still cannot be created with the same error. Only when I reinstall HNC (`make deploy`), the object reconciler will be created.
> 
> This is a bug, and I'm pretty sure we can find a way to fix it - let's just file an issue for now. For example, we can add a watch onto CRD types - or maybe just use the webhook to stop it from happening. You can also always just kill the manager pod to restart HNC, you don't need to fully reinstall it.

Fired #488 to track this issue.

PR #478, opened 2020-03-02
Update HNS and HC reconcilers to allow hns creation

yiqigao217:
Create namespace and hierarchyconfig instances of the self-serve
subnamespace in the HNS reconciler.

Make changes to the HC reconciler to use the list of hns objects in the
namespace as requiredChildren instead of using the
spec.requiredChildren.

Tested on GKE cluster. First, added the
`enable-hierarchicalnamespace-reconciler` flag into the
config/default/manager_auth-proxy_patch.yaml file. Used `kubectl hns
create2 -n parent child` and found the hc and the hns objects were
created in the `parent` namespace. `child` namespace and its hc object
were also created. I also saw the test log in Stackdriver saying that
the `requiredChildOf` field of the subnamespace in the forest was set to
the `parent`.

Part of #457

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchical_namespace.go
+++ b/incubator/hnc/api/v1alpha1/hierarchical_namespace.go
@@ -19,8 +19,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Constant for the hierarchicalnamespaces resource type.
-const HierarchicalNamespaces string = "hierarchicalnamespaces"
+// Constants for the HierarchicalNamespace resource, kind and API version.
+const (
+	HierarchicalNamespaces           = "hierarchicalnamespaces"
+	HierarchicalNamespacesKind       = "HierarchicalNamespace"
+	HierarchicalNamespacesAPIVersion = "hnc.x-k8s.io/v1alpha1"
+)
 
 // HNSState describes the state of a hierarchical namespace. The state could be
 // "missing", "ok", "conflict" or "forbidden". The definitions will be described below.
--- a/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
@@ -16,8 +16,11 @@ limitations under the License.
 package reconcilers
 
 import (
+	"context"
+
 	"github.com/go-logr/logr"
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -30,7 +33,7 @@ type HierarchicalNamespaceReconciler struct {
 }
 
 // Reconcile sets up some basic variables and then calls the business logic.
-// It currently only generates some logs for testing.
+// It currently handles basic creation of namespace and hierarchyconfig.
 func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// TODO report error state if the webhook is bypassed - see issue
 	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/459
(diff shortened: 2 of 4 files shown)


incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -30,20 +33,72 @@ type HierarchicalNamespaceReconciler struct {
...
+	nm := req.Name
+	pnm := req.Namespace
+
+	// We want to trigger the hierarchyConfig reconciliation on the self-serve namespace, since
+	// the parent's 'requiredChildren' field will be deprecated. The HierarchyConfig Reconciler
+	// will do the real job of updating the hierarchy in the forest and creating namespaces and
+	// hierarchyconfig (hc) instances.
+	//
+	// Just enqueuing an hc instance in the memory for hc reconciliation will lose its parent
+	// information since the hc reconciler will get an empty instance from the apiserver.
```
sophieliu15: Maybe specify what "its" mean here? Does it mean the current namespace?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -30,20 +33,72 @@ type HierarchicalNamespaceReconciler struct {
...
 	if ex[req.Namespace] {
 		return ctrl.Result{}, nil
 	}
 
+	ctx := context.Background()
 	log := r.Log.WithValues("trigger", req.NamespacedName)
 	log.Info("Reconciling HNS")
 
+	nm := req.Name
+	pnm := req.Namespace
```
sophieliu15: Does "pnm" stand for "parent name"?  IIUC, req.Name is the name of an object and req.Namespace is the namespace of an object? In this sense, "pnm" might be confusing? Feel free to correct me if I am wrong.

yiqigao217: Yes. The `req` here stands for the request to reconcile an hns object, of which the namespace is the parent namespace name and the name is the self-serve subnamespace name.

I will add a comment to explain this. LMK if you think other names would be better :)

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: SG, the comment is updated!

It means the parent field in the hc object of the current self-serve subnamespace.

incubator/hnc/api/v1alpha1/hierarchical_namespace.go
```diff
@@ -19,8 +19,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Constant for the hierarchicalnamespaces resource type.
-const HierarchicalNamespaces string = "hierarchicalnamespaces"
+// Constants for the HierarchicalNamespace resource, kind and API version.
+const (
+	HierarchicalNamespaces           = "hierarchicalnamespaces"
+	HierarchicalNamespacesKind       = "HierarchicalNamespace"
+	HierarchicalNamespacesAPIVersion = "hnc.x-k8s.io/v1alpha1"
```
adrianludwin: I think the "hnc.x-k8s.io" constant is defined somewhere else, maybe use that here as well? I think you can say something like `foo = MetaGroup + "/v1alpha1"`

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -30,20 +33,74 @@ type HierarchicalNamespaceReconciler struct {
...
+	ctx := context.Background()
 	log := r.Log.WithValues("trigger", req.NamespacedName)
 	log.Info("Reconciling HNS")
 
+	// Names of the self-serve subnamespace and the parent namespace.
+	nm := req.Name
+	pnm := req.Namespace
+
+	// We want to trigger the hierarchyConfig reconciliation on the self-serve subnamespace, since
+	// the parent's 'requiredChildren' field will be deprecated. The HierarchyConfig Reconciler
```
adrianludwin: This comment won't make much sense in about a week or two - maybe rewrite it as though `requiredChildren` never existed?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -30,20 +33,74 @@ type HierarchicalNamespaceReconciler struct {
...
+	// will do the real job of updating the hierarchy in the forest and creating namespaces and
+	// hierarchyconfig (hc) instances.
+	//
+	// Please note that the self-serve subnamespace and its hc object don't exist on apiserver for now.
+	// Just enqueuing an in-memory hc instance for hc reconciliation will lose the hc's parent field
+	// since the hc reconciler will get nothing from the apiserver and create an empty in-memory hc.
```
adrianludwin: Hmm, this isn't what I had in mind. I thought we were going to do this:

* In the HNS reconciler:
  * Create the child namespace in the forest (_not_ on the apiserver) and set its owner to the parent namespace
  * Enqueue the child namespace for reconciliation
* In the HC reconciler:
  * Notice that the child namespace doesn't exist on the apiserver, but that it _does_ exist in the forest
  * Create all apiserver objects

This centralizes all the writing of HCs and namespaces in one place (the HC reconciler) so the HNS reconciler only needs to read/write the HNS.

Does this make sense? Or is there a reason this doesn't work?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -99,8 +100,8 @@ func (r *HierarchyConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 }
 
 func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logger, nm string) error {
-	// Get the singleton and namespace.
-	inst, nsInst, err := r.getInstances(ctx, log, nm)
+	// Get the singleton, the namespace and a list of the names of the hns instances in the namespace.
+	inst, nsInst, hnsnms, err := r.getInstances(ctx, log, nm)
```
adrianludwin: If the HNS reconciler updates the in-memory forest, there shouldn't be any need to read it here.

yiqigao217:
Closing this since it's redone by #485 using the approach suggested by https://github.com/kubernetes-sigs/multi-tenancy/pull/478#discussion_r387019984

PR #491, opened 2020-03-05
Remove duplicate code in reconcilers_test package

sophieliu15:
This PR removes duplicate code in reconcilers_test package. Specifically, the duplicate code in methods that make an object (e.g., `makeSecret`), check if an object exists (e.g, `hasSecret`), and get the namespace where an object is inherited from (e.g., `secretInheritedFrom`) are removed.

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hnc_config_test.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config_test.go
@@ -8,10 +8,8 @@ import (
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 var _ = Describe("HNCConfiguration", func() {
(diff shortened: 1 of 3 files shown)


sophieliu15:
Hi @adrianludwin and @yiqigao217 , this is the test refactoring PR I mentioned in our standup today. PTAL.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -120,36 +118,18 @@ func hasTypeWithMode(apiVersion, kind string, mode api.SynchronizationMode, conf
 }
 
 func makeSecret(ctx context.Context, nsName, secretName string) {
-	sec := &corev1.Secret{}
-	sec.Name = secretName
-	sec.Namespace = nsName
-	ExpectWithOffset(1, k8sClient.Create(ctx, sec)).Should(Succeed())
+	makeObject(1, ctx, GVKs["Secret"], nsName, secretName)
```
adrianludwin: This is a great cleanup! But I wonder if it could be even better - get rid of all the type-specific functions, and also allow you to pass in the string directly instead of using the `GVKs` map? Eg, instead of calling `makeSecret(ctx, nnm, nm)`, just call `makeObject(ctx, "Secret", nnm, nm)`? That will make this change a bit messier as you modify all the call sites, but much better after that.

sophieliu15: Good idea. I agree it is better to get rid of them once and for all. Please see the updated code.

PR #494, opened 2020-03-05
[Incubator][VC]Avoid using Secret/SA name in Labels

Fei-Guo:
SA and secret name can be as long as 253 characters but the Label value max length is limited to 63 characters. We should not use name in Labels to identify pSecret that represents tenant service account secret. 

In this change, we switch to use sa/secret UID to identify pSecret that represents tenant service account secret.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
@@ -107,7 +107,7 @@ func PodMutateDefault(vPod *v1.Pod, SASecret *v1.Secret, services []*v1.Service,
 		}
 
 		for i, volume := range p.pPod.Spec.Volumes {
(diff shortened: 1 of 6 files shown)


incubator/virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -374,9 +374,16 @@ func (s *saSecretMutator) Mutate(vSecret *v1.Secret, clusterName string) {
 	if labels == nil {
 		labels = make(map[string]string)
 	}
-	labels[constants.LabelSecretName] = vSecret.Name
-	labels[constants.LabelServiceAccountName] = vSecret.GetAnnotations()[v1.ServiceAccountNameKey]
+	annotations := s.pSecret.GetAnnotations()
+	if annotations == nil {
+		labels = make(map[string]string)
```
zhuangqh: s/labels/annotations/

Issue #487, opened 2020-03-03
HNC: Properly update the HierarchicalNamespace.Status.State


yiqigao217:
Part of #457 

The changes of the state include 
1) missing -> ok when the subnamespace is created; 
2) missing -> conflict when any conflict is detected;
etc.

We will also move the update of the `RequiredChildConflict` from `syncChildren()` in HC reconciler to HNS reconciler.

yiqigao217:
Fixed by #496

Issue #459, opened 2020-02-26
HNC: Report error state if hns bypassed validation webhook


yiqigao217:
Part of #457 

In HNS reconciliation, if the hns has somehow bypassed validation webhook, change the `hns.status.state` to report error state and early exit.

yiqigao217:
Fixed by #496

PR #496, opened 2020-03-05
Update transient & report invalid HNS states

yiqigao217:
Update transient HNS states and report invalid states if bypassing the
admission controller. Add a channel in the HNS reconciler to enqueue
HNS for reconciliation. Add HNS reconciler and HC reconciler into each
other so that they can enqueue the other resource. Sync the HNS
instances in HNS reconciler, instead of syncing the
spec.requiredChildren in "syncChildren()" func in HC reconciler. Add
integration tests to test all different states.

Tested with integration tests ("make test" & "make test HNS=1") and
on GKE cluster. To test manually, I added the "enable-hns-reconciler"
flag and commented out the webhook "Prevent changing parent of a
required child" part, since it uses the "RequiredChildOf" field in the
forest. I tested deleting the subnamespace, changing the annotation,
change the parent of the subnamespace. They all worked as expected.

Part of #457 , #459 , #487 .

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -32,13 +32,15 @@ const (
 )
 
 // Condition codes. *All* codes must also be documented in the comment to Condition.Code.
+// TODO change condition codes to CamelCase strings. See issue:
+//  https://github.com/kubernetes-sigs/multi-tenancy/issues/500
 const (
-	CritParentMissing     Code = "CRIT_PARENT_MISSING"
-	CritParentInvalid     Code = "CRIT_PARENT_INVALID"
-	CritAncestor          Code = "CRIT_ANCESTOR"
-	RequiredChildConflict Code = "REQUIRED_CHILD_CONFLICT"
-	CannotUpdate          Code = "CANNOT_UPDATE_OBJECT"
-	CannotPropagate       Code = "CANNOT_PROPAGATE_OBJECT"
+	CritParentMissing    Code = "CRIT_PARENT_MISSING"
+	CritParentInvalid    Code = "CRIT_PARENT_INVALID"
+	CritAncestor         Code = "CRIT_ANCESTOR"
+	SubnamespaceConflict Code = "SUBNAMESPACE_CONFLICT"
+	CannotUpdate         Code = "CANNOT_UPDATE_OBJECT"
+	CannotPropagate      Code = "CANNOT_PROPAGATE_OBJECT"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
--- a/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
@@ -16,11 +16,19 @@ limitations under the License.
 package reconcilers
 
 import (
+	"context"
+
 	"github.com/go-logr/logr"
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // HierarchicalNamespaceReconciler reconciles HierarchicalNamespace CRs to make sure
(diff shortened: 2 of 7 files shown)


incubator/hnc/pkg/kubectl/root.go
```diff
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/reconcilers"
```
adrianludwin: Can you move this to the next "paragraph" on line 34? We've tried to keep the imports in three sections:

* Standard library
* External dependencies
* Internal imports

Thanks!

incubator/hnc/pkg/kubectl/root.go
```diff
@@ -43,6 +44,7 @@ type realClient struct{}
 type Client interface {
 	getHierarchy(nnm string) *api.HierarchyConfiguration
 	updateHierarchy(hier *api.HierarchyConfiguration, reason string)
+	hasNamespace(nnm string) bool
```
adrianludwin: This is never called through the `Client` interface, so it doesn't need to be here. Just having it on `realClient` struct is enough.

incubator/hnc/pkg/kubectl/root.go
```diff
@@ -135,10 +137,35 @@ func (cl *realClient) updateHierarchy(hier *api.HierarchyConfiguration, reason s
...
+	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/458
+	if reconcilers.EX[nnm] {
+		hns.Status.State = api.Forbidden
+	} else {
+		if cl.hasNamespace(hnnm) {
+			hns.Status.State = api.Conflict
+		} else {
+			hns.Status.State = api.Missing
+		}
+	}
```
adrianludwin: Clients are never supposed to set the status of an object, only the spec - only reconcilers can set the status, because they can change at any time. So I don't think you should make any changes to kubectl at all - the client is just doing what the user asked it to, and it's up to the admission controller to reject it, or the reconciler to mark it as invalid.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
+	}
+
+	// Report "Forbidden" state and early exist if the namespace is not allowed to self-serve
+	// namespaces but has bypassed the webhook and successfully created the hns instance.
+	// TODO refactor/split the EX map for 1) reconciler exclusion and 2) self-serve not allowed
+	//  purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
+	if EX[pnm] {
+		inst.Status.State = api.Forbidden
+		err := r.writeInstance(ctx, log, inst)
+		return ctrl.Result{}, err
```
adrianludwin: Nit: if you want, you can say `return ctrl.Result{}, r.writeInstance(ctx, log, inst)`. Up to you.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
+
+	// Get the self-serve namespace's hierarchyConfig and namespace instances.
+	hcInst, nsInst, err := r.hcr.GetInstances(ctx, log, nm)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	r.syncWithForest(log, inst, nsInst, hcInst)
+
+	// TODO report the "RequiredChildConflict" in the HNS reconciliation. See issue:
```
adrianludwin: We should change `RequiredChildConflict` to a better name, eg `SubnamespaceConflict` or something like that. We don't use the term "Required Child" in the new UX.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
+	defer r.forest.Unlock()
+
+	// Names of the hierarchical namespace and the current namespace.
+	nm := inst.Name
+	pnm := inst.Namespace
+
+	// Get the namespace instance in memory.
+	ns := r.forest.Get(nm)
+
+	// If the real namespace instance doesn't exist yet, update forest and enqueue the namespace
```
adrianludwin: This function's getting a bit long. It might be nice to say something like:

```
switch {
case ns.Inst.Name == "":
  r.syncMissing(...)
case nsInst.Annotations[api.AnnotationOwner] != pnm:
  r.syncConflicting(...)
default:
  r.syncExisting(...)
}
```

That cleanly exposes the top-level logic of this function and allows each sub-case to be handled in its own function. sgty?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
 
-	// TODO sync with forest to report conflicts in hns.Status.State. See issue:
-	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/487
+		// Enqueue the not-yet existent self-serve subnamespace. The HierarchyConfig reconciler will
+		// create the namespace and the HierarchyConfig instances on apiserver accordingly.
+		reason := "new/updated hierarchical namespace"
+		r.hcr.enqueueAffected(log, reason, nm)
 
-	return ctrl.Result{}, nil
+		inst.Status.State = api.Missing
```
adrianludwin: 👍 This is where we should be updating the state, not in the client.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
+		return
+	}
+
+	log.Info("The self-serve subnamespace already exists", "namespace", nsInst.Name)
+	// If the namespace is not a self-serve namespace or it's a self-serve namespace of another
+	// namespace. Report the conflict.
+	if nsInst.Annotations[api.AnnotationOwner] != pnm {
+		log.Info("The namespace cannot be used as a self-serve subnamespace")
```
adrianludwin: Maybe be more specific here? Include a log about what the conflict was (eg the value of the annotation).

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
+	}
+
+	log.Info("The namespace has the correct owner annotation", "annotation", nsInst.Annotations[api.AnnotationOwner])
+	switch hcInst.Spec.Parent {
+	case "":
+		log.Info("Parent is not set", "namespace", nm)
+		// This case is rare. It means the namespace is created with the right annotation but
+		// no HC. This could be a transient state before HC reconciler finishes creating the HC
+		// or a human manually created the namespace with the right annotation but no HC.
+		// Both cases meant to create this namespace as HNS.
```
adrianludwin: Nice comment!

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,176 @@ type HierarchicalNamespaceReconciler struct {
...
+		inst.Status.State = api.Ok
+	default:
+		log.Info("Self-serve subnamespace is already owned by another parent", "child", nm, "otherParent", hcInst.Spec.Parent)
+		inst.Status.State = api.Conflict
+	}
+}
+
+// enqueueAffected enqueues all affected hierarchicalNamespace instances for later reconciliation. This occurs
+// in a goroutine so the caller doesn't block; since the reconciler is never garbage-collected, this is safe.
+func (r *HierarchicalNamespaceReconciler) enqueueAffected(log logr.Logger, nm, pnm, reason string) {
```
adrianludwin: Maybe instead of `enqueueAffected`, just say `enqueue`? This only enqueues a single HNS object, it doesn't go looking for other stuff (like the other `enqueueAffected` methods).

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -56,15 +57,79 @@ var _ = Describe("Hierarchy", func() {
 	It("should set the self-serve subnamespace's owner annotation to the current namespace", func() {
 		Eventually(getNamespaceAnnotation(ctx, barName, api.AnnotationOwner)).Should(Equal(fooName))
 	})
+
+	It("should set the hns.status.state to Ok if the above tests all pass", func() {
```
adrianludwin: What do you mean by "the above tests?" All the tests are supposed to be independent.

Re-reading this, I think all the tests from lines 35-63 should be combined into one big test, and maybe line 31-32 from `BeforeEach` as well.

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -56,15 +57,79 @@ var _ = Describe("Hierarchy", func() {
...
+		kube_system_hns_bar := newHierarchicalNamespace(barName, "kube-system")
+		updateHierarchicalNamespace(ctx, kube_system_hns_bar)
+		Eventually(getHNSState(ctx, "kube-system", barName)).Should(Equal(api.Forbidden))
+	})
+
+	It("should set the hns.status.state to Missing if the self-serve subnamespace doesn't exist", func() {
+		bazName := createNSName("baz")
+		// This is a trick to disable hc reconciler on "baz" namespace by having "baz" in the excluded namespace list.
+		// Therefore the "baz" namespace won't be created even if the HNS reconciler enqueues the not-yet existing
+		// "baz" namespace for hc reconciler to reconcile and create.
```
adrianludwin: Nice trick :)

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -33,12 +33,12 @@ const (
 
 // Condition codes. *All* codes must also be documented in the comment to Condition.Code.
 const (
-	CritParentMissing     Code = "CRIT_PARENT_MISSING"
-	CritParentInvalid     Code = "CRIT_PARENT_INVALID"
-	CritAncestor          Code = "CRIT_ANCESTOR"
-	RequiredChildConflict Code = "REQUIRED_CHILD_CONFLICT"
-	CannotUpdate          Code = "CANNOT_UPDATE_OBJECT"
-	CannotPropagate       Code = "CANNOT_PROPAGATE_OBJECT"
+	CritParentMissing    Code = "CRIT_PARENT_MISSING"
```
adrianludwin: Can you file an issue to change these all to CamelCase error strings? Thanks!

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,182 @@ type HierarchicalNamespaceReconciler struct {
...
 
-	// Names of the hierarchical namespace and the current namespace.
+	// Get names of the hierarchical namespace and the current namespace.
 	nm := req.Name
 	pnm := req.Namespace
 
-	// Set RequiredChildOf (Owner) in the forest of the hierarchical namespace to the
-	// current namespace.
+	// Get instance from apiserver. If the instance doesn't exist, we don't want to reconcile
+	// it since it may trigger the HC reconciler to recreate the namespace that was just deleted.
```
adrianludwin: Will we expand on this for cascading deletion?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,182 @@ type HierarchicalNamespaceReconciler struct {
...
 	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/469
-	r.forest.Get(nm).RequiredChildOf = pnm
+	ns.RequiredChildOf = pnm
 
-	// Enqueue the in-momery hierarchyConfig instance of the hierarchical namespace.
-	// The hierarchyConfig reconciler will create the namespace and hierarchyConfig
-	// instances on apiserver accordingly.
+	// Enqueue the not-yet existent self-serve subnamespace. The HierarchyConfig reconciler will
+	// create the namespace and the HierarchyConfig instances on apiserver accordingly.
 	reason := "new/updated hierarchical namespace"
```
adrianludwin: Shouldn't this just be "new" (not "updated")?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,182 @@ type HierarchicalNamespaceReconciler struct {
...
+		// no HC. This could be a transient state before HC reconciler finishes creating the HC
+		// or a human manually created the namespace with the right annotation but no HC.
+		// Both cases meant to create this namespace as HNS.
+		log.Info("Setting the subnamespace's owner in the forest", "owner", pnm, "namespace", nm)
+		ns.RequiredChildOf = pnm
 
-	// TODO sync with forest to report conflicts in hns.Status.State. See issue:
-	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/487
+		reason := "new/updated hierarchical namespace"
+		r.hcr.enqueueAffected(log, reason, nm)
```
adrianludwin: Shouldn't this just be "updated" (not "new")?

Also feel free to combine into one line if you like: `r.hcr.enqueueAffected(log, "updated hierarchical namespace", nm)`

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -25,46 +26,124 @@ var _ = Describe("Hierarchy", func() {
 
 		fooName = createNS(ctx, "foo")
 		barName = createNSName("bar")
+	})
 
-		// Create "bar" hns in "foo" namespace
+	It("should set the hns.status.state to Ok if the subnamespace is created succussfully", func() {
```
adrianludwin: This test now does a lot of things, maybe generalize the description?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -31,44 +39,182 @@ type HierarchicalNamespaceReconciler struct {
...
 
-	return ctrl.Result{}, nil
+		// We will set it as "Conflict" though it's just a short transient state. Once the hc is
+		// reconciled, this HNS will be enqueued and then set the state to "Ok".
+		inst.Status.State = api.Conflict
+	case pnm:
+		log.Info("Setting the HierarchicalNamespace state to Ok")
+		inst.Status.State = api.Ok
+	default:
+		log.Info("Self-serve subnamespace is already owned by another parent", "child", nm, "otherParent", hcInst.Spec.Parent)
```
adrianludwin: Maybe include the intended parent as well?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: Yes, created #501 and added a todo with more details.

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
yiqigao217: Do you mean in a comment above this or in the "should..." string?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
sophieliu15: Looks from the code, if instance does not exist and we return an error instead of nil, `Reconcile()` of `HierarchicalNamespaceReconciler` (this method) will be called again, which will get the instance, and we are trapped in an infinite loop. Why it will trigger `HC reconciler to recreate the namespace that was just deleted.`?  Feel free to correct me if I miss anything. :)

yiqigao217: Ah, line 67-70 did return a `nil` to prevent a retry.

It "may" trigger the HC reconciler to recreate the namespace is the namespace is missing, see `syncMissing`.

sophieliu15: Thanks so much for the explanation offline.

yiqigao217: You are welcome and thanks for bring this up!

PR #502, opened 2020-03-06
Add `allowCascadingDelete` to HC.spec and enable `kubectl`

### LINKED ISSUE #480: HNC: Add `allowCascadingDelete` to HC spec and toggle it with `kubectl`
Part of #457 

`allowCascadingDelete` has to be set true before deleting self-serve subnamespace. This will be used in the webhook #458 . We may also need to figure out a way to set it back from true to false after the user's actions are done in case they forget to set it back to false.

yiqigao217:
Add `allowCascadingDelete` to HC.spec. Enable `kubectl hns set parent
--allowCascadingDelete=true/false` to set the spec.

Tested manually on GKE.

Fixes #480 . Part of #457

### DIFF
--- a/incubator/hnc/pkg/kubectl/set.go
+++ b/incubator/hnc/pkg/kubectl/set.go
@@ -28,10 +28,11 @@ import (
 
 //hcUpdates struct stores name of namespaces against type of flag passed
 type hcUpdates struct {
-	root             bool
-	parent           string
-	requiredChildren []string
-	optionalChildren []string
+	root                 bool
+	parent               string
+	allowCascadingDelete bool
+	requiredChildren     []string
+	optionalChildren     []string
 }
 
 var setCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


incubator/hnc/pkg/kubectl/set.go
```diff
@@ -174,6 +190,7 @@ func normalizeStringArray(in []string) []string {
 func newSetCmd() *cobra.Command {
 	setCmd.Flags().Bool("root", false, "Turns namespace into root namespace")
 	setCmd.Flags().String("parent", "", "Parent namespace")
+	setCmd.Flags().Bool("allowCascadingDelete", false, "Allows cascading deletion of its self-serve subnamespaces. The values can only be true or false.")
```
sophieliu15: Is this necessary? Will there errors from internal library saying the flag is boolean if user enters a non boolean value?

adrianludwin: +1

yiqigao217: It's true that if the value is not true or false, it will say "Invalid argument/syntax" followed by the `-help` info.

This text will show up in the `-help` or when users input invalid arguments, so it may be helpful having it here.

adrianludwin: Personally I'd take that out - almost all `allow` flags are booleans, and I don't think we need to document them all every time.

yiqigao217: sg. The "value" part is taken out.

PR #503, opened 2020-03-06
Fix a flaky object controller test

yiqigao217:
The flaky error is that when an initial propagation is not done, the
test expects to get the propagated object with success and see if it's
modified from the source but it failed when getting the object.

Added 1 second sleep to fix the flaky test.

### DIFF
--- a/incubator/hnc/pkg/reconcilers/object_test.go
+++ b/incubator/hnc/pkg/reconcilers/object_test.go
@@ -118,6 +118,8 @@ var _ = Describe("Secret", func() {
 	It("should overwrite the propagated ones if the source is updated", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, barName)
+		// Wait 1 second to make sure the source get propagated.
+		time.Sleep(1 * time.Second)
 		Eventually(isModified(ctx, fooName, "foo-role")).Should(BeFalse())
 		Eventually(hasObject(ctx, "Role", barName, "foo-role")).Should(BeTrue())
 		Eventually(isModified(ctx, barName, "foo-role")).Should(BeFalse())


incubator/hnc/pkg/reconcilers/object_test.go
```diff
@@ -118,6 +118,8 @@ var _ = Describe("Secret", func() {
 	It("should overwrite the propagated ones if the source is updated", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, barName)
+		// Wait 1 second to make sure the source get propagated.
+		time.Sleep(1 * time.Second)
 		Eventually(isModified(ctx, fooName, "foo-role")).Should(BeFalse())
```
adrianludwin: Why is `Eventually` not enough here? Why does this test (but only this test) need an additional sleep?

sophieliu15: Looks like `isModified` is nore turning `func() bool`, instead it returns `bool`. IIUC, `Eventually` will not retry in this case ([details](https://moddoc.marwan.io/github.com/onsi/gomega/@v/v1.4.2#Eventually)).

> If Eventually is passed an actual that is a function taking no arguments and returning at least one value, then Eventually will call the function periodically and try the matcher against the function's first return value.

Try wrap it with `func() bool` to see if it works.

yiqigao217: Thanks for the comments! The `isModified` is changed to return `func() bool` in #508

yiqigao217: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_multi-tenancy/502/pull-hnc-test/1236025480545570816

This is the error mentioned in the commit message.

adrianludwin: For the record (since that log file will eventually be deleted):

```
• Failure [0.079 seconds]
Secret
/home/prow/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/reconcilers/object_test.go:16
  should overwrite the propagated ones if the source is updated [It]
  /home/prow/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/reconcilers/object_test.go:118
  Expected success, but got an error:
      
      <*errors.StatusError | 0xc0002f2ab0>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {SelfLink: "", ResourceVersion: "", Continue: ""},
              Status: "Failure",
              Message: "Role.rbac.authorization.k8s.io \"foo-role\" not found",
              Reason: "NotFound",
              Details: {
                  Name: "foo-role",
                  Group: "rbac.authorization.k8s.io",
                  Kind: "Role",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 404,
          },
      }
      Role.rbac.authorization.k8s.io "foo-role" not found
  /home/prow/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/reconcilers/object_test.go:125
```

Issue #488, opened 2020-03-04
HNC: Avoid restart HNC after adding a new CRD


sophieliu15:
Right now when HNC is running and we add a new CRD and hope to config the new type, there are two problems:
- Reconciliation of `ConfigReconciler` will not be triggered automatically. We need to change `config` object to trigger reconciliation.
- Even after manually triggering the reconciliation from `ConfigReconciler`, `ObjectReconciler` will not recognize the CRD and as a result corresponding object reconciler cannot be created.

In order to create the object reconciler for the new type, we need to kill the manager pod to restart HNC. We should either avoid restarting HNC after adding a new CRD, or prevent users from adding a new CRD after starting HNC.

sophieliu15:
Discussion here: https://github.com/kubernetes-sigs/multi-tenancy/pull/476#issuecomment-594901438

sophieliu15:
The first problem can be resolved by letting `ConfigReconciler` watch for changes in crds. I will address this problem after resolving the second problem (see below).

The cause of the second problem is that the controller manager's client's `RESTMapper` is initialized at startup and never updates when adding new CRDs to the apiserver. Therefore, the client is only aware of CRDs installed before starting the controller manager. This issue was address in [kubernetes-sigs/controller-runtime#321](https://github.com/kubernetes-sigs/controller-runtime/issues/321) and was fixed in [kubernetes-sigs/controller-runtime#544](https://github.com/kubernetes-sigs/controller-runtime/pull/554). However, after updating `controller-runtime` from current version (v0.2.2) to the latest version (v0.5.0), the object reconciler could be created successfully even when the CRD is not installed. It then received an error complaining the CRD was not found when the `Reconcile` method was called. I haven't found a way to check CRD before creating the reconciler or figured out the reason why the reconciler creation could succeed after the update. See discussion about this issue in kubernetes-sigs/controller-runtime#840. Working code can be found [here](https://github.com/kubernetes-sigs/multi-tenancy/compare/master...sophieliu15:crd_issue?expand=1).

sophieliu15:
Add @adrianludwin and @yiqigao217 in the loop.

adrianludwin:
Nice digging! If controller-runtime doesn't reject it, we'll just have to use the Discovery API to figure out if a type exists ourselves before we create the object reconciler. I have no idea how to do that but I suspect it's not _too_ terribly difficult.

sophieliu15:
[...]

PR #504, opened 2020-03-06
Upgrade controller-runtime from v0.2.2 to v0.5.0 and its dependencies.

sophieliu15:
Reason for upgrade: 
The new version uses [DynamicRESTMapper](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/apiutil/dynamicrestmapper.go) as default `RESTMapper` for controller runtime manager. `DynamicRESTMapper` will "reload the delegated `meta.RESTMapper` on a cache miss" (see kubernetes-sigs/controller-runtime#554), which can solve problem that we need to restart HNC after adding a new CRD to create corresponding object reconciler when using controller-runtime v0.2.2 (see details in #488).

Notice that in v0.5.0, when the GVK does not exist, the controller-runtime manager will create an object reconciler without failure. Instead, the object reconciler and manager will terminate after starting the object reconciler. This is expected (see kubernetes-sigs/controller-runtime#840). We will address this issue in a follow-up PR by explicitly checking if a GVK exists before creating the corresponding object reconciler.  

This PR also solved following incompatibility issues caused by controller-runtime upgrade:

- Upgrade go version in `go.mod` and `Dockerfile` to 1.13. The [errors.As](https://golang.org/pkg/errors/#As) in [dynamicrestmapper.go](https://github.com/kubernetes-sigs/controller-runtime/blob/bfc982769615817ee15db8a543662652d900d27b/pkg/client/apiutil/dynamicrestmapper.go#L48) requires go 1.13.
- A higher version for k8s.io/cli-runtime and k8s.io/client-go are required after upgrading controller-runtime to v0.5.0
- Version changes of other packages in `go.mod` are updated automatically. 
- serializer.DirectCodecFactory was renamed to [serializer.WithoutConversionCodecFactory](https://godoc.org/k8s.io/apimachinery/pkg/runtime/serializer#WithoutConversionCodecFactory) after k8s.io/apimachinery 1.16 (see [here](https://github.com/kubernetes/apimachinery/commit/ed8af178d620cdb55580cddff6bfef9d3b00a7d7), and [here](https://github.com/kubernetes/apimachinery/commit/4fac835274d8f4a7de6666ac9544f793d2884b84))
- Default [LeaderElectionID](https://github.com/kubernetes-sigs/controller-runtime/blob/bfc982769615817ee15db8a543662652d900d27b/pkg/leaderelection/leader_election.go#L46) in controller-runtime manager is removed after kubernetes-sigs/controller-runtime#446.
- [NewlineReporter](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest/printer) is moved from `sigs.k8s.io/controller-runtime/pkg/envtest/` to `https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest/printer` by [this](https://github.com/kubernetes-sigs/controller-runtime/commit/748f55d736d60ce218913bf4129ebfe6cecaa051#diff-42de1d59fbe8f8b90154f01dd01f5748) commit.
- Upgrade controller-tools/cmd/controller-gen in `ci-test.sh` because we upgraded k8s.io/api to v0.17.3 which did not have `Webhook` in [v1beta1](https://godoc.org/k8s.io/api/admissionregistration/v1beta1). If we do not upgrade controller-gen, we will get error `undefined: v1beta1.Webhook` in `controller-tools@v0.2.1` in `pull-hnc-test`.
- Upgrade `kustomize` in `ci-test.sh` because otherwise it will downgrade `k8s.io/client-go` to `v11.0.0`, which will cause an error `not enough arguments in call to watch.NewStreamWatcher` from `rest/request.go:598:31` in `pull-hnc-test`

Tested:

- Unit tests.
- Went through [demo script](https://docs.google.com/document/d/1tKQgtMSf0wfT3NOGQx9ExUQ-B8UkkdVZB6m4o3Zqn64/edit#) to make sure HNC behaves as expected on a GKE cluster.

This partly solve:  #488


sophieliu15:
@yiqigao217 Feel free to take a look.

sophieliu15:
/assign @adrianludwin cancel

Got errors:

```
# sigs.k8s.io/controller-tools/pkg/webhook
/home/prow/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/webhook/parser.go:98:29: undefined: v1beta1.Webhook
/home/prow/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/webhook/parser.go:129:9: undefined: v1beta1.Webhook
/home/prow/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/webhook/parser.go:161:21: undefined: v1beta1.Webhook
/home/prow/go/pkg/mod/sigs.k8s.io/controller-tools@v0.2.1/pkg/webhook/parser.go:162:23: undefined: v1beta1.Webhook
make: *** [Makefile:123: controller-gen] Error 2
```

Looks like we might need to update kubebuilder in `hack/ci-test.sh` to make it work (from a comment [here](https://github.com/alphagov/gsp/pull/884#issuecomment-579223635)). I am still experimenting to see if it works. I will try to unassign (not sure if this works) you first.

incubator/hnc/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.12.5 as builder
+FROM golang:1.13 as builder
```
adrianludwin: The current version is 1.14 - any reason not to try upgrading to the latest?

incubator/hnc/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubernetes-sigs/multi-tenancy/incubator/hnc
 
-go 1.12
+go 1.13
```
adrianludwin: General question - are we using the latest possible version of everything in this file?

incubator/hnc/go.mod
```diff
@@ -10,22 +10,21 @@ require (
...
-	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
-	sigs.k8s.io/controller-runtime v0.2.2
+	k8s.io/api v0.17.3
+	k8s.io/apimachinery v0.17.3
+	k8s.io/cli-runtime v0.17.3
+	k8s.io/client-go v0.17.3
+	sigs.k8s.io/controller-runtime v0.5.0
+	sigs.k8s.io/controller-tools v0.2.5 // indirect
 	sigs.k8s.io/kustomize v2.0.3+incompatible // indirect
+	sigs.k8s.io/kustomize/kustomize/v3 v3.5.4 // indirect
```
adrianludwin: Do you know what's going on here? What are the two versions kustomize?

incubator/hnc/hack/ci-test.sh
```diff
@@ -32,8 +32,11 @@ wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.0.0-alp
 tar -zxvf kubebuilder_2.0.0-alpha.1_linux_amd64.tar.gz
 mv kubebuilder_2.0.0-alpha.1_linux_amd64 /usr/local/kubebuilder
 
+echo "Installing 'controller-gen'"
+GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5
+
```
adrianludwin: Why did we not need this before, but we need it now? Shouldn't the fact that it's in `go.mod` be enough?

The fewer steps and versions we have in this file, the better IMO.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -169,6 +169,17 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 func (r *ConfigReconciler) createObjectReconciler(gvk schema.GroupVersionKind, mode api.SynchronizationMode, inst *api.HNCConfiguration) {
 	r.Log.Info("Creating an object reconciler", "gvk", gvk, "mode", mode)
 
+	// After updating sigs.k8s.io/controller-runtime version to v0.5.0, we can create
```
adrianludwin: This is all great, but can we split it into a separate PR? It's nice IMO if the PR called "upgrade controller-runtime" does nothing _except_ upgrade controller-runtime (and its other similar dependencies).

sophieliu15: If we do not do the check below, the HNC behaves different compared with what it is now (before upgrade) and it will crash whenever we try to create an object reconciler with an undefined resource. Is that ok?

adrianludwin: Yes, I think so - this feature's under development and isn't expected to work until we say it's ready.

incubator/hnc/Dockerfile
sophieliu15: I tried to upgrade to 1.14 but running into [this problem ](https://github.com/golang/go/issues/37436):

`runtime: mlock of signal stack failed: 1`

I took a quick look at the thread. Looks like there are some bugs in the Ubuntu kernel and they are going to include the patch in the kernel for release 20.04 ([see comment here](https://github.com/golang/go/issues/37436#issuecomment-595836976)). I am not sure how long it will take for our glinux workstation to pickup the updates.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Sure! Reverted this file. Will add it in a later PR.

incubator/hnc/hack/ci-test.sh
sophieliu15: Looks like if I do not explicitly get it here, when we run pull-hnc-test, it will get v0.2.1 version of sigs.k8s.io/controller-tools/cmd/controller-gen. Even though we set it to v0.2.5 in go.mod. I am not sure the reason. Looks like when running pull-hnc-test does not get all packages listed in go.mod (e.g., it doesn't get controller-runtime from the log below):

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_multi-tenancy/504/pull-hnc-test/1237032233525579778/build-log.txt

incubator/hnc/go.mod
sophieliu15: I think the v2.0.3+incompatible was the old version we used before? v3.5.4 was the new version being automatically added to go.mod. I deleted the v2.0.3+incompatible version and it was not added back automatically.

incubator/hnc/go.mod
sophieliu15: The ones we updated in this PR were the latest version.

sophieliu15:
> Mainly a bunch of questions. One general comment - we're upgrading a bunch of things, not just controller-runtime, so maybe let's make the title of the commit and PR more general (e.g. "Upgrade golang, controller-runtime and other dependencies").

@adrianludwin 

The reason we upgraded everything else except controller-runtime is because we upgraded controller-runtime. How about "Upgrade controller-runtime and its dependencies"?

incubator/hnc/hack/ci-test.sh
adrianludwin: (I looked at the log and noticed that it was the first line after calling `make`, so I checked the makefile)

adrianludwin:
> The reason we upgraded everything else except controller-runtime was because we upgraded controller-runtime. How about "Upgrade controller-runtime and its dependencies"?

The new title looks good

incubator/hnc/hack/ci-test.sh
sophieliu15: Ah! That's the problem! Thanks a lot for pointing it out!

PR #508, opened 2020-03-09
Update `isModified()` to remove sleeps in tests

yiqigao217:
Update `isModified()` to return a function for `Eventually` to retry.
Change `ExpectWithOffset` to `EventuallyWithOffset` to get the object.
This prevents wierd cases of failure when the object does exist but the
first attempts of getting the object fails.

Tested with "make test" for 20+ times to make sure it's not flaky.

### DIFF
--- a/incubator/hnc/pkg/reconcilers/object_test.go
+++ b/incubator/hnc/pkg/reconcilers/object_test.go
@@ -108,8 +108,6 @@ var _ = Describe("Secret", func() {
 		Eventually(hasObject(ctx, "Role", bazName, "foo-role")).Should(BeTrue())
 
 		removeRole(ctx, fooName, "foo-role")
-		// Wait 1 second to make sure the propagated objects are removed.
-		time.Sleep(1 * time.Second)
 		Eventually(hasObject(ctx, "Role", fooName, "foo-role")).Should(BeFalse())
 		Eventually(hasObject(ctx, "Role", barName, "foo-role")).Should(BeFalse())
 		Eventually(hasObject(ctx, "Role", bazName, "foo-role")).Should(BeFalse())
@@ -118,8 +116,7 @@ var _ = Describe("Secret", func() {
 	It("should overwrite the propagated ones if the source is updated", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, barName)
-		// Wait 1 second to make sure the source get propagated.
-		time.Sleep(1 * time.Second)
+		Eventually(hasObject(ctx, "Role", fooName, "foo-role")).Should(BeTrue())
 		Eventually(isModified(ctx, fooName, "foo-role")).Should(BeFalse())
 		Eventually(hasObject(ctx, "Role", barName, "foo-role")).Should(BeTrue())
 		Eventually(isModified(ctx, barName, "foo-role")).Should(BeFalse())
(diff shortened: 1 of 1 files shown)


adrianludwin:
I'd rather not add a new parameter to `isModified` - we should always have called `hasObject` first if we're not certain the object doesn't exist. (If `hasObject` and `isModified` are _disagreeing_ on whether the object exists, we need to look into why).

However, `isModified` should _definitely_ return a function, so that part is good.

yiqigao217:
> I'd rather not add a new parameter to `isModified` - we should always have called `hasObject` first if we're not certain the object doesn't exist. (If `hasObject` and `isModified` are _disagreeing_ on whether the object exists, we need to look into why).
> 
> However, `isModified` should _definitely_ return a function, so that part is good.

This commit is updated addressing the comments.
Here's the test log that `hasObject` and `isModified` had conflicts: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_multi-tenancy/502/pull-hnc-test/1236025480545570816. This is also the error #503 mentioned in the commit message.

sophieliu15:
> > I'd rather not add a new parameter to `isModified` - we should always have called `hasObject` first if we're not certain the object doesn't exist. (If `hasObject` and `isModified` are _disagreeing_ on whether the object exists, we need to look into why).
> > However, `isModified` should _definitely_ return a function, so that part is good.
> 
> SG. The changes are made and repushed.
> 
> Here's the test log that `hasObject` and `isModified` had conflicts: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_multi-tenancy/502/pull-hnc-test/1236025480545570816.
> 
> This is also the error #503 mentioned in the commit message.
> 
> The conflict was between line 124 and 125. Line 125 failed as mentioned in the test log.
> L124 `Eventually(hasObject(ctx, "Role", bazName, "foo-role")).Should(BeTrue())`
> L125 `Eventually(isModified(ctx, bazName, "foo-role")).Should(BeFalse())`
> 
> Please correct me if I misunderstood the failed test log.

@yiqigao217 

Can you try removing ExpectWithOffset...

Just

```
return func() bool {
...
k8sClient.Get(ctx, nnm, role)
...
return ok
}
```

I suspect the Expect failed because we did not get the object in the first try and it will not trigger retry here.

adrianludwin:
> Can you try removing ExpectWithOffset...
> 
> Just
> 
> ```
> return func() bool {
> ...
> k8sClient.Get(ctx, nnm, role)
> ...
> return ok
> }
> ```
> 
> I suspect the Expect failed because we did not get the object in the first try and it will not trigger retry here.

No, I _do_ think we want the Expect, because it's assumed that we've already waited for `hasObject` before we call `isModified`. Otherwise, it's unclear what it means if we `return false` - does it mean we've verified that the object wasn't modified, or just that the object wasn't seen yet?

sophieliu15:
> > Can you try removing ExpectWithOffset...
> > Just
> > ```
> > return func() bool {
> > ...
> > k8sClient.Get(ctx, nnm, role)
> > ...
> > return ok
> > }
> > ```
> > 
> > 
> > I suspect the Expect failed because we did not get the object in the first try and it will not trigger retry here.
> 
> No, I _do_ think we want the Expect, because it's assumed that we've already waited for `hasObject` before we call `isModified`. Otherwise, it's unclear what it means if we `return false` - does it mean we've verified that the object wasn't modified, or just that the object wasn't seen yet?

I see. Cool! Thanks for the explanation!

yiqigao217:
@adrianludwin @sophieliu15 , PTAL. The `client.Get()` is now wrapped with `EventuallyWithOffset()` to reduce/eliminate the flaky false negative error.

rjbez17:
@adrianludwin sounds good. Do we need a separate issue to 'keep an eye' on the issue or are we pretty confident now that we have a retry? Regardless, shouldn't hold up the merge.

PR #509, opened 2020-03-09
Set an object reconciler to "ignore" mode if its corresponding type is removed from the HNCConfiguration Spec

sophieliu15:
If a type is removed from the HNCConfiguration Spec, we will set the corresponding object reconciler to "ignore" mode. 

Ideally, we should shut down the corresponding object reconciler. Gracefully terminating an object reconciler is still under
development (https://github.com/kubernetes-sigs/controller-runtime/issues/764). Once the feature is released, we will see if we can shut down the object reconciler instead of setting it to "ignore" mode.

Tested: GKE, unit tests

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hnc_config.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config.go
@@ -151,6 +151,20 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 	r.Forest.Lock()
 	defer r.Forest.Unlock()
 
+	if err := r.syncActiveReconcilers(ctx, inst); err != nil {
+		return err
+	}
+
+	if err := r.syncRemovedReconcilers(ctx, inst); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// syncActiveReconcilers syncs object reconcilers for types that are in the Spec. If an object reconciler exists, it sets
+// its mode according to the Spec; otherwise, it creates the object reconciler.
+func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.HNCConfiguration) error {
 	for _, t := range inst.Spec.Types {
 		gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
 		if ts := r.Forest.GetTypeSyncer(gvk); ts != nil {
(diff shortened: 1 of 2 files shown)


incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -162,6 +162,29 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 		}
 	}
 
+	// If a type exists in the forest but not exists in the Spec, we will
```
adrianludwin: Can we start breaking this function up into multiple ones? It's getting a bit long now.

Maybe something like:
```
func syncObjectReconcilers(...) {
  r.Forest.Lock()
  defer r.Forest.Unlock()
  syncActiveReconcilers(...)
  syncRemovedReconcilers(...)
}
```

rjbez17:
lgtm after @adrianludwin is happy with refactor.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -162,6 +162,29 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
...
+		}
+		if exist {
+			continue
+		}
+		// The type does not exist in the Spec. Ignore subsequent reconciliations.
+		if err := ts.SetMode(ctx, api.Ignore, r.Log); err != nil {
+			return err // retry the reconciliation
+		}
+	}
+
```
yiqigao217: Just another thought from how HC did it but your call whether it's clearer either way.

Have a `map[TypeSyncer]bool` to represent if the `ts` exists in forest. Traverse `inst.Spec.Types` to do L154-163 and remove `ts` from map. Then traverse the `ts` remained in the map to do L183-185.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion! PTAL.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestions! It will definitely help us get rid of an extra loop, but as a tradeoff, we need to store an additional map in the forest. I feel the extra loop might be ok in this case because we are not likely to have too many types to configure. But if it is not the case, or if we find it causes performance issues in future, I will refactor it as you suggested in a later PR.

adrianludwin: You could also build the map each time instead of storing it. Turns O(n^2) into O(n). But the types are probably small enough that big-O doesn't really matter for now

PR #511, opened 2020-03-10
Validate RBAC roles and bindings generated by tenant reconciler

shivi28:
1. start tenant and tenantnamespace controller
2. create tenant admin and nonadmin ser
3. validate tenantadmin is able to create tenantnamespace, whereas nonadmin user is not

### DIFF
--- a/tenant/pkg/controller/tenantnamespace/tenantnamespace_helper.go
+++ b/tenant/pkg/controller/tenantnamespace/tenantnamespace_helper.go
@@ -62,25 +62,24 @@ func getTemplateContent(kubeConfigTmpl string, context interface{}) (string, err
 	return writer.String(), nil
 }
 
+//findSecretNameOfSA: get secret name of tenant admin service account
 func findSecretNameOfSA(c client.Client, saName string) (string, error) {
-	//Get service account list, to fetch above sa because secret generation is async
-	saList := corev1.ServiceAccountList{}
-	if err := c.List(context.TODO(), &client.ListOptions{}, &saList); err != nil {
+	var saSecretName string
+	secretList := corev1.SecretList{}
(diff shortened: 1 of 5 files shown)


tenant/pkg/controller/tenantnamespace/tenantnamespace_helper.go
```diff
@@ -0,0 +1,86 @@
...
+- context:
+    cluster: {{ .cluster }}
+    user: {{ .username }}
+  name: default
+current-context: default
+preferences: {}
+`
+)
+
+func GenerateCfgStr(clusterName string, ip string, ca, token []byte, username string) (string, error) {
```
shivi28: This is different from VC code

tenant/pkg/controller/tenantnamespace/tenantnamespace_helper.go
```diff
@@ -0,0 +1,86 @@
...
+	saList := corev1.ServiceAccountList{}
+	if err := c.List(context.TODO(), &client.ListOptions{}, &saList); err != nil {
+		return "", err
+	}
+
+	//Get secret name of tenant admin service account
+	var saSecretName string
+	for _, eachSA := range saList.Items {
+		for _, each := range eachSA.Secrets {
+			if strings.Contains(each.Name, saName) {
```
Fei-Guo: You should do more strict check.  First the secret type is type: kubernetes.io/service-account-token and 
the sa name is in its annotation:
annotations:
    kubernetes.io/service-account.name: t1-user1
    kubernetes.io/service-account.uid: e6ed084c-fc32-11e9-8f10-0800278e01e3

shivi28:
Deleting this PR and opened another one https://github.com/kubernetes-sigs/multi-tenancy/pull/569

PR #514, opened 2020-03-11
Make ConfigReconciler watches for custom resources

### LINKED ISSUE #488: HNC: Avoid restart HNC after adding a new CRD

sophieliu15:
This PR makes ConfigReconciler watches for custom resources and triggers reconciliation of the `config` singleton when any custom resource is created/updated.

closes #488


incubator/hnc/cmd/manager/main.go
```diff
@@ -24,6 +24,7 @@ import (
 	prom "github.com/prometheus/client_golang/prometheus"
 	"go.opencensus.io/stats/view"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
```
adrianludwin: Maybe make a note that this should be changed to v1 when we only need to support 1.17 and higher?

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -293,10 +295,21 @@ func (r *ConfigReconciler) forceInitialReconcile(log logr.Logger, reason string)
...
 // SetupWithManager builds a controller with the reconciler.
 func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Whenever a CRD is created/updated, we will send a request to reconcile the
+	// singleton again, in case the singleton has configuration for the resource.
+	crdMapFn := handler.ToRequestsFunc(
+		func(a handler.MapObject) []reconcile.Request {
+			return []reconcile.Request{
+				{NamespacedName: types.NamespacedName{
+					Name: api.HNCConfigSingleton,
+				}},
```
adrianludwin: Nit: this looks a bit unbalanced to me. Maybe just split it into two lines?

```
nnm := types.NamespacedName{Name: api.HNCConfigSingleton}
return []reconcile.Request{{NamespacedName: nnm}}
```

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -293,10 +295,21 @@ func (r *ConfigReconciler) forceInitialReconcile(log logr.Logger, reason string)
...
+			}
+		})
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&api.HNCConfiguration{}).
 		Watches(&source.Channel{Source: r.Igniter}, &handler.EnqueueRequestForObject{}).
-		Complete(r)
+		Watches(&source.Kind{Type: &v1beta1.CustomResourceDefinition{}},
+			&handler.EnqueueRequestsFromMapFunc{ToRequests: crdMapFn}).Complete(r)
```
adrianludwin: Nit: leave `Complete(r)` on its own line?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -212,6 +213,25 @@ var _ = Describe("HNCConfiguration", func() {
...
+		addToHNCConfig(ctx, "stable.example.com/v1", "CronTab", api.Propagate)
+
+		// The corresponding object reconciler should not be created because the type does not exist.
+		Eventually(hasHNCConfigurationConditionWithMsg(ctx, api.ObjectReconcilerCreationFailed, "stable.example.com/v1, Kind=CronTab")).Should(BeTrue())
+
+		// Add the CRD for CronTab to the apiserver.
+		createCronTabCRD(ctx)
+		setParent(ctx, barName, fooName)
+		makeObject(ctx, "CronTab", fooName, "foo-crontab")
+
```
adrianludwin: Is there a reason you're also calling `setParent` and `makeObject` _before_ checking that the condition has been cleared?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -312,3 +332,18 @@ func removeHNCConfigType(ctx context.Context, apiVersion, kind string) {
 		return updateHNCConfig(ctx, c)
 	}).Should(Succeed())
 }
+
+func createCronTabCRD(ctx context.Context) {
+	crontab := v1beta1.CustomResourceDefinition{}
+	crontab.Kind = "CronTab"
+	crontab.APIVersion = "stable.example.com/v1"
```
adrianludwin: Really? Are you sure this isn't `CustomResourceDefinition` and `apiextensions.k8s.io/v1beta` (or whatever) respectively? Eg see line 262 on this file.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -312,3 +332,18 @@ func removeHNCConfigType(ctx context.Context, apiVersion, kind string) {
...
+	crontab.Name = "crontabs.stable.example.com"
+	crontab.Spec.Group = "stable.example.com"
+	crontab.Spec.Versions = append(crontab.Spec.Versions, v1beta1.CustomResourceDefinitionVersion{Name: "v1", Served: true, Storage: true})
+	crontab.Spec.Names.Kind = "CronTab"
+	crontab.Spec.Names.Plural = "crontabs"
+	crontab.Spec.Names.Singular = "crontab"
```
adrianludwin: Consider using the Go struct initializer, eg see line 262 in this file.

incubator/hnc/pkg/reconcilers/test_helpers_test.go
```diff
@@ -24,6 +24,7 @@ var GVKs = map[string]schema.GroupVersionKind{
 	"ResourceQuota": {Group: "", Version: "v1", Kind: "ResourceQuota"},
 	"LimitRange":    {Group: "", Version: "v1", Kind: "LimitRange"},
 	"ConfigMap":     {Group: "", Version: "v1", Kind: "ConfigMap"},
+	"CronTab":       {Group: "stable.example.com", Version: "v1", Kind: "CronTab"},
```
adrianludwin: This one's probably worth a comment :)

sophieliu15: Sure! Done.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Thanks for the suggestion! Done.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Interesting. It should be `CustomResourceDefinition` and `apiextensions.k8s.io/v1beta1` (I used the definition from [here](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition)). When I tried to install the CRD on the apiserver, I got some errors complaining about invalid definition, but I never got an error for that.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: I was trying to group "actions" together then put "expectations" together. We can also do the following:

```go
	
 // Add the CRD for CronTab to the apiserver.
createCronTabCRD(ctx)

// The object reconciler for CronTab should be created successfully.
Eventually(hasHNCConfigurationConditionWithMsg(ctx, api.ObjectReconcilerCreationFailed, "stable.example.com/v1, Kind=CronTab")).Should(BeFalse())

makeObject(ctx, "CronTab", fooName, "foo-crontab")

// "foo-crontab" should be propagated from foo to bar.
Eventually(hasObject(ctx, "CronTab", barName, "foo-crontab")).Should(BeTrue())
Expect(objectInheritedFrom(ctx, "CronTab", barName, "foo-crontab")).Should(Equal(fooName))
```
Actually I think this is more clear. I will change to that for now. Feel free to let me know if there are any suggestions. :)

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Sure!

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Sure!

incubator/hnc/cmd/manager/main.go
sophieliu15: Done.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
adrianludwin: Could be that the Go client just overwrote the (obviously bad) fields for you.

incubator/hnc/cmd/manager/main.go
rjbez17: Do we have a stated goal of our supported k8s version? Are we by default following k8s (3 versions back, I know there is work being done for LTS as well) or something different? May be good to have an official statement in docs somewhere.

sophieliu15: Just chatted with Adrian about supported k8s version. 

From Adrian, we are not completely sure what k8s versions we should support at this stage. 

One idea is that we should at least support whatever k8s versions that are supported by GKE (e.g., my GKE cluster is using 1.13). For the other k8s versions, we may not test if it is compatibility with HNC. 

Another idea is that we can support k8s versions that are released in the past 18 months (e.g., from 1.11 to 1.17). 

Feel free to let us know if you have suggestions!

Issue #397, opened 2020-02-11
MTB: Add test for Block access to other tenant resources


JimBugwadia:
Automated test for: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/benchmarks/e2e/tests/block_other_tenant_resources/README.md

JimBugwadia:
closing as the project is being restructured

PR #515, opened 2020-03-11
#397 added block other tenant resources test

phoenixking25:


benchmarks/e2e/config/config.go
```diff
@@ -29,3 +29,15 @@ func (c *BenchmarkConfig) GetValidTenant() (TenantSpec, error) {
...
 	return c.TenantB, nil
 }
+
+func (c *BenchmarkConfig) ValidateTenant(t TenantSpec) (error) {
+	if c == nil {
+		return errors.New("Please fill in a valid/non-empty config.yaml")
+	}
+
+	if !reflect.DeepEqual(t, TenantSpec{}) {
+		return nil
```
JimBugwadia: shouldn't this return an error?

phoenixking25: It was confusing to me so i tested it and it was working well, no problem i will run again the test and get back to you

phoenixking25: I am agreed that it should give error but its working according to required behavior only when we compliment it by `not !` in front of it. Actually if we see https://github.com/kubernetes-sigs/multi-tenancy/blob/a99b6dbd22693bf23f119226184df87301549f32/benchmarks/e2e/config/config.go#L26 this is also behaving exact like this! Lets say i remove `not !` from  front of it and fill the Tenant A and comment out the Tenant B in `config.yaml` then it shows error.

PR #522, opened 2020-03-18
Implement a validation controller for hncconfigurations resource.

sophieliu15:
This PR implements a validation controller for hncconfigurations resource. 

Specifically, following requests will be denied by the controller:
 - Missing RBAC configuration
 - Non-default RBAC mode
 - At least one type in the configuration does not exist in the cluster
 - Invalid object name (the name should be config)
 - Unrecognizable mode
 - Multiple configurations for the same type
 - Delete `config` object

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/validators/hncconfig.go
+++ b/incubator/hnc/pkg/validators/hncconfig.go
@@ -1,0 +1,182 @@
+package validators
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"k8s.io/api/admission/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+// ConfigServingPath is where the validator will run. Must be kept in sync with the
+// kubebuilder markers below.
+const (
+	ConfigServingPath = "/validate-hnc-x-k8s-io-v1alpha1-hncconfigurations"
+)
+
+// Note: the validating webhook FAILS CLOSE. This means that if the webhook goes down, all further
+// changes are denied.
+//
+// +kubebuilder:webhook:path=/validate-hnc-x-k8s-io-v1alpha1-hncconfigurations,mutating=false,failurePolicy=fail,groups="hnc.x-k8s.io",resources=hncconfigurations,verbs=create;update;delete,versions=v1alpha1,name=hncconfigurations.hnc.x-k8s.io
+
+type HNCConfig struct {
+	Log       logr.Logger
+	validator gvkValidator
+	decoder   *admission.Decoder
+}
+
+// gvkValidator checks if a resource exists. The check should typically be performed against the apiserver,
+// but need to be stubbed out during unit testing.
+type gvkValidator interface {
+	// Exists takes a GVK, and returns true if the GVK exists in the apiserver.
+	Exists(ctx context.Context, gvk schema.GroupVersionKind) (bool, error)
+}
+
+type gvkSet map[schema.GroupVersionKind]bool
+
+func (c *HNCConfig) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if isHNCServiceAccount(&req.AdmissionRequest.UserInfo) {
+		return allow("HNC SA")
+	}
+
+	if req.Operation == v1beta1.Delete {
+		if req.Name == "config" {
+			return deny(metav1.StatusReasonForbidden, "Deleting the 'config' object is forbidden")
+		} else {
+			// We allow deleting other objects. If the validation controller has always been running, we should not
+			// enter this case because objects of other names cannot be created. If users somehow bypass the
+			// validation controller and create objects of other names, we will allow them to delete the objects.
+			return allow("")
+		}
+	}
+
+	inst := &api.HNCConfiguration{}
+	if err := c.decoder.Decode(req, inst); err != nil {
+		c.Log.Error(err, "Couldn't decode request")
+		return deny(metav1.StatusReasonBadRequest, err.Error())
+	}
+
+	resp := c.handle(ctx, inst)
+	c.Log.Info("Handled", "allowed", resp.Allowed, "code", resp.Result.Code, "reason", resp.Result.Reason, "message", resp.Result.Message)
+	return resp
+}
+
+// handle implements the validation logic of this validator for Create and Update operations,
+// allowing it to be more easily unit tested (ie without constructing a full admission.Request).
+func (c *HNCConfig) handle(ctx context.Context, inst *api.HNCConfiguration) admission.Response {
+	if inst.GetName() != "config" {
+		return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Wrong singleton name: %s; the name should be 'config'", inst.GetName()))
+	}
+
+	roleExist := false
+	roleBindingExist := false
+	ts := gvkSet{}
+	for _, t := range inst.Spec.Types {
+		if rp := c.isTypeConfigured(t, ts); !rp.Allowed {
+			return rp
+		}
+
+		if c.isRBAC(t, "Role") {
+			if rp := c.validateRBAC(t.Mode, "Role"); !rp.Allowed {
+				return rp
+			}
+			roleExist = true
+		} else if c.isRBAC(t, "RoleBinding") {
+			if rp := c.validateRBAC(t.Mode, "RoleBinding"); !rp.Allowed {
+				return rp
+			}
+			roleBindingExist = true
+		} else {
+			if rp := c.validateType(ctx, t, ts); !rp.Allowed {
+				return rp
+			}
+		}
+	}
+	if !roleExist {
+		return deny(metav1.StatusReasonInvalid, "Configuration for Role is missing")
+	}
+	if !roleBindingExist {
+		return deny(metav1.StatusReasonInvalid, "Configuration for RoleBinding is missing")
+	}
+	return allow("")
+}
+
+// Validate if the configuration of a type already exists. Each type should only have one configuration.
+func (c *HNCConfig) isTypeConfigured(t api.TypeSynchronizationSpec, ts gvkSet) admission.Response {
+	gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+	if exists := ts[gvk]; exists {
+		return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Duplicate configurations for %s", gvk))
+	}
+	ts[gvk] = true
+	return allow("")
+}
+
+func (c *HNCConfig) isRBAC(t api.TypeSynchronizationSpec, kind string) bool {
+	return t.APIVersion == "rbac.authorization.k8s.io/v1" && t.Kind == kind
+}
+
+// The mode of Role and RoleBinding should be either unset or set to the propagate mode.
+func (c *HNCConfig) validateRBAC(mode api.SynchronizationMode, kind string) admission.Response {
+	if mode == api.Propagate || mode == "" {
+		return allow("")
+	}
+	return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Invalid mode of %s; current mode: %s; expected mode %s", kind, mode, api.Propagate))
+}
+
+// validateType validates a non-RBAC type.
+func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationSpec, ts gvkSet) admission.Response {
+	gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+
+	// Validate if the GVK exists in the apiserver.
+	exists, err := c.validator.Exists(ctx, gvk)
+	if !exists {
+		return deny(metav1.StatusReasonInvalid,
+			fmt.Sprintf("Cannot find the %s in the apiserver with error: %s", gvk, err.Error()))
+	}
+
+	// The mode of a type should be either unset or set to one of the supported modes.
+	if t.Mode != api.Propagate && t.Mode != api.Remove && t.Mode != api.Ignore && t.Mode != "" {
+		return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Unrecognized mode '%s' for %s", t.Mode, gvk))
+	}
+
+	return allow("")
+}
+
+// realGVKValidator implements gvkValidator, and is not used during unit tests.
+type realGVKValidator struct {
+	client client.Client
+}
+
+// Exists validates if a given GVK exists in the apiserver. The method tries to get an object of the
+// GVK from the apiserver. If the error returned from the apiserver is not IsNotFound, we will assume the
+// GVK does not exist.
+func (r *realGVKValidator) Exists(ctx context.Context, gvk schema.GroupVersionKind) (bool, error) {
+	inst := &unstructured.Unstructured{}
+	inst.SetGroupVersionKind(gvk)
+	err := r.client.Get(ctx, types.NamespacedName{Name: "nm"}, inst)
+	// We try to get an object of the given GVK with name "nm". It is possible that the
+	// object does not exist. Therefore, we will ignore the IsNotFound error.
+	if err != nil && !errors.IsNotFound(err) {
+		return false, err
+	}
+	return true, nil
+}
+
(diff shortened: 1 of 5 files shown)


incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+		c.Log.Error(err, "Couldn't decode request")
+		return deny(metav1.StatusReasonBadRequest, err.Error())
+	}
+
+	resp := c.handle(ctx, inst)
+	c.Log.Info("Handled", "allowed", resp.Allowed, "code", resp.Result.Code, "reason", resp.Result.Reason, "message", resp.Result.Message)
+	return resp
+}
+
+// handle implements the non-webhook-y businesss logic of this validator, allowing it to be more
```
yiqigao217: What does `non-webhook-y` mean? Also a typo for business here.

sophieliu15: Interesting. I think I got it from the comment of `handle` in `object.go`. I thought it was a standard comment so I did not look it closely. Updated both. Feel free to take a look!

incubator/hnc/pkg/validators/config_test.go
```diff
@@ -0,0 +1,239 @@
...
+			AdmissionRequest: v1beta1.AdmissionRequest{
+				Operation: v1beta1.Delete,
+				Name:      "config",
+			}}
+		config := &Config{}
+
+		got := config.Handle(context.Background(), req)
+
+		logResult(t, got.AdmissionResponse.Result)
+		g.Expect(got.AdmissionResponse.Allowed).Should(BeFalse())
```
yiqigao217: It seems allowed to delete config object according to the doc but here it's expected to be not allowed?

yiqigao217: It complies with the validator's behavior you have above though.

yiqigao217:
LGTM after @sophieliu15 answers the delete config object question. It's likely the change is needed on the doc to say it's not allowed.

/assign @rjbez17

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
+package validators
```
adrianludwin: Should this filename be hncconfig? Otherwise it's ambiguous with hierarchyconfig. Ditto for all the names below.

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+// ConfigServingPath is where the validator will run. Must be kept in sync with the
+// kubebuilder markers below.
+const (
+	ConfigServingPath = "/validate-hnc-x-k8s-io-v1alpha1-hncconfigurations"
+)
+
+// Note: the validating webhook FAILS OPEN. This means that if the webhook goes down, all further
+// changes to the objects are allowed.
+//
+// +kubebuilder:webhook:path=/validate-hnc-x-k8s-io-v1alpha1-hncconfigurations,mutating=false,failurePolicy=fail,groups="hnc.x-k8s.io",resources=hncconfigurations,verbs=create;update;delete,versions=v1alpha1,name=hncconfigurations.hnc.x-k8s.io
```
adrianludwin: This isn't true - `failurePolicy=fail` means it "fails closed," which means that if the webhook goes down, all further changes are denied. I think that is the correct thing to do, so we just need to change the comment.

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+
+type Config struct {
+	Log     logr.Logger
+	client  validationClient
+	decoder *admission.Decoder
+}
+
+// validationClient checks if a resource exists. The check should typically be performed against the apiserver,
+// but need to be stubbed out during unit testing.
+type validationClient interface {
```
adrianludwin: Can we give this a more unique name, like gvkValidator?

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+	// ResourceExists takes a GVK, and returns true if the corresponding resource of the GVK exists in the apiserver.
+	ResourceExists(ctx context.Context, gvk schema.GroupVersionKind) (bool, error)
```
adrianludwin: "Resource" usually means an object; you just mean the type (GVK), right? You could also just change this method name to `Exists`.

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type set map[schema.GroupVersionKind]bool
```
adrianludwin: `set` is a bit _too_ terse, maybe `gvkSet`?

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+	if !roleBindingExist {
+		return deny(metav1.StatusReasonForbidden, "Configuration for RoleBinding is missing")
+	}
+	return allow("")
+}
+
+// Validate if the configuration of a type already exists. Each type should only have one configuration.
+func (c *Config) isTypeConfigured(t api.TypeSynchronizationSpec, ts set) admission.Response {
+	gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+	if _, ok := ts[gvk]; ok {
```
adrianludwin: `if ts[gvk]` should work; missing map values default to the zero-value, which is `false` for bools.

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+type set map[schema.GroupVersionKind]bool
+
+func (c *Config) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if isHNCServiceAccount(&req.AdmissionRequest.UserInfo) {
+		return allow("HNC SA")
+	}
+
+	if req.Operation == v1beta1.Delete {
+		if req.Name == "config" {
+			return deny(metav1.StatusReasonForbidden, "Deleting the 'config' object is forbidden")
```
adrianludwin: "Forbidden" usually means "the current user is not authorized to take that action," which isn't the case here. See https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#StatusReason for the possible reasons. Maybe `StatusReasonBadRequest`?

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+	resp := c.handle(ctx, inst)
+	c.Log.Info("Handled", "allowed", resp.Allowed, "code", resp.Result.Code, "reason", resp.Result.Reason, "message", resp.Result.Message)
+	return resp
+}
+
+// handle implements the validation logic of this validator for Create and Update operations,
+// allowing it to be more easily unit tested (ie without constructing a full admission.Request).
+func (c *Config) handle(ctx context.Context, inst *api.HNCConfiguration) admission.Response {
+	if inst.GetName() != "config" {
+		return deny(metav1.StatusReasonForbidden, fmt.Sprintf("Wrong singleton name: %s; the name should be 'config'", inst.GetName()))
```
adrianludwin: StatusReasonBadRequest? Or maybe StatusReasonInvalid?

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+			}
+			roleBindingExist = true
+		} else {
+			if rp := c.validateType(ctx, t, ts); !rp.Allowed {
+				return rp
+			}
+		}
+	}
+	if !roleExist {
+		return deny(metav1.StatusReasonForbidden, "Configuration for Role is missing")
```
adrianludwin: StatusReasonInvalid?

incubator/hnc/pkg/validators/config.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	if !roleBindingExist {
+		return deny(metav1.StatusReasonForbidden, "Configuration for RoleBinding is missing")
```
adrianludwin: StatusReasonInvalid? I'll let you work out the rest of these.

incubator/hnc/pkg/validators/config_test.go
sophieliu15: Thanks for pointing it out! I updated the doc. It should deny deleting the `config` object.

incubator/hnc/pkg/validators/config.go
sophieliu15: Good idea! Changed to `hncconfig` along with the struct name.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the explanation! Updated.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the suggestion! Changed to `gvkValidator`. It is definitely more relevant.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the explanation! Changed to `Exists`.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the suggestion! Done.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the explanation! Done.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the link. From the documentation, IIUC `not authorized` is covered by `StatusReasonUnauthorized `  not `StatusReasonForbidden` :

> StatusReasonUnauthorized means the server can be reached and understood the request, but requires the user to present appropriate authorization credentials  in order for the action to be completed. 

`StatusReasonBadRequest` might also not appropriate. That code means the request does not make any sense. However, in our case, we understand the request. Here is the documentation for `StatusReasonBadRequest`:

>   StatusReasonBadRequest means that the request itself was invalid, because the request doesn't make any sense, for example deleting a read-only object. API calls that return BadRequest can never succeed.

`StatusReasonInvalid` is also not appropriate here. From the documentation, it is for create and update operations and it means there is invalid data in the request, which does not apply here because we are doing a delete operation and the request itself is valid. 

> StatusReasonInvalid means the requested create or update operation cannot be completed due to invalid data provided as part of the request. The client may
need to alter the request. 

IMO `StatusReasonForbidden` is appropriate here. It means the server understand the request but deny the request:

> StatusReasonForbidden means the server can be reached and understood the request, but refuses to take any further action.  It is the result of the server being configured to deny access for some reason to the requested resource by the client.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the suggestion! Looks like `StatusReasonInvalid` is more appropriate. Changed to that.

incubator/hnc/pkg/validators/config.go
sophieliu15: Thanks for the suggestion. Changed to `StatusReasonInvalid`. Same below and the rest of the code.

Looks like for the `StatusReasonInvalid`, the webhook will only show `The request is invalid` text message when applying an invalid request. The detailed message can be found in the log. I suspect this is intended but it is less convenient than `StatusReasonForbidden`, which will show the error message after apply a request, e.g.,

> Error from server (Forbidden): admission webhook "hncconfigurations.hnc.x-k8s.io" denied the request: Deleting the 'config' object is forbidden

incubator/hnc/pkg/validators/config.go
sophieliu15: ditto.

incubator/hnc/pkg/validators/config.go
adrianludwin: I thought there was some way to send more custom text back to the user? The log isn't really good enough, IMO.

incubator/hnc/pkg/validators/config.go
adrianludwin: sgtm

incubator/hnc/pkg/validators/hncconfig.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Invalid mode of %s; current mode: %s; expected mode %s", kind, mode, api.Propagate))
+}
+
+// validateType validates a non-RBAC type.
+func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationSpec, ts gvkSet) admission.Response {
+	gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+
+	// Validate if the GVK exists in the apiserver.
+	exists, err := c.validator.Exists(ctx, gvk)
```
adrianludwin: Nit: is `err` always non-nil if `exists` is false? If not, why not:

```
if err := c.validator.Exists(ctx, gvk); err != nil {
  return deny(...)
}
```

incubator/hnc/pkg/validators/hncconfig.go
```diff
@@ -0,0 +1,182 @@
...
+	if !exists {
+		return deny(metav1.StatusReasonInvalid,
+			fmt.Sprintf("Cannot find the %s in the apiserver with error: %s", gvk, err.Error()))
+	}
+
+	// The mode of a type should be either unset or set to one of the supported modes.
+	if t.Mode != api.Propagate && t.Mode != api.Remove && t.Mode != api.Ignore && t.Mode != "" {
+		return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Unrecognized mode '%s' for %s", t.Mode, gvk))
+	}
```
adrianludwin: Nit: maybe better to write this as a switch statement?

incubator/hnc/pkg/validators/hncconfig.go
sophieliu15: It is possible that err returned by Exists is `errors.IsNotFound`. In this case exists is true. It means the GVK exists in the cluster but the object is not found. It is likely possible that the object does not exist, since we are querying a random object of the GVK. If GVK does not exist in the cluster, the error is not nil but also not `errors.IsNotFound`

See line: 168.

incubator/hnc/pkg/validators/config.go
sophieliu15: From the documentation of [StatusReasonInvalid](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#StatusReason) we can add "cases" to the `Details` field of the [Status](https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#Status). I also tried that but it still did not show the custom text. I am wondering if there is something specific about `StatusReasonInvalid` that the webhook ignore all the custom text. However, I haven't found any related logic under [controller-runtime/pkg/webhook](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/webhook).

We can ask the cotroller-runtime team about that to see if what might be cause of that. In the meantime, I am thinking about switching the reason back to `StatusReasonForbidden` since I feel it is the second appropriate one and it shows the custom text. WDYT?

incubator/hnc/pkg/validators/hncconfig.go
sophieliu15: Thanks for the suggestion! Changing to the switch should be more readable. I am thinking about updating this in the same PR that changes `StatusReasonInvalid` (see my comment above), if we are going to change it. Otherwise, I can sent out a small PR to change this to the switch statement.

adrianludwin: Whichever you prefer is fine by me :)

incubator/hnc/pkg/validators/hncconfig.go
adrianludwin: That seems like a bit of an odd implementation detail to leak out of this function. The caller of `c.validator.Exists` isn't asking if an arbitrarily named object exists, so returning an error that says "not found" doesn't really seem relevant? I'd rather have something inside Exists like:

```
if err == errors.IsNotFound {
  return nil // the type exists
}
return err // if err is nil, that means the object was found, which means the type exists
```

Alternatively, it could _just_ return the boolean, since the text of the error message itself is unlikely to be interesting. In which case the logic will be:

```
if err == nil || err == errors.IsNotFound {
  return true // the type exists
}
return false
```

incubator/hnc/pkg/validators/hncconfig.go
sophieliu15: Thanks for the suggestion! I changed it to `switch` in #525

incubator/hnc/pkg/validators/hncconfig.go
sophieliu15: Thanks for the suggestion! Only returning an error makes sense. I will still output the error message just in case users are interested and in case there are some transient errors occur when querying the apiserver. See #525

incubator/hnc/pkg/validators/config.go
sophieliu15: As we discussed offline, I changed it back to `StatusReasonForbidden` in #525. I will ask the controller runtime team to see if this behavior is expected. Thanks for the suggestion!

PR #524, opened 2020-03-19
Handle missing or non-default specs for Role and RoleBinding in ConfigReconciler.

sophieliu15:
This PR handles missing or non-default specs for Role and RoleBinding in ConfigReconciler. 

Specifically, it implements the following two features:

- If the specs for Role and/or RoleBinding are missing, the ConfigReconciler will insert default configuration in the spec for each of them
- If the specs for Role and/or RoleBinding are non-default, the ConfigReconciler will modify the spec to the default configuration for each of them

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hnc_config.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config.go
@@ -91,13 +91,68 @@ func (r *ConfigReconciler) getSingleton(ctx context.Context) (*api.HNCConfigurat
 		if !errors.IsNotFound(err) {
 			return nil, err
 		}
-
-		// It doesn't exist - initialize it to a default value.
-		inst.Spec = config.GetDefaultConfigSpec()
-		inst.ObjectMeta.Name = api.HNCConfigSingleton
 	}
 
+	r.validateSingleton(inst)
 	return inst, nil
+}
+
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -91,13 +91,68 @@ func (r *ConfigReconciler) getSingleton(ctx context.Context) (*api.HNCConfigurat
...
+	roleExists, roleBindingsExists := false, false
+
+	// Check the mode of Role and RoleBinding. The mode can be either the default mode
+	// or not set; otherwise, we will change the mode to the default mode.
+	for i := 0; i < len(inst.Spec.Types); i++ {
+		t := &inst.Spec.Types[i]
+		if r.isRole(*t) {
+			mode := config.GetDefaultRoleSpec().Mode
+			if t.Mode != mode && t.Mode != "" {
+				r.Log.Info(fmt.Sprintf("Invalid mode for Role. Change the mode from %s to %s", t.Mode, mode))
```
adrianludwin: "Changing the mode" is better, because it shows the action that HNC has taken

sophieliu15: Thanks for the suggestion! Updated "change" along with two other verbs ("add" and "set") in this PR.

PR #529, opened 2020-03-20
Handle multiple configurations of the same type in ConfigReconciler

sophieliu15:
This PR handles multiple configurations of the same type in ConfigReconciler. If there are multiple configurations for the same type, only the first configuration will be followed, the rest will be ignored.

Tested:unit tests, GKE cluster

Issue: #411

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hnc_config.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config.go
@@ -44,13 +44,14 @@ type ConfigReconciler struct {
 	HierarchyConfigUpdates chan event.GenericEvent
 }
 
+type gvkSet map[schema.GroupVersionKind]bool
+
 // Reconcile sets up some basic variable and logs the Spec.
 // TODO: Updates the comment above when adding more logic to the Reconcile method.
 func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
 
 	// Validate the singleton name.
-	// TODO: Add a validating admission controller to prevent the problem in the first place.
 	if err := r.validateSingletonName(ctx, req.NamespacedName.Name); err != nil {
 		r.Log.Error(err, "Singleton name validation failed")
 		return ctrl.Result{}, nil
@@ -222,7 +223,14 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 // syncActiveReconcilers syncs object reconcilers for types that are in the Spec. If an object reconciler exists, it sets
 // its mode according to the Spec; otherwise, it creates the object reconciler.
 func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.HNCConfiguration) error {
+	// exist keeps track of existing types in the `config` singleton.
+	exist := gvkSet{}
 	for _, t := range inst.Spec.Types {
+		// If there are multiple configurations of the same type, we will follow the first
+		// configuration and ignore the rest.
+		if !r.ensureNoDuplicateTypeConfigurations(inst, t, exist) {
+			continue
+		}
 		gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
 		if ts := r.Forest.GetTypeSyncer(gvk); ts != nil {
 			if err := ts.SetMode(ctx, t.Mode, r.Log); err != nil {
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -235,6 +241,25 @@ func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.
...
+	gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+	if mp[gvk] {
+		specMsg := fmt.Sprintf("APIVersion: %s, Kind: %s, Mode: %s", t.APIVersion, t.Kind, t.Mode)
+		r.Log.Info(fmt.Sprintf("Ignoring the configuration: %s", specMsg))
+		condition := api.HNCConfigurationCondition{
+			Code: api.MultipleConfigurationsForOneType,
+			Msg: fmt.Sprintf("Ignore the configuration: %s because the configuration of the type already exists; "+
+				"only the first configuration will be applied", specMsg),
+		}
+		inst.Status.Conditions = append(inst.Status.Conditions, condition)
```
yiqigao217: Just confirming, if there are 3 configurations for the same type, the `Status.Conditions` will have 2 same condition for the 2nd and 3rd configurations, right?

sophieliu15: Yes. The message will show what 2nd and 3rd configurations are.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -222,7 +221,14 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 // syncActiveReconcilers syncs object reconcilers for types that are in the Spec. If an object reconciler exists, it sets
 // its mode according to the Spec; otherwise, it creates the object reconciler.
 func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.HNCConfiguration) error {
+	// exist keeps track of existing types in the `config` singleton.
+	exist := map[schema.GroupVersionKind]bool{}
 	for _, t := range inst.Spec.Types {
+		// If there are multiple configurations of the same type, we will follow the first
+		// configuration and ignore the rest.
+		if r.isTypeConfigured(inst, t, exist) {
```
adrianludwin: Nit: this doesn't _just_ return an answer, it both checks for something invalid happening, and _sets_ a condition. Maybe `r.ensureNoDuplicateTypeConfigurations`?

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -222,7 +221,14 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 // syncActiveReconcilers syncs object reconcilers for types that are in the Spec. If an object reconciler exists, it sets
 // its mode according to the Spec; otherwise, it creates the object reconciler.
 func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.HNCConfiguration) error {
+	// exist keeps track of existing types in the `config` singleton.
+	exist := map[schema.GroupVersionKind]bool{}
```
adrianludwin: Maybe make a type out of this since you mention it multiple times?
`type gvkSet map[schema.GroupVersionKind]bool`?

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -235,6 +241,25 @@ func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.
 	return nil
 }
 
+// isTypeCofigured keeps track of whether the configuration of a type already exists.
+func (r *ConfigReconciler) isTypeConfigured(inst *api.HNCConfiguration, t api.TypeSynchronizationSpec,
+	mp map[schema.GroupVersionKind]bool) bool {
+	gvk := schema.FromAPIVersionAndKind(t.APIVersion, t.Kind)
+	if mp[gvk] {
```
adrianludwin: Probably more clear to invert this:
```
if !mp[gvk] {
  mp[gvk] = true
  return false
}
... everything else ...
```

go/gocomments#indent-error-flow (Google-internal link) is not _exactly_ what we mean here since this isn't an "error" condition, but the fundamental guideline of "reduce levels of indentation" does apply here.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion! Changed to `r.ensureNoDuplicateTypeConfigurations`.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Sure! Thanks for the suggestion!

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the pointer. Updated the code to follow the guidance.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -222,7 +223,14 @@ func (r *ConfigReconciler) syncObjectReconcilers(ctx context.Context, inst *api.
 // syncActiveReconcilers syncs object reconcilers for types that are in the Spec. If an object reconciler exists, it sets
 // its mode according to the Spec; otherwise, it creates the object reconciler.
 func (r *ConfigReconciler) syncActiveReconcilers(ctx context.Context, inst *api.HNCConfiguration) error {
+	// exist keeps track of existing types in the `config` singleton.
+	exist := gvkSet{}
 	for _, t := range inst.Spec.Types {
+		// If there are multiple configurations of the same type, we will follow the first
+		// configuration and ignore the rest.
+		if r.ensureNoDuplicateTypeConfigurations(inst, t, exist) {
```
adrianludwin: I think this would look cleaner if it was `!r.ensure()` instead of `r.ensure()`. That is, return false on "failure" so we skip. This makes it look like "we've successfully shown there's no conflict so we're going to do nothing" :)

sophieliu15: Thanks for the suggestion! Done.

PR #537, opened 2020-03-22
Implement cascading delete for self-serve namespaces

yiqigao217:
This allows cascading deletion on an HNS or an owner namespace based on
if an owner or the namespace has spec.allowCascadingDelete field set to
true. If not, the owned namespace will have HNS_MISSING condition.

Additional fields are added to the forest to indicate if the namespace
is being deleted and if it allows cascading delete. Finalizers are added
to all HC and HNS instances, which will only be removed if the cascading
deletion is finished from that point.

The creation/deletion of namespaces are handled by HC Reconciler.

Manually tested on GKE cluster with three levels of owned namespaces and
unowned namespaces. Owned namespaces are created using 'k hns create2'
and the allowCascadingDelete is toggled by 'k hns set <ns>
--allowCascadingDelete='. The owned namespaces were cascading deleted
when allowCascadingDelete is set. Otherwise, they got HNS_MISSING
condition.

Part of #457

### DIFF
--- a/incubator/hnc/pkg/forest/forest.go
+++ b/incubator/hnc/pkg/forest/forest.go
@@ -140,11 +140,13 @@ const Local = ""
 // Namespace represents a namespace in a forest. Other than its structure, it contains some
 // properties useful to the reconcilers.
 type Namespace struct {
-	forest   *Forest
-	name     string
-	parent   *Namespace
-	children namedNamespaces
-	exists   bool
+	forest               *Forest
+	name                 string
+	parent               *Namespace
+	children             namedNamespaces
+	exists               bool
+	deleting             bool
+	allowCascadingDelete bool
 
 	// originalObjects store the objects created by users, identified by GVK and name.
 	// It serves as the source of truth for object controllers to propagate objects.
@@ -176,12 +178,15 @@ func (ns *Namespace) SetExists() bool {
 	return changed
 }
 
-// UnsetExists marks this namespace as missing, returning true if it previously existed. It also
-// removes it from its parent, if any, since a nonexistent namespace can't have a parent.
+// UnsetExists marks this namespace as missing, returning true if it previously existed. It
+// removes it from its parent, if any, since a nonexistent namespace can't have a parent. It
+// also clears the deleting status and allowCascadingDelete flag in the namespace.
 func (ns *Namespace) UnsetExists() bool {
 	changed := ns.exists
 	ns.SetParent(nil) // Unreconciled namespaces can't specify parents
 	ns.exists = false
+	ns.deleting = false
+	ns.allowCascadingDelete = false
 	ns.clean() // clean up if this is a useless data structure
 	return changed
 }
@@ -195,6 +200,33 @@ func (ns *Namespace) clean() {
 
 	// Remove from the forest.
 	delete(ns.forest.namespaces, ns.name)
+}
+
+// Deleting returns true if the namespace is being deleted.
+func (ns *Namespace) Deleting() bool {
+	return ns.deleting
+}
+
+// SetDeleting marks this namespace as being deleted.
+func (ns *Namespace) SetDeleting() {
+	ns.deleting = true
+}
+
+// UpdateAllowCascadingDelete updates if this namespace allows cascading deletion.
+func (ns *Namespace) UpdateAllowCascadingDelete(acd bool) {
+	ns.allowCascadingDelete = acd
+}
+
+// AllowCascadingDelete returns if the namespace's or any of the owner ancestors'
+// allowCascadingDelete field is set to true.
+func (ns *Namespace) AllowCascadingDelete() bool {
+	if ns.allowCascadingDelete == true {
+		return true
+	}
+	if ns.Owner == "" {
+		return false
+	}
+	return ns.forest.Get(ns.Owner).AllowCascadingDelete()
 }
 
 // SetParent attempts to set the namespace's parent. This includes removing it from the list of
@@ -263,14 +295,25 @@ func (ns *Namespace) ChildNames() []string {
 	return nms
 }
 
-// OwnedNames returns a list of names of the owned namespaces or nil if there's none.
+// OwnedNames returns a list of names of the namespaces whose owner is set to
+// this namespace.
 func (ns *Namespace) OwnedNames() []string {
-	if len(ns.forest.namespaces) == 0 {
-		return nil
-	}
 	nms := []string{}
 	for nm, ns := range ns.forest.namespaces {
 		if ns.Owner == ns.name {
+			nms = append(nms, nm)
+		}
+	}
+	return nms
+}
+
+// OwnedHNS returns a list of names of the children whose owner is set to this
+// namespace and they don't have conditions related to this owner (HNS_MISSING
+// specifically).
+func (ns *Namespace) OwnedHNS() []string {
+	nms := []string{}
+	for nm, cns := range ns.children {
+		if cns.Owner == ns.name && len(cns.GetCondition(ns.name)) == 0 {
 			nms = append(nms, nm)
 		}
 	}
--- a/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
@@ -64,11 +64,24 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/501
 	inst, err := r.getInstance(ctx, pnm, nm)
 	if err != nil {
-		if errors.IsNotFound(err) {
-			// If the instance doesn't exist, return nil to prevent a retry.
+		return ctrl.Result{}, err
+	}
+
+	// If the instance is being deleted, remove the finalizer or cascading delete.
+	if !inst.DeletionTimestamp.IsZero() {
+		if len(inst.ObjectMeta.Finalizers) == 0 {
 			return ctrl.Result{}, nil
 		}
-		return ctrl.Result{}, err
+		updateFinalizer := r.syncDeleting(log, inst)
+		if updateFinalizer {
+			r.writeInstance(ctx, log, inst)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// If the instance doesn't exist, return nil to prevent a retry.
+	if inst.CreationTimestamp.IsZero() {
+		return ctrl.Result{}, nil
 	}
 
 	// Report "Forbidden" state and early exist if the namespace is not allowed to self-serve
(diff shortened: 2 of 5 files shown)


incubator/hnc/pkg/forest/forest.go
```diff
@@ -197,6 +202,33 @@ func (ns *Namespace) clean() {
...
+}
+
+// UpdateAllowCascadingDelete updates if this namespace allows cascading deletion.
+func (ns *Namespace) UpdateAllowCascadingDelete(acd bool) {
+	ns.allowCascadingDelete = acd
+}
+
+// AllowCascadingDelete returns if the namespace's or any of the owner ancestors'
+// allowCascadingDelete field is set to true.
+func (ns *Namespace) AllowCascadingDelete() bool {
```
adrianludwin: Nit: this looks like a command. Maybe consider `AllowsCascadingDelete` or `CascadingDeletionAllowed`

incubator/hnc/pkg/forest/forest.go
```diff
@@ -197,6 +202,33 @@ func (ns *Namespace) clean() {
...
+	if ns.allowCascadingDelete == true {
+		return true
+	}
+	if ns.Owner == "" {
+		return false
+	}
+	return ns.forest.Get(ns.Owner).AllowCascadingDelete()
```
adrianludwin: What happens if the parent is missing here? Is it worth commenting?

incubator/hnc/pkg/forest/forest.go
```diff
@@ -277,6 +307,19 @@ func (ns *Namespace) OwnedNames() []string {
 	return nms
 }
 
+// OwnedHNS returns a list of names of the children whose owner is set to this
+// namespace and they don't have conditions related to this owner (HNS_MISSING
+// specifically).
+func (ns *Namespace) OwnedHNS() []string {
```
adrianludwin: This seems a little weird IMO. Why do we have both `OwnedHNS` and `OwnedNames`?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -64,11 +64,24 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 	inst, err := r.getInstance(ctx, pnm, nm)
 	if err != nil {
-		if errors.IsNotFound(err) {
-			// If the instance doesn't exist, return nil to prevent a retry.
+		return ctrl.Result{}, err
+	}
+
+	// If the instance is being deleted, remove the finalizer or cascading delete.
+	if !inst.DeletionTimestamp.IsZero() {
+		if len(inst.ObjectMeta.Finalizers) == 0 {
```
adrianludwin: Comment: We've finished process this, nothing to do.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -64,11 +64,24 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 		}
-		return ctrl.Result{}, err
+		updateFinalizer := r.syncDeleting(log, inst)
+		if updateFinalizer {
+			r.writeInstance(ctx, log, inst)
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// If the instance doesn't exist, return nil to prevent a retry.
```
adrianludwin: Let's say we never call syncDeleting because there were no finalizers. Do we need to do anything else here?

I think the comment should say something like "Someone enqueued this nonexisting instance by mistake, or else it's just been deleted. Either way, there's nothing we need to do." We shouldn't say things like "return nil to prevent a retry" because that just says _what_ the code is doing, and not _why_ the code is doing it. We should return nil because "there's no error, or we've already handled the error," not because "we don't want a retry."

adrianludwin: Update: we just discussed this in our 1:1. It's fine to leave it for this PR, but in the future we should handle the "the HNS has _been_ deleted here" case.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -91,9 +104,58 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+
+	state := inst.Status.State
+	cnm := inst.Name
+	pnm := inst.Namespace
+	cns := r.forest.Get(cnm)
+	updateFinalizers = false
+
+	switch {
+	// If the owned namespace doesn't exist, remove the finalizer. The missing state
+	// is excluded here in case it's being created by the HC reconciler concurrently.
```
adrianludwin: This will get much simpler once HNSR creates/deletes the namespace :) Feel free to take some shortcuts for now and comment that fact.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -91,9 +104,58 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+	case state == api.Conflict || state == api.Forbidden || !cns.Exists():
+		r.removeFinalizers(log, inst, "the owned namespace doesn't exist")
+		updateFinalizers = true
```
adrianludwin: Would "return true" make more sense here? We don't do anything after this switch statement.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -91,9 +104,58 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+	case !cns.AllowCascadingDelete():
+		r.removeFinalizers(log, inst, "the owned namespace doesn't allow cascading delete")
+		updateFinalizers = true
+		cns.SetCondition(pnm, api.HNSMissing, "missing HNS in the owner namespace")
+		r.hcr.enqueueAffected(log, "updating HNS_MISSING condition", cnm)
+	// If the owned namespace exists and allow cascading deletion, delete it.
+	default:
+		log.Info("setting ns.deleting to true in the forest")
+		cns.SetDeleting()
+		r.hcr.enqueueAffected(log, "cascading deleting this owned namespace from the owner", cnm)
```
adrianludwin: TODO: directly delete the namespace here (right?)

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -91,9 +104,58 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+func (r *HierarchicalNamespaceReconciler) removeFinalizers(log logr.Logger, inst *api.HierarchicalNamespace, reason string) {
+	log.Info("Removing the finalizers", "reason", reason)
+	inst.ObjectMeta.Finalizers = []string{}
+
+	// Only enqueue the owner to update its HC finalizers if the owner is also being deleted.
+	pnm := inst.Namespace
+	pns := r.forest.Get(pnm)
+	if pns.Deleting() {
+		r.hcr.enqueueAffected(log, "updating the finalizers since an HNS instance is deleted", pnm)
+	}
```
adrianludwin: I'm not sure this belongs in a function called `removeFinalizers` - can you explain? I would have thought we should only enqueue the owner _after_ this object has been deleted, not before...?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -145,6 +156,70 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 	return r.updateObjects(ctx, log, nm)
 }
 
+// syncDeleting returns true if the namespace or the HC instance is being deleted.
+func (r *HierarchyConfigReconciler) syncDeleting(ctx context.Context, log logr.Logger, inst *api.HierarchyConfiguration, nsInst *corev1.Namespace) (bool, error) {
+	r.Forest.Lock()
```
adrianludwin: Normally, this would seem quite dangerous (eg no `defer unlock`) but I think once we simplify this code we'll be able to put it back. Can you put a TODO here to add a defer unlock?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -145,6 +156,70 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 	return r.updateObjects(ctx, log, nm)
 }
 
+// syncDeleting returns true if the namespace or the HC instance is being deleted.
+func (r *HierarchyConfigReconciler) syncDeleting(ctx context.Context, log logr.Logger, inst *api.HierarchyConfiguration, nsInst *corev1.Namespace) (bool, error) {
```
adrianludwin: General comment - this function's really complicated, it has a lot of different cases and it's hard for me to understand how they all fit together. I think it's fine for now since when we move the namespace creation/deletion to the HNSR, this will get a lot simpler, but it's probably worth a comment.

In the future, I think you can simplify this by checking all the `DeletionTimestamp.IsZero()` conditions _outside_ the "sync" function, and then call different "sync" functions for each one. Eg:

```
switch {
case nsInst.CreationTimestamp.IsZero():
  return r.syncMissingNamespace()
case nsInst.DeletionTimestamp.IsZero():
  return r.syncDeletingNamespace()
case inst.CreationTimestamp.IsZero():
  return r.syncMissingHC()
case inst.DeletionTimestamp.IsZero():
  return r.syncDeletingHC()
default:
  return r.syncHC()
}
```

You can keep this in mind when making the next few PRs.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -145,6 +156,70 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
...
+	case ns.Deleting() && nsInst.DeletionTimestamp.IsZero():
+		r.Forest.Unlock()
+		r.deleteNamespace(ctx, log, nsInst)
+	// The namespace is being deleted. Wait for any existing HNS instances to be gone.
+	default:
+		r.Forest.Unlock()
+	}
+
+	// Get the HNSes from the forest so that we don't need to keep looking until
+	// the HNSes are purged by apiserver.
```
adrianludwin: TODO: this will be fixed soon?

yiqigao217:
Closing this because we moved the subnamespace creation/deletion to HNSR from HCR. See #552 

This one is the implementation of cascadingDelete in the old version where subnamespace creation is in HCR.

PR #542, opened 2020-03-24
Compute NumPropagatedObjects in HNCConfiguration status

sophieliu15:
This PR computes NumPropagatedObjects as a part of HNCConfiguration status. NumPropagatedObjects indicates the number of propagated objects of a specific type created by HNC. 

Tested:unit tests, GKE cluster

Issue: #411


incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -96,25 +110,18 @@ var _ = Describe("HNCConfiguration", func() {
 			return updateHNCConfig(ctx, config)
 		}).Should(Succeed())
 
-		Eventually(hasHNCConfigurationConditionWithName(ctx, api.CritSingletonNameInvalid, nm)).Should(BeTrue())
-	})
-
-	It("should set ObjectReconcilerCreationFailed condition if an object reconciler creation fails", func() {
```
sophieliu15: Remove this test case because it is tested in the following test case.

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -290,6 +323,7 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 		return write, srcInst
 	}
 
+	r.addToNumPropagated(log, inst.GetNamespace(), inst.GetName())
```
adrianludwin: Why is this here? Shouldn't we only call this in line 473, below?

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -539,6 +580,34 @@ func (r *ObjectReconciler) exclude(log logr.Logger, inst *unstructured.Unstructu
 	}
 }
 
+// addToNumPropagated adds a propagated object to the PropagatedObjects map if it does not exist.
+func (r *ObjectReconciler) addToNumPropagated(log logr.Logger, namespace, name string) {
+	// We need to lock the map because there can be concurrent reconciliation threads reading/writing
+	// the map.
+	r.lock.Lock()
```
adrianludwin: Use `defer`

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -539,6 +580,34 @@ func (r *ObjectReconciler) exclude(log logr.Logger, inst *unstructured.Unstructu
...
+		r.PropagatedObjects[nnm] = true
+	}
+	r.lock.Unlock()
+}
+
+// removeFromNumPropagated removes a propagated object to the PropagatedObjects map.
+func (r *ObjectReconciler) removeFromNumPropagated(log logr.Logger, namespace, name string) {
+	// We need to lock the map because there can be concurrent reconciliation threads reading/writing
+	// the map.
+	r.lock.Lock()
```
adrianludwin: Use `defer`

incubator/hnc/pkg/reconcilers/object_test.go
```diff
@@ -44,13 +44,13 @@ var _ = Describe("Secret", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, barName)
 
-		Eventually(hasObject(ctx, "Role", barName, "foo-role")).Should(BeTrue())
+		Eventually(hasObject(ctx, "Role", barName, "foo-role"), 4*time.Second).Should(BeTrue())
```
adrianludwin: Save these magic frequencies somewhere as named constants

adrianludwin: Wait, why are we waiting this long? The object should show up right away, it only takes 3s for it to show up in the HNC Config status.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -12,14 +13,24 @@ import (
...
 )
 
 var _ = Describe("HNCConfiguration", func() {
-	// sleepTime is the time to sleep for objects propagation to take effect.
+	// objectPropagationTime is the time to sleep for objects propagation to take effect.
 	// From experiment it takes ~0.015s for HNC to propagate an object. Setting
 	// the sleep time to 1s should be long enough.
 	// We may need to increase the sleep time in future if HNC takes longer to propagate objects.
-	const sleepTime = 1 * time.Second
+	const objectPropagationTime = 1 * time.Second
```
adrianludwin: We should specify that we only use this when we're expecting something _not_ to happen. Otherwise, we just use Eventually.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -12,14 +13,24 @@ import (
...
+	// hncConfigReconciliationTime is the time to wait for reconciliation of the config
+	// object happens. It is the same as the sleep time at the beginning of the Reconcile method
+	// of HNCConfiguration.
+	const hncConfigReconciliationTime = 3 * time.Second
+	// hncConfigStatusUpdateTime is the time to wait for updating HNCConfiguration status on
+	// the apiserver. From experiment, this is the minimal integer seconds needed for the
+	// status to be updated successfully on the apiserver.
+	// We may need to increase the time in future if it takes longer to update status.
+	const hncConfigStatusUpdateTime = 7 * time.Second
```
adrianludwin: This seems too high to me. From the time we make our _last_ change to an object, I would expect it would take no more than three seconds plus a little bit extra (e.g. four seconds should be enough): 3s for the HNC Config sleep to end, plus a bit of time for it to do its work. So if we said:

`SetDefaultEventuallyTimeout(hncConfigReconciliationTime + 2 * time.Second)`

Then `Eventually` will wait up to five seconds and that _should_ be more than enough time. (Or don't change the global default and just pass this into each `Eventually`) call.

Do you agree with this logic? And if so, do you know why it doesn't work?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -37,6 +51,7 @@ var _ = Describe("HNCConfiguration", func() {
 		Eventually(func() error {
 			return resetHNCConfigToDefault(ctx)
 		}).Should(Succeed())
+		SetDefaultEventuallyTimeout(2 * time.Second)
```
adrianludwin: Is there a way to restore this to what it used to be (e.g. in case it wasn't 2s for some reason)?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -28,6 +39,9 @@ var _ = Describe("HNCConfiguration", func() {
 	)
 
 	BeforeEach(func() {
+		// Increase the timeout of `Eventually` for test cases in this file because the `Reconcile`
+		// method of ConfigReconciler sleeps at the very beginning of the method.
+		SetDefaultEventuallyTimeout(4 * time.Second)
```
adrianludwin: You're already overriding Eventually with the constants you've defined above, why do you need this too?

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: Done

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: Done

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: If there are propagated objects on the apiserver before we install HNC, we will return `ignore` action here and those objects will not be counted by line 473 ([write](https://github.com/kubernetes-sigs/multi-tenancy/blob/ca59e4a66a81d819183188e7d4c49f754302ff03/incubator/hnc/pkg/reconcilers/object.go#L414) method will not be called). 

For example, if we create namespace a and b, set a to be parent of b, add an object in a. The `NumPropagatedObjects` will be 1 as expected. Then if we reinstall (or restart) HNC, `NumPropagatedObjects` will be 0 if we don't have this line.

Added a comment to explain that.

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -111,15 +120,22 @@ func (r *ObjectReconciler) SetMode(ctx context.Context, mode api.Synchronization
...
-		err := r.enqueueAllObjects(ctx, r.Log)
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+	// For propagate and remove mode, we enqueue all objects to update objects in the cluster (e.g., propagate or
+	// remove existing objects). For ignore mode, we enqueue objects so that reconciliations of objects can trigger
+	// the config reconciler to update the status of the `config` singleton. Triggering object reconciliations in the
+	// ignore mode is necessary because only an object reconciler can compute NumPropagatedObjects for a given type.
```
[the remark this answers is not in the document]
sophieliu15: There might be some complications here. config reconciler can set the mode to anything (even an invalid mode). Object reconciler will validate the mode and change it to a valid mode if it is invalid (see [getValidateMode](https://github.com/kubernetes-sigs/multi-tenancy/blob/ca59e4a66a81d819183188e7d4c49f754302ff03/incubator/hnc/pkg/reconcilers/object.go#L129)). 

Ideally,  the config reconciler can get a valid mode from an object reconciler by adding& calling `GetMode` via the `TypeSyncer` interface. However, if a config reconciler creates (instead of syncs) an object reconciler, the `Mode` field of the object reconciler will be set to the valid mode when the (first) reconcile is [called](https://github.com/kubernetes-sigs/multi-tenancy/blob/ca59e4a66a81d819183188e7d4c49f754302ff03/incubator/hnc/pkg/reconcilers/object.go#L161). If config reconciler calls `GetMode` before any reconcile happens, it will still get an invalid mode.

This might not be an issue if we only want to know if the mode is `propagate` in config reconciler (check for `propagate` or unset mode), but it will be an issue if we want to know if the mode is `ignore`. To let config reconciler knows the valid mode, we can either move `getValidateMode` back to config reconciler so config reconciler will set a valid mode to an object reconciler. Or we can see if we can set the `Mode` field to a valid mode when creating the object reconciler, so that we don't need to wait for the first reconciler to happen.

WDYT?

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -180,9 +193,29 @@ func (r *ObjectReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			return resp, err
 		}
 	}
+
+	// We need to compute NumPropagatedObjects and show it in the status of the `config`
+	// singleton even in ignore mode.
```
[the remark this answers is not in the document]
sophieliu15: I might misunderstand that. We agreed on only reporting `NumSourceObjects` for `propagate` mode.  I thought that meant we would report `NumPropagatedObjects` for all modes. I will delete the logic if we decide to follow the same rule and only report  `NumPropagatedObjects` for `propagate` mode as well.

WDYT?

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -111,15 +120,22 @@ func (r *ObjectReconciler) SetMode(ctx context.Context, mode api.Synchronization
...
+	// For propagate and remove mode, we enqueue all objects to update objects in the cluster (e.g., propagate or
+	// remove existing objects). For ignore mode, we enqueue objects so that reconciliations of objects can trigger
+	// the config reconciler to update the status of the `config` singleton. Triggering object reconciliations in the
+	// ignore mode is necessary because only an object reconciler can compute NumPropagatedObjects for a given type.
+	err := r.enqueueAllObjects(ctx, r.Log)
+	return err
+}
+
+// GetNumPropagatedObjects returns the number of propagated objects of the GVK handled by this object reconciler.
+func (r *ObjectReconciler) GetNumPropagatedObjects() int32 {
```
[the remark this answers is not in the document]
sophieliu15: Yes, we need `int32` when we write to the status. Changed to `int` here and converted to `int32` in the config reconciler.

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -111,15 +120,22 @@ func (r *ObjectReconciler) SetMode(ctx context.Context, mode api.Synchronization
...
+	// We need to lock the map because there can be concurrent reconciliation threads reading/writing
+	// the map.
+	r.lock.Lock()
+	n := len(r.PropagatedObjects)
+	r.lock.Unlock()
```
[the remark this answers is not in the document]
sophieliu15: Done

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -51,10 +53,14 @@ const (
 	ignore
 )
 
+type namespacedNameSet map[types.NamespacedName]bool
+
 // ObjectReconciler reconciles generic propagated objects. You must create one for each
 // group/version/kind that needs to be propagated and set its `GVK` field appropriately.
 type ObjectReconciler struct {
 	client.Client
+	lock sync.Mutex
```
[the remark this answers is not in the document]
sophieliu15: Changed to `propagatedObjectsLock`. Thanks for the suggestion!

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -51,10 +53,14 @@ const (
 	ignore
 )
 
+type namespacedNameSet map[types.NamespacedName]bool
```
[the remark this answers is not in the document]
sophieliu15: Done.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -362,21 +374,46 @@ func (r *ConfigReconciler) validateSingletonName(ctx context.Context, nm string)
 	return fmt.Errorf("Error while validating singleton name: %s", msg)
 }
 
-// forceInitialReconcile forces reconciliation to start after setting up the
-// controller with the manager. This is used to create a default singleton if
-// there is no singleton in the cluster. This occurs in a goroutine so the
+// addTypeStatus adds Status.Types for types configured in the spec.
+func (r *ConfigReconciler) addTypeStatus(inst *api.HNCConfiguration) {
+	for _, ts := range r.Forest.GetTypeSyncers() {
```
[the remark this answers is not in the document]
sophieliu15: Cool! Sorted in alphabetical order based on APIVersion.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -33,22 +34,31 @@ type ConfigReconciler struct {
 	// Forest is the in-memory data structure that is shared with all other reconcilers.
 	Forest *forest.Forest
 
-	// Igniter is a channel of event.GenericEvent (see "Watching Channels" in
+	// Trigger is a channel of event.GenericEvent (see "Watching Channels" in
```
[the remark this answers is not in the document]
sophieliu15: Thank you!

incubator/hnc/pkg/forest/forest.go
```diff
@@ -86,6 +98,14 @@ func (f *Forest) GetTypeSyncer(gvk schema.GroupVersionKind) TypeSyncer {
...
 }
 
+func (f *Forest) AddConfigSyncer(c ConfigSyncer) {
+	f.config = c
+}
+
+func (f *Forest) GetConfigSyncer() ConfigSyncer {
+	return f.config
+}
+
```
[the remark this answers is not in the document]
sophieliu15: Sure!

incubator/hnc/pkg/forest/forest.go
```diff
@@ -32,6 +32,14 @@ type TypeSyncer interface {
...
 	// The method also syncs objects in the cluster for the type handled by the reconciler if necessary.
 	SetMode(context.Context, api.SynchronizationMode, logr.Logger) error
+	// GetNumPropagatedObjects returns the number of propagated objects on the apiserver.
+	GetNumPropagatedObjects() int32
+}
+
+// ConfigSyncer syncs the status of the HNCConfiguration object. ConfigReconciler implements the interface so that
+// it can be called by an ObjectReconciler if the status needs to be updated.
+type ConfigSyncer interface {
+	SyncHNCConfigStatus(logr.Logger)
```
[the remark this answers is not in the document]
sophieliu15: Thanks for the suggestion, changed to `NumPropagatedObjectsSyncer`.

sophieliu15:
> Eh, now I see the downside to putting the Sleep at the beginning of the Reconciler - it means that the regular business logic (e.g. adding/removing types) is also delayed by 3s. I'm not sure we want that. Sorry I missed that.
> 
> How about we write some logic to intercept the calls from the object reconcilers and make sure we only _trigger_ one reconciliation every N seconds? That way, most of the reconciler will still work ASAP. You'll have to write the deduping logic yourself but that should be pretty simple. Talk to Yiqi about how to trigger a periodic event using a long-running goroutine. It'll look something like this in the Setup function:
> 
> ```
> go func() {
>   for { // have some kind of way to break the loop when the controller exits
>     time.Sleep(3 * time.Second) // use constant instead
>     if r.objCountChanged {
>       r.objCountChanged = false
>       r.enqueueReconcile()
>     }
>   }
> }()
> ```

Thanks for the suggestion! I will resolve the comments about tests later after I change to this method. Hopefully it won't take that long after.

incubator/hnc/pkg/reconcilers/object.go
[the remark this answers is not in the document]
sophieliu15: Actually, I think it makes sense to not record `NumPropagatedObjects` in `ignore` mode since we also use that to "terminate" an object reconciler.

I guess the only question is that if we want to record `NumPropagatedObjects` in `remove` mode. The number should go to 0 eventually. I am not sure if users are interested in the process.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Ditto. Resolve this comment since it is outdated.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Ditto. Resolve this comment since it is outdated.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Ditto. Resolve this comment since it is outdated.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Done. Thanks for the suggestion!

incubator/hnc/pkg/reconcilers/object_test.go
sophieliu15: Turns out we don't need to increase wait time at all after switching to the goroutine approach. I will resolve outdated test related comments for now.

incubator/hnc/pkg/reconcilers/object.go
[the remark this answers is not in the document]
sophieliu15: Discussed offline. I changed the method to a function.

incubator/hnc/pkg/reconcilers/object.go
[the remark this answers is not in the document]
sophieliu15: As discussed offline. I supported `NumPropagatedObjects` in `propagate` and `remove` modes.

sophieliu15:
Hi @adrianludwin 

Thanks so much for the review. I updated the code and replied to all your comments. Feel free to take a look!

Thank you!

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -69,10 +69,11 @@ type TypeSynchronizationStatus struct {
 	// Kind to be configured.
 	Kind string `json:"kind,omitempty"`
 
-	// Tracks the number of original objects that are being propagated to descendant namespaces.
+	// Tracks the number of objects that are being propagated to descendant namespaces. The propagated
+	// objects are created by HNC.
 	// +kubebuilder:validation:Minimum=0
 	// +optional
-	NumPropagated *int32 `json:"numPropagated,omitempty"`
+	NumPropagatedObjects *int32 `json:"numPropagatedObjects,omitempty"`
```
adrianludwin: Why int32 and not just int?

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -362,21 +383,88 @@ func (r *ConfigReconciler) validateSingletonName(ctx context.Context, nm string)
...
-		r.Igniter <- event.GenericEvent{Meta: inst}
+		r.Trigger <- event.GenericEvent{Meta: inst}
 	}()
 }
 
+func (r *ConfigReconciler) getShouldReconcile() bool {
+	r.shouldReconcileLock.Lock()
+	defer r.shouldReconcileLock.Unlock()
+
+	return r.shouldReconcile
```
adrianludwin: Booleans don't need to be guarded by a lock if they're just being set/unset in isolation of anything else, which is true here. So you can get rid of the lock, and these functions, and just access the bool directly.

incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -362,21 +383,88 @@ func (r *ConfigReconciler) validateSingletonName(ctx context.Context, nm string)
...
+}
+
+// SyncNumPropagatedObjects will be called by object reconcilers to signal config
+// reconciler to reconcile when an object is reconciled successfully and the status of
+// the `config` object might need to be updated.
+func (r *ConfigReconciler) SyncNumPropagatedObjects(log logr.Logger) {
+	r.shouldReconcileLock.Lock()
+	defer r.shouldReconcileLock.Unlock()
+
+	log.Info("Signalling config reconciler for reconciliation.")
```
adrianludwin: This will be very noisy - I'd suggest log.V(1).Info

incubator/hnc/pkg/forest/forest.go
```diff
@@ -53,6 +63,10 @@ type Forest struct {
 	// We can also move the lock out of the forest and pass it to all reconcilers that need the lock.
 	// In that way, we don't need to put the list in the forest.
 	types []TypeSyncer
+
+	// config is the ConfigReconciler that an object reconciler can call if the status of the HNCConfiguration
+	// object needs to be updated.
+	Config NumPropagatedObjectsSyncer
```
adrianludwin: I'd suggest changing Config to something like ObjectStatsSyncer

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -284,6 +283,30 @@ var _ = Describe("HNCConfiguration", func() {
 		Eventually(hasObject(ctx, "CronTab", barName, "foo-crontab")).Should(BeTrue())
 		Expect(objectInheritedFrom(ctx, "CronTab", barName, "foo-crontab")).Should(Equal(fooName))
 	})
+
+	It("should set NumPropagatedObjects back to 0 after deleting the source object in propagate mode", func() {
+		addToHNCConfig(ctx, "v1", "LimitRange", api.Propagate)
+		setParent(ctx, barName, fooName)
+		makeObject(ctx, "LimitRange", fooName, "foo-lr")
+
+		Eventually(getNumPropagatedObjects(ctx, "v1", "LimitRange"), 4*time.Second).Should(Equal(int32(1)))
```
adrianludwin: Replace `4*time.Second` with a named constant with an explanation of how you picked that number

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -284,6 +283,30 @@ var _ = Describe("HNCConfiguration", func() {
 		Eventually(hasObject(ctx, "CronTab", barName, "foo-crontab")).Should(BeTrue())
 		Expect(objectInheritedFrom(ctx, "CronTab", barName, "foo-crontab")).Should(Equal(fooName))
 	})
+
+	It("should set NumPropagatedObjects back to 0 after deleting the source object in propagate mode", func() {
```
adrianludwin: This also checks that the propagated count _increases_ to non-zero, right?

incubator/hnc/pkg/reconcilers/object.go
adrianludwin: sgtm, thanks!

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -539,6 +585,32 @@ func (r *ObjectReconciler) exclude(log logr.Logger, inst *unstructured.Unstructu
...
+	r.propagatedObjectsLock.Lock()
+	defer r.propagatedObjectsLock.Unlock()
+
+	nnm := types.NamespacedName{
+		Namespace: namespace,
+		Name:      name,
+	}
+	if !r.propagatedObjects[nnm] {
+		r.propagatedObjects[nnm] = true
+	}
```
adrianludwin: You don't need the `if`, you can just always write.

incubator/hnc/api/v1alpha1/hnc_config.go
adrianludwin: Changing this to "int" would let you get rid of a lot of casts in the rest of the code.

incubator/hnc/pkg/reconcilers/object.go
sophieliu15: Done. Thanks for the suggestion!

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Yes. That's why it says `back to 0`. Feel free to let me know if I need to make the description more clear. I am heading off now but I can update it in a separate PR.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
sophieliu15: Good catch! Done!

incubator/hnc/pkg/forest/forest.go
sophieliu15: Sure! Thanks for the suggestion!

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Sure! Done.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Thanks for the suggestion! Done!

incubator/hnc/api/v1alpha1/hnc_config.go
sophieliu15: I thought we cannot use `int` from the documentation [here](https://book.kubebuilder.io/cronjob-tutorial/api-design.html), but looks like we can. Changed to `int`.

PR #545, opened 2020-03-24
Corrects path of config.yaml

### LINKED ISSUE #543: [MTB] Error: stat ./config.yaml: no such file or directory
The path of the config file is wrong and it is causing all the tests to fail. The breaking changes were introduced by #499

SomtochiAma:
This pull request fixes #543

### DIFF
--- a/benchmarks/e2e/config/config.go
+++ b/benchmarks/e2e/config/config.go
@@ -4,7 +4,7 @@ import (
 	"reflect"
 )
 
-const ConfigPath = "./config.yaml"
+const ConfigPath = "../../config.yaml"
 
 type BenchmarkConfig struct {
 	Adminkubeconfig string     `yaml:"adminKubeconfig"`


phoenixking25:
@SomtochiAma this fix has already  merged.

SomtochiAma:
@Fei-Guo @JimBugwadia, Following the [readme](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/benchmarks/documentation/run.md) for running the benchmark test.
The two commands for running the test requires different config path
Running locally
`go test ./e2e/tests`
uses this path -> `../../config.yaml`

while this command
`go test -c ./e2e/tests && ./tests.test`
uses this path -> `./config.yaml`

We might need to restructure the code so the user can pass in the path for the config or in such a  way that both commands use the same config path

I am ready to work on either solution. I would like to  hear your thoughts. Thank you!

rjbez17:
@JimBugwadia  can you comment on @SomtochiAma question above?

JimBugwadia:
While we are in heavy development, seems to me that the local test option i.e. `go test ./e2e/tests` is simpler and easier to manage. Lets remove the second option, if you agree.

Longer term, we can consider the best packaging and distribution of the tests for flexibility.

Issue #479, opened 2020-03-02
HNC: Clean up before setting HNS reconciler as default


yiqigao217:
Part of #457 

Todo:
1) remove `enable-hierarchicalnamespace-reconciler` flag
2) remove `hc.Spec.RequiredChildren`

yiqigao217:
Fixed by #550

PR #550, opened 2020-03-24
Update HCR & HNSR to move NS creation to HNSR

yiqigao217:
This commit removes the hc.Spec.RequiredChildren field and the
enable-hns-reconciler flag. It also moves the creation of the owned
(previously requiredChild) namespaces from HC Reconciler (HCR) to HNS
Reconciler (HNSR). The logic in HCR is cleaned up to adapt this change.

Tested with the current and updated integration tests by "make test".
Also manually tested the 'k hns describe'. The HNS_MISSING condition was
set when I deleted the HNS instance in the owner. The owned namespace
got recreated after I deleted it, which means the new watch in HNSR is
working.

Fully described in #552 
Part of #457, #479


incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -98,7 +104,7 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 // the namespace instances to update its HNS state. This will be the only place the "Owner"
 // field in the forest is set. Therefore, the "Owner" field can be used in the forest to get
 // all the HNS objects of a namespace.
-func (r *HierarchicalNamespaceReconciler) syncWithForest(log logr.Logger, inst *api.HierarchicalNamespace, nsInst *corev1.Namespace, hcInst *api.HierarchyConfiguration) {
+func (r *HierarchicalNamespaceReconciler) syncWithForest(log logr.Logger, inst *api.HierarchicalNamespace, nsInst *corev1.Namespace, hcInst *api.HierarchyConfiguration) (createNS bool) {
```
adrianludwin: Maybe replace `inst` with `hnsInst` because there are three instances floating around here with nearly equal importance.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -115,6 +121,7 @@ func (r *HierarchicalNamespaceReconciler) syncWithForest(log logr.Logger, inst *
 		// to HierarchyConfig reconciler.
 		log.Info("The self-serve subnamespace does not exist", "namespace", nm)
 		r.syncMissing(log, inst, ns)
```
adrianludwin: If we simply delete the `syncMissing` function, and just replace it with `inst.Status.State = api.Missing` - here's what I _think_ will happen:

* HNSR creates the namespace with the annotation (because we return true). This only succeeds if the namespace doesn't already exist, so there's no possibility of conflict.
* HCR is invoked on the newly created (child) namespace
  * HCR reads the annotation and updates hc.Parent
  * HCR syncs the forest (we'd move the update to `ns.Owner` here)
  * The normal HCR flow will enqueue HCR on the parent (not further discussed here)
  * HCR also re-enqueues HNSR on this object (we're already doing this)
* HNSR runs again, observes that the namespace exists, and clears the "missing" state

So everything in `syncMissing` _seems_ to be unnecessary (given the new changes). Does that seem reasonable?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -210,6 +217,19 @@ func (r *HierarchicalNamespaceReconciler) writeInstance(ctx context.Context, log
 	return nil
 }
 
+func (r *HierarchicalNamespaceReconciler) writeNamespace(ctx context.Context, log logr.Logger, nm, pnm string) error {
+	inst := &corev1.Namespace{}
+	inst.ObjectMeta.Name = nm
+	metadata.SetAnnotation(inst, api.AnnotationOwner, pnm)
+
+	log.Info("Creating namespace on apiserver")
+	if err := r.Create(ctx, inst); err != nil {
```
adrianludwin: Might be worth leaving a few comments here:
* If the namespace already exists, it means someone else has created a namespace while this reconciler was running, and there's likely a conflict. Returning an error will ensure that this reconciler runs again and is able to report the conflict.
* Any other transient problem should also be retried.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
adrianludwin: Also - do we even need to see the forest in this reconciler, or enqueue the HCR at all? Would everything work if we just do this:

* Try to load hnsInst and nsInst, but not hcInst
* hnsInst exists:
  * If nsInst doesn't exist, try to create it and set state=Missing
  * If nsInst exists and its annotation is correct, state=OK. Don't check hcInst.Parent, that's up to HCR to ensure that it works.
  * Otherwise, state=Conflict.
* hnsInst is being deleted (i.e. exists but DeletionTimestamp != zero)
  * If nsInst exists and has the right annotation, delete it.
  * Otherwise, remove finalizers

I'll cover setting HNS_MISSING in another comment below, but I think we can move that to the HCR pretty easily, since HCR already needs to load the full list of HNSes anyway so it knows when it can clear its finalizers.

This is much simpler than what we have now. Is there a reason it won't work?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -89,20 +84,22 @@ func (r *HierarchyConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-	// Output a log for testing.
-	log.Info("HC will be reconciled with", "HNSReconcilerEnabled", r.HNSReconcilerEnabled)
-
 	return ctrl.Result{}, r.reconcile(ctx, log, ns)
 }
 
 func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logger, nm string) error {
 	// Get the singleton and namespace.
 	inst, nsInst, err := r.GetInstances(ctx, log, nm)
+	hnsnms, err := r.getHierarchicalNamespaceNames(ctx, nm)
```
adrianludwin: Should we add a watch on the HNSes as well?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -184,25 +182,17 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, ns *fore
...
 
 	// Remove it from the forest and notify its relatives
 	r.enqueueAffected(log, "relative of deleted namespace", ns.RelativesNames()...)
 	// Enqueue the HNS if the owned namespace is deleted.
-	if r.HNSReconcilerEnabled {
-		if ns.Owner != "" {
-			r.hnsr.enqueue(log, ns.Name(), ns.Owner, "hns for the deleted owned namespace")
-		}
+	if ns.Owner != "" {
+		r.hnsr.enqueue(log, ns.Name(), ns.Owner, "hns for the deleted owned namespace")
```
adrianludwin: An alternative option: have the HNSR put a watch on namespaces as well? Then K8s would invoke HNSR directly and we wouldn't have to do it here?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -221,41 +211,47 @@ func (r *HierarchyConfigReconciler) markExisting(log logr.Logger, ns *forest.Nam
...
-// (the spec itself will be synced next), or removes the owner value if there's a problem
-// and notifies the would-be parent namespace.
-func (r *HierarchyConfigReconciler) syncOwner(log logr.Logger, inst *api.HierarchyConfiguration, ns *forest.Namespace) {
-	if ns.Owner == "" {
+// syncOwner propagates the owner value from the annotation to the forest. If
+// the owner changed, the HNS instances will be enqueued to update state. If
+// there's a conflict with the owner and the parent, the owner namespace will be
+// enqueued to update the "SubnamespaceConflict" condition.
+func (r *HierarchyConfigReconciler) syncOwner(log logr.Logger, inst *api.HierarchyConfiguration, nsInst *corev1.Namespace, ns *forest.Namespace) {
+	oldOwner := ns.Owner
```
adrianludwin: Now that we're syncing these together, do we still need ns.Owner as a separate field from ns.Parent()? Maybe just a boolean `ns.IsOwned` would be good enough, and it just means "we set ns.Parent() based on the annotation in the namespace?"

incubator/hnc/pkg/reconcilers/hierarchy_config.go
adrianludwin: Also: if we add these to the forest (eg `ns.SetHNSes(hnsnms)`), then it's easy to set and clear the `HNS_MISSING` condition on the child here in the HCR. All we need to do is re-invoke the HCR on any children when the HNS is added or removed, eg something like:

```
for changedHNS := range ns.SetHNSes(hnsNms) {
  r.enqueueAffected(changedHNS)
}
```
Make sense?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -294,142 +290,117 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
...
-	// updated when a parent namespace syncChildren().
-	// TODO report the "SubnamespaceConflict" in the HNS reconciliation. See issue:
-	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/490
-	if !r.HNSReconcilerEnabled {
-		// We'll reset any of these conditions if they occur.
-		ns.ClearAllConditions(api.SubnamespaceConflict)
-	}
+	// We'll first clear SubnamespaceConflict condition and set it if any owned
+	// subnamespaces are missing or claimed by a different parent.
+	ns.ClearAllConditions(api.SubnamespaceConflict)
```
adrianludwin: Is `SubnamespaceConflict` still possible? The annotation is now the source-of-truth and K8s will ensure that there's no conflict there.

The conflict is now reported in the HNS object itself, not as a condition.

I think this lets us delete most of the rest of this function too, right?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: Comment 2 sgtm to simply HNSR. I think the current HCR in this PR is already implemented in the way suggested in comment 1.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: sg, will do!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: yes, good idea!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: Yes, that sg. We can have a mapping func to trigger HNSR on related HNSes.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: I think ns.Owner and ns.Parent are the forest version of nsInst's owner annotation and hcInst.Spec.Parent. What if the webhook is bypassed and the parent is different from the owner annotation? CMIIW

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: The following is implemented based on if the annotation or the parent of the owned namespace can be changed. I am not sure if it's the case (e.g. webhook is bypassed, etc.). We can discuss on it!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: As discussed offline, we will use owner annotation as source of truth. The parent will always match the owner annotation.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: As discussed offline, since we use owner annotation as source of truth and always set the owner as parent, this should not happen. Say a child changed its spec.parent to something else, the HCR will force the parent to be owner (if applicable).

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: Removed `hcInst` in HNSR, so kept using `inst` here in the HNSR.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -98,7 +104,7 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 // the namespace instances to update its HNS state. This will be the only place the "Owner"
 // field in the forest is set. Therefore, the "Owner" field can be used in the forest to get
```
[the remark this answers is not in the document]
yiqigao217: Discussed offline and agreed.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: I kept `ns.Owner` for now.

adrianludwin: sgtm but can you explain why?

yiqigao217: Ah I will remove it in a next PR since it's used a lot in `hierarchy_config.go` and `hierarchy.go` and I don't want to make this PR more complex (mostly for me to test). I will update #552 and put a todo comment in the forest.

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -140,10 +139,10 @@ func (v *Hierarchy) checkForest(hc *api.HierarchyConfiguration) ([]authzReq, adm
...
 	}
 
-	resp = v.checkRequiredChildren(ns, hc.Spec.RequiredChildren)
-	if !resp.Allowed {
-		return nil, resp
-	}
+	//resp = v.checkRequiredChildren(ns, hc.Spec.RequiredChildren)
+	//if !resp.Allowed {
+	//	return nil, resp
+	//}
```
adrianludwin: Just delete this completely?

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -20,67 +19,55 @@ var _ = Describe("HierarchicalNamespace", func() {
...
-		reconcilers.EX[barName] = true
-
-		// Create "bar" hns in "foo" namespace after the HC reconciler is "disabled" (only for "bar" namespace).
-		foo_hns_bar := newHierarchicalNamespace(barName, fooName)
-		updateHierarchicalNamespace(ctx, foo_hns_bar)
-
-		// We should then have the HNS state stays at "Missing"
-		Eventually(getHNSState(ctx, fooName, barName)).Should(Equal(api.Missing))
-	})
+	//It("should set the hns.status.state to Missing if the self-serve subnamespace doesn't exist", func() {
```
adrianludwin: Can just delete this completely

incubator/hnc/pkg/reconcilers/hierarchy_config.go
adrianludwin: sg!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -184,31 +174,60 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, ns *fore
...
+		log.Info("Owner annotation was updated", "old", ns.Owner, "new", onm)
+		ns.Owner = nsInst.Annotations[api.AnnotationOwner]
+	}
+
+	if inst.Spec.Parent != onm {
+		if ns.Parent() == nil && inst.Spec.Parent == "" {
+			// This is the first time an owned namespace is reconciled in HCR. We need
+			// to enqueue the HNS so it can update the state from 'Missing' to 'Ok'.
+			r.hnsr.enqueue(log, nm, onm, "the missing owned namespace is found")
+		}
```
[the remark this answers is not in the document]
yiqigao217: Ah right, I should put the `r.hnsr.enqueue()` to the `markExisting()` when `ns.Owner != ""`. We still need this `hnsr.enqueue` since the `nsInst` doesn't change in this hcr, thus it won't trigger the hnsr by watch naturally, right?

incubator/hnc/api/v1alpha1/hierarchical_namespace.go
```diff
@@ -21,8 +21,10 @@ import (
 
 // Constants for the hierarchicalnamespaces resource type and namespace annotation.
 const (
-	HierarchicalNamespaces = "hierarchicalnamespaces"
-	AnnotationOwner        = MetaGroup + "/owner"
+	HierarchicalNamespaces           = "hierarchicalnamespaces"
+	HierarchicalNamespacesKind       = "HierarchicalNamespace"
+	HierarchicalNamespacesAPIVersion = MetaGroup + "/v1alpha1"
+	AnnotationOwner                  = MetaGroup + "/owner"
```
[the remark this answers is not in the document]
yiqigao217: sg, thanks for sharing!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -184,31 +174,60 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, ns *fore
...
+		log.Info("The parent doesn't match the owner. Setting the owner as the parent.", "parent", inst.Spec.Parent, "owner", onm)
+		inst.Spec.Parent = onm
+	}
+
+	// Look up the HNSes in the owner namespace. Set HNS_MISSING condition if it's
+	// not there.
+	for _, hnsnm := range ons.HNSes {
+		if hnsnm == nm {
+			return
```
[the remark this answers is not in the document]
yiqigao217: Thanks for the tips!

incubator/hnc/pkg/validators/hierarchy.go
yiqigao217: sg, will do.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -294,6 +276,18 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
...
 
+func (r *HierarchyConfigReconciler) syncChildren(log logr.Logger, inst *api.HierarchyConfiguration, ns *forest.Namespace) {
+	inst.Status.Children = ns.ChildNames()
+}
+
+// syncHNSes updates the HNS list. If any HNS is created/deleted, it will enqueue
+// the child to update its HNS_MISSING condition.
+func (r *HierarchyConfigReconciler) syncHNSes(log logr.Logger, ns *forest.Namespace, hnsnms []string) {
+	for _, changedHNS := range ns.SetHNSes(hnsnms) {
+		r.enqueueAffected(log, "the HNS instance is created/deleted", changedHNS)
```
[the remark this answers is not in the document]
yiqigao217: sg. I will take option #2 that add the child ns to the map of HNS in HCR watch.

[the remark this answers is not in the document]
yiqigao217: Somehow I got issue in manual testing. The child is not enqueued when an HNS is deleted.

This is how I tested it:
`k create ns parent`
`k hns create2 child -n parent`
`k get ns` - I can see they are both created
`k get hns -n parent -o yaml` - the HNS state is ok
`k delete hns child -n parent` - the HNS is deleted
`k hns describe child` -  no HNS_MISSING condition
`k get hierarchyconfigurations.hnc.x-k8s.io -n child -o yaml` - no HNS_MISSING condition
I doubted if the child HC is not reconciled, so I did:
`k hns set child --allowCascadingDelete=true` to write HC instance, then I can see the HNS_MISSING condition is updated.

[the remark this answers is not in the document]
yiqigao217: I will use option#1 to add a comment to the original enqueue children of changed HNSes.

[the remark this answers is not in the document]
adrianludwin: sgtm, we can try revisiting this later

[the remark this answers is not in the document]
yiqigao217: sg!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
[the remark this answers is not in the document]
adrianludwin: Right, sorry, I misread `markExisting`. Your new code lgtm.

Issue #513, opened 2020-03-11
HNC: Set finalizer in HC reconciler


yiqigao217:
Part of #457, #501 .

We need finalizers on HC and HNS instances to make the cascading deletion work. The finalizers allow the reconcilers to trigger further deletion and make sure the related instances are purged before the current instances can be purged. Here's the implementation detail:
1. Add a finalizer to all HNS instances and remove it only when the subnamespace is deleted and purged for a cascading deletion OR if it's a single deletion (`HNS_MISSING` condition will be set on the subnamespace then).
2. Add a finalizer to HC instances only if the namespace has HNS instances. It will be removed when the HNS instances are all deleted and purged for a cascading deletion OR if it's a single deletion (a new default HC instance will be created by the next HC reconciliation).

PR #564, opened 2020-03-27
Implement HNS cascading delete

### LINKED ISSUE #501: HNC: Implement cascadingDelete in HC and HNS reconcilers
Part of #457 

An owned namespace can be cascading deleted if the `hc.spec.allowCascadingDelete` field of the direct owners or itself is set to `true`. Cascading deletion can be initiated by a deletion of an HNS instance or an owner (since it will trigger deletion of the HNSes in that namespace natually). 

If the HNS instance in the owner namespace is missing and the owned namespace doesn't allow cascading delete, an `HNS_MISSING` condition will be set in the subnamespace. It's not a `CRIT_` condition and it won't trigger the `CRIT_ANCESTOR` conditions in the descendants.

yiqigao217:
Add a finalizer to all HNS instances and remove it only when the
subnamespace is deleted and purged for a cascading deletion OR if it's a
single deletion (HNS_MISSING condition will be set on the subnamespace
then).

Add a finalizer to HC instances only if the namespace has HNS instances.
It will be removed when the HNS instances are all deleted and purged for
a cascading deletion OR if it's a single deletion (a new default HC
instance will be created by the next HC reconciliation).

Add allowCascadingDelete field to forest. This is updated by the
hc.Spec.AllowCascadingDelete field.

Add allowCascadingDelete() func to forest. It indicates if the namespace
can be cascading deleted.

Update and move onMissingNamespace() to the top of HCR when a namespace
doesn't exist.

Add a testing script since the purge is not available in the integration
test.

Tested the testing script on GKE cluster by ". hack/test-issue-501.sh".

Fixes #501, #513 
Part of #457

### DIFF
--- a/incubator/hnc/hack/test-issue-501.sh
+++ b/incubator/hnc/hack/test-issue-501.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+bold=$(tput bold)
+normal=$(tput sgr0)
+
+# Set up namespace structure
+echo "----------------------------------------------------"
+echo "${bold}Setting up a 3-level tree with 'parent' as the root${normal}"
+echo "Creating the root 'parent'"
+kubectl create ns parent
+sleep 1
+
+echo "${bold}----Creating the 1st branch of owned namespaces----${normal}"
+echo "Creating owned namespace 'ochild1' for 'parent'"
+kubectl hns create ochild1 -n parent
+sleep 1
+echo "Creating owned namespace 'ochild1ofochild1' for 'ochild1'"
+kubectl hns create ochild1ofochild1 -n ochild1
+echo "Creating owned namespace 'ochild2ofochild1' for 'ochild1'"
+kubectl hns create ochild2ofochild1 -n ochild1
+
+echo "${bold}----Creating the 2nd branch of owned namespaces----${normal}"
+echo "Creating owned namespace 'ochild2' for 'parent'"
+kubectl hns create ochild2 -n parent
+sleep 1
+echo "Creating owned namespace 'ochildofochild2' for 'ochild2'"
+kubectl hns create ochildofochild2 -n ochild2
+
+echo "${bold}----Creating the 3rd branch of a mix of unowned and owned namespaces----${normal}"
+echo "Creating a namespace 'nochild' (not-owned-child) and set its parent to 'parent'"
+kubectl create ns nochild
+sleep 1
+kubectl hns set nochild --parent parent
+echo "Creating owned namespace 'ochildofnochild' for 'nochild'"
+kubectl hns create ochildofnochild -n nochild
+sleep 1
+
+echo "${bold}Here's the outcome of the tree hierarhy:${normal}"
+kubectl hns tree parent
+
+echo "----------------------------------------------------"
+echo "${bold}Test-1:${normal} If the owned namespace doesn't allow cascadingDelete and the HNS is missing in the owner namespace, it should have 'HNS_MISSING' condition while its descendants shoudn't have any conditions."
+echo "${bold}Operation:${normal} delete 'ochild1' hns in 'parent' - kubectl delete hns ochild1 -n parent"
+kubectl delete hns ochild1 -n parent
+sleep 1
+echo "${bold}Expected:${normal} 'ochild1' namespace is not deleted and should have 'HNS_MISSING' condition; no other conditions."
+echo "${bold}Result:${normal}"
+kubectl hns tree parent
+
+echo "----------------------------------------------------"
+echo "${bold}Test-2:${normal} If the HNS is not missing, it should unset the 'HNS_MISSING' condition in the owned namespace."
+echo "${bold}Operation:${normal} recreate the 'ochild1' hns in 'parent' - kubectl hns create ochild1 -n parent"
+kubectl hns create ochild1 -n parent
+sleep 1
+echo "${bold}Expected:${normal} no conditions."
+echo "${bold}Result:${normal}"
+kubectl hns tree parent
+
+echo "----------------------------------------------------"
+echo "${bold}Test-3:${normal} If the owned namespace allows cascadingDelete and the HNS is deleted, it should cascading delete all directly owned namespaces."
+echo "${bold}Operation:${normal} 1) allow cascadingDelete in 'ochid1' - kubectl hns set ochild1 --allowCascadingDelete=true"
+kubectl hns set ochild1 --allowCascadingDelete=true
+echo "2) delete 'ochild1' hns in 'parent' - kubectl delete hns ochild1 -n parent"
+kubectl delete hns ochild1 -n parent
+echo "Waiting 3s for the namespaces to be deleted..."
+sleep 3
+echo "${bold}Expected:${normal} 'ochild1', 'ochild1ofochild1', 'ochild2ofochild1' should all be gone"
+echo "${bold}Result:${normal}"
+kubectl hns tree parent
+
+echo "----------------------------------------------------"
+echo "${bold}Test-4:${normal} If the owner namespace allows cascadingDelete and it's deleted, all its directly owned namespaces should be cascading deleted."
+echo "${bold}Operation:${normal} 1) allow cascadingDelete in 'parent' - kubectl hns set parent --allowCascadingDelete=true"
+kubectl hns set parent --allowCascadingDelete=true
+echo "2) delete 'parent' namespace - kubectl delete ns parent"
+kubectl delete ns parent
+echo "Waiting 3s for the namespaces to be deleted..."
+sleep 3
+echo "${bold}Expected:${normal} only 'nochild' and 'ochildofnochild' should be left and they should have CRIT_ conditions related to missing 'parent'"
+echo "${bold}Result:${normal}"
+kubectl hns tree parent
+kubectl hns tree ochild2
+kubectl hns tree nochild
+
+echo "----------------------------------------------------"
+echo "${bold}Cleaning up${normal}"
+kubectl hns set nochild --allowCascadingDelete=true
+kubectl delete ns nochild
--- a/incubator/hnc/pkg/forest/forest.go
+++ b/incubator/hnc/pkg/forest/forest.go
@@ -154,11 +154,12 @@ const Local = ""
 // Namespace represents a namespace in a forest. Other than its structure, it contains some
 // properties useful to the reconcilers.
 type Namespace struct {
-	forest   *Forest
-	name     string
-	parent   *Namespace
-	children namedNamespaces
-	exists   bool
+	forest               *Forest
+	name                 string
+	parent               *Namespace
+	children             namedNamespaces
+	exists               bool
+	allowCascadingDelete bool
 
 	// originalObjects store the objects created by users, identified by GVK and name.
 	// It serves as the source of truth for object controllers to propagate objects.
@@ -212,6 +213,24 @@ func (ns *Namespace) clean() {
 
 	// Remove from the forest.
 	delete(ns.forest.namespaces, ns.name)
+}
+
+// UpdateAllowCascadingDelete updates if this namespace allows cascading deletion.
+func (ns *Namespace) UpdateAllowCascadingDelete(acd bool) {
+	ns.allowCascadingDelete = acd
+}
+
+// AllowsCascadingDelete returns if the namespace's or any of the owner ancestors'
+// allowCascadingDelete field is set to true.
+func (ns *Namespace) AllowsCascadingDelete() bool {
+	if ns.allowCascadingDelete == true {
+		return true
+	}
+	if !ns.IsOwned {
+		return false
+	}
+	// If the owner is missing, it will return the default false.
+	return ns.Parent().AllowsCascadingDelete()
 }
 
 // SetParent attempts to set the namespace's parent. This includes removing it from the list of
--- a/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
@@ -40,7 +40,6 @@ type HierarchicalNamespaceReconciler struct {
 	Log logr.Logger
 
 	forest *forest.Forest
-	hcr    *HierarchyConfigReconciler
 
 	// Affected is a channel of event.GenericEvent (see "Watching Channels" in
 	// https://book-v1.book.kubebuilder.io/beyond_basics/controller_watches.html) that is used to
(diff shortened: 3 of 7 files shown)


incubator/hnc/pkg/forest/forest.go
```diff
@@ -109,11 +109,12 @@ func (f *Forest) Get(nm string) *Namespace {
...
 	ns = &Namespace{
-		forest:          f,
-		name:            nm,
-		children:        namedNamespaces{},
-		conditions:      conditions{},
-		originalObjects: objects{},
+		forest:               f,
+		name:                 nm,
+		children:             namedNamespaces{},
+		allowCascadingDelete: false,
```
adrianludwin: This isn't needed - `false` is the default value for booleans

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +59,97 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 	nm := req.Name
 	pnm := req.Namespace
 
-	// Get instance from apiserver. If the instance doesn't exist, we don't want to reconcile
-	// it since it may trigger the HC reconciler to recreate the namespace that was just deleted.
-	// TODO expand on this to check the owner's hc.spec.allowCascadingDelete. If it's set to
-	//  true, we still want to reconcile the hns instance. See issue:
-	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/501
+	// Get instance from apiserver. If the instance doesn't exist, we don't want
+	// to reconcile it to create the HNS instance or the namespace.
```
adrianludwin: I'm not sure I understand this comment. If it doesn't exist, it means it's been deleted - how should we handle deletions?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +59,97 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+	cnsInst, err := r.getNamespace(ctx, nm)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
-	switch {
-	case nsInst.Name == "":
-		log.Info("The owned subnamespace does not exist", "subnamespace", nm)
-		inst.Status.State = api.Missing
+	// If the instance is being deleted, remove the finalizers or cascading delete.
```
adrianludwin: It's hard to tell from the code in this function where the finalizers are being removed. Maybe slightly rewriting this logic could make it a bit more clear?

```
var err error
switch {
case len(inst.ObjectMeta.Finalizers) == 0:
  // nothing
case removeFinalizers():
  err = r.writeInstance(ctx, log, inst)
default:
  r.deleteNamespace(ctx, log, cnsInst)
}
return ctrl.Result{}, nil
```

Also, `Reconcile()` is getting a bit long - maybe factor this all out into a function?
```
if deleting, err := r.onDeleting(...); deleting {
   return ctrl.Result{}, err
}
```

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +59,97 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+			return ctrl.Result{}, nil
+		}
+		if deleteNamespace := r.deleting(log, inst, cnsInst); deleteNamespace {
+			r.deleteNamespace(ctx, log, cnsInst)
+		}
+		return ctrl.Result{}, r.writeInstance(ctx, log, inst)
+	}
+
+	r.updateState(log, inst, cnsInst)
+	if inst.Status.State == api.Missing {
```
adrianludwin: Maybe comment why we write the NS if the state is missing? (I get it but it might be worth spelling it out)

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +59,97 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 		if err := r.writeNamespace(ctx, log, nm, pnm); err != nil {
 			return ctrl.Result{}, err
 		}
-	case nsInst.Annotations[api.AnnotationOwner] != pnm:
-		log.Info("The owner annotation of the subnamespace doesn't match the owner", "annotation", nsInst.Annotations[api.AnnotationOwner])
+	}
+
+	// Add finalizers on all non-forbidden HNSes.
```
adrianludwin: Maybe say _why_ we're adding the finalizer (e.g. to ensure it's not deleted until after the child namespace is deleted)

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -101,24 +115,17 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
...
-	if !inst.GetDeletionTimestamp().IsZero() || !nsInst.GetDeletionTimestamp().IsZero() {
-		log.Info("Singleton or namespace are being deleted; will not update")
-		update = false
+	// Set finalizers first. Remove them in the deleting() later if applicable.
+	inst.ObjectMeta.Finalizers = []string{api.MetaGroup}
+
+	// If the instance is being deleted, remove the finalizers accordingly.
+	if !inst.DeletionTimestamp.IsZero() {
+		r.updateFinalizers(ctx, log, inst, nsInst, hnsnms)
 	}
```
adrianludwin: Alternatively, `updateFinalizers` could contain all this logic? Nice simplification btw!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -136,6 +143,31 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
...
+		r.enqueueAffected(log, "relative of deleted namespace", ns.RelativesNames()...)
+		ns.UnsetExists()
+		log.Info("Removed namespace")
+	}
+}
+
+func (r *HierarchyConfigReconciler) updateFinalizers(ctx context.Context, log logr.Logger, inst *api.HierarchyConfiguration, nsInst *corev1.Namespace, hnsnms []string) {
+	switch {
+	case nsInst.DeletionTimestamp.IsZero():
+		// If the namespace is not being deleted, just remove finalizers.
```
adrianludwin: Can you explain a bit better _why_ we're doing this (both comments in this function)?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: Updated with 'If the instance doesn't exist, do nothing and early exist because HCR watches HNS instances and (if applicable) has already updated the owner HC when this HNS was purged.'

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: Added 'Update the state. If the owned child namespace doesn't exist, create it. (This is how a self-service namespace is created through a CR)'

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: Done. Put all above into `updateFinalizers()`.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: This happens when it's not a cascading deletion and user just deletes the HC instance. We should remove the finalizer so it can be purged. A new default HC instance will then be recreated by a next HC reconciliation.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: Done.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +59,121 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+		log.Info("Do nothing in this reconciliation.")
+		return true, nil
+	case r.removeFinalizers(log, inst, cnsInst):
+		log.Info("Remove the finalizers.")
+		return true, r.writeInstance(ctx, log, inst)
+	default:
+		// The owned namespace is not being deleted and it allows cascadingDelete.
+		// Delete the owned namespace.
+		return true, r.deleteNamespace(ctx, log, cnsInst)
+	}
```
adrianludwin: Nit: I feel like this might be easier to read in the opposite direction:

* Is there a child namespace that isn't being deleted? Delete it.
* If there's no namespace, are there finalizers? If so, remove them.
* Default: there must not be anything to do - wait for something to happen (the default case is usually the do-nothing case).

This (roughly) matches the order in which things will actually happen too.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +59,121 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
 	pnm := req.Namespace
 
-	// Get instance from apiserver. If the instance doesn't exist, we don't want to reconcile
-	// it since it may trigger the HC reconciler to recreate the namespace that was just deleted.
-	// TODO expand on this to check the owner's hc.spec.allowCascadingDelete. If it's set to
-	//  true, we still want to reconcile the hns instance. See issue:
-	//  https://github.com/kubernetes-sigs/multi-tenancy/issues/501
+	// Get instance from apiserver. If the instance doesn't exist, do nothing and
+	// early exist because HCR watches HNS instances and (if applicable) has
+	// already updated the owner HC when this HNS was purged.
```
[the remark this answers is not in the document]
yiqigao217: What I meant in the comment is that when this HNS is purged, HCR will be triggered to update the HC instance (possilbly remove the finalizers if there's no HNSes). Since this is when the instance `IsNotFound()`, which means it's already purged, that's why I put that comment. WDYT?

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: sg, will update it.

yiqigao217:
> I didn't see where we called `hnsr.enqueue()` when the namespace has been deleted - did I miss it? Otherwise, looks good other than a few nits.

Ah that's replaced by the HNSR watching for namespaces with mapping function from `NS` to `HNS{Name: NS.Name(), Namespace: NS.OwnerAnnotation()}`. BTW this reminded me to remove `hnsr` or `hcr` if there's no enqueue needed in the other reconciler.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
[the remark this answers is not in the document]
adrianludwin: sg!

adrianludwin:
> Ah that's replaced by the HNSR watching for namespaces with mapping function from `NS` to `HNS{Name: NS.Name(), Namespace: NS.OwnerAnnotation()}`. This makes sure the HNS finalizers get removed when the subnamespace is purged.
> 
> BTW this reminded me to remove `hnsr` or `hcr` if there's no enqueue needed in the other reconciler.

Right, I forgot about that! Thanks.

yiqigao217:
> /assign @adrianludwin
> Please note the integration tests are not passed because of the finalizers set on HC instances. I'm still look into it but the manual testing on GKE cluster all worked out.
> 
> The failed tests are all saying the 'hierarchy' already exists for this[ line](https://github.com/kubernetes-sigs/multi-tenancy/blob/a90b8dd86c38ef1523bfbfc267bd9844b6907b0b/incubator/hnc/pkg/reconcilers/hierarchy_config_test.go#L300) in the test. I guess it's related to the finalizers. I will look into it.

Just keep a record that the integration test failure issue is resolved by only adding the HC finalizers if there's HNS(es) in the namespace, instead of adding the finalzers to all HCs. Adding finalizers to all HCs causes previous unnecessary HCs to be created.

incubator/hnc/hack/test-issue-501.sh
```diff
@@ -0,0 +1,88 @@
...
+echo "${bold}Expected:${normal} only 'nochild' and 'ochildofnochild' should be left and they should have CRIT_ conditions related to missing 'parent'"
+echo "${bold}Result:${normal}"
+kubectl hns tree parent
+kubectl hns tree ochild2
+kubectl hns tree nochild
+
+echo "----------------------------------------------------"
+echo "${bold}Cleaning up${normal}"
+kubectl hns set nochild --allowCascadingDelete=true
+kubectl delete ns nochild
```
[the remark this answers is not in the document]
yiqigao217: When addressing this comment, somehow the newline I added didn't show up in a force push. I added a tab in the new line and it showed up.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +58,113 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+	case len(inst.ObjectMeta.Finalizers) == 0:
+		// We've finished process this, nothing to do.
+		log.Info("Do nothing since the finalizers are already gone.")
+		return true, nil
+	case cnsInst.DeletionTimestamp.IsZero() && r.allowsCascadingDelete(cnm):
+		// The owned namespace is not being deleted and it allows cascadingDelete.
+		// Delete the owned namespace.
+		log.Info("The subnamespace is not being deleted but it allows cascading deletion.")
+		return true, r.deleteNamespace(ctx, log, cnsInst)
+	case r.removeFinalizers(log, inst, cnsInst):
```
adrianludwin: This is basically a nested `switch` statement, which I find a bit hard to read. For example, if the child namespace isn't being deleted yet, I need to look in `onDeleting` to see what happens if we're allowing cascading deletion, but `removeFinalizers` to see what happens if it's not.

Combining this into one larger switch statement might make more sense here, but I'll approve this (and send it to Ryan) so it's up to you if you want to make this change or not. I'd do something like this:

```
done := false
switch{
case cnsInst.Name == "":
  log.Info("Namespace has been fully deleted")
  done = true

case cnsInst.GetAnnotations()[api.AnnotationOwner] != pnm:
  log.Info("Skipping cascading deletion: namespace has conflicting owner")
  done = true

case cnsInst.DeletionTimestamp.IsZero(): // not being deleted yet
  if r.allowsCascadingDelete(cnm) {
    if err := r.deleteNamespace(ctx, log, cnsInst); err != nil {
      return true, err
    }
  } else {
    log.Info("Skipping cascading deletion: allowsCD is false")
    done = true
  }

default: // already being deleted
  log.Info("Waiting for cascading deletion to finish")
}

if done {
  inst.ObjectMeta.Finalizers = nil
  return true, r.writeInstance(ctx, log, inst)
}

return true, nil
```

yiqigao217:
Please hold until I make the last change @adrianludwin suggested.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
yiqigao217: After another thought, I decided to keep it as is to have `removeFinalizers()` separated as suggested in https://github.com/kubernetes-sigs/multi-tenancy/pull/564#discussion_r399524967.

Also there will be an additional `len(inst.ObjectMeta.Finalizers) == 0` case in the proposed large switch here, which may be too long and  hard to read.

adrianludwin: sgtm! Thanks.

rjbez17:
One nit, but lgtm. I'm am a bit worried about the nested switch situation discussed but can't think of an immediate better solution. Feel free to remove hold if you want to merge without the nit.

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -59,51 +58,113 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
...
+
+	cnm := inst.Name
+	log.Info("The HNS instance is being deleted.")
+	switch {
+	case len(inst.ObjectMeta.Finalizers) == 0:
+		// We've finished process this, nothing to do.
+		log.Info("Do nothing since the finalizers are already gone.")
+		return true, nil
+	case cnsInst.DeletionTimestamp.IsZero() && r.allowsCascadingDelete(cnm):
+		// The owned namespace is not being deleted and it allows cascadingDelete.
```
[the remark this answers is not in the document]
yiqigao217: Updated.

yiqigao217:
Ah I need to make some changes since `ns.Owner` field no longer exists after #581 is merged.

yiqigao217:
Retested with `make test` and manual test:
```
make deploy
. hack/test-issue-501.sh
```

incubator/hnc/pkg/forest/forest.go
```diff
@@ -214,6 +215,24 @@ func (ns *Namespace) clean() {
...
+	if ns.allowCascadingDelete == true {
+		return true
+	}
+	if !ns.IsOwned {
+		return false
+	}
+	// If the owner is missing, it will return the default false.
+	return ns.Parent().AllowsCascadingDelete()
+}
+
```
yiqigao217, line 235: @adrianludwin this is where the `ns.Ower` => `ns.IsOwned` change is.

adrianludwin: Thanks!

PR #571, opened 2020-03-29
Add better HNC Config observability and fix flaky test

adrianludwin:
One of the HNC config observability tests was failing because we didn't
wait for the HNC config reconciler to pick up a change before verifying
that it was working:

* We changed a sync mode from "propagate" to "ignore"
* We created a new object
* We verified that this object wasn't created

Occasionally (about 1/4 of the time), the object reconciler would run before
the HNC config reconciler had a chance to change it to "ignore." The fix
is to add a "Mode" to the HNC Config status, so we can tell when the
reconciler has actually picked it up. This is also useful for when the
spec doesn't include an explicit mode.

This change also includes a few other minor improvements:

* If there's an extra HNC Config object, and we can't write it back due
to an apiserver error, retry the failure (previously we wouldn't retry).
* Fully deterministically sort the types in the Status (had only been
sorting by apiVersion)
* Added comments throughout
* Make the Gomega pattern matching more explicit for easier-to-read
tests and better error messages (e.g. when the sync mode is wrong, the
bad value will now be printed out)
* Other minor test improvements

Tested: ran `make test` 10x

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hnc_config.go
+++ b/incubator/hnc/pkg/reconcilers/hnc_config.go
@@ -61,32 +61,30 @@ type gvkSet map[schema.GroupVersionKind]bool
 // `config` singleton.
 const checkPeriod = 3 * time.Second
 
-// Reconcile sets up some basic variable and logs the Spec.
-// TODO: Updates the comment above when adding more logic to the Reconcile method.
+// Reconcile is the entrypoint to the reconciler.
 func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
 
-	// Validate the singleton name.
-	if err := r.validateSingletonName(ctx, req.NamespacedName.Name); err != nil {
-		r.Log.Error(err, "Singleton name validation failed")
-		return ctrl.Result{}, nil
+	// Validate the singleton name, and early exit if we're not validating the real one (note that the
+	// bad singleton will have a condition set, but the main one will not be affected).
+	if ok, err := r.validateSingletonName(ctx, req.NamespacedName.Name); !ok || err != nil {
+		r.Log.Error(err, "An incorrectly-named HNC Config exists", "name", req.NamespacedName.Name)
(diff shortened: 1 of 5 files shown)


incubator/hnc/pkg/reconcilers/hnc_config.go
```diff
@@ -372,36 +376,49 @@ func (r *ConfigReconciler) validateSingletonName(ctx context.Context, nm string)
...
+		if !r.activeGVKs[ts.GetGVK()] {
+			continue
 		}
+
+		// Initialize status
+		apiVersion, kind := gvk.ToAPIVersionAndKind()
+		status := api.TypeSynchronizationStatus{
+			APIVersion: apiVersion,
+			Kind:       kind,
+			Mode:       ts.GetMode(), // may be different from the spec if it's implicit
```
yiqigao217: Question: How is this GetMode() up to date?

adrianludwin: On line 329, we validate the mode (e.g. change it from the empty string to `propagate`) for a new controller. When we update an existing controller, it happens on line 250. Does that make sense?

yiqigao217: Thank you for the explanation! Ah I see, this is the forest version and it may not be up to date (because of the lag in between the mode change) but it will eventually match the spec. It always reflects the mode Object controller is using.

adrianludwin: That's exactly it. This was the cause of the flakiness in the tests - we set the spec to "ignore," but then _before_ that could take effect, we ran a test that assumed the object controller was already ignoring stuff. But it wasn't.

yiqigao217: Got you, yes, that's why you added `typeHasStatus()` before checking the object propagation.

adrianludwin: Yup!

Issue #347, opened 2020-01-15
HNC: propagate deleted source when conditions are gone


yiqigao217:
Currently when critical conditions are cleared for a namespace, the previously paused deletion of an object will not propagate to children namespaces.

The issue is that since there’s a critical condition in the namespace, the `syncDeletedSource()` is skipped. What the `syncDeleteSource()` func does is deleting the source in the forest and enqueuing all descendant copies.

Since the source is never deleted in the forest, even when the critical conditions are gone for the namespace, the should-be-deleted propagated objects are still consistent with the forest thinking that the source is still there in the ancestor by looking at the forest. The source in the forest will never be deleted because the real object has been deleted and will never be reconciled again to sync with forest.

adrianludwin:
@yiqigao217 Is this still an open issue?

yiqigao217:
I tested manually again and it's still an open issue. Here's my testing:
`k create ns zparent`
`k create ns zchild`
`k create ns zgrand`
`k hns set zchild --parent zparent`
`k hns set zgrand --parent zchild`
`k hns tree zparent`
zparent
|__zchild
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp; |__zgrand
`k -n zchild create configmap configmap-zchild --from-literal key=value`
The source is created in `zchild` and it's propagated to `zgrand`.
`k delete ns zparent`
This causes `zchild` and `zgrand` to have `CRIT` condition. `zgrand` got msg 'Propagation paused in the subtree of zchild due to a critical condition'
`k delete configmap configmap-zchild -n zchild`
The propagated `configmap-zchild` in `zgrand` is not deleted as expected.
`k create ns zparent`
Now the `CRIT` conditions are all gone, but the propagated `configmap-zchild` in `zgrand` is still there.

adrianludwin:
Thanks, I'll look at this along with #328

adrianludwin:
Here's what _should_ be happening here:
* The critical condition is removed, and all descendents are enqueued
* As each descendent is reconciled, it notices that it changed, so it calls `updateObjects`
* In `updateObjects`, it should notice that the `configmap-zchild` needs to be deleted

Currently looking into why this isn't happening.

adrianludwin:
Ah - when there's a crit condition, we skip the entire object reconciler so we never update the forest to know that the object's been removed. This doesn't seem right - we should only skip _actions_ if there's a crit condition, not our internal reflection of the apiserver state.

[...]

Issue #292, opened 2019-12-05
HNC: Use IsZero() uniformly (may require changes to k/k)


adrianludwin:
See: https://github.com/kubernetes-sigs/multi-tenancy/pull/282#discussion_r353772980.

It appears that the k8s client API isn't consistent across unstructured.Unstructured and regular runtime.Objects?

PR #574, opened 2020-03-29
Always sync objects to forest, even in namespaces with critical conditions

adrianludwin:
See #347. When an object is removed, we need to remove it from the
in-memory forest even when the object is in a namespace with a critical
condition, because when the condition is resolved, we'll use the forest
to re-sync all descendants of that namespace, and if the forest has
stale objects in it, the propagated copies won't be removed.

Added several other cleanups including:

* Made several names more unique so they can be more easily searched
(e.g. replaced `delete()` with `deleteObject()`)
* Factored out a new syncMissingObject() functionout of Reconcile()
* Some cosmetic renames plus a bunch of longer comments

TESTED: new integ test fails without the fix and works with it; also
tested manually on GKE.

### DIFF
--- a/incubator/hnc/pkg/reconcilers/object.go
+++ b/incubator/hnc/pkg/reconcilers/object.go
@@ -41,16 +41,16 @@ import (
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 )
 
-// action is the action to take after Reconcile syncs with the in-memory forest.
+// syncAction is the action to take after Reconcile syncs with the in-memory forest.
 // This is introduced to consolidate calls with forest lock.
-type action int
+type syncAction int
 
 const (
-	// Start with an “unknown” to be sure of enum’s initialization.
-	unknown action = iota
-	remove
-	write
-	ignore
+	// Start with an “actionUnknown” to be sure of enum’s initialization.
+	actionUnknown syncAction = iota
+	actionRemove
+	actionWrite
+	actionNop
 )
 
 // namespacedNameSet is used to keep track of existing propagated objects of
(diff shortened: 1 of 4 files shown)


adrianludwin:
Fixes #347

rjbez17:
/lgtm
/approve
/hold

@yiqigao217 feel free to remove hold when satisfied

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -271,33 +263,69 @@ func (r *ObjectReconciler) ignore(ctx context.Context, log logr.Logger, inst *un
...
-// all descendant copies will be enqueued. If a propagated object is obsolete, it will be deleted.
-// Otherwise, it will be overwritten by the source if they are different.
-func (r *ObjectReconciler) syncObject(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) (action, *unstructured.Unstructured) {
+// syncMissingObject handles the case where the object we're reconciling doesn't exist. If it can
+// determine a final action to take, it returns that action, otherwise it returns actionUnknown
+// which indicates that we need to call the regular syncObject method. Note that this method may
+// modify `inst`.
+func (r *ObjectReconciler) syncMissingObject(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) syncAction {
+	// If the object exists, skip
+	if inst.GetCreationTimestamp() != (v1.Time{}) {
```
yiqigao217: Replace it with `!inst.GetCreationTimestampe().isZero()` to be consistent across?

incubator/hnc/pkg/reconcilers/object_test.go
```diff
@@ -129,6 +129,32 @@ var _ = Describe("Secret", func() {
...
+		fooHier.Spec.Parent = brumpfName
+		updateHierarchy(ctx, fooHier)
+		Eventually(hasCondition(ctx, fooName, api.CritParentMissing)).Should(BeTrue())
+		Eventually(hasCondition(ctx, barName, api.CritAncestor)).Should(BeTrue())
+
+		// Delete the object from `foo` and verify it isn't deleted from `bar` because of the critical
+		// condition
+		deleteObject(ctx, "Role", fooName, "foo-role")
+		time.Sleep(1 * time.Second) // todo: merge with similar constants elsewhere
+		Expect(hasObject(ctx, "Role", barName, "foo-role")()).Should(BeTrue())
```
yiqigao217: Maybe add this before it?
```
Expect(hasObject(ctx, "Role", fooName, "foo-role")()).Should(BeFaulse())
```

incubator/hnc/pkg/reconcilers/object.go
adrianludwin: This is one of those annoying cases where we can't call IsZero() because it expects a pointer receiver, and `inst.GetCreationTimestamp()` doesn't return one. See #292 .

incubator/hnc/pkg/reconcilers/object_test.go
adrianludwin: Good idea, thanks. Done.

incubator/hnc/pkg/reconcilers/object.go
yiqigao217: Thanks for filing that issue earlier for easy referencing now!

PR #575, opened 2020-03-30
Fix #333. Fail a release on duplicate image and tag

### LINKED ISSUE #333: HNC: release should not re-push an existing image by default
See https://github.com/kubernetes-sigs/multi-tenancy/pull/332#discussion_r360651870. By default, the release should refuse to push an image if it already exists on the repo, although we'll need a way to override (eg to push :latest or to test).

adrianchung:
Do not allow overwriting of an image and tag if it already exists in the
repository. This prevents us from being able to overwrite released
images. Pass in FORCE_RELEASE=true to override this behavior.


adrianchung:
Poke. Check to see if it's detecting signed CLA...

PR #576, opened 2020-03-30
Address #333. Fail a release on duplicate image and tag

adrianchung:
Do not allow overwriting of an image and tag if it already exists in the
repository. This prevents us from being able to overwrite released
images. Pass in FORCE_RELEASE=true to override this behavior.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -17,6 +17,13 @@ ifeq ($(CONFIG),kind)
 HNC_IMG ?= controller:kind-local
 else
 HNC_IMG ?= "gcr.io/${PROJECT_ID}/hnc/controller:${HNC_IMG_TAG}"
+# We don't want to overwrite an existing docker image and tag so we query the repository
+# to see if the image already exists. This can be overridden by setting FORCE_RELEASE=true
+ifeq ($(FORCE_RELEASE), true)
+	IMG_READ_ERROR=1
+else
+	IMG_READ_ERROR=$(shell gcloud container images describe $(HNC_IMG) > /dev/null 2>&1; echo $$?)
+endif
 endif
 
 HNC_RELEASED_IMG ?= "gcr.io/k8s-staging-multitenancy/hnc/controller:${HNC_IMG_TAG}"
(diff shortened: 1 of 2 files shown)


incubator/hnc/Makefile
```diff
@@ -198,6 +205,9 @@ ifndef HNC_PAT
 endif
 ifndef HNC_RELEASE_ID
 	$(error HNC_RELEASE_ID is undefined; ${ERR_MSG})
+endif
+ifeq ($(IMG_ALREADY_EXISTS), 0)
```
adrianludwin: I think this means "the return code from calling gcloud was 0, which means it exists" but the variable makes it sound like a boolean, which has the opposite meaning. I was a bit confused reading this.

Can we either invert the value in this variable, or else give it a more descriptive name (e.g. `IMG_READ_ERROR`)?

adrianludwin:
Can you please amend your commit message to describe how you tested this change? (Examples: [1](https://github.com/kubernetes-sigs/multi-tenancy/commit/f312c5b4e8d71b9dfcf92c7413541577d3f606d0#diff-a75a5601d28a16c6aa2432163fc30534), [2](https://github.com/kubernetes-sigs/multi-tenancy/commit/b2871269b52f1f78c1f2b46e8804fa38cf88784f#diff-a75a5601d28a16c6aa2432163fc30534))

incubator/hnc/Makefile
adrianchung: Absolutely I'll make the change.

adrianchung:
I forced pushed a new commit message to get the commit message with how I tested the change. [Commit message](https://github.com/kubernetes-sigs/multi-tenancy/pull/576/commits/2897dee1953ccd1e22629396cab9202dee122790)

adrianludwin:
Ah sorry I forgot to mention this - do you mind squashing your two commits? We generally try to have one commit per PR, it keeps the history a bit cleaner.

adrianludwin:
lgtm after the squash

adrianchung:
Yep, I didn't squash yet because of https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#6-squashing-and-commit-titles. I'll get that squashed for the final ok.

adrianludwin:
> Yep, I didn't squash yet because of https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#6-squashing-and-commit-titles. I'll get that squashed for the final ok.

Fascinating, I'd never seen that before. It's pretty good advice!

adrianludwin:
/lgtm
/approve

Fixes #333

PR #579, opened 2020-03-31
Reduce default verbosity of validators

adrianludwin:
I had a hard time reading the logs with all the logspam from the
validators, which very rarely have anything useful to say IMO.

Tested: observed that the logs were easier to read


adrianludwin:
/assign @yiqigao217 

Hi, can you please give me a quick lgtm on this? No need for external approval.

Issue #328, opened 2019-12-19
HNC: object conditions on source namespaces may not be cleared


adrianludwin:
If an object cannot be propagated to a child namespace, in some cases the resulting condition could get "stuck" on the source namespace. Here's the scenario:

* An object cannot be propagated into namespace X - eg, because it's a rolebinding with higher privileges than HNC has, or because there's some transient error on the apiserver, or because some other validating webhook is forbidding it.
* The object will have a CANNOT_UPDATE condition added to namespace X, but also a CANNOT_PROPAGATE condition added to the source namespace, which we'll call namespace Y.
* Now the parent of namespace X is changed, such that Y is no longer an ancestor.

In this case, the condition will never be cleared from namespace Y until the controller is restarted.

adrianludwin:
This problem is being introduced in #322 but we decided not to hold that PR up until this is fixed.

adrianludwin:
/assign

I'll have a look at this one

PR #580, opened 2020-03-31
Use API objects to store conditions

### LINKED ISSUE #326: HNC: fewer propagations while objects have conditions
Once #322 is merged, there's a known problem where every time the object is re-reconciled, it will cause any namespaces with a condition pointing to it to be re-reconciled as well. This should be relatively rare - objects with conditions won't be reconciled very often - but in some cases (eg can't write an object) this could be called repeatedly.

adrianludwin:
Special thanks to @yiqigao217 for recommending some key parts of this
approach.

Our handling of conditions in the forest has been pretty lax - we
basically use strings with magical formatting. This change tightens up
the interface by allowing us to use the API's AffectedObject struct
instead of the string keys. This also makes conversions and
manipulations much easier, and is used by two followon PRs to address
 #326 and #328.

This change also includes some minor cleanups, e.g. removing the term
"keys" from the original objects in the forest and replacing it with
"names" instead.

Tested: existing tests plus confirmed that conditions from #328 still
look good via `kubectl hns tree`.

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -16,7 +16,11 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"fmt"
+	"sort"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 // Constants for types and well-known names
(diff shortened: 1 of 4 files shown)


incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -145,6 +149,52 @@ type AffectedObject struct {
 	Name      string `json:"name,omitempty"`
 }
 
+func NewAffectedNamespace(ns string) AffectedObject {
+	return AffectedObject{
+		Version: "v1",
+		Kind:    "Namespace",
+		Name:    ns,
+	}
```
yiqigao217: Just confirming, is this a change on how to represent namespaces? Previously, we planned to use only the name field to represent namespaces, right?

This may be better in that we don't need to specifically document the representation.

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -145,6 +149,52 @@ type AffectedObject struct {
...
+}
+
+func (a AffectedObject) String() string {
+	if a.Name == "" {
+		return ""
+	}
+	if a.Namespace != "" && a.Name == "" {
+		return a.Namespace
+	}
```
yiqigao217: Is this a bug? Seems it should be?
```
if a.Namespace == "" && a.Name != "" {
    return a.Name
}
```

incubator/hnc/api/v1alpha1/hierarchy_types.go
adrianludwin: At the API level, we always represented namespaces this way - which is the same way that Kubernetes itself represents namespaces. Internally, we didn't, but now we're using the same representation internally and externally.

incubator/hnc/api/v1alpha1/hierarchy_types.go
adrianludwin: Nice catch, thanks!

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: Thanks for the explanation!

PR #584, opened 2020-04-01
Compute NumSourceObjects in HNCConfiguration status

sophieliu15:
This PR computes NumSourceObjects as a part of HNCConfiguration status. NumSourceObjects indicates the number of source objects of a specific type created by users.

Issue: #411


sophieliu15:
/assign @adrianludwin 
/assign @yiqigao217 

@adrianchung Feel free to take a look. :)

incubator/hnc/pkg/forest/forest.go
```diff
@@ -142,6 +142,11 @@ func (f *Forest) GetNamespaceNames() []string {
 	return keys
 }
 
+// GetNamespaces returns the namespaces in the forest.
+func (f *Forest) GetNamespaces() namedNamespaces {
+	return f.namespaces
+}
+
```
adrianludwin, line 149: We shouldn't directly return the map like this, since the caller could then modify it directly. This is why `f.namespaces` is unexported (i.e., it has a lowercase first letter) plus the type is unexported too. I'm actually surprised Go let you do this but I guess you never refer to the type `namedNamespaces` directly outside of this package, so it was allowed. 

Instead, the client should use `GetNamespaceNames()`:

```
for _, nsnm := range r.forest.GetNamespaceNames() {
  ns := r.forest.Get(nsnm)
  ...
}

File: incubator/hnc/pkg/reconcilers/hnc_config.go [hunk 2]
```diff
@@ -399,12 +399,22 @@ func (r *ConfigReconciler) setTypeStatuses(inst *api.HNCConfiguration) {
...
 		if ts.GetMode() != api.Ignore {
 			numProp := ts.GetNumPropagatedObjects()
 			status.NumPropagatedObjects = &numProp
 		}
 
+		// Only add NumSourceObjects if we are propagating objects of this type.
+		if ts.GetMode() == api.Propagate {
+			numSrc := 0
+			nms := r.Forest.GetNamespaces()
+			for _, ns := range nms {
```
adrianludwin, line 412: Does it make sense to count the source objects in the reconciler at the same time we count the propagated objects instead? e.g. in addition to `recordPropagatedObject` and `removePropagatedObject`, we could have a `<record|remove>SourceObject` as well?

incubator/hnc/pkg/reconcilers/hnc_config_test.go
```diff
@@ -324,6 +324,42 @@ var _ = Describe("HNCConfiguration", func() {
 
 		Eventually(getNumPropagatedObjects(ctx, "v1", "LimitRange"), statusUpdateTime).Should(Equal(0))
 	})
+
+	It("should set NumSourceObjects for a type in propagate mode", func() {
+		addToHNCConfig(ctx, "v1", "LimitRange", api.Propagate)
+		makeObject(ctx, "LimitRange", fooName, "foo-lr")
+
+		// There should be 2 source objects of LimitRange: 1 added in this test case, the other added in the previous test case.
```
adrianludwin, line 332: Tests shouldn't depend on each other, I don't think we even guarantee the order in which tests are run. Is there a way we can do cleanups? E.g. `makeObject` stores each object in a map, and then `AfterEach` (at the top of this file) deletes them?

sophieliu15: Good idea! Thanks for the suggestion! I manually deleted those objects for the `NumSourceObjects` test cases. I added a TODO and fired an issue (#586) for this. I will address this issue in next PR just in case I need to change other test cases.

incubator/hnc/pkg/reconcilers/hnc_config.go
sophieliu15: Did you mean we call `<record|remove> SourceObject` next to `<record|remove>PropagatedObject>` in the places where `<record|remove>PropagatedObject>` are called? 

I don't think they should be counted at the same place. Looks like propagated objects and source objects are handled in different code paths.  For example, if we only have one namespace and we add a source object, if we only rely on `<record|remove>PropagatedObject>` to signal the config reconciler, the `NumSourceObjects` won't be updated because we will not go through the code path that invokes any of the `<record|remove>PropagatedObject>`.

Feel free to let me know if I miss anything or I misunderstand your comment.

incubator/hnc/pkg/forest/forest.go
sophieliu15: Thanks for the explanation. Updated.

incubator/hnc/pkg/reconcilers/hnc_config.go
adrianludwin: Yeah not in exactly the same places, I just mean the general idea of storing an up-to-date count in the object reconciler.

incubator/hnc/pkg/reconcilers/hnc_config_test.go
adrianludwin: sg

incubator/hnc/pkg/reconcilers/hnc_config.go
adrianludwin: Actually, it's pretty clear the way it is now, maybe best just to leave it.

PR #585, opened 2020-04-01
Set object conditions atomically (#326)

adrianludwin:
The object reconciler locks the forest twice: once when determining what
to do with an object (i.e. propagate or delete it), and once after the
propagation/deletion has been attempted so we can update the condition.
Prior to this change, we cleared all conditions associated with an
object during the first lock, and set them (if necessary) during the
second lock.

Unfortunately, in cases where objects failed to be propagated, this set
up neverending spin loop:

1. During the first lock, we'd clear the condition. Because the
conditions changed, we'd re-enqueue the HC reconciler.
2. If the HCR managed to run before the second lock was acquired (which
was usually the case), it would see that its conditions had changed and
update the HC.
3. During the second lock, we'd put the object condition back to what it
was before, also re-enqueng the HCR since it looks like another
condition change.
4. HCR would run again, update the HC, and re-enqueue all local objects
to be reconciled again, including this one (it also did this in step 2,
but that's not as relevant).

This led to three consequences:
* The HC was constantly changing, making it hard for clients like
`kubectl hns set` to modify it successfully.
* The conditions appeared to keep appearing and disappearing.
* Spinlocks are probably a bad idea as they DOS the apiserver.

This new change updates the object reconciler to _never_ modify the
conditions during the first lock, only the second. This ensures that we
only modify the conditions if we absolutely have to.

Tested: while this issue was previously known as a theoretical
possibility, it became an impedement to testing and fixing #328. I
developed both fixes together but am submitting them as separate PRs.
Without this fix, my test script for #328 (plus the fix) sometimes fails
because we can't update the hierarchy, sometimes fails because the
conditions aren't present or cleared correctly, and the logs always show
HNC spinning. With this change, the fix for #328 works every time and
the logs show that HNC is not spinning.

### DIFF
--- a/incubator/hnc/pkg/reconcilers/object.go
+++ b/incubator/hnc/pkg/reconcilers/object.go
@@ -43,14 +43,13 @@ import (
 
 // syncAction is the action to take after Reconcile syncs with the in-memory forest.
 // This is introduced to consolidate calls with forest lock.
-type syncAction int
+type syncAction string
 
 const (
-	// Start with an “actionUnknown” to be sure of enum’s initialization.
-	actionUnknown syncAction = iota
-	actionRemove
-	actionWrite
-	actionNop
+	actionUnknown syncAction = "<hnc internal error - unknown action>"
+	actionRemove  syncAction = "remove"
+	actionWrite   syncAction = "write"
+	actionNop     syncAction = "no-op"
 )
 
 // namespacedNameSet is used to keep track of existing propagated objects of
@@ -226,11 +225,6 @@ func (r *ObjectReconciler) syncWithForest(ctx context.Context, log logr.Logger, 
 	if r.skipNamespace(ctx, log, inst) {
 		return actionNop, nil
 	}
-
-	// Clear any existing conditions associated with this object. TODO: this will never be called if
-	// the source namespace holding this object goes away (i.e., if this namespace gets a new parent).
-	// See https://github.com/kubernetes-sigs/multi-tenancy/issues/328.
-	r.clearConditions(log, inst)
 
 	// If the object's missing and we know how to handle it, return early.
 	if missingAction := r.syncMissingObject(ctx, log, inst); missingAction != actionUnknown {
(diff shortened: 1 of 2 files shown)


adrianludwin:
Fixes #326 

/assign @yiqigao217 
/assign @rjbez17

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -504,58 +503,72 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+			r.enqueueNamespace(log, ns.Name(), "Set CannotPropagate since it has finalizers")
+		}
 
-func getAO(inst *unstructured.Unstructured) api.AffectedObject {
-	gvk := inst.GetObjectKind().GroupVersionKind()
-	return api.NewAffectedObject(gvk, inst.GetNamespace(), inst.GetName())
-}
+	case err != nil:
+		// There was an error updating this object
+		msg := fmt.Sprintf("Could not %s: %s", act, err.Error())
```
yiqigao217, line 544: The msg may not look right for `actionUnknown syncAction = ""`?

BTW Question: when would `actionUnknown` happen?

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -462,18 +456,21 @@ func (r *ObjectReconciler) enqueuePropagatedObjects(ctx context.Context, log log
...
-		return r.deleteObject(ctx, log, inst)
+		err = r.deleteObject(ctx, log, inst)
 	case actionWrite:
-		return r.writeObject(ctx, log, inst, srcInst)
+		err = r.writeObject(ctx, log, inst, srcInst)
 	default:
-		// Generate log for any unset action.
-		log.Error(nil, "ACTION UNSET!!")
-		return nil
+		err = fmt.Errorf("HNC couldn't determine how to update this object (desired action: %s)", act)
```
yiqigao217, line 469: `(desired action: %s)` would be 'actionUnknown`?

incubator/hnc/pkg/reconcilers/object.go
adrianludwin: It should never happen :) I'll make that a bit more clear elsewhere.

incubator/hnc/pkg/reconcilers/object.go
adrianludwin: Yeah, I should make it more clear. This _should_ never happen but if it does, printing an empty string as the reason doesn't really make sense.

Issue #281, opened 2019-11-28
HNC: Store source objects by GK in the forest


yiqigao217:
Currently, the source objects in the forest are stored by GVK:
`type objects map[schema.GroupVersionKind]map[string]*unstructured.Unstructured`
Since users can deal with the same object with different versions, the identifier here should just be Group and Kind.

adrianchung:
I'm going to be picking this issue up

adrianludwin:
/close

We decided not to do anything more in #589.

PR #589, opened 2020-04-02
Address #281. Add comment to GVK object map in forest

adrianchung:
Remove the link to the GitHub issue and give an explanation for the type
in the forst in-memory map

### DIFF
--- a/incubator/hnc/pkg/forest/forest.go
+++ b/incubator/hnc/pkg/forest/forest.go
@@ -144,7 +144,10 @@ func (f *Forest) GetNamespaceNames() []string {
 
 type namedNamespaces map[string]*Namespace
 
-// TODO Store source objects by GK in the forest - https://github.com/kubernetes-sigs/multi-tenancy/issues/281
+// While storing the V in GVK is not strictly necessary to match what's in the HNC type configuration,
+// as a client of the API server, HNC will be to be reading and writing versions of the API to communicate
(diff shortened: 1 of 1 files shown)


adrianchung:
cc: @sophieliu15 @yiqigao217

incubator/hnc/pkg/forest/forest.go
```diff
@@ -144,7 +144,12 @@ func (f *Forest) GetNamespaceNames() []string {
 
 type namedNamespaces map[string]*Namespace
 
-// TODO Store source objects by GK in the forest - https://github.com/kubernetes-sigs/multi-tenancy/issues/281
+// These lookups remain as GVK instead of GK as the interface we get out of Unstructured
+// naturally lends ourselves to using GVK. This also gives us more flexibility to handling
```
adrianludwin, line 148: Rather than saying it's because of the interface from Unstructured - which is an implementation detail - I think I'd suggest pointing to the root cause: HNC is a client of the apiserver, and as a client, it needs to pick specific versions both on the reading and writing side. So while we technically don't need to store the version because it will always match what's in the HNC type configuration, we may as well because we need the GVK anyway while both reading and writing.

Does that make sense?

adrianchung: Absolutely that's a good point. I'll make the change and update

Issue #454, opened 2020-02-25
HNC: Design & Implement a CLI for creating and updating HNCConfiguration object


sophieliu15:
Currently we can only create/update HNCConfiguration object via yaml files. We can create a `kubectl hns` command to achieve that, which will make the interaction more user-friendly and efficiently.

sophieliu15:
/close

This issue is fixed.

PR #592, opened 2020-04-02
Add kubectl commands for manipulating and displaying the HNC configuration

sophieliu15:
This PR adds a command (`kubectl hns config`) for manipulating the HNC configuration and a child command  (`kubectl hns config describe`) for displaying the HNC configuration.

Specifically, `kubectl hns config describe` will display current HNC configuration. Following is an example output:

```
Spec:
  Types:
    APIVersion: rbac.authorization.k8s.io/v1
    Kind: Role
    Mode: 

    APIVersion: rbac.authorization.k8s.io/v1
    Kind: RoleBinding
    Mode: 

    APIVersion: v1
    Kind: CronTab
    Mode: propagate

Status:
  Types:
    APIVersion: rbac.authorization.k8s.io/v1
    Kind: Role
    Mode: propagate

    APIVersion: rbac.authorization.k8s.io/v1
    Kind: RoleBinding
    Mode: propagate

  Conditions:
    Code: objectReconcilerCreationFailed
    Message: Couldn't create ObjectReconciler for type /v1, Kind=CronTab: no matches for kind "CronTab" in version "v1"
```
Issue: #454


adrianludwin:
For `describe`, I don't think we need both the spec and status. People can use `get -o yaml` if they want that... I think just the status should be fine, and you don't even need to _call_ is the status. Eg maybe go for something like this:

```
$ k hns config descript
Synchronized types:
* Propagating: Role (rbac.authorization.k8s.io/v1)
* Propagating: RoleBinding (rbac.authorization.k8s.io/v1)

Conditions:
* Couldn't create ObjectReconciler for type /v1, Kind=CronTab: no matches for kind "CronTab" in version "v1" (ObjectReconcilerCreationFailed)
```

incubator/hnc/pkg/kubectl/config.go
```diff
@@ -0,0 +1,34 @@
...
+package kubectl
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "Manipulates the HNC configuration",
+	Args:  cobra.MinimumNArgs(1),
```
adrianludwin, line 25: Why do we need this line? What happens if we leave it out? Do we get a description of all the valid subcommands (that's what we want, I think, ideally).

incubator/hnc/pkg/kubectl/config_describe.go
```diff
@@ -0,0 +1,58 @@
+/*
```
adrianludwin, line 1: Go filenames usually don't use underscores except for meaningful suffixes like `_test`. So this should be `configdescribe.go`.

incubator/hnc/pkg/kubectl/config.go
sophieliu15: Good catch! I think I mixed up "argument" and "subcommand". We can leave this line out. It will show valid subcommands with or without this line, like below:

```
Manipulates the HNC configuration

Usage:
  kubectl config [command]

Available Commands:
  describe    Displays information about the HNC configuration
```

incubator/hnc/pkg/kubectl/config_describe.go
sophieliu15: Done. Thanks for the suggestion!

sophieliu15:
> For `describe`, I don't think we need both the spec and status. People can use `get -o yaml` if they want that... I think just the status should be fine, and you don't even need to _call_ is the status. Eg maybe go for something like this:
> 
> ```
> $ k hns config describe
> Synchronized types:
> * Propagating: Role (rbac.authorization.k8s.io/v1)
> * Propagating: RoleBinding (rbac.authorization.k8s.io/v1)
> 
> Conditions:
> [ObjectReconcilerCreationFailed] Couldn't create ObjectReconciler for
> type /v1, Kind=CronTab: no matches for kind "CronTab" in version "v1"
> ```

Thanks for the suggestion. Please see the updated code.

sophieliu15:
> This PR says "manipulating and displaying" but so far it just displays, but doesn't manipulate :)

Good catch! Changed to "displaying".

incubator/hnc/pkg/kubectl/configdescribe.go
```diff
@@ -0,0 +1,54 @@
...
+			action := ""
+			switch t.Mode {
+			case api.Propagate:
+				action = "Propagating"
+			case api.Remove:
+				action = "Removing"
+			default:
+				action = "Ignoring"
+			}
+			fmt.Printf("* %s: %s(%s)\n", action, t.Kind, t.APIVersion)
```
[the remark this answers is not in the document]
adrianludwin: I was wondering about that, I had trouble commenting too.

adrianludwin:
/lgtm

I'm happy is @rjbez17 is too!

incubator/hnc/pkg/kubectl/configdescribe.go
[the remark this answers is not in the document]
sophieliu15: Looks like git is working now. Just added a space between Kind and APIVersion.

PR #593, opened 2020-04-02
HNC: Change condition codes to CamelCase strings.

### LINKED ISSUE #500: HNC: Change all condition code to CamelCase
For example,  change `CritParentMissing    Code = "CRIT_PARENT_MISSING"` to `CritParentMissing    Code = "CritParentMissing"`.

jiachengxu:
fixes #500 

Tested: I added lines like:
```go
fmt.Println(string(api.CritParentMissing))
```
in cmd/manager/main.go to check if the condition code is printed in initial-capitalized CameCase string.

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -37,15 +37,13 @@ const (
 )
 
 // Condition codes. *All* codes must also be documented in the comment to Condition.Code.
-// TODO change condition codes to CamelCase strings. See issue:
-//  https://github.com/kubernetes-sigs/multi-tenancy/issues/500
 const (
-	CritParentMissing Code = "CRIT_PARENT_MISSING"
-	CritParentInvalid Code = "CRIT_PARENT_INVALID"
-	CritAncestor      Code = "CRIT_ANCESTOR"
-	HNSMissing        Code = "HNS_MISSING"
-	CannotUpdate      Code = "CANNOT_UPDATE_OBJECT"
-	CannotPropagate   Code = "CANNOT_PROPAGATE_OBJECT"
+	CritParentMissing Code = "CritParentMissing"
+	CritParentInvalid Code = "CritParentInvalid"
+	CritAncestor      Code = "CritAncestor"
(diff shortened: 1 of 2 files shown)


jiachengxu:
/assign @jimbugwadia
I noticed that the condition codes in the HNC api folder are not unified, for example, in the `HNCConfigurationCode`: https://github.com/jiachengxu/multi-tenancy/blob/master/incubator/hnc/api/v1alpha1/hnc_config.go#L48-L50,
The code strings start with lower case letter, should I  change them to capitalize letter?

adrianludwin:
> I noticed that the condition codes in the HNC api folder are not unified, for example, in the `HNCConfigurationCode`: https://github.com/jiachengxu/multi-tenancy/blob/master/incubator/hnc/api/v1alpha1/hnc_config.go#L48-L50,
> The code strings start with lower case letter, should I change them to capitalize letter?

That would be great, thanks!

adrianludwin:
Can you please amend your existing commit to include any new changes (rather than have multiple commits in one PR), and also edit the commit description to say how you tested it? Thanks!

adrianludwin:
Eg see #579 and #580 for examples of what a testing paragraph looks like.

jiachengxu:
@adrianludwin Can you please take a look again? I have updated the PR and also added a "tested" paragraph in the PR description.

adrianludwin:
Thanks! FYI that's not really the kind of testing we prefer (though this is a low-risk change so I'll accept it anyway). For UI changes like this, we usually want to see a manual end-to-end test, such as:

* Install HNC on a cluster
* Create a hierarchy
* Delete the parent, causing a MissingParent condition
* Using `kubectl get -n child hierarchyconfigurations hierarchy -o yaml`, verify that the condition is stored using camel case.
* Using `kubectl hns describe child`, verify that the condition is printed using camel case.

Without this test, you're not sure whether you've fully solved the problem. Maybe there was some conversion between the constants and the YAML that you didn't know about? Maybe the kubectl plugin is doing its own conversions?

In this case, as I said, the risk is low, but there's usually nothing like manually verifying what the end-user will see.

PR #595, opened 2020-04-02
Add a kubectl command to set the HNC configuration of a specific type

sophieliu15:
This PR adds a kubectl command to set the HNC configuration of a specific type. Specifically it implements the following features:
- If the type exists in the spec, it will update the configuration of the type. 
- If the type does not exist in the spec, it will add the configuration in the spec.

Example usage: `kubectl hns config --apiVersion v1 --kind Secret ignore`

Issue: #454

### DIFF
--- a/incubator/hnc/pkg/kubectl/configset.go
+++ b/incubator/hnc/pkg/kubectl/configset.go
@@ -1,0 +1,69 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubectl
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+var setTypeCmd = &cobra.Command{
+	Use: fmt.Sprintf("set-type --apiVersion X --kind Y <%s|%s|%s>",
+		api.Propagate, api.Remove, api.Ignore),
+	Short: "Sets the HNC configuration of a specific resources type",
+	Example: fmt.Sprintf("  # Set configuration of a core type\n" +
+		"  kubectl hns config set-type --apiVersion v1 --kind Secret ignore\n\n" +
+		"  # Set configuration of a custom type\n" +
+		"  kubectl hns config set-type --apiversion stable.example.com/v1 --kind CronTab propagate"),
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		mode := api.SynchronizationMode(args[0])
+		flags := cmd.Flags()
+		apiVersion, _ := flags.GetString("apiVersion")
+		kind, _ := flags.GetString("kind")
+		config := client.getHNCConfig()
+
+		exist := false
+		for i := 0; i < len(config.Spec.Types); i++ {
+			t := &config.Spec.Types[i]
+			if t.APIVersion == apiVersion && t.Kind == kind {
+				t.Mode = mode
+				exist = true
+				break
+			}
+		}
+
+		if !exist {
+			config.Spec.Types = append(config.Spec.Types,
+				api.TypeSynchronizationSpec{
+					APIVersion: apiVersion,
+					Kind:       kind,
+					Mode:       mode,
+				})
+		}
+
+		client.updateHNCConfig(config)
+	},
+}
+
+func newSetTypeCmd() *cobra.Command {
+	setTypeCmd.Flags().String("apiVersion", "", "API version of the kind")
+	setTypeCmd.Flags().String("kind", "", "Kind to be configured")
+	return setTypeCmd
+}
--- a/incubator/hnc/pkg/validators/hierarchy.go
+++ b/incubator/hnc/pkg/validators/hierarchy.go
@@ -381,13 +381,26 @@ func deny(reason metav1.StatusReason, msg string) admission.Response {
 			},
 		}}
 	} else {
-		// metav1.StatusReasonInvalid shows the custom message in the Details field instead of
-		// Message field of metav1.Status.
+		// We need to set the custom message in both Details and Message fields.
+		//
+		// When manipulating the HNC configuration object via kubectl directly, kubectl
+		// ignores the Message field and displays the Details field if an error is
+		// StatusReasonInvalid (see implementation here: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/util/helpers.go#L145-L160).
+		//
+		// When manipulating the HNC configuration object via the hns kubectl plugin,
+		// if an error is StatusReasonInvalid, only the Message field will be displayed. This is because
+		// the Error method (https://github.com/kubernetes/client-go/blob/cb664d40f84c27bee45c193e4acb0fcd549b0305/rest/request.go#L1273)
+		// calls FromObject (https://github.com/kubernetes/apimachinery/blob/7e441e0f246a2db6cf1855e4110892d1623a80cf/pkg/api/errors/errors.go#L100),
+		// which generates a StatusError (https://github.com/kubernetes/apimachinery/blob/7e441e0f246a2db6cf1855e4110892d1623a80cf/pkg/api/errors/errors.go#L35) object.
+		// *StatusError implements the Error interface using only the Message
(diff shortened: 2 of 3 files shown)


sophieliu15:
/hold

Will make this PR on hold till #592 is merged.

sophieliu15:
/assign @adrianludwin 
/assign @yiqigao217 
/cc  @adrianchung 

Thanks for reviewing my previous PR. I will resolve comments in that PR first and ping you again once that one is merged.

incubator/hnc/pkg/kubectl/configset.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+var setTypeCmd = &cobra.Command{
+	Use: fmt.Sprintf("set-type <%s|%s|%s>", string(api.Propagate), string(api.Remove),
```
adrianludwin, line 27: Is the `string(x)` cast really needed?

Also, the usage should be something more like `set-type --kind X --apiVersion Y <propagate|remove|ignore>`, right?

incubator/hnc/pkg/kubectl/configset.go
```diff
@@ -0,0 +1,65 @@
...
+					Mode:       mode,
+				})
+		}
+
+		client.updateHNCConfig(config)
+	},
+}
+
+func newSetTypeCmd() *cobra.Command {
+	setTypeCmd.Flags().String("apiVersion", "", "API version of the kind")
```
adrianludwin, line 62: Some examples here might be nice, e.g. "v1" for core objects or "example.com/v1beta1" for a contributed type.

incubator/hnc/pkg/kubectl/configset.go
sophieliu15: It doesn't look like we need the `string(x)` cast. Removed them. Changed to the usage to contain the flags.

incubator/hnc/pkg/kubectl/configset.go
sophieliu15: Thanks for the suggestion! Done. PTAL.

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -382,13 +382,16 @@ func deny(reason metav1.StatusReason, msg string) admission.Response {
 			},
 		}}
 	} else {
-		// metav1.StatusReasonInvalid shows the custom message in the Details field instead of
-		// Message field of metav1.Status.
+		// We need to set the custom message in both Message and Details fields.
+		// When interacting via the kutectl plugin, the custom message
+		// in the Message field will be shown when an error occurs. When interacting
+		// without the kubectl plugin, the custom message in the Details field will be shown.
```
adrianludwin, line 388: Eh? That's odd. Is this behaviour documented anywhere? Or is our plugin not treating the Details field properly?

incubator/hnc/pkg/kubectl/configset.go
```diff
@@ -0,0 +1,67 @@
...
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+var setTypeCmd = &cobra.Command{
+	Use: fmt.Sprintf("set-type --apiVersion X --kind Y <%s|%s|%s>",
+		api.Propagate, api.Remove, api.Ignore),
+	Short: "Sets the HNC configuration of a specific type",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
```
adrianludwin, line 31: If you type in `kubectl help create role`, you get a lot of examples. Can we add some examples here as well? I don't know if they belong in the "Use" field or somewhere else (e.g., is ther a "Long" help field?).

incubator/hnc/pkg/validators/hierarchy.go
sophieliu15: I don't think it is related to our plugin. When tracing back the call tree of [Error()](https://github.com/kubernetes-sigs/multi-tenancy/blob/f6a9f3e724d0ec5f9bbbc5a6e9351bc7a40719c8/incubator/hnc/pkg/kubectl/root.go#L150), it [calls ](https://github.com/kubernetes/client-go/blob/cb664d40f84c27bee45c193e4acb0fcd549b0305/rest/request.go#L1291) [FromObject](https://github.com/kubernetes/apimachinery/blob/7e441e0f246a2db6cf1855e4110892d1623a80cf/pkg/api/errors/errors.go#L100), which generates a [StatusError](https://github.com/kubernetes/apimachinery/blob/7e441e0f246a2db6cf1855e4110892d1623a80cf/pkg/api/errors/errors.go#L35) from `metav1.Status` (this is the type of [AdmissionResponse.Result](https://github.com/kubernetes/api/blob/464a0b5499224886bb4a6eeb41b33b79be9af1c8/admission/v1beta1/types.go#L127)). The `StatusError` implements the Error interface using [Message](https://github.com/kubernetes/apimachinery/blob/7e441e0f246a2db6cf1855e4110892d1623a80cf/pkg/api/errors/errors.go#L49) field from the status. IIUC, that's why when we print `Error()` it only prints the `Message` field.

incubator/hnc/pkg/kubectl/configset.go
sophieliu15: Sure! I added two examples in the [Example](https://github.com/spf13/cobra/blob/138b98f39dedc35ae01b827fb9bceacaf30b0880/command.go#L56) field and removed the examples in the flag definitions. Feel free to take a look.

incubator/hnc/pkg/kubectl/configset.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+var setTypeCmd = &cobra.Command{
+	Use: fmt.Sprintf("set-type --apiVersion X --kind Y <%s|%s|%s>",
+		api.Propagate, api.Remove, api.Ignore),
+	Short: "Sets the HNC configuration of a specific type",
+	Example: fmt.Sprintf("  # Set configuration of a core type\n" +
```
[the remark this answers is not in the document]
adrianludwin: Oh, I just thought there might be something special for `Example` because it's a builtin feature of Cobra. But that's fine, thanks!

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: No, I don't think that's need, but can you please update this comment to include everything we've discussed? Because it is super-confusing. And don't try to write a short comment, make it as long and rambling as you like (e.g. https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/pkg/reconcilers/hierarchy_config.go#L306) - give any future reader as much of a chance as you can that they'll be able to figure out why we did this.

sophieliu15: Thanks for the suggestion! I added a detailed the comment here. PTAL. I also added code links in the comment, feel free to let me know if it makes the comment less readable or any suggestions to improve the readability. Thank you!

adrianludwin: Perfect :)

sophieliu15:
I changed the commit and PR message to refer to the correct issue (454 instead of 545).

incubator/hnc/pkg/kubectl/configset.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/spf13/cobra"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+var setTypeCmd = &cobra.Command{
+	Use: fmt.Sprintf("set-type --apiVersion X --kind Y <%s|%s|%s>",
+		api.Propagate, api.Remove, api.Ignore),
+	Short: "Sets the HNC configuration of a specific type",
```
rjbez17, line 29: `type` in this context isn't that obvious IMO. Maybe `resources type` is better? Thoughts?

sophieliu15: Hi Ryan! 

Thanks for the suggestion! I feel `set-resources-type` might be a little bit long (e.g., the whole command will be like `kubectl hns config set-resources-type [flags] propagate`) but I don't have a better suggestion about the naming. Are there other kinds of "types" that users might be confused with "resource types"? 

I don't have preference between `type` and `resources-type`. If you think `resources-type` is not too long, we can go for it. :) 

@adrianludwin  WDYT?

rjbez17: I was referring just to the  short description, not the actual command names. I agree it'd be too long for the name.

sophieliu15: Cool! Thanks for the clarification. Changed to "resources type" in the short description.

adrianludwin: This lgtm!

PR #602, opened 2020-04-03
Implement hns validation webhook

### LINKED ISSUE #458: HNC: Implement validation webhook for HierarchicalNamespace (HNS)
This is part of #457 

The webhook should validate:
* The parent namespace is allowed to create subnamespaces.
* The HNS name is not conflict with currently existing namespaces.

yiqigao217:
Add HierarchicalNamespace and Namespace webhooks. They will allow all
requests from HNC SA.

HNS webhook will deny:
1) creating hns in a forbidden (excluded) namespace;
2) creating hns with an existing namespace name;
3) deleting hns if the owned namespace doesn't allow cascading deletion.

NS webhook will deny:
1) deleting ns if it's owned, no matter if it allows cascading deletion,
since users should delete the hns instance in the owner instead;
2) deleting ns if it has owned ns and it doesn't allow cascading
deletion.

Tested with new unit tests and manually tested on GKE cluster to see if
it works as expected and no conflict with cascading deletion implemented
earlier.

Fixes #458 
Part of #457


yiqigao217:
/assign @adrianludwin 
/assign @sophieliu15 
cc @adrianchung

incubator/hnc/pkg/validators/hierarchical_namespace.go
```diff
@@ -0,0 +1,122 @@
...
+	v.Forest.Lock()
+	defer v.Forest.Unlock()
+
+	pnm := req.hns.Namespace
+	cnm := req.hns.Name
+	cns := v.Forest.Get(cnm)
+
+	if req.op == v1beta1.Create {
+		if reconcilers.EX[pnm] {
+			msg := fmt.Sprintf("The namespace %s is not allowed to create subnamespaces. Please create hns in a different namespace.", pnm)
```
sophieliu15, line 73: Maybe use the full name instead of acronym? Or jut use "subnamespaces" as the first sentance.

incubator/hnc/pkg/validators/hierarchical_namespace.go
```diff
@@ -0,0 +1,122 @@
...
+			log.V(1).Info(msg)
+			return deny(metav1.StatusReasonForbidden, msg)
+		}
+
+		if cns.Exists() {
+			msg := fmt.Sprintf("The requested namespace %s already exists. Please use a different name.", cnm)
+			log.V(1).Info(msg)
```
sophieliu15, line 80: Do we need to log the message here? IIUC, it will be logged in line 56?

incubator/hnc/pkg/validators/hierarchical_namespace.go
```diff
@@ -0,0 +1,122 @@
...
+		return nil, err
+	}
+
+	return &hnsRequest{
+		hns: hns,
+		op:  in.Operation,
+	}, nil
+}
+
+func (v *HierarchicalNamespace) InjectDecoder(d *admission.Decoder) error {
```
adrianchung, line 119: Is this function used in this PR? I'd expect to see a test for something like this, but don't see anything associated with it -- either that or as a part of an interface definition. Am I missing some understanding here?

adrianludwin: You mean `InjectDecoder`? It's part of the boilerplate code that's called by `controller-runtime`, I'm pretty sure nothing will compile if it's not there.

incubator/hnc/pkg/validators/namespace.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (v *Namespace) checkForest(req *nsRequest) admission.Response {
+	v.Forest.Lock()
+	defer v.Forest.Unlock()
+
+	ns := v.Forest.Get(req.ns.Name)
+
+	// Check whether the deleting namespace has owned namespaces if it doesn't allow cascading deletion.
+	if req.op == v1beta1.Delete && !ns.AllowsCascadingDelete() {
```
adrianludwin, line 79: This might be a bit easier to read if you had more early exits:

```
if req.op != v1beta1.Delete {
  return allow("") // we only care about deletions
}

if ns.AllowsCascadingDelete {
  return allow("") // go ahead, make my day
}

// Check to see if there are any HNSes that can't be deleted
...
```
```

File: incubator/hnc/pkg/validators/namespace.go [hunk 5]
```diff
@@ -0,0 +1,121 @@
...
+		hnses := []string{}
+		for _, cnm := range ns.ChildNames() {
+			cns := v.Forest.Get(cnm)
+			if cns.IsOwned {
+				hnses = append(hnses, cnm)
+			}
```
adrianludwin, line 85: Shouldn't this be `if cns.IsOwned && !cns.AllowsCascadingDelete()`? If so, maybe change `hnses` to `cantDelete` or something like that, to make it clear why things get added to that list.

If I'm correct, and this is a bug, maybe add a unit test for this too?

incubator/hnc/pkg/validators/namespace.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		if len(hnses) != 0 {
+			msg := fmt.Sprintf("The namespace has owned namespace(s): %s. Please set allowCascadingDelete flag to true first.", hnses)
```
adrianludwin, line 88: Maybe `Please set allowCascadingDelete first either in the parent namespace or in all the owned children`?

incubator/hnc/pkg/validators/hierarchical_namespace_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	tests := []struct {
+		name string
+		op   v1beta1.Operation
+		pnm  string
+		cnm  string
+		fail bool
+	}{
+		{name: "ok-create", op: v1beta1.Create, pnm: "foo", cnm: "brumpf"},
+		{name: "ok-delete", op: v1beta1.Delete, pnm: "foo", cnm: "baz"},
```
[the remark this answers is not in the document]
yiqigao217: FYI updated in line 21 by `createOwnedNamespace()`.

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: The documentation of the interface DecoderInjector is here: https://godoc.org/sigs.k8s.io/controller-runtime/pkg/webhook/admission#DecoderInjector.

incubator/hnc/pkg/validators/namespace.go
yiqigao217: Yes good catch it's a bug. Say the owner doesn't allowCD but all its owned NS allows, then it should return `allow()`. Good catch on this corner case, will add the unit test on that and manual test it too.

incubator/hnc/pkg/validators/hierarchical_namespace.go
adrianchung: I tried just deleting `InjectDecoder` and `InjectClient` methods out of .pkg/validators/hierarchy.go as a test and the binary still compiles and deploys. I'm not one to argue with the compiler. As an experiment can you see what'd happen if we took this out and see if there's any side effects? I don't want to block the review or anything -- you're just likely closest to trying this out!

adrianludwin: Hmm. Are you sure the webhook is running correctly after this? Without `InjectDecoder`, I'm not sure how the `decoder` field is set, which means it should cause a nil pointer exception the first time we try to decode something. Perhaps it just looks fine because you never created a HierarchicalNamespace object, but as soon as you do, HNC will crash?

I would have _thought_ that `InjectDecoder` was checked at compile time, but perhaps it's only checked at runtime with a [type switch](https://tour.golang.org/methods/16).

incubator/hnc/pkg/validators/hierarchical_namespace.go
```diff
@@ -0,0 +1,119 @@
...
+// HierarchicalNamespaceServingPath is where the validator will run. Must be kept in sync with the
+// kubebuilder markers below.
+const (
+	HierarchicalNamespaceServingPath = "/validate-hnc-x-k8s-io-v1alpha1-hierarchicalnamespaces"
+)
+
+// Note: the validating webhook FAILS CLOSE. This means that if the webhook goes down, all further
+// changes are forbidden.
+//
+// +kubebuilder:webhook:path=/validate-hnc-x-k8s-io-v1alpha1-hierarchicalnamespaces,mutating=false,failurePolicy=fail,groups="hnc.x-k8s.io",resources=hierarchicalnamespaces,verbs=create;update;delete,versions=v1alpha1,name=hierarchicalnamespaces.hnc.x-k8s.io
```
jiachengxu, line 26: why the `update` is needed? In your `Handle` method, it only handles `create` and `delete` operations.

adrianludwin: Nice catch! We should probably reject any updates to HNSes (it doesn't have any real fields anyway), or else take `update` out of this list.

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: Thanks for bringing this up! I am also curious so I tested with removing the `InjectDecoder` in namespace webhook and `make deploy`. I got errors when trying to delete a namespace. The errors was the same as when I had dereference of an nil pointer similar to #603 .

incubator/hnc/pkg/validators/namespace.go
yiqigao217: Done, I put first two `if`s into one.

incubator/hnc/pkg/validators/namespace.go
yiqigao217: Added new unit tests and manually tested with setting `allowCascadingDelete`s (the webhook and cascading deletion all worked as expected).

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: Good catch, thanks! Removed.

incubator/hnc/pkg/validators/namespace.go
yiqigao217: Done, I also added a list of owned namespaces with `allowCascadingDelete` unset.

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: I manually tested with `update` removed. Seems to reject all `update` requests, we need to have it listed here and in `Handle` reject all `update` requests.

adrianludwin: How did you try to update it? You mean with labels and stuff? Yeah, perhaps we should allow those.

I'm surprised that an action that we _didn't_ list as part of the validator would be blocked. What was the error message with `update` removed from the webhook config?

yiqigao217:
PTAL, I updated the namespace webhook to check if all owned namespaces allowCD and added the new unit tests. I also updated the HNS webhook to deny all `update` requests and added one unit test case. @adrianludwin , @jiachengxu

incubator/hnc/pkg/validators/hierarchical_namespace.go
adrianludwin: Note that `update` is still present here.

incubator/hnc/pkg/validators/hierarchical_namespace.go
```diff
@@ -0,0 +1,123 @@
...
+// based on the current in-memory state of the forest.
+func (v *HierarchicalNamespace) handle(req *hnsRequest) admission.Response {
+	v.Forest.Lock()
+	defer v.Forest.Unlock()
+
+	pnm := req.hns.Namespace
+	cnm := req.hns.Name
+	cns := v.Forest.Get(cnm)
+
+	if req.op == v1beta1.Update {
```
adrianludwin, line 71: Maybe restructure this as a switch statement, with any unexpected op resulting in `log.Error`? We should let unexpected ops through though, since it's probably just a webhook misconfiguration.

I'm thinking we _should_ allow `Update` though in case people want to put labels or annotations onto them. You can't change the namespace or name, which are the only properties we care about anyway.

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: > How did you try to update it? You mean with labels and stuff? Yeah, perhaps we should allow those.

I used `kubectl edit hns child -n parent`
> I'm surprised that an action that we _didn't_ list as part of the validator would be blocked. What was the error message with `update` removed from the webhook config?

The action not listed will be `allowed`, not blocked. When `update` removed, I could successfully edit the hns instance.

incubator/hnc/pkg/validators/hierarchical_namespace.go
adrianludwin: Why didn't you need the decoder when trying to _create_ a namespace?

incubator/hnc/pkg/validators/hierarchical_namespace.go
adrianludwin: Ah, I see. I think we should allow them to let labels and annotations be added, wdyt?

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: I was just testing if it would work if the `InjectDecoder` is removed from the code. And it turned out it will fail, so we need it.

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: But that's by HNC SA, right? Not by users? We only deny `update` if it's not from SA.

incubator/hnc/pkg/validators/hierarchical_namespace.go
adrianludwin: 👍

incubator/hnc/pkg/validators/hierarchical_namespace.go
adrianludwin: No, if a user wants to go add a couple of labels to the HNS objects (_not_ the namespaces), I don't see why we'd stop them.

yiqigao217: Got you. Sure, then I just need to remove the `update` verb and clean it up in `handle`.

incubator/hnc/pkg/validators/hierarchical_namespace.go
yiqigao217: As discussed above, i will remove all `update` related code since we will allow `update` requests.

yiqigao217:
`update` verb is removed in HNS webhook and the `update` request handling and the unit test are all removed. @adrianludwin

PR #603, opened 2020-04-03
Fix dereference of a possibly nil pointer

yiqigao217:
The decoded could possibly be empty, e.g. when the request operation is
DELETE, the decoded will be empty.

Tested by 'make test'.


yiqigao217:
/assign @adrianludwin 
/assign @sophieliu15 
cc @adrianchung

PR #606, opened 2020-04-03
Add a kubectl command to delete the HNC configuration of a specific type

sophieliu15:
This PR adds a kubectl command to delete the HNC configuration of a specific type. 

Example usage: `kubectl hns config delete-type --apiVersion v1 --kind Secret `

Issue: #454

### DIFF
--- a/incubator/hnc/pkg/kubectl/configdelete.go
+++ b/incubator/hnc/pkg/kubectl/configdelete.go
@@ -1,0 +1,63 @@
+/*
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kubectl
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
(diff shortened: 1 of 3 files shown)


sophieliu15:
/assign @adrianludwin 
/assign @yiqigao217 
cc: @adrianchung

incubator/hnc/pkg/kubectl/configdelete.go
```diff
@@ -0,0 +1,61 @@
...
+		kind, _ := flags.GetString("kind")
+		config := client.getHNCConfig()
+
+		i := 0
+		for _, t := range config.Spec.Types {
+			if t.APIVersion != apiVersion || t.Kind != kind {
+				config.Spec.Types[i] = t
+				i++
+			}
+		}
```
adrianludwin, line 43: I feel like the more Go-idiomatic way of doing this would be:

```
newTypes := []api.Whatever{}
found := false
for _, t := range config.Spec.Types {
  if t.APIVersion == apiVersion && t.Kind == kind {
    found := true
  } else {
    newTypes = append(newTypes, t)
  }
}
if !found { ... error ... }
config.Spec.Types = newTypes
```

The current way has a "loop" variable that isn't really incremented all the time plus you're modifying a list as you step over it, which I'm pretty sure works fine but always makes me closely re-read it to make sure I'm not missing something important.

sophieliu15: Thanks so much for the suggestion and explanation! Changed to more Go-idiomatic way.

sophieliu15:
I changed the commit and PR message to refer to the correct issue (454 instead of 545).

sophieliu15:
> /lgtm
> /approve
> 
> @sophieliu15 does this need to merge after #595 ?

Hi @rjbez17 ! No it doesn't need to be merged after #595. Both are independent.

PR #610, opened 2020-04-04
[Incubator][VC] Using the signature of the tenant apiserver certificate to identify the request sender

charleszheng44:
Using the signature of the tenant apiserver certificate to identify the sender of a request received by vn-agent

1. add the annotation of the certificate's signature to VC CR after it
is created.
2. when receiving a request, vn-agent extracts the signature from the
client certificate and find the VC that has the same signature.
4. implement the `GetClusterCertificate` function for
`master_provisioner_native`
3. add RBAC rights to vn-agent-role for accessing virtualcluster CR

### DIFF
--- a/incubator/virtualcluster/pkg/controller/constants/constants.go
+++ b/incubator/virtualcluster/pkg/controller/constants/constants.go
@@ -1,0 +1,5 @@
+package constants
+
+const (
+	AnnoX509SignatureBase64 = "tenancy.x-k8s.io/apiserver.sig"
+)
--- a/incubator/virtualcluster/config/setup/all_in_one.yaml
+++ b/incubator/virtualcluster/config/setup/all_in_one.yaml
@@ -338,6 +338,14 @@ rules:
   - get
   - list
   - create
+- apiGroups:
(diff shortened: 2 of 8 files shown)


incubator/virtualcluster/pkg/controller/constants/constants.go
```diff
@@ -0,0 +1,5 @@
+package constants
```
zhuangqh, line 1: please add the file header

zhuangqh:
lgtm overall

the kube-apiserver certificate is the same as kubelet-client certificate in minikube cluster, right?

Fei-Guo:
> lgtm overall
> 
> the kube-apiserver certificate is the same as kubelet-client certificate in minikube cluster, right?

It seems not to be the case. We need to find some other methods.

charleszheng44:
> lgtm overall
> 
> the kube-apiserver certificate is the same as kubelet-client certificate in minikube cluster, right?

Yes. But if we use cloud mode, the `kube-apiserver` certificate and the `kubelet-client-certificate` are different.

PR #609, opened 2020-04-04
Rename kubectl-hierarchical_namespaces to kubectl-hns

adrianchung:
Issue #594.

TESTED: ran `make release` all the way to the point to publish to
k8s-staging-multitenancy, but I didn't have permissions to push there

Issue #601, opened 2020-04-03
HNC: kubectl plugin does not show `hierarchical-namespaces` in the Usage section


sophieliu15:
`hierarchical-namespaces` is added in the Usage section in the [root command](https://github.com/kubernetes-sigs/multi-tenancy/blob/2946d71727d081f8be651fc86d68b948414bbabc/incubator/hnc/pkg/kubectl/root.go#L59). However, when typing `kubectl hns`, the output shows:

```
Manipulates hierarchical namespaces provided by HNC

Usage:
  kubectl [command]

Available Commands:
  config      Manipulates the HNC configuration
  create      Creates a subnamespace under the given parent.
  ...
```

Ideally, the Usage section should be `kubectl hierarchical-namespaces [command]`

This also impacts the usage section of subcommands. For example, when typing `kubectl hns create`, the usage section is:

```
Usage:
  kubectl create -n PARENT CHILD [flags]
```

Ideally, it should be `kubectl hierarchical-namespaces create -n PARENT CHILD [flags]`

P.S. We should probably also change `kubectl hierarchical-namespaces` to `kubectl hns` once we fix the issue described above.

sophieliu15:
cc:
@adrianludwin 
@yiqigao217 
@adrianchung

sophieliu15:
This is fixed in #615

PR #615, opened 2020-04-06
Change the Use field of the root command to "kubectl-hns" for kubectl plugin 

sophieliu15:
This PR changes the `Use` field of the root command of kubectl plugin from `kubectl hierarchical-namespaces` to `kubectl-hns` because:
- We switched from "kubectl-hierarchical_namespaces" to "kubectl-hns" in #609 
- Only the first word of the `Use` field will be displayed in the "Usage" section of a root command, we set it to "kubectl-hns" here so that both "kubectl" and "hns" will be shown in the "Usage" section.

Issue: #601


sophieliu15:
/assign @adrianludwin 
@adrianchung

PR #616, opened 2020-04-06
Finalizer removed from tenant namespace controller

shivi28:
This is done because resources were in terminating stage for some time even after deletion.

reported correctly.


Fei-Guo:
The problem is fixed by #627

PR #618, opened 2020-04-07
[Incubator][VC]Restructure syncer namespace mock test

Fei-Guo:
This change restructures the mock test so that the RunDownwardSync method can be reused by
the mock tests of other resources.

To achieve this, this change makes a few abstractions in the controller creation codepath. 
Basically, the manager.Controller interface is used in the util.RunDownwardSync method instead of a specific controller implementation, making the mock test infra code reusable for all synced resources.


zhuangqh:
great job

PR #619, opened 2020-04-07
Update validators to respect excluded namespaces

### LINKED ISSUE #607: HNC: Add exclusions to hierarchy validation webhook
The hierarchy validator should respect the namespaces that are [excluded](https://github.com/adrianchung/multi-tenancy/blob/ad111630877a2a46498a5efc8c39367bab495696/incubator/hnc/pkg/reconcilers/setup.go#L15). 

The webhook will deny:

Create a namespace with a parent in EX
Reparent a child namespace to a parent in EX

adrianchung:
Fixes #607. Parent namespaces in the excluded namespaces should be
denied. Objects in parent namespaces should be ignored.

### DIFF
--- a/incubator/hnc/pkg/validators/hierarchy.go
+++ b/incubator/hnc/pkg/validators/hierarchy.go
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/config"
 	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
 )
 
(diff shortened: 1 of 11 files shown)


adrianchung:
/assign @adrianludwin 
/assign @yiqigao217 
cc @sophieliu15

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -141,6 +142,11 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
 		return allow("parent unchanged")
 	}
 
+	if reconcilers.EX[newParent.Name()] {
```
adrianludwin, line 145: Do you mind moving EX out of `reconcilers` and into `pkg/config`? It doesn't make sense for the validators to import all the reconcilers just to get this one map, IMO.

adrianchung: Absolutely! I left in default_config.go as it seemed most appropriate. I suppose if we wanted to expand on the exclusion concept we can find another home for it.

adrianludwin:
No, looks like there's a missed reference to the "old" copy of EX.

yiqigao217:
> LGTM after taking `EX` map out into config. Can you see the failed test? I suspect it's the flaky test in HNS test but couldn't see the failed test detail.

Oh, the error is ```vet: pkg/validators/hierarchical_namespace.go:72:18: EX not declared by package reconcilers
make: *** [Makefile:107: vet] Error 2
make: Leaving directory '/home/prow/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc'``` You will need to replace the pkg one more place.

adrianchung:
@yiqigao217 One moment. When I was working on this PR your PR with respect to the hierarchical_namespace.go validator didn't land so I didn't catch that in my PR. I'll get rebased with master and push within the next hour

adrianchung:
Ok, I've rebased from master and should be good to go now

yiqigao217:
> @yiqigao217 One moment. When I was working on this PR your PR with respect to the hierarchical_namespace.go validator didn't land so I didn't catch that in my PR. I'll get rebased with master and push within the next hour

Ah I see, right, it's the newly checked-in! Thanks for the explanation!

PR #624, opened 2020-04-07
Don't create HC in namespaces being deleted

adrianludwin:
I noticed while doing a cascading deletion that the HCR was trying to
write back the HC, because after it's deleted, it looks like we want to
sync it with the namespace's owner attribute. This was causing a lot of
errors in the log because you can't create an object in a namespace
that's being deleted. This change allows HNC to continue writing back
_existing_ HCs (e.g. to remove their finalizers) but will never write
back a _new_ HC to a namespace being deleted.

I also took out the code that allows the HCR to _create_ namespaces
since this is now owned entirely by the HNSR.

Tested: watched logs during cascading deletion before and after this
change. Before, there were tons of error messages (although it did
successfully delete the NS); after, the log was clean.

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchy_config.go
@@ -423,28 +423,34 @@ func (r *HierarchyConfigReconciler) enqueueAffected(log logr.Logger, reason stri
 }
 
 func (r *HierarchyConfigReconciler) writeInstances(ctx context.Context, log logr.Logger, oldHC, newHC *api.HierarchyConfiguration, oldNS, newNS *corev1.Namespace) (bool, error) {
-	ret := false
-	if updated, err := r.writeHierarchy(ctx, log, oldHC, newHC); err != nil {
+	isDeletingNS := !newNS.DeletionTimestamp.IsZero()
+	updated := false
+	if up, err := r.writeHierarchy(ctx, log, oldHC, newHC, isDeletingNS); err != nil {
 		return false, err
 	} else {
-		ret = ret || updated
+		updated = updated || up
 	}
 
-	if updated, err := r.writeNamespace(ctx, log, oldNS, newNS); err != nil {
+	if up, err := r.writeNamespace(ctx, log, oldNS, newNS); err != nil {
 		return false, err
 	} else {
-		ret = ret || updated
+		updated = updated || up
 	}
-	return ret, nil
+	return updated, nil
 }
 
-func (r *HierarchyConfigReconciler) writeHierarchy(ctx context.Context, log logr.Logger, orig, inst *api.HierarchyConfiguration) (bool, error) {
+func (r *HierarchyConfigReconciler) writeHierarchy(ctx context.Context, log logr.Logger, orig, inst *api.HierarchyConfiguration, isDeletingNS bool) (bool, error) {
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -423,28 +423,34 @@ func (r *HierarchyConfigReconciler) enqueueAffected(log logr.Logger, reason stri
...
+	updated := false
+	if up, err := r.writeHierarchy(ctx, log, oldHC, newHC, isDeletingNS); err != nil {
 		return false, err
 	} else {
-		ret = ret || updated
+		updated = updated || up
 	}
 
-	if updated, err := r.writeNamespace(ctx, log, oldNS, newNS); err != nil {
+	if up, err := r.writeNamespace(ctx, log, oldNS, newNS); err != nil {
```
yiqigao217, line 434: I am not sure if we still need to update namespace if `isDeletingNS == true`, since the only thing that could change on a namespace is tree labels. I am not sure if dropping the update of a deleting namespace's tree labels could cause any issue for hierarchical resource quota since it may use tree labels.

adrianludwin: I'm a little nervous about stopping this update. I agree that the labels could disappear from the namespace shortly before it gets deleted, but this should be transient and very rare.

Maybe we can file an issue to revisit?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -423,28 +423,34 @@ func (r *HierarchyConfigReconciler) enqueueAffected(log logr.Logger, reason stri
...
 func (r *HierarchyConfigReconciler) writeInstances(ctx context.Context, log logr.Logger, oldHC, newHC *api.HierarchyConfiguration, oldNS, newNS *corev1.Namespace) (bool, error) {
-	ret := false
-	if updated, err := r.writeHierarchy(ctx, log, oldHC, newHC); err != nil {
+	isDeletingNS := !newNS.DeletionTimestamp.IsZero()
+	updated := false
+	if up, err := r.writeHierarchy(ctx, log, oldHC, newHC, isDeletingNS); err != nil {
 		return false, err
 	} else {
-		ret = ret || updated
+		updated = updated || up
```
rjbez17, line 431: isn't `updated` always false at this point?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
adrianludwin: Ping - @yiqigao217 sgty?

incubator/hnc/pkg/reconcilers/hierarchy_config.go
adrianludwin: Yes, I just wanted it to look the same as the one below :)

rjbez17: Thats what I figured, thanks!

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: SGTM, filed #635 .

PR #626, opened 2020-04-07
Better flags, output and examples for hns-set

adrianludwin:
The main change here is to add --forbidCascadingDelete, as well as some
short flags (-p for --parent, -r for --root, -a for
--allowCascadingDelete and -f for --forbidCascadingDelete). It also adds
an "examples" section to the documentation and prints out fewer messages
if allowCD is unchanged.

Tested: manual testing for all the modified variants of 'set'; verified
that new example documentation looks good.

### DIFF
--- a/incubator/hnc/pkg/kubectl/set.go
+++ b/incubator/hnc/pkg/kubectl/set.go
@@ -27,25 +27,47 @@ import (
 
 //hcUpdates struct stores name of namespaces against type of flag passed
 type hcUpdates struct {
-	root                 bool
-	parent               string
-	allowCascadingDelete bool
+	root     bool
+	parent   string
+	allowCD  bool
+	forbidCD bool
 }
 
 var setCmd = &cobra.Command{
 	Use:   "set NAMESPACE",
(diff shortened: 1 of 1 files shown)


incubator/hnc/pkg/kubectl/set.go
```diff
@@ -130,8 +166,9 @@ func normalizeStringArray(in []string) []string {
 }
 
 func newSetCmd() *cobra.Command {
-	setCmd.Flags().Bool("root", false, "Turns namespace into root namespace")
-	setCmd.Flags().String("parent", "", "Parent namespace")
-	setCmd.Flags().Bool("allowCascadingDelete", false, "Allows cascading deletion of its self-serve subnamespaces.")
+	setCmd.Flags().BoolP("root", "r", false, "Turns namespace into root namespace")
```
yiqigao217, line 169: Nit not sure if it would be clearer to mention "removes current parent" for "root" flag.

adrianludwin: Updated the descriptions, ptal!

PR #631, opened 2020-04-08
Break cyclical dependency when tenant CR is deleted

Fei-Guo:
When tenant object is removed, all the tenantnamespace CRs should be deleted since their namepsace, aka, the tenant admin namespace is deleted. The Tenantnamespace reconciler needs to first retrieve the tenant list. The reconciler will keep reporting err since the tenant object has gone. Hence the finalizer of tenantnamespace CR can never be removed.

This change makes sure that if tenantnamespace CR is deleted and the tenant object is not found,
the finalizer should be removed so that the tenantnamespace CR can be removed from apiserver eventually.

A UT is added to test this scenario.

### DIFF
--- a/tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
+++ b/tenant/pkg/controller/tenantnamespace/tenantnamespace_controller.go
@@ -112,6 +112,20 @@ func (r *ReconcileTenantNamespace) updateNamespace(ns *corev1.Namespace, tenantA
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


rjbez17:
Is this related to https://github.com/kubernetes-sigs/multi-tenancy/pull/616?

Fei-Guo:
> Is this related to #616?

#616 should be closed. I notice this problem while fixing the problem that #616 tackles.

PR #632, opened 2020-04-09
Flush obsolete conditions on descendants

### LINKED ISSUE #605: HNC: CannotUpdate conditions might not be cleared in descendants when a hierarchy changes
See https://github.com/kubernetes-sigs/multi-tenancy/pull/597#discussion_r403301854

yiqigao217:
As described in issue 605, obsolete CannotUpdate condition was not
cleared on the descendants namespaces of a namespace who just changes
its parent. This commit changes the key of a CannotUpdate condition from
the local object to the source object if it exists. Therefore, we can
flush obsolete conditions on descendants by clearing the conditions by
ancestor namespaces, the same way as how we flush obsolete conditions on
ancestors.

Tested with "make test" and running the test script
"hack/test-issue-605.sh' seeing the CannotUpdate condition was gone
immediately in all 10 attempts. See [results](https://gist.github.com/yiqigao217/a99afd57f5a5e3084d90b56fad31c99d) and [logs](https://gist.github.com/yiqigao217/424eafae7ac4900753b432e7768543e0).

Fixes #605

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchy_config.go
@@ -296,10 +296,25 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
 // flushObsoleteObjectConditions looks for object conditions from objects that were in one of the
 // namespaces that's been removed as a descendant, and enqueues any affected namespaces.
 func (r *HierarchyConfigReconciler) flushObsoleteObjectConditions(log logr.Logger, ns, oldAns *forest.Namespace) {
-	// Create the set of all relevant names
+	// There should be no obsolete object conditions if the old parent is empty.
+	if oldAns == nil {
+		return
+	}
+
+	// To clear all descendants conditions introduced by ancestors, create a set of all ancestor names.
+	wasAnc := map[string]bool{}
+	for _, anc := range oldAns.AncestryNames(nil) {
+		wasAnc[anc] = true
+	}
+
+	// Remove conditions referring to its ancestors from all its descendants. Also create a set of
+	// all descendant names to clear ancestors conditions introduced by descendants later.
 	wasDesc := map[string]bool{ns.Name(): true}
 	for _, desc := range ns.DescendantNames() {
 		wasDesc[desc] = true
+		if r.Forest.Get(desc).ClearConditionsByNamespace(log, wasAnc) {
+			r.enqueueAffected(log, "cleared obsolete conditions", desc)
+		}
 	}
 
 	// Remove conditions referring to this namespace and its descendants from all old ancestors.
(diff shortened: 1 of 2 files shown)


incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -296,7 +296,25 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
...
+	}
+
+	// To clear all descendants conditions introduced by ancestors, create a set of all ancestor names.
+	wasAnc := map[string]bool{}
+	for _, anc := range oldAns.AncestryNames(nil) {
+		wasAnc[anc] = true
+	}
+
+	// Remove conditions referring to its ancestors from all its descendants.
+	for _, desc := range ns.DescendantNames() {
```
adrianludwin, line 311: You may as well build the `wasDesc` set in this loop, instead of having a separate loop

incubator/hnc/pkg/reconcilers/object.go
```diff
@@ -548,11 +550,16 @@ func (r *ObjectReconciler) setConditions(log logr.Logger, srcInst, inst *unstruc
 		}
 
 	case err != nil:
-		// There was an error updating this object; set a local condition.
+		// There was an error updating this object; set a condition.
 		msg := fmt.Sprintf("Could not %s: %s", act, err.Error())
-		if ns.SetCondition(ao, api.CannotUpdate, msg) {
+		if !hasPropagatedLabel(inst) {
```
adrianludwin, line 555: Can this happen? I feel like this would be a bug, we should never try to propagate a source object.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: sg, combined!

incubator/hnc/pkg/reconcilers/object.go
yiqigao217: It should never happen. If `!hasPropagatedLabel(inst)`, it will only return `actionNop` in our current logic. And `actionNop` won't cause any error. I will remove the `hasPropagatedLabel()` check here and have a comment that "we never take actions on source objects so it should never get into this error case"

yiqigao217:
> lgtm with a few nits. Can you please attach a before/after gist of the output of test-605 so I can see what the new conditions look like?

The [results](https://gist.github.com/yiqigao217/a99afd57f5a5e3084d90b56fad31c99d) and [logs](https://gist.github.com/yiqigao217/424eafae7ac4900753b432e7768543e0) are added to the PR msg too.

incubator/hnc/pkg/reconcilers/object.go
adrianludwin: sgtm. In the future I'd recommend doing something like `log.Error` in case something that _shouldn't_ ever happen actually _does_ (e.g. because someone else changes some logic in the future) but for now I think this is fine. Thanks!

PR #636, opened 2020-04-09
Fix a flaky test on changing namespace annotation

### LINKED ISSUE #560: HNC: Fix flaky test on changing the namespace annotation
Part of #457 

In`hierarchical_namespace_test.go` to test the case of 'Conflict', we want to change the owned namespace's owner annotation to trigger the 'Conflict' state. Most of the times, the tests will pass. However, sometimes it may fail because it was not able to change the namespace annotation.

The error I got is:
Operation cannot be fulfilled on hierarchyconfigurations.hnc.x-k8s.io "hierarchy": the object has been modified; please apply your changes to the latest version and try again

yiqigao217:
Instead of changing the annotation of the namespace created by the hns
instance, this PR creates a conflict hns instance with the name of an
already-existing namespace.

Tested by 'make test'.

Fixes #560

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
@@ -2,13 +2,12 @@ package reconcilers_test
 
 import (
 	"context"
-	"time"
 
-	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+
+	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
 )
 
 var _ = Describe("HierarchicalNamespace", func() {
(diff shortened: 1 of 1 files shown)


yiqigao217:
/assign @adrianludwin 
cc @sophieliu15 , @adrianchung

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -57,24 +56,14 @@ var _ = Describe("HierarchicalNamespace", func() {
...
-		time.Sleep(1 * time.Second)
-
-		// Clear the owner annotation and the HNS state should be set to "Conflict".
-		clearAnnotations(ctx, barName)
-		Eventually(getHNSState(ctx, fooName, barName)).Should(Equal(api.Conflict))
+		// Create "baz" namespace.
+		bazName := createNS(ctx, "baz")
+		// Create "baz" hns in "foo" namespace. Even "baz" namespace has already existed, we can
+		// still create the hns instance using HNC SA to bypass the webhook here. However, we
+		// should get conflict state on the newly created "baz" hns instance.
```
adrianchung, line 63: I see that the end result here is that we end up with an api conflict. However, I'm unclear how this ties into the test name because I don't see anything in here about annotations anymore.

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -57,24 +56,14 @@ var _ = Describe("HierarchicalNamespace", func() {
...
-		//	 https://github.com/kubernetes-sigs/multi-tenancy/issues/560
-		// Sleep 1 second to avoid updating the namespace instance too quickly.
-		time.Sleep(1 * time.Second)
-
-		// Clear the owner annotation and the HNS state should be set to "Conflict".
-		clearAnnotations(ctx, barName)
-		Eventually(getHNSState(ctx, fooName, barName)).Should(Equal(api.Conflict))
+		// Create "baz" namespace.
+		bazName := createNS(ctx, "baz")
+		// Create "baz" hns in "foo" namespace. Even "baz" namespace has already existed, we can
```
adrianchung, line 61: Nit.. grammar. "Even though "baz" namespace already exists"

adrianchung:
Less flaky tests = great :D Thanks for doing this Yiqi!

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
```diff
@@ -57,24 +56,14 @@ var _ = Describe("HierarchicalNamespace", func() {
 	})
 
 	It("should set the hns.status.state to Conflict if the namespace's owner annotation is wrong", func() {
```
sophieliu15, line 58: Maybe specify why the owner annotation is wrong, e.g., IIUC, 
"should set the hns.status.state to Conflict if a namespace of the same name already exists"

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
yiqigao217: Thanks will update it!

incubator/hnc/pkg/reconcilers/hierarchical_namespace_test.go
yiqigao217: Yes, for the hns instance perspective, the conflict is only when the annotation doesn't match. It could be the reason you mentioned or someone with higher privileges changes the owner annotation. Here I wanted to introduce a mismatch to trigger the conflict.

yiqigao217:
Thanks for the reviews! I overthought about it in my first commit. Now I just simplified the comments and it should be easy to explain or understand now. @sophieliu15 @adrianchung

adrianludwin:
Ugh, I can't Git today

/lgtm
/approve

PR #637, opened 2020-04-09
Update condition code documentation

### LINKED ISSUE #622: HNC: Inconsistent documentation for Condition.Code
[Condition.Code](https://github.com/kubernetes-sigs/multi-tenancy/blob/9f46e6aac110f62fcba9bc4d8d204e68d1760acd/incubator/hnc/api/v1alpha1/hierarchy_types.go#L130) does not contain documentation for [HNSMissing](https://github.com/kubernetes-sigs/multi-tenancy/blob/9f46e6aac110f62fcba9bc4d8d204e68d1760acd/incubator/hnc/api/v1alpha1/hierarchy_types.go#L44), [CannotUpdate](https://github.com/kubernetes-sigs/multi-tenancy/blob/9f46e6aac110f62fcba9bc4d8d204e68d1760acd/incubator/hnc/api/v1alpha1/hierarchy_types.go#L45), [CannotPropagate](https://github.com/kubernetes-sigs/multi-tenancy/blob/9f46e6aac110f62fcba9bc4d8d204e68d1760acd/incubator/hnc/api/v1alpha1/hierarchy_types.go#L46). It also has an extra documentation for [REQUIRED_CHILD_CONFLICT](https://github.com/kubernetes-sigs/multi-tenancy/blob/9f46e6aac110f62fcba9bc4d8d204e68d1760acd/incubator/hnc/api/v1alpha1/hierarchy_types.go#L126), which is not defined [here](https://github.com/kubernetes-sigs/multi-tenancy/blob/9f46e6aac110f62fcba9bc4d8d204e68d1760acd/incubator/hnc/api/v1alpha1/hierarchy_types.go#L40).

yiqigao217:
Add documentation on new codes and delete the obsolete codes.

Tested with "make test".

Fixes #622

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -42,8 +42,8 @@ const (
 	CritParentInvalid Code = "CritParentInvalid"
 	CritAncestor      Code = "CritAncestor"
 	HNSMissing        Code = "HNSMissing"
-	CannotUpdate      Code = "CannotUpdateObject"
 	CannotPropagate   Code = "CannotPropagateObject"
+	CannotUpdate      Code = "CannotUpdateObject"
 )
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
@@ -104,7 +104,7 @@ type Condition struct {
 	// shown below, but new values may be added over time. This field is always present in a
 	// condition.
 	//
-	// All codes that begin with the prefix `CRIT_` indicate that all HNC activities (e.g. propagating
+	// All codes that begin with the prefix `Crit` indicate that all HNC activities (e.g. propagating
 	// objects, updating labels) have been paused in this namespaces. HNC will resume updating the
 	// namespace once the condition has been resolved. Non-critical conditions typically indicate some
 	// kind of error that HNC itself can ignore, but likely indicates that the hierarchical structure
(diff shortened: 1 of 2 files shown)


yiqigao217:
@sophieliu15 @adrianchung please feel free to take a look and lmk if anything's not clear in the condition code documentation. Thank you Sophie to bring this up!

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -116,17 +116,22 @@ type Condition struct {
...
-	// - "CRIT_ANCESTOR": a critical error exists in an ancestor namespace, so this namespace is no
+	// - "CritAncestor": a critical error exists in an ancestor namespace, so this namespace is no
 	// longer being updated either.
 	//
-	// - "REQUIRED_CHILD_CONFLICT": this namespace has a required child, but a namespace of the same
-	// name already exists and is not a child of this namespace. Note that the condition is _not_
-	// annotated onto the other namespace; it is considered an error _only_ for the would-be parent
-	// namespace.
+	// - "HNSMissing": this namespace is an owned namespace (created by reconciling an hns instance),
+	// but the hns instance is missing in its owner (referenced in its owner annotation) or the owner
```
sophieliu15, line 127: Will there be more information about which case (i.e., hns instance missing or owner namespace missing) causes this problem when a namespace has this error condition?

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -116,17 +116,22 @@ type Condition struct {
...
+	// namespace is missing.
+	//
+	// - "CannotUpdateObject": this namespace has an error when updating a propagated object. The
+	// affected object will point to the source object.
+	//
+	// - "CannotPropagateObject": this namespace has an error when propagating a source object. The
+	// affected objects will point to all the propagated objects that have an error to update.
```
sophieliu15, line 134: > all the propagated objects that have an error to update.

I am wondering what's the difference between this code and the `CannotUpdateObject` above? From the documentation, they seem to be both about problems when updating propagated objects?

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: No, we only look for the hns instance. If it doesn't exist, this condition will be set, no matter the owner namespace exists or not. The two cases here are actually the two possible cases of a missing hns instance.

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: Yes, they are two conditions triggered by the same error. However, `CannotPropagateObject` is set on the namespace with the source object and `CannotUpdateObject` is set on the namespace with the propagated object.

adrianludwin: Maybe: "This object contains an object that couldn't be propagated to one or more of its descendants. The condition's affect objects will include a list of the copies that couldn't be updated."

yiqigao217: Thanks all, updated!

PR #646, opened 2020-04-14
[Incubator][VC]Introduce Patroller to abstract periodic checker

Fei-Guo:
This change introduces Patroller struct to abstract the current periodic checker mechanism used by all resource syncers. This is the prerequisite for implementing checker mock tests. 

A new method PatrollerDo is added to manager.Controller interface and the StartPeriodChecker method is renamed to StartPatrol.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/patrol/patroller.go
+++ b/incubator/virtualcluster/pkg/syncer/patrol/patroller.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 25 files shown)


incubator/virtualcluster/pkg/syncer/patrol/patroller.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/virtualcluster/pkg/syncer/reconciler"
+)
+
+type Patroller struct {
+	name string
+
+	Options
+}
+
+// Options are the arguments for creating a new UpwardController.
```
zhuangqh, line 35: s/UpwardController/Patroller

PR #668, opened 2020-04-22
[Incubator][VC] fix syncer command line option bug

charleszheng44:
fix invalid memory address error caused by empty '--super-master' command line option

### DIFF
--- a/incubator/virtualcluster/cmd/syncer/app/options/options.go
+++ b/incubator/virtualcluster/cmd/syncer/app/options/options.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
 	clientgokubescheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


incubator/virtualcluster/cmd/syncer/app/options/options.go
```diff
@@ -235,24 +236,25 @@ func getInClusterNamespace() (string, error) {
 // createClients creates a meta cluster kube client and a super master custer client from the given config and masterOverride.
 func createClients(config componentbaseconfig.ClientConnectionConfiguration, masterOverride string, timeout time.Duration) (clientset.Interface,
 	vcclient.Interface, clientset.Interface, error) {
-	if len(config.Kubeconfig) == 0 && len(masterOverride) == 0 {
-		klog.Warningf("Neither --kubeconfig nor --master was specified. Using in-cluster API client.")
+	if len(config.Kubeconfig) == 0 {
+		klog.Warningf("--super-master-kubeconfig was not specified. Using in-cluster API client.")
```
zhuangqh, line 240: the condition of this warning should be the same as the follow one.
or just delete it?

PR #679, opened 2020-04-28
Integrate GK cert as the default internal cert

### LINKED ISSUE #653: HNC: let HNC run without cert-manager
We've had some discussions at Google about HNC, and some of us were concerned by the runtime dependency on `cert-manager`. For example, we've already seen cases where different versions of `cert-manager` can break us - what happens if someone has the wrong version installed on their cluster? 

By contrast, Gatekeeper is a similar project to HNC, but has the optional ability to generate its own [self-signed certs for its webhooks](https://github.com/open-policy-agent/gatekeeper/blob/master/pkg/webhook/certs.go). We thought it would be nice to include this capability in HNC as well.

We had a chat with the `controller-runtime` folks about moving Gatekeeper's cert management there, but they're not really interested in maintaining it. For now, the best solution we can think of is simply to copy Gatekeeper's manager (which is a single code file, plus a test file) into HNC, and update it periodically with fixes. @yiqigao217 has already been working with Gatekeeper to get some (unrelated) fixes into their cert manager so she has a good understanding of how that code works.

Hopefully this is an interim solution until we move that file to a proper library location, but in the meantime, we think this will make HNC easier to use in more environments.

yiqigao217:
Fork the cert management from open-policy-agent/gatekeeper repo to serve
as the default internal cert management for HNC.

Add an "enable-internal-cert-management" flag to switch between using
the internal cert (integrated Gatekeeper cert) or the external cert
"cert-manager".

Update main.go to start the controller manager with internal certs first
and once the cert files are ready, add other controllers and webhooks if
webhook is enabled and internal certs are used.

Refactor the cert generation and validator startup into
pkg/validators/setup.go.

To use external cert-manager, update config/default/kustomization.yaml
and manager_auth_proxy_patch.yaml as instructed. Run 'make deploy-cm'
before 'make deploy'.

Tested on a GKE cluster. The manager restarts 0 times and the webhooks
are working as expected.

Fixes #653

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -43,7 +43,7 @@ all: test docker-build
 
 # Run tests
 test: build
-	go test ./api/... ./cmd/... ./pkg/... -coverprofile cover.out
+	go test ./api/... ./cmd/... ./pkg/... ./third_party/... -coverprofile cover.out
 
 # Builds all binaries (manager and kubectl) and manifests
 build: generate fmt vet manifests
@@ -128,15 +128,15 @@ endif
 # deleting the CRDs will cause all the existing hierarchy configs to be wiped
 # away and b) if we don't delete the deployment, a new image won't be pulled
 # unless the tag changes.
-deploy: deploy-prereq docker-push kubectl manifests
+deploy: docker-push kubectl manifests
 	-kubectl -n hnc-system delete deployment hnc-controller-manager
 	kubectl apply -f manifests/hnc-manager.yaml
 
 deploy-watch:
 	kubectl logs -n hnc-system --follow deployment/hnc-controller-manager manager
 
-# Installs prerequisites. See https://cert-manager.io/docs/installation/kubernetes/ for why we're not validating our YAML.
-deploy-prereq:
+# Installs cert-manager. See https://cert-manager.io/docs/installation/kubernetes/ for why we're not validating our YAML.
+deploy-cm:
 	@kubectl cluster-info
 	-kubectl create namespace cert-manager
 	kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v0.11.0/cert-manager.yaml
@@ -166,7 +166,7 @@ kind-reboot:
 # Creates a local kind cluster, destroying the old one if necessary. It's not
 # *necessary* to call this wih CONFIG=kind but it's not a bad idea either so
 # the correct manifests get created.
-kind-reset: kind-reboot deploy-prereq
+kind-reset: kind-reboot
 	@echo "If this didn't work, ensure you ran 'source devenv' to point kubectl at kind'"
 
 # Convenience target to deploy specifically for kind
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -32,7 +32,6 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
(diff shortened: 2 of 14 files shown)


incubator/hnc/Makefile
```diff
@@ -128,14 +135,14 @@ endif
...
-deploy: deploy-prereq docker-push kubectl manifests
+deploy: docker-push kubectl manifests
 	-kubectl -n hnc-system delete deployment hnc-controller-manager
 	kubectl apply -f manifests/hnc-manager.yaml
 
 deploy-watch:
 	kubectl logs -n hnc-system --follow deployment/hnc-controller-manager manager
 
-# Installs prerequisites. See https://cert-manager.io/docs/installation/kubernetes/ for why we're not validating our YAML.
+# Installs external certs before deploying HNC. See https://cert-manager.io/docs/installation/kubernetes/ for why we're not validating our YAML.
```
adrianludwin, line 145: Maybe change to deploy-cm, since that's now all it does? They're not really prereqs anymore

incubator/hnc/cmd/manager/main.go
```diff
@@ -131,6 +150,41 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Make sure certs are generated and valid if cert rotation is enabled.
```
adrianludwin, line 153: Maybe move this to its own function? main() is getting very long

incubator/hnc/config/default/kustomization.yaml
```diff
@@ -17,7 +17,8 @@ bases:
 - ../rbac
 - ../manager
 - ../webhook
-- ../certmanager
```
adrianludwin, line 20: Comment this out rather than remove it completely? Likewise below

incubator/hnc/config/oldDefaultExternalCert/kustomization.yaml
```diff
@@ -0,0 +1,67 @@
+# Adds namespace to all resources.
```
adrianludwin, line 1: Rather than duplicate all these files, for now I'd just comment out the relevant parts in the main one. That's what kubebuilder gives us in the first place.

incubator/hnc/config/default/kustomization.yaml
```diff
@@ -17,7 +17,10 @@ bases:
...
+# Add cert management. Only one of the ../internalcert and the
+# ../certmanager should be enabled.
```
adrianludwin, line 21: Also mention where `--enable-internal-cert-management` should be added or removed.

incubator/hnc/config/default/manager_auth_proxy_patch.yaml
```diff
@@ -26,3 +26,4 @@ spec:
         - "--leader-election-id=hnc-controller-leader-election-helper"
         - "--max-reconciles=10"
         - "--apiserver-qps-throttle=50"
+        - "--enable-internal-cert-management=true"
```
adrianludwin, line 29: Do we need this line in any other yaml file? Eg I think this list of arguments is included in another file too.

yiqigao217: I don't think so. I just double checked the PRs [1](https://github.com/kubernetes-sigs/multi-tenancy/commit/29c17f2757803f0e28e07e99ceaba0f1c59b57e2), [2](https://github.com/kubernetes-sigs/multi-tenancy/pull/377/commits/39010d8bb51978fc392e5e99b2c56e86292d5c14) for `max-reconciles` and `apiserver-qps-throttle` to confirm that they are only used here.

incubator/hnc/config/default/kustomization.yaml
yiqigao217: Good point! Done, thanks!

incubator/hnc/config/default/manager_auth_proxy_patch.yaml
adrianludwin: sgtm, thanks for checking.

PR #686, opened 2020-05-01
Reduce default cardinality of metrics

adrianludwin:
We got a few comments internally that we shouldn't be putting too many
tags on our metrics by default. I also doubt that detailed stats would
be that useful unless we're doing performance testing, so I made the
GroupKind object tags flag-enabled and disabled them by default

Tested: running HNC on StackDriver, I observed that by default,
GroupKind tags are not displayed on our metrics, but restarting HNC with
--suppress-object-tags=false brought them back.

### DIFF
--- a/incubator/hnc/pkg/stats/metrics.go
+++ b/incubator/hnc/pkg/stats/metrics.go
@@ -15,6 +15,12 @@ import (
 // ReportingInterval is the exporter reporting period.
 const ReportingInterval = 1 * time.Minute
 
+var (
+	// SuppressObjectTags, if true, will prevent any GroupKind tags from being annotated onto object
+	// metrics.
+	SuppressObjectTags bool
+)
+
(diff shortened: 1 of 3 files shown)


incubator/hnc/pkg/stats/metrics.go
```diff
@@ -122,10 +128,13 @@ func recordMetric(m counter, ms *ocstats.Int64Measure) {
 	ocstats.Record(context.Background(), ms.M(int64(m)))
 }
 
-// recordTagMetric inserts a tag to the context before recording a metric.
+// recordObjectMetric inserts a KeyGroupKind tag to the context before recording a metric.
```
yiqigao217, line 131: Maybe add "if the object tag is not suppressed"?

adrianludwin: Done, also passing in the `GroupKind` directly now.

PR #689, opened 2020-05-04
Make HNC container run as non root

yiqigao217:
Make the HNC container image run as non root. Add securityContext in the
HNC pod to make containers in it run as non root.

Add a port flag that defaults to 443 in the code. Since the container is
run as non root, it's not able to use ports below 1024 reserved for root
only. Therefore, in the pod manifest, set port flag to 9443 and have the
webhook service map the 443 port to targetPort 9443 in the conainer.

Tested on a GKE cluster with PSP of notrunasroot is enabled.

### DIFF
--- a/incubator/hnc/Dockerfile
+++ b/incubator/hnc/Dockerfile
@@ -18,9 +18,12 @@ COPY third_party/ third_party/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager ./cmd/manager/main.go
 
+# Copied from kubebuilder scaffold to run as nonroot at
+# https://github.com/kubernetes-sigs/kubebuilder/blob/7af89cb00c224c57ece37dc14ea37caf1eb769db/pkg/scaffold/v2/dockerfile.go#L60
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:latest
+FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
+USER nonroot:nonroot
 ENTRYPOINT ["/manager"]
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -57,6 +57,7 @@ var (
 	testLog              bool
 	internalCert         bool
 	qps                  int
+	webhookServerPort    int
 )
 
 func init() {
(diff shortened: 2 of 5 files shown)


incubator/hnc/Dockerfile
```diff
@@ -20,7 +20,8 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:latest
+FROM gcr.io/distroless/static:nonroot
```
adrianludwin, line 23: Neat! Where did you find this? Can we add a comment pointing to the docs?

incubator/hnc/cmd/manager/main.go
```diff
@@ -81,6 +82,7 @@ func main() {
 	flag.IntVar(&maxReconciles, "max-reconciles", 1, "Number of concurrent reconciles to perform.")
 	flag.IntVar(&qps, "apiserver-qps-throttle", 50, "The maximum QPS to the API server.")
 	flag.BoolVar(&stats.SuppressObjectTags, "suppress-object-tags", true, "If true, suppresses the kinds of object metrics to reduce metric cardinality.")
+	flag.IntVar(&port, "port", 443, "port for the server. defaulted to 443 if unspecified ")
```
adrianludwin, line 85: Use sentence case (eg capitalize the first word). Also say _which_ service this is for - e.g. the webhook. Maybe also change the flag name from `--port` to `--admission-webhook-port` or something like that (ditto for the internal variable).

incubator/hnc/config/default/manager_auth_proxy_patch.yaml
```diff
@@ -8,6 +8,10 @@ metadata:
 spec:
   template:
     spec:
+      securityContext:
+        fsGroup: 2000
+        runAsNonRoot: true
+        runAsUser: 1000
```
adrianludwin, line 14: Can you comment where these values came from?

yiqigao217: Added!

incubator/hnc/cmd/manager/main.go
yiqigao217: sg. Updated to `webhook-server-port` to match the controller-runtime description of the port.

incubator/hnc/Dockerfile
yiqigao217: Added the [link](https://github.com/kubernetes-sigs/kubebuilder/blob/7af89cb00c224c57ece37dc14ea37caf1eb769db/pkg/scaffold/v2/dockerfile.go#L60) in the comment.

incubator/hnc/Dockerfile
```diff
@@ -18,9 +18,12 @@ COPY third_party/ third_party/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager ./cmd/manager/main.go
 
+# Copied from kubebuilder scaffold to run as nonroot at
```
adrianludwin, line 21: Maybe: "Following the latest kubebuilder recommendation to use nonroot"? If we re-used kubebuilder today, we would never have used "static:latest"

Issue #667, opened 2020-04-22
HNC: update all uses of the term "self-service namespace" and "owned namespace" to "subnamespace"


adrianludwin:
In the [user guide](http://bit.ly/38YYhE0), we exclusively use the term "subnamespace" rather than the (clunkier) "self-service namespace," but several webhooks and code use the "self-serve" term. We should go fix these.

PR #697, opened 2020-05-07
Remove the term "self-serve (sub)namespace"

adrianludwin:
The documentation now uses the term "subnamespaces" so all internal and
external code should use that as well.

Tested: none

Addresses #667


yiqigao217:
/lgtm
/approve
/hold

cc @rjbez17

PR #698, opened 2020-05-07
Change "owned" namespaces to subnamespaces

### LINKED ISSUE #667: HNC: update all uses of the term "self-service namespace" and "owned namespace" to "subnamespace"

adrianludwin:
Part of #667. The term "owned" namespace has been changed throughout to
"subnamespace"; the annotation changed from `ownerReference` to
`subnamespaceOf`; the term "owner namespace" has usually been changed to
simply "parent."

Tested: ran hack/test-issue-501.sh; everything works as well as expected
given that the new webhooks prevent us from doing the wrong thing. Also
ran the HNC demo script.

Fixes #667

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -123,9 +123,8 @@ type Condition struct {
 	// - "CritAncestor": a critical error exists in an ancestor namespace, so this namespace is no
 	// longer being updated either.
 	//
-	// - "HNSMissing": this namespace is an owned namespace (created by reconciling an hns instance),
-	// but the hns instance is missing in its owner (referenced in its owner annotation) or the owner
-	// namespace is missing.
+	// - "HNSMissing": this namespace is a subnamespace, but the hierarchicalnamespace instance
+	// referenced in its `subnamespaceOf` annotation does not exist in the parent.
 	//
 	// - "CannotPropagateObject": this namespace contains an object that couldn't be propagated to
 	// one or more of its descendants. The condition's affect objects will include a list of the
--- a/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
+++ b/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
@@ -35,8 +35,8 @@ spec:
           description: HierarchySpec defines the desired state of Hierarchy
           properties:
             allowCascadingDelete:
-              description: AllowCascadingDelete indicates if the self-serve subnamespaces
-                of this namespace are allowed to cascading delete.
+              description: AllowCascadingDelete indicates if the subnamespaces of
+                this namespace are allowed to cascading delete.
               type: boolean
             parent:
               description: Parent indicates the parent of this namespace, if any.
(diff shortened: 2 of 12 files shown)


incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -95,10 +95,9 @@ spec:
                       would cause a cycle) \n - \"CritAncestor\": a critical error
                       exists in an ancestor namespace, so this namespace is no longer
                       being updated either. \n - \"HNSMissing\": this namespace is
-                      an owned namespace (created by reconciling an hns instance),
-                      but the hns instance is missing in its owner (referenced in
-                      its owner annotation) or the owner namespace is missing. \n
-                      - \"CannotPropagateObject\": this namespace contains an object
+                      a subnamespace, but the hierarchicalnamespace instance referenced
+                      in its `ownerReference` annotation does not exist in the parent.
```
rjbez17, line 99: Didn't you change this to `subnamespaceOf` above?

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -123,9 +123,8 @@ type Condition struct {
 	// - "CritAncestor": a critical error exists in an ancestor namespace, so this namespace is no
 	// longer being updated either.
 	//
-	// - "HNSMissing": this namespace is an owned namespace (created by reconciling an hns instance),
-	// but the hns instance is missing in its owner (referenced in its owner annotation) or the owner
-	// namespace is missing.
+	// - "HNSMissing": this namespace is a subnamespace, but the hierarchicalnamespace instance
+	// referenced in its `ownerReference` annotation does not exist in the parent.
```
yiqigao217, line 127: r/ownerReference/subnamespaceOf

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
```diff
@@ -80,32 +80,32 @@ func (r *HierarchicalNamespaceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, r.writeInstance(ctx, log, inst)
 	}
 
-	// Get the owned child namespace instance. If it doesn't exist, initialize one.
-	cnsInst, err := r.getNamespace(ctx, nm)
+	// Get the subnamespace instance. If it doesn't exist, initialize one.
```
yiqigao217, line 83: Replace it with "Get the namespace instance of the subnamespace"? In case the `hierarchicalnamespace` resource will be renamed to `subnamespace` in the future.

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
adrianludwin: Good catch, fixed.

incubator/hnc/api/v1alpha1/hierarchy_types.go
adrianludwin: Good catch

incubator/hnc/pkg/reconcilers/hierarchical_namespace.go
adrianludwin: Good call. Changed to "Get the subnamespace."

rjbez17:
lgtm after nits

Issue #666, opened 2020-04-22
HNC: Add critical condition to both culprit namespaces when they cause a cycle.


sophieliu15:
Right now if setting the parent of a namespace causes a cycle, only the parent will get `CritParentInvalid` condition. When restarting HNC, it is nondeterministic which namespace in the cycle will have `CritParentInvalid` condition. For details, see discussions in #660.

Ideally, if we set parent of `a` to `b` and it causes a cycle, the condition should be on both `a` and `b` and not on other namespaces in the cycle. Furthermore, the locations of the condition should not change when restarting HNC.

Since it is possible that `b` might be a descendant of `a` but not a child of `a`, we will change the condition name to exclude the word "parent", e.g., `CritCycle`.

sophieliu15:
cc:
@adrianludwin 
@adrianchung 
@yiqigao217

PR #699, opened 2020-05-08
Stable removal of obsolete conditions

adrianludwin:
I never liked the fix to #328 since there was basically only one chance
to notice that conditions were now obsolete - when the parent was being
changed. This felt weird, and when I started looking at (fittingly)
issue #666, which requires us to allow cycles in the in-memory forest,
it became difficult to see how to make the existing logic work.

This new approach, by contrast, annotates each condition code with the
knowledge of whether it can be caused by an ancestor or descendant
namespace, making it easy to look at the *current* state of the forest
(and not of any one change) and determine whether the condition should
be removed. This will make future logic changes easier and just seems
cleaner overall.

This also makes a minor improvement to kubectl-hns-describe. It also
removes the AffectedObject from the HNSMissing condition, because it's
not really necessary (it always points to the immediate parent) and
because it made the logic significantly easier.

Finally, I noticed that markExisting() wouldn't always do the right
thing for subnamespaces because it's called before syncParent(), so I
fixed this. I doubt this was causing any problem (I think we always
re-sync a namespace at least once) but we may as well be accurate.

Tested: re-ran hack/test-issue-328.sh (it worked). Played around a bit
by hand and noticed that everything seemed to be working well.

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -36,7 +36,8 @@ const (
 	FinalizerHasOwnedNamespace = MetaGroup + "/hasOwnedNamespace"
 )
 
-// Condition codes. *All* codes must also be documented in the comment to Condition.Code.
+// Condition codes. *All* codes must also be documented in the comment to Condition.Code, and must
+// also have an entry in ClearConditionCriteria, set in init() in this file.
 const (
 	CritParentMissing         Code = "CritParentMissing"
 	CritParentInvalid         Code = "CritParentInvalid"
@@ -45,6 +46,31 @@ const (
 	CannotPropagate           Code = "CannotPropagateObject"
 	CannotUpdate              Code = "CannotUpdateObject"
 )
+
+// ClearConditionCriterion describes when a condition should be automatically cleared based on
+// forest changes. See individual constants for better documentation.
+type ClearConditionCriterion int
+
+const (
+	CCCUnknown ClearConditionCriterion = iota
+
+	// CCCManual indicates that the condition should never be cleared automatically, based on the
+	// structure of the forest. Instead, the reconciler that sets the condition is responsible for
+	// clearing it as well.
+	CCCManual
+
+	// CCCAncestor indicates that the condition should always exist in the namespace's ancestors, and
+	// should be cleared if this is no longer true.
+	CCCAncestor
+
+	// CCCSubtree indicates that the condition should always exist in the namespace's subtree (that
+	// is, the namespace itself or any of its descendants), and should be cleared if this is no longer
+	// true.
+	CCCSubtree
+)
+
+// ClearConditionCriteria is initialized in init(). See ClearConditionCriterion for more details.
+var ClearConditionCriteria map[Code]ClearConditionCriterion
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
@@ -208,4 +234,24 @@ func SortAffectedObjects(objs []AffectedObject) {
 
 func init() {
 	SchemeBuilder.Register(&HierarchyConfiguration{}, &HierarchyConfigurationList{})
+	ClearConditionCriteria = map[Code]ClearConditionCriterion{
+		// All conditions on namespaces are set/cleared manually by the HCR
+		CritParentMissing:         CCCManual,
+		CritParentInvalid:         CCCManual,
+		CritAncestor:              CCCManual,
+		SubnamespaceAnchorMissing: CCCManual,
+
+		// A source object can cause the CannotPropagate condition in two ways: if it cannot be
+		// propagated *out* of its original namespace (e.g. because it has a finalizer), or if it cannot
+		// be propagated *into* a descendant namespace. In the former case, the affected object is in
+		// the source namespace; in the latter, it's in a descendant (and the descendant namespace will
+		// have a CannotUpdate condition).
+		CannotPropagate: CCCSubtree,
+
+		// A propagated object can cause the CannotUpdate object in the destination namespace if there's
+		// a reason it cannot be copied into that namespace (e.g. transient errors, HNC has insufficient
+		// privileges, etc). The affected object will always be the source object, which will always be
+		// in an ancestor namespace.
+		CannotUpdate: CCCAncestor,
+	}
 }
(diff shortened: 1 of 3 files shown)


incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -46,6 +47,21 @@ const (
 	CannotUpdate      Code = "CannotUpdateObject"
 )
 
+// ClearConditionCriterion describes when a condition should be cleared: never, when it refers to an
+// object in a namespace that's no longer an ancestor, or when it refers to an object that's no
+// longer a descendant.
```
yiqigao217, line 52: A condition should *never* be cleared in these two cases? Shouldn't it be the opposite?

incubator/hnc/pkg/forest/forest.go
```diff
@@ -461,79 +461,70 @@ func (ns *Namespace) HasCritCondition() bool {
...
-	if !ns.HasCondition(obj, code) {
+	if _, hasObj := ns.conditions[obj]; !hasObj {
+		// No conditions for this object
 		return false
 	}
 
 	if code == "" {
 		delete(ns.conditions, obj)
-	} else {
-		delete(ns.conditions[obj], code)
```
yiqigao217, line 492: Was this a bug that you fixed by returning the right value?

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -208,4 +224,12 @@ func SortAffectedObjects(objs []AffectedObject) {
 
 func init() {
 	SchemeBuilder.Register(&HierarchyConfiguration{}, &HierarchyConfigurationList{})
+	ClearConditionCriteria = map[Code]ClearConditionCriterion{
+		CritParentMissing: CCCNone,
+		CritParentInvalid: CCCNone,
+		CritAncestor:      CCCNone,
+		HNSMissing:        CCCNone,
+		CannotPropagate:   CCCNotDescendant,
+		CannotUpdate:      CCCNotAncestor,
```
yiqigao217, line 233: It's a little hard to understand `CCCNotDescendant` and `CCCNotDescendant`. Also it's not clear *who* is not descendant or ancestor.

Either change the const name to e.g. `CCCAONotDescendant` or add comments in the const declarations in L58-59.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -186,59 +186,58 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
...
 }
 
-// syncOwner sets the parent to the owner and updates the HNSMissing condition if the HNS instance
-// is missing in the parent namespace according to the forest.  The namespace owner annotation is
-// the source of truth of the ownership (e.g. being a subnamespace), since modifying a namespace has
-// higher privilege than what HNC users can do.
-func (r *HierarchyConfigReconciler) syncOwner(log logr.Logger, inst *api.HierarchyConfiguration, nsInst *corev1.Namespace, ns *forest.Namespace) {
-	// Clear the HNSMissing condition if this is not a subnamespace or to
-	// reset it for the updated condition later.
-	ns.ClearConditionsByCode(log, api.HNSMissing)
```
yiqigao217, line 214: Seems `HNSMissing` will never be cleared once set if it's not cleared here. For example, when an hns was missing but a correct hns is somehow added again.

incubator/hnc/api/v1alpha1/hierarchy_types.go
adrianludwin: Changed from "never" to "manual" and rewrote all the documentation - thanks for pointing that out.

incubator/hnc/api/v1alpha1/hierarchy_types.go
adrianludwin: Added docs, thanks for keeping me honest :D

incubator/hnc/pkg/forest/forest.go
adrianludwin: No, the logic's still exactly the same - line 492 got moved to line 484. But we no longer rely on the `HasCondition` method - it was only being used here, so I deleted it.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
adrianludwin: Because HNSMissing is now a local condition (see line 240), it gets cleared with all the other local conditions at the beginning of `syncForest` (line 193).

adrianludwin:
@yiqigao217 FYI I've fixed the test failures if you want to have another look

incubator/hnc/pkg/forest/forest.go
yiqigao217: Right, I was confirming that it was a previous bug here to not `return false` if the condition wasn't there already at the first place. Your code moving it out and `return false` when `!hasCode` fixed this bug seems.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: Thanks for the clarification!

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: Nice doc, very clear!!

incubator/hnc/api/v1alpha1/hierarchy_types.go
yiqigao217: Nice doc!

yiqigao217:
Nice doc!
/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/pkg/forest/forest.go
adrianludwin: Didn't the old line 485 catch that case?

```
if !ns.HasCondition(obj, code) {
  return false
```

I needed to add a bunch of `return false` statements because that one line no longer caught them all.

adrianludwin:
I've discovered I want to make a change to this, please do not submit.

incubator/hnc/pkg/forest/forest.go
adrianludwin: Update: I'm reverting this part of the change since I have another change which needs the `HasCondition` method. Please LMK what you think the bug is!

yiqigao217: Ah NVM, then there was no bug in the old one. I missed the return in line 485.

adrianludwin:
/hold cancel

/cc @yiqigao217 
/cc @rjbez17 

Can you please re-review? The server checks used to be part of this change but I factored them out; otherwise, not _too_ much has changed.

PR #702, opened 2020-05-10
More flexible way to create and test forests

adrianludwin:
I removed some of the tests about missing intermediate namespaces since they're not actually possible anymore, and my next PR will remove the ability to express them.

### DIFF
--- a/incubator/hnc/pkg/validators/hierarchy_test.go
+++ b/incubator/hnc/pkg/validators/hierarchy_test.go
@@ -10,16 +10,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	api "github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/api/v1alpha1"
-	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/forest"
+	"github.com/kubernetes-sigs/multi-tenancy/incubator/hnc/pkg/foresttest"
 )
 
 func TestStructure(t *testing.T) {
-	f := forest.NewForest()
-	// Create the following forest: foo -> bar
-	foo := createNS(f, "foo", nil)
-	bar := createNS(f, "bar", nil)
-	createNS(f, "baz", nil)
-	bar.SetParent(foo)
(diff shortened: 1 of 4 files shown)


incubator/hnc/pkg/validators/hierarchy_test.go
```diff
@@ -10,16 +10,11 @@ import (
...
 )
 
 func TestStructure(t *testing.T) {
-	f := forest.NewForest()
-	// Create the following forest: foo -> bar
-	foo := createNS(f, "foo", nil)
-	bar := createNS(f, "bar", nil)
-	createNS(f, "baz", nil)
-	bar.SetParent(foo)
+	f := foresttest.Create("-a-") // a -> b; c
```
yiqigao217, line 17: nit: s/a->b/a<-b to match what in the `foresttest.Create` examples (arrow pointing to the parent).

adrianludwin: Good catch, fixed

yiqigao217:
> I removed some of the tests about missing intermediate namespaces since they're not actually possible anymore, and my next PR will remove the ability to express them.

Question: why missing intermediate namespaces was possible before and is not possible now?

adrianludwin:
> Question: why missing intermediate namespaces was possible before and is not possible now?

Before, we could have namespace `A` with requiredChild `B`, and namespace `C` could have parent `B`, but there could be some problem that prevented `B` from being created. That is, `B` is an intermediate namespace between `C` and `A`, and it would appear in the in-memory forest as both a child of `A` and a parent of `C`, just with `ns.Exists()` returning false.

But now, when we create an anchor in `A`, we don't actually create an entry for `B` in the in-memory forest until the Anchor reconciler successfully creates the namespace. So `B` will never appear to have parent `A` until it actually gets created.

By contrast, if `C` reference parent `B` and it doesn't exist, we _will_ create it in-memory. But it won't appear to be an _intermediate_ namespace anymore - it will just look like a root.

Does this make sense?

yiqigao217:
I see makes sense, thanks for the explanation!

/lgtm
/approve
/hold cancel

PR #703, opened 2020-05-10
More robust server checks

adrianludwin:
Now that we have subnamespaces, we can't have missing namespaces in the
middle of a tree, only at the root. Simplify the authz checks to reflect
this, and better handle the two reasons why the namespace might be
missing (HNC is starting up vs the namespace is truly orphaned).

Tested: manually confirmed that we can fix namespaces with the
CritParentMissing condition, either by making it a root or recreating
the missing parent. Also manually confirmed that we can't change the
properties of a namespace with a critical ancestor. Added logs and
verified that we're checking authz on the correct namespaces in multiple
cases. Also added new tests to confirm all of this automatically.

### DIFF
--- a/incubator/hnc/pkg/forest/forest.go
+++ b/incubator/hnc/pkg/forest/forest.go
@@ -450,15 +450,16 @@ func (ns *Namespace) HasLocalCritCondition() bool {
 	return ns.HasCondition(api.AffectedObject{}, "")
 }
 
-// HasCritCondition returns if the namespace or any of its ancestors has any critical condition.
-func (ns *Namespace) HasCritCondition() bool {
+// GetCritAncestor returns the name of the first ancestor with a critical condition, or the empty
+// string if there are no such ancestors. It *can* return the name of the current namespace.
+func (ns *Namespace) GetCritAncestor() string {
 	if ns.HasLocalCritCondition() {
-		return true
+		return ns.name
 	}
 	if ns.Parent() == nil {
-		return false
+		return ""
 	}
-	return ns.Parent().HasCritCondition()
+	return ns.Parent().GetCritAncestor()
 }
 
 // HasCondition returns true if there's a condition with the given object and code. If code is the
--- a/incubator/hnc/pkg/validators/hierarchy.go
+++ b/incubator/hnc/pkg/validators/hierarchy.go
@@ -9,7 +9,10 @@ import (
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	authnv1 "k8s.io/api/authentication/v1"
 	authzv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -33,13 +36,16 @@ const (
 type Hierarchy struct {
 	Log     logr.Logger
 	Forest  *forest.Forest
-	authz   authzClient
+	server  serverClient
 	decoder *admission.Decoder
 }
 
-// authzClient represents the authz checks that should typically be performed against the apiserver,
-// but need to be stubbed out during unit testing.
-type authzClient interface {
+// serverClient represents the checks that should typically be performed against the apiserver, but
+// need to be stubbed out during unit testing.
+type serverClient interface {
+	// Exists returns true if the given namespace exists.
+	Exists(ctx context.Context, nnm string) (bool, error)
+
 	// IsAdmin takes a UserInfo and the name of a namespace, and returns true if the user is an admin
 	// of that namespace (ie, can update the hierarchical config).
 	IsAdmin(ctx context.Context, ui *authnv1.UserInfo, nnm string) (bool, error)
@@ -65,11 +71,15 @@ type request struct {
 // mistakenly allowed).  False negatives can easily be retried and so are not a significant problem,
 // since (by definition) we expect the problem to be transient.
 //
-// False positives are a more serious concern, but the reconciler has been designed to assume that
-// the validator is _never_ running, and any illegal configuration that makes it into K8s will
+// False positives are a more serious concern, and fall into two categories: structural failures,
+// and authz failures. Regarding structural failures, the reconciler has been designed to assume
+// that the validator is _never_ running, and any illegal configuration that makes it into K8s will
 // simply be reported via HierarchyConfiguration.Status.Conditions. It's the admins'
 // responsibilities to monitor these conditions and ensure that, transient exceptions aside, all
-// namespaces are condition-free.
+// namespaces are condition-free. Note that even if the validator is working perfectly, it's still
+// possible to introduce structural failures, as described in the user docs.
+//
+// Authz false positives are prevented as described by the comments to `getServerChecks`.
 func (v *Hierarchy) Handle(ctx context.Context, req admission.Request) admission.Response {
 	log := v.Log.WithValues("ns", req.Namespace, "user", req.UserInfo.Username)
 	decoded, err := v.decodeRequest(req)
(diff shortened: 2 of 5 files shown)


adrianludwin:
/hold

DO NOT SUBMIT until after #702 

/cc @yiqigao217 
/cc @rjbez17

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -172,97 +176,110 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
+// If this webhook is working correctly, a namespace can never be deliberately assigned to a parent
+// that doesn't exist (in Gitops flows, this means that the client is expected to create all
+// namespaces before syncing HierarchyConfiguration objects, or at least be able to keep retrying
+// until after all namespaces have been created). Therefore, there are only three cases where an
+// ancestor might be missing:
+//
+// 1. HNC is just starting up and the parent hasn't been synced yet, so we can't determine the tree
+// root. In these cases, we should just reject the request and ask the user to try again.
+// 2. The parent has been deleted, and the namespace is orphaned. In this case, the namespace will
+// have a CritParentMissing condition, and we should treat the namespace as though it's already a
```
yiqigao217, line 203: In the case 1, when the parent is not synced yet, the namespace will have `CritParentMissing` too, right? How are case 1 and 2 handled differently? Could you please point it to me where case 1 is denied in the code?

incubator/hnc/pkg/validators/hierarchy_test.go
```diff
@@ -69,6 +69,8 @@ func TestAuthz(t *testing.T) {
 		{name: "dst and cur parent (but not root) across trees", forest: "-a-", from: "b", to: "c", authz: "bc", fail: true}, // a <- b; c
 		{name: "dst and cur root across trees", forest: "-a-", from: "b", to: "c", authz: "ac"},                              // a <- b; c
 		{name: "mrca in tree", forest: "-abb", from: "c", to: "d", authz: "b"},                                               // a <- b <- (c, d)
+		{name: "dest but missing parent", forest: "-", from: "z", to: "a", authz: "a"},                                       // a
+		{name: "dest but missing ancestor", forest: "z-", from: "a", to: "b", authz: "ab", fail: true},                       // z <- a; b (z is missing)
```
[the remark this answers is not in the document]
adrianludwin: The first one is case 2 (parent is missing) and the second one is case 3 (ancestor is missing):

```
// 2. The parent has been deleted, and the namespace is orphaned. In this case, the namespace will
// have a CritParentMissing condition, and we should treat the namespace as though it's already a
// root to allow admins to resolve the problem.
// 3. An ancestor has been deleted, but not the parent. In this case, the namespace will have a
// CritAncestor condition, and HNC will freeze propagation. I think it makes sense to freeze all
// changes to the subtree as well; the admin of the namespace with the missing parent needs to take
// action first.
```

Does this make sense?

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: Good catch! You're right, for some reason I was thinking that the condition wouldn't be set, but of course that's not true.

What do you think we should do? I can't think of a good way to tell the difference in our current framework. Do you think it's fine to simply not support this case?

yiqigao217: Do you mean just deny both case 1 and 2? I couldn't think of a good way to handle case 2 either. Denying them (both case 1 and 2) sgtm.

incubator/hnc/pkg/validators/hierarchy_test.go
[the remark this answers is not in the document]
yiqigao217: Ah yes, and just confirming from the other comment, both of these will be denied after the change, right?

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: I'd meant the opposite - if there's a missing parent, let an admin resolve the problem. If we go with "always deny," how do you think an admin should fix it?

yiqigao217: Ah ok, sgtm. Then the only concern is moving a namespace when the parent is not synced, right? Just confirming.

incubator/hnc/pkg/validators/hierarchy_test.go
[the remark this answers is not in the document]
yiqigao217: NVM since the change will be the opposite in the other comment.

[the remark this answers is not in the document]
adrianludwin: This logic has now been fully revamped, PTAL!

incubator/hnc/pkg/validators/hierarchy.go
adrianludwin: This logic's been fully revamped as discussed, PTAL

adrianludwin:
This is ready for another review - thanks!

/cc @yiqigao217 
/cc @rjbez17

adrianludwin:
Failure appears to be an unrelated flake; will file PR to fix the flake

/retest

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -222,57 +296,45 @@ func (v *Hierarchy) needAuthzOn(curParent, newParent *forest.Namespace) []authzR
...
-	}
-	return existing
-}
-
-// checkAuthz executes the list of requested checks.
-func (v *Hierarchy) checkAuthz(ctx context.Context, ui *authnv1.UserInfo, reqs []authzReq) admission.Response {
-	if v.authz == nil {
+// checkServer executes the list of requested checks.
+func (v *Hierarchy) checkServer(ctx context.Context, log logr.Logger, ui *authnv1.UserInfo, reqs []serverCheck) admission.Response {
+	if v.server == nil {
```
yiqigao217, line 308: Just curious, when would `v.server == nil`?

incubator/hnc/pkg/validators/hierarchy_test.go
```diff
@@ -55,20 +55,23 @@ func TestStructure(t *testing.T) {
 func TestAuthz(t *testing.T) {
 	tests := []struct {
 		name   string
-		authz  fakeAuthz
+		server fakeServer
 		forest string
 		from   string
 		to     string
-		fail   bool
+		code   int32
```
yiqigao217, line 62: Nice change!

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -33,13 +36,16 @@ const (
...
 }
 
-// authzClient represents the authz checks that should typically be performed against the apiserver,
-// but need to be stubbed out during unit testing.
-type authzClient interface {
+// serverClient represents the checks that should typically be performed against the apiserver, but
+// need to be stubbed out during unit testing.
+type serverClient interface {
+	// Exists returns true if the given namespace exists.
+	Exists(ctx context.Context, nnm string) (bool, error)
```
yiqigao217, line 47: One nit after reviewing the code: it may be clearer to rename this function to differentiate from `forest.ns.Exists`. I am actually not sure after another scan. Maybe it's fine to leave it as is.

incubator/hnc/pkg/forest/forest.go
```diff
@@ -476,6 +477,11 @@ func (ns *Namespace) HasCondition(obj api.AffectedObject, code api.Code) bool {
 	return exists
 }
 
+// HasLocalCondition is a specialization of HasCondition
+func (ns *Namespace) HasLocalCondition(code api.Code) bool {
+	return ns.HasCondition(api.AffectedObject{}, code)
+}
+
```
yiqigao217, line 484: Seems this is not used anywhere?

incubator/hnc/pkg/validators/hierarchy.go
```diff
@@ -33,13 +36,16 @@ const (
 type Hierarchy struct {
 	Log     logr.Logger
 	Forest  *forest.Forest
-	authz   authzClient
+	server  serverClient
```
yiqigao217, line 39: Maybe still call it `client` and call `v.getServerChecks()` `v.checkWithServer()`?

incubator/hnc/pkg/forest/forest.go
adrianludwin: Nice catch - gone.

incubator/hnc/pkg/validators/hierarchy_test.go
adrianludwin: Thanks!

yiqigao217:
/lgtm
/approve
/hold

Since this change is large, I feel safer to have another eye on this.

PR #704, opened 2020-05-10
Change subnamespace object to SubnamespaceAnchor

adrianludwin:
This is a breaking change; the HierarchicalNamespace object is replaced
with the SubnamespaceAnchor object. Otherwise, there are no changes
(including to the UX).

Tested: manual + all tests continue to pass


rjbez17:
If this is a breaking change, shouldn't we at a minimum rev to `v1alpha2`?

adrianludwin:
We've made breaking changes in "v1alpha1" in every version of HNC up until
now. We should probably be more cautious in the future (e.g. once we move
to our own repo and start having real users) but personally I'd like to
take advantage of the fact that, AFAIK, no-one will be badly impacted if we
do this today.

rjbez17:
Yeah, I was just thinking about the (hopefully soon) move to a real repo and wondered when we should start following those practices. I'm good with waiting until then.

adrianludwin:
sgtm!

PR #710, opened 2020-05-12
Fix a test flake

adrianludwin:
Observed this flake on #703, made it retryable.

Tested: ran the tests six times, they all passed

### DIFF
--- a/incubator/hnc/pkg/reconcilers/anchor_test.go
+++ b/incubator/hnc/pkg/reconcilers/anchor_test.go
@@ -75,10 +75,12 @@ var _ = Describe("Anchor", func() {
 
 		// Change the bar's parent. Additionally change another field to reflect the
 		// update of the HC instance (hc.Spec.AllowCascadingDelete).
-		barHier := getHierarchy(ctx, barName)
-		barHier.Spec.Parent = "other"
-		barHier.Spec.AllowCascadingDelete = true
-		updateHierarchy(ctx, barHier)
+		Eventually(func() error {
+			barHier := getHierarchy(ctx, barName)
+			barHier.Spec.Parent = "other"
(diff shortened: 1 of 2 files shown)


incubator/hnc/pkg/reconcilers/anchor_test.go
```diff
@@ -75,10 +75,13 @@ var _ = Describe("Anchor", func() {
...
 		// update of the HC instance (hc.Spec.AllowCascadingDelete).
-		barHier := getHierarchy(ctx, barName)
-		barHier.Spec.Parent = "other"
-		barHier.Spec.AllowCascadingDelete = true
-		updateHierarchy(ctx, barHier)
+		Eventually(func() bool {
+			barHier := getHierarchy(ctx, barName)
+			barHier.Spec.Parent = "other"
+			barHier.Spec.AllowCascadingDelete = true
+			updateHierarchy(ctx, barHier) // can fail if made too quickly
```
yiqigao217, line 82: Hmm, seems the test is still flaky. Was the error you got is something like "the object is updated, please update on a new version"?

/hold

adrianludwin:
Yup. I'll have another look.

adrianludwin:
Looks like you can't nest "Eventually"s. Fixed in a better way.

yiqigao217:
> Looks like you can't nest "Eventually"s. Fixed in a better way.

Did you run the test more than once? If so maybe update the commit/PR message?

adrianludwin:
> Did you run the test more than once? If so maybe update the commit/PR message?

No, only once. It's so slow :(

adrianludwin:
Sigh, you're right :D Set it up to run five times, but heading out for dinner now. You can wait to approve it you want!

adrianludwin:
Updated comment, thanks for making me do the right thing!

PR #719, opened 2020-05-13
Better handle namespace cycles

### LINKED ISSUE #666: HNC: Add critical condition to both culprit namespaces when they cause a cycle.

adrianludwin:
If a user bypasses the admission controller, they can create namespace
ancestry cycles, but these have never been correctly represented
in-memory in HNC, and therefore were not correctly reported. For
example, only _one_ namespace in a cycle would have the
CritInvalidParent condition, as described in #666.

This commit includes several tightly-coupled changes to fix this:
* The Forest no longer prevents you from creating cycles. Instead,
Namespace.CycleNames() returns a non-nil list of a namespace is in a
cycle, and other methods like AncestryNames() and DescendantNames() are
now cycle-safe.
* The kubectl-hns plugin correctly handles cycles
* The HC reconciler ensures that other cycle members are enqueued when a
cycle is created or broken, and uses the new CritCycle condition on all
members of the cycle instead of the CritInvalidParent condition on only
a single member.

Tested: updated various unit tests. Also verified the following:
* With the admission controller disabled, I can create cycles. All
members of the cycle get the CritCycle condition annotated or removed as
appropriate.
* Descendants of namespaces in a cycle continue to get the CritAncestor
condition.
* With the admission controller *enabled*, I can resolve cycles. I
didn't test anything fancy with the authz checks as the recommended mode
to solve these problems is "I have admin so I can do anything," but the
new unit tests do verify that you need authz to break a cycle.

Fixes #666

### DIFF
--- a/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
+++ b/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
@@ -89,21 +89,35 @@ spec:
                       webhooks are working properly, there should typically not be
                       any conditions on any namespaces, although some may appear transiently
                       when the HNC controller is restarted. These should quickly resolve
-                      themselves (<30s). \n Currently, the supported values are: \n
-                      - \"CritParentMissing\": the specified parent is missing \n
-                      - \"CritParentInvalid\": the specified parent is invalid (e.g.,
-                      would cause a cycle) \n - \"CritAncestor\": a critical error
+                      themselves (<30s). However, validation webhooks are not perfect,
+                      especially if multiple users are modifying the same namespace
+                      trees quickly, so it's important to monitor for critical conditions
+                      and resolve them if they arise. See the user guide for more
+                      information. \n Currently, the supported values are: \n - \"CritParentMissing\":
+                      the specified parent is missing and the namespace is an orphan.
+                      \n - \"CritCycle\": the namespace is a member of a cycle. For
+                      example, if namespace B says that its parent is namespace A,
+                      but namespace A says that its parent is namespace B, then A
+                      and B are in a cycle with each other and both of them will have
+                      the CritCycle condition. \n - \"CritAncestor\": a critical error
                       exists in an ancestor namespace, so this namespace is no longer
                       being updated either. \n - \"SubnamespaceAnchorMissing\": this
                       namespace is a subnamespace, but the anchor referenced in its
                       `subnamespaceOf` annotation does not exist in the parent. \n
                       - \"CannotPropagateObject\": this namespace contains an object
-                      that couldn't be propagated to one or more of its descendants.
-                      The condition's affect objects will include a list of the copies
-                      that couldn't be updated. \n - \"CannotUpdateObject\": this
-                      namespace has an error when updating a propagated object from
-                      its source. The condition's affected object will point to the
-                      source object."
+                      that couldn't be propagated *out* of this namespace, to one
+                      or more of this namespace's descendants. If the object couldn't
+                      be propagated to *any* descendants - for example, because it
+                      has a finalizer on it (HNC can't propagate objects with finalizers),
+                      the `Affects` field will point to the object in this namespace.
+                      Otherwise, if it couldn't be propagated to *some* descendants,
+                      `Affects` will contain a list of the objects in those descendants
+                      that couldn't be created or updated. \n - \"CannotUpdateObject\":
+                      this namespace has an object that couldn't be propagated *into*
+                      this namespace - that is, it couldn't be created in the first
+                      place, or it couldn't be updated. The `Affects` field will point
+                      to the source object, which will always be in a namespace that's
+                      an ancestor of this namespace."
                     type: string
                   msg:
                     description: A human-readable description of the condition, if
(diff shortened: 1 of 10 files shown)


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 
/cc @sophieliu15 

Also CC @adrianchung

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -91,13 +91,12 @@ spec:
...
                       themselves (<30s). \n Currently, the supported values are: \n
                       - \"CritParentMissing\": the specified parent is missing \n
-                      - \"CritParentInvalid\": the specified parent is invalid (e.g.,
-                      would cause a cycle) \n - \"CritAncestor\": a critical error
-                      exists in an ancestor namespace, so this namespace is no longer
-                      being updated either. \n - \"SubnamespaceAnchorMissing\": this
-                      namespace is a subnamespace, but the anchor referenced in its
-                      `subnamespaceOf` annotation does not exist in the parent. \n
-                      - \"CannotPropagateObject\": this namespace contains an object
+                      - \"CritCycle\": the namespace is a member of a cycle \n - \"CritAncestor\":
```
rjbez17, line 94: nit: The term `cycle` could be confusing to many people. I'm not sure the typical sysadmin would understand what this means given the context.

adrianludwin:
Will elaborate tomorrow, thanks!

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
adrianludwin: Done, PTAL

incubator/hnc/pkg/reconcilers/hierarchy_config.go
```diff
@@ -269,23 +269,21 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
...
+	ns.SetParent(curParent)
+
+	// Finally, enqueue all other namespaces that could be directly affected. The old and new parents
+	// have just gained/lost a child, while the descendants need to have their tree labels updated and
+	// their objects resynced.
+	//
+	// If we've just created a cycle, all the members of that cycle will be listed as the descendants,
+	// so enqueuing them will ensure that the conditions show up in all members of the cycle.
+	r.enqueueAffected(log, "removed as parent", oldParent.Name())
+	r.enqueueAffected(log, "set as parent", curParent.Name())
```
yiqigao217, line 286: Question: why the nil pointer case is impossible here?

incubator/hnc/pkg/forest/forest.go
```diff
@@ -225,18 +225,20 @@ func (ns *Namespace) AllowsCascadingDelete() bool {
...
-func (ns *Namespace) SetParent(p *Namespace) error {
-	if reason := ns.CanSetParent(p); reason != "" {
-		return errors.New(reason)
-	}
+	// This is a subnamespace so it must have a non-nil parent. If the parent is missing, it will
+	// return the default false.
+	//
+	// Subnamespaces can never be involved in cycles, since those can only occur at the "top" of a
+	// tree and subnamespaces cannot be roots by definition. So this line can't cause a stack
+	// overflow.
```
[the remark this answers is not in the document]
yiqigao217: Yes, thanks for the explanation!

incubator/hnc/pkg/forest/forest.go
```diff
@@ -250,22 +252,33 @@ func (ns *Namespace) SetParent(p *Namespace) error {
...
+	// the proposed parent. Start at the end of the ancestry (e.g. at the proposed parent) and work
+	// our way up to the root.
+	ancestors := p.AncestryNames()
+	cycle := []string{}
+	found := false
+	for i := len(ancestors) - 1; !found && i >= 0; i-- {
+		cycle = append(cycle, ancestors[i])
+		found = (ancestors[i] == ns.name)
+	}
+	if found {
```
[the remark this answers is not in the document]
yiqigao217: Ah right I see, right, this is in `CanSetParent()` to predict.

incubator/hnc/pkg/forest/forest.go
```diff
@@ -608,14 +650,33 @@ func (ns *Namespace) Conditions() []api.Condition {
...
+	ds := map[string]bool{}
+	ns.populateDescendants(ds)
+	if len(ds) == 0 {
+		return nil
+	}
+	d := []string{}
+	for k, _ := range ds {
+		d = append(d, k)
+	}
+	sort.Strings(d)
```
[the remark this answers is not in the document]
yiqigao217: sg.

incubator/hnc/pkg/reconcilers/hierarchy_config.go
yiqigao217: nice thanks.

yiqigao217:
/lgtm
/approve

Great job @adrianludwin :D

PR #720, opened 2020-05-14
[incubator][VC] add max retry rate about dws

zhuangqh:


thousandhu:
In our cluster, we adds initcontainers to pod and causs the pod dws fails. The failure will retry forever and let the dws slower. So we think the max retry will also add to the  dws logic.

PR #721, opened 2020-05-14
[incubator][VC] fix: container diff if webhook add to super pod

zhuangqh:


thousandhu:
In multi-tenancy cluster，we may not want to sync containers to tenant master because the container added by super may not want to known by tenant. For example, some sidecar container or some initcontainer about the security. The tenant should only care about the containers they submitted which relate to they business logic.

PR #722, opened 2020-05-14
Switch to Go 1.14 and vendored dependencies

adrianludwin:
Other projects similar to HNC, such as Gatekeeper, includ a local copy
of all dependencies, managed by `go mod vendor`. There are two reasons
for this:
1. To protect ourselves in case a third-party dependency goes away (e.g.
someone deletes their repo).
2. To make remote builds faster (e.g. we no longer need to download all
our dependencies from the internet).

This change was generated as follows:
* Update go.mod to use Go 1.14 (required for automatic detection of
  vendored dependencies).
* `go mod tidy` (removed one dependency from go.mod)
* `go mod vendor` (created the vendor/ directory)
* Manually updated Makefile to avoid running fmt and vet on vendor/
* Manually updated Dockerfile to use the new vendor/ directory.

Tested:
* Ran `go clean -modcache` and verified that we can build locally
  outside of Docker without $GOPATH/pkg/mod being repopulated (note:
  this didn't work until I upgraded to Go 1.14, as expected).
* Verified that we can build locally inside Docker as well.
* Deployed to a test cluster and made sure I can still create
  subnamespaces, look at trees, etc.
* I can't verify that this works as part of the release process until
  it's checked in, since the release process only works with checked-in
  code.


adrianludwin:
/assign @rjbez17 

Hey Ryan, do you have any concerns with this? It's based on what our colleagues at Gatekeeper are doing, so we thought it would be useful to align with their best practices. Thanks!

adrianludwin:
cc @adrianchung

rjbez17:
No issues here, this is what I/we do internally as well.

I'm giving a LGTM after reviewing the makefile and dockerfile only, i'm trusting there isn't something fishy going on somewhere between files 1000 and 3000 :)

/lgtm
/approve

adrianludwin:
Seems like a flake

/retest

PR #723, opened 2020-05-14
Move user guide to markdown

adrianludwin:
Also rename the `doc` directory to `docs` (more standard).

Tested: read the manual; you can see everything staged at
https://github.com/adrianludwin/multi-tenancy/tree/docs/incubator/hnc/docs


adrianludwin:
You don't need to read this carefully, it's easy to fix :)

PR #724, opened 2020-05-14
Update tree labels even with critical conditions

### LINKED ISSUE #660: HNC: Update namespace labels when a namespace has a critical condition
Currently [namespace labels](https://docs.google.com/document/d/1R4rwTweYBWYDTC9UC-qThaMkPk8gtflr_ycHOOqzBQE/edit#heading=h.40wgqgbzpspe) of a namespace are not updated if the namespace has a [critical condition](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/api/v1alpha1/hierarchy_types.go#L41-L43) (see [here](https://github.com/kubernetes-sigs/multi-tenancy/blob/cd26b9a0002743920fad427918058929544aee08/incubator/hnc/pkg/reconcilers/hierarchy_config.go#L361)).

Ideally, when a namespace has a critical condition, we should update the labels to reflect the namespace hierarchy that we are certain about. We propose the following namespace label updating strategies for each critical condition:

*   [CritParentMissing
](https://github.com/kubernetes-sigs/multi-tenancy/blob/cd26b9a0002743920fad427918058929544aee08/incubator/hnc/api/v1alpha1/hierarchy_types.go#L119)

`CritParentMissing` means the parent of a namespace is missing (e.g., the parent is deleted). In this case, we propose to view the namespace that has the condition as a root of a namespace tree and it should only contain `depth=0` label. Namespace labels of the descendants of the namespace should include the path from a descendant up till the namespace. Following is an example:

Suppose we have a tree as below where r is the parent of a; a is the parent of b; b is the parent of c. 

r
|
a
|
b
|
c

If we delete r, each namespace will have following conditions and proposed labels:

a: Condition: CritParentMissing; Proposed Label: a/depth=0
b: Condition: CritAncestor; Proposed Labels: b/depth=0; a/depth=1
c: Condition: CritAncestor; Proposed Labels: c/depth=0; b/depth=1; a/depth=2

The hierarchy based on proposed labels becomes:

a
|
b
|
c 

*   [CritParentInvalid
](https://github.com/kubernetes-sigs/multi-tenancy/blob/cd26b9a0002743920fad427918058929544aee08/incubator/hnc/api/v1alpha1/hierarchy_types.go#L121)

`CritParentInvalid` means the parent of a namespace is invalid and causes a cycle. In this case, we propose to view a namespace that has  `CritParentInvalid` as a root and it should only contain the `depth=0` label. Namespace labels of the descendants of the namespace should include the path from a descendant up till the namespace. Following is an example:

Suppose we have a tree as below where r is the parent of a; a is the parent of b and c; b is the parent of d. 

r
|
a - b - d
|
c

If we set c to be parent of a, each namespace will have following conditions and proposed labels:

r: Condition: nil; Proposed Label: r/depth=0
a: Condition: CritParentInvalid; Proposed Label: a/depth=0
b: Condition: CritAncestor; Proposed Labels: b/depth=0; a/depth=1
d: Condition: CritAncestor; Proposed Labels: d/depth=0; b/depth=1; a/depth=2
c: Condition: CritAncestor; Proposed Labels: c/depth=0; a/depth=1

The hierarchy based on proposed labels becomes:

r

a - b - d
|
c

* [CritAncestor](https://github.com/kubernetes-sigs/multi-tenancy/blob/bbbc8c751a40317f9754080185b73f57631cf516/incubator/hnc/api/v1alpha1/hierarchy_types.go#L123)

`CritAncestor` means a critical error (i.e., `CritParentMissing` and/or `CritParentInvalid`) exists in an ancestor namespace. In this case, the namespace labels include the path from the current namespace till the root as described above.

adrianludwin:
See #660. Even when a namespace has a critical condition, we should
still update its tree labels - both because there are things we *can*
say about a namespace's ancestry, and also to *remove* labels that are
no longer valid (e.g. missing parents, ancestors in cycles).

Tested: For cycles, new unit tests plus manual checks. For orphans, I
couldn't add a unit test since the test env can't delete namespaces (I
had another go at this but failed). But I created an orphan manually and
verified that the missing parent's labels were removed from all
descendants, but other labels were not.

Fixes #660

### DIFF
--- a/incubator/hnc/pkg/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/pkg/reconcilers/hierarchy_config.go
@@ -299,11 +299,6 @@ func (r *HierarchyConfigReconciler) syncAnchors(log logr.Logger, ns *forest.Name
 }
 
 func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespace, ns *forest.Namespace) {
-	// Depth label only makes sense if there's no error condition.
-	if ns.GetCritAncestor() != "" {
-		return
-	}
-
 	// Pre-define label depth suffix
 	labelDepthSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
 
@@ -314,12 +309,19 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 		}
 	}
 
-	// AncestryNames includes the namespace itself.
-	ancestors := ns.AncestryNames()
-	for i, ancestor := range ancestors {
-		l := ancestor + labelDepthSuffix
-		dist := strconv.Itoa(len(ancestors) - i - 1)
-		metadata.SetLabel(nsInst, l, dist)
+	// Look for all ancestors. Stop as soon as we find a namespaces that has a critical condition in
+	// the forest (note that CritAncestor is never included in the forest). This should handle orphans
+	// and cycles.
+	anc := ns
+	depth := 0
+	for anc != nil {
+		l := anc.Name() + labelDepthSuffix
+		metadata.SetLabel(nsInst, l, strconv.Itoa(depth))
+		if anc.HasLocalCritCondition() {
+			break
+		}
+		anc = anc.Parent()
+		depth++
 	}
 }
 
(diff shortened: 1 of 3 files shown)


adrianludwin:
Last one! Thanks Yiqi and Ryan!

/assign @rjbez17 
/assign @yiqigao217 

/cc @sophieliu15 
cc @adrianchung

PR #725, opened 2020-05-14
Break up forest.go

adrianludwin:
This file was getting huge, so I broke it up into five smaller ones.

Tested: HNC still builds and tests still pass.


adrianludwin:
/assign @yiqigao217 

cc @rjbez17 in case you care but you can skip this one.

rjbez17:
A quick look this lgtm, I'll do a more thorough look tomorrow unless @yiqigao217 beats me to it.

PR #729, opened 2020-05-15
[VC]: Switch to distroless image

christopherhein:
This switches the container used to build the VC components to `distroless` to follow  https://github.com/kubernetes/enhancements/blob/8fd69422d913706c0badb30b78ac07698c9aaecd/keps/sig-release/1729-rebase-images-to-distroless/README.md KEP. This shaved about 30mb per container.

#### Before

PNG image 2

#### After

PNG image

### DIFF
--- a/incubator/virtualcluster/hack/lib/docker-image.sh
+++ b/incubator/virtualcluster/hack/lib/docker-image.sh
@@ -59,7 +59,13 @@ create_docker_image() {
     IFS=$oldifs
 
     local binary_name="$1"
-    local base_image="$2"
+    local base_image=$2
+    local image_user=""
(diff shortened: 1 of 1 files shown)


christopherhein:
/hold

Discussing this with @Fei-Guo offline, there may be restrictions in mainland china getting access to gcr.io domain based images. I'm going to change this so you can optionally supply an image but leave the defaults the same.

christopherhein:
/remove hold

@Fei-Guo this has been updated so that if you run:

```
BASE_IMAGE=distroless make build-images
```

You end up with the `distroless` based container and if you don't set that then you will get the original `debian` image.

PR #732, opened 2020-05-15
Add Apache license from root to HNC

adrianchung:
This will need to be done as a prerequisite before moving HNC into its
own repository

### DIFF
--- a/incubator/hnc/LICENSE
+++ b/incubator/hnc/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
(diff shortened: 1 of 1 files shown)


incubator/hnc/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright {yyyy} {name of copyright owner}
```
adrianludwin, line 189: Incredibly, even k/k doesn't actually fill this in! Sigh.

PR #734, opened 2020-05-15
Change all local golang imports to sigs.k8s.io

adrianludwin:
Use the vanity domain instead of github.com/kubernetes-sigs.

Tested: can still build via make and Dockerfile

Inspired by: #727 

/assign @rjbez17


christopherhein:
I love these vanity urls, it really helps to trim down what you have to import. Glad it inspired this PR.  I'm considering doing it to the tenant controller as well.

PR #739, opened 2020-05-19
Prevent anyone from importing HNC internals

adrianludwin:
Following Golang standards, this change moves `pkg/` to `internal/`. If
we ever have any packages that can be imported from outside HNC, we can
reintroduce `pkg/` at that time.

Tested: ran `make test`; deployed to my test cluster.

/assign @rjbez17 
/cc @yiqigao217 
/cc @sophieliu15 
cc @adrianchung


adrianludwin:
/assign @yiqigao217 

Hey Yiqi, do you mind having a look at this? The rationale of the `internal/` package is described here: https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

cc @sophieliu15, @adrianchung FYI

rjbez17:
/lgtm
/approve
/hold

@yiqigao217 Feel free to remove hold

adrianludwin:
I don't think this needs two approvers :)

/hold cancel

PR #742, opened 2020-05-20
enable workqueue prometheus metrics

thousandhu:
Enable prometheus metrics for uws queue, mccontroller queue and syncer queue.
These metrics is helpful for location problem that sync slowly or create virtual cluster slowly.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/mccontroller/mccontroller.go
+++ b/incubator/virtualcluster/pkg/syncer/mccontroller/mccontroller.go
@@ -117,7 +117,7 @@ func NewMCController(name string, objectType runtime.Object, options Options) (*
 	}
 
 	if c.Queue == nil {
-		c.Queue = workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+		c.Queue = workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), name)
 	}
 
 	return c, nil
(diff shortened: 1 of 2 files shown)


zhuangqh:
thanks for your first submission to vc project.
can we put this import to `metrics` pkg? maybe import once is enough?

Fei-Guo:
> thanks for your first submission to vc project.
> can we put this import to `metrics` pkg? maybe import once is enough?

One more thing. It will be nice to list all the counters that are added by the change in the change description.

Fei-Guo:
Also, note that we have 12 mccontroller workerqueues (one per resource type). Make sure they are not going to flood the counters.

thousandhu:
@Fei-Guo @zhuangqh Thanks for your advice. I've moved the import to metrics.go. In this issue, workqueues won't share the counters, so i don't think the counter will flood.

Here is the metrics information, I also add it to the commit log.

This issue will enable k8s official work queue promethues metrics. For each work queue, which in our project are mccontrollers for 12 resources, uws controller for 6 resources and work queue for vc controller, this issue will adds following metrics.
- depth: Current depth of workqueue
- adds_total: Total number of adds handled by workqueue
- queue_duration_seconds: How long in seconds an item stays in workqueue before being requested
- work_duration_seconds: How long in seconds processing an item from workqueue takes
- unfinished_work_seconds: How many seconds of work has done that is in progress and hasn't been observed by work_duration. Large values indicatestuck threads. One can deduce the number of stuck threads by observing
the rate at which this increases.
- longest_running_processor_seconds: How many seconds has the longest running processor for workqueue been running.
- retries_total: Total number of retries handled by workqueue

The metrics are named as following formate:
workqueue_work_${METRICS_NAME}{name="${CONTROLLER NAME}"}

For example, the depth of virtual_cluster controller is workqueue_depth{name="virtual_cluster"} 0

Issue #736, opened 2020-05-18
HNC: provide metrics for conditions


adrianludwin:
Conditions are currently only visible on the HCs in individual namespaces, but the cluster admin should have some way to be able to monitor if there are problems across the entire cluster. However, metrics are poor places for high-cardinality data such as the individual namespace names. In addition, metrics are often fairly widely shared and namespace names can be sensitive.

Therefore, I'd like to recommend the following changes:

* The HNCConfig object (a per-cluster singleton) should get a new `.status.namespaceCondition` field which, for each condition, gives a list of namespace names affected by that condition.
* Upon generating this list, the HNCConfig reconciler will also update a gauge metric `hnc/namespace_conditions` which, for each condition, tracks the number of namespaces affected by that condition.
  * We'll also add two pseudo-tags called "critical" and "noncritical".

/cc @yiqigao217 
/cc @sophieliu15 
/cc @rjbez17 

Does this all sgty?

PR #743, opened 2020-05-20
Simpler HNCConfig reconciler triggers

adrianludwin:
Rather than going through the forest, keep an in-memory singleton and
let anyone ask for the HNC Config to be re-reconciled (in preparation
for work on #736).

Tested: unit tests; verified that this worked properly with the fix to
issue #736.

/assign @sophieliu15 
/cc @yiqigao217

### DIFF
--- a/incubator/hnc/internal/reconcilers/hnc_config.go
+++ b/incubator/hnc/internal/reconcilers/hnc_config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 	"time"
 
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
@@ -23,6 +24,10 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/config"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 )
+
+// hnccrSingleton stores a pointer to the cluster-wide config reconciler so anyone can
+// request that it updates itself (see UpdateHNCConfig, below).
+var hnccrSingleton *ConfigReconciler
 
 // ConfigReconciler is responsible for determining the HNC configuration from the HNCConfiguration CR,
 // as well as ensuring all objects are propagated correctly when the HNC configuration changes.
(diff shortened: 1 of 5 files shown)


adrianludwin:
To see how this new way of invoking the HNC Config Reconciler is used, see https://github.com/adrianludwin/multi-tenancy/commit/ebbb8ea4dc49f3e56a2d9233c3a026b99b08a2ac.

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -437,57 +442,61 @@ func (r *ConfigReconciler) setTypeStatuses(inst *api.HNCConfiguration) {
...
-func (r *ConfigReconciler) enqueueSingleton(log logr.Logger, reason string) {
-	go func() {
-		log.Info("Enqueuing for reconciliation", "reason", reason)
-		// The watch handler doesn't care about anything except the metadata.
-		inst := &api.HNCConfiguration{}
-		inst.ObjectMeta.Name = api.HNCConfigSingleton
-		r.Trigger <- event.GenericEvent{Meta: inst}
-	}()
+// RequestReconcile records that the reconciler needs to be reinvoked.
+func (r *ConfigReconciler) RequestReconcile(reason string) {
```
sophieliu15, line 446: Do we need to make this method exported? Are we expecting this to be called outside the package?

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -24,6 +25,10 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 )
 
+// HNCConfigReconcilerSingleton stores a pointer to the cluster-wide config reconciler so anyone can
```
sophieliu15, line 28: For "anyone", do we intent code outside `reconcilers` package to call this?

sophieliu15:
LGTM. Just have a question about exporting the method. See me comments above.

adrianludwin:
> LGTM. Just have a question about exporting the method. See me comments above.

Yeah, we don't really need it to be exported outside of reconcilers, but I wanted to indicate that this method was "more public" than the others. I'm thinking that we should break up the `reconcilers/` package at some point to enforce more separation.

I can make everything private if you like, wdyt?

incubator/hnc/internal/reconcilers/hnc_config.go
adrianludwin: (As discussed in another comment) no and no :) This is mainly about saying that this method _should_ be called outside of the ConfigReconciler.

incubator/hnc/internal/reconcilers/hnc_config.go
adrianludwin: (As discussed in another comment) nope :)

sophieliu15:
> > LGTM. Just have a question about exporting the method. See me comments above.
> 
> Yeah, we don't really need it to be exported outside of reconcilers, but I wanted to indicate that this method was "more public" than the others. I'm thinking that we should break up the `reconcilers/` package at some point to enforce more separation.
> 
> I can make everything private if you like, wdyt?

How about making it private for now and making it public when we break up the package? Just in case someone misuse it by accident. Thank you!

adrianludwin:
> How about making it private for now and making it public when we break up the package? Just in case someone misuse it by accident. Thank you!

Done!

PR #744, opened 2020-05-20
Add namespace conditions to HNCConfig and metrics

### LINKED ISSUE #736: HNC: provide metrics for conditions

adrianludwin:
This change adds a new namespaceConditions field to HNCConfig, as well
as as populating a namespace_conditions metric tagged by both the
condition as well as whether or not it's critical. This allows admins to
set up monitoring for bad behaviour.

Tested: induced the CritParentMissing condition and observed that all
affected namespaces were listed in HNCConfig, and that the metric counts
and tags were correct in Stackdriver.

Fixes #736 

/assign @rjbez17 
/assign @yiqigao217 
cc @sophieliu15 @adrianchung


adrianludwin:
> 1. The `namesapceConditions` field has the list of affected namespaces, while the metrics only has the number of the affected namespaces. Do you think it necessary to have the namespace names in the metrics too?

We've been trying to reduce the cardinality of our metrics already, so I think _not_ adding namespace names unless there's a clear need is the right way to go. Metrics are mainly there to alert a human that there's a problem, and sending them to look at the `HNCConfiguration` object seems like a good first action for them to take.

> 2. For the two tags in the metrics, seems the `keyNamesapceCritical` is not necessary since the condition code has already included `crit` prefix.

This just makes it easier to build an alert in systems like Stackdriver or Grafana. I'm not sure if either of them let you say something like "build an alert off of any tag that matches the following regex," but even if they do, I think it's nicer to have that behaviour out-of-the-box.

yiqigao217:
Thanks for the explanations! SGTM

/lgtm
/approve
/hold

adrianludwin:
/hold cancel

@rjbez17 if you have any concerns with this just LMK and I'll go make the fixes. But we're off tomorrow and you're off on Monday so I'd kind of like to get this in :)

PR #747, opened 2020-05-20
Allow namespace deletion in K8s 1.14

### LINKED ISSUE #688: HNC: Allow NS deletion if the request cannot be decoded in K8s versions <1.15
See #680 

A workaround to make it safer before the issue is fixed is to: Allow namespace deletion if the request cannot be properly [decoded](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/pkg/validators/namespace.go#L107) in the namespace webhook.

This workaround makes it possible to delete namespaces in <1.15 but won't prevent deleting self-service namespaces by mistake. It shouldn't affect any HNC features in >=1.15.

adrianludwin:
See https://github.com/kubernetes-sigs/multi-tenancy/issues/688. If we
can't validate the deletion because OldObject doesn't exist, we need to
just allow the deletion.

In the case of a namespace with children and allowCascadingDelete
*unset*, this will delete the parent but *not* the children; they will
be left with two conditions (CritMissingParent and
MissingSubnamespaceAnchor). If allowCascadingDelete *is* set, everything
will work properly.

Tested: verified that everything works as expected in 1.14 and that
there are no changes in 1.16.

Fixes #688 

/assign @yiqigao217 
/assign @rjbez17 
cc @sophieliu15 @adrianchung

### DIFF
--- a/incubator/hnc/internal/validators/anchor.go
+++ b/incubator/hnc/internal/validators/anchor.go
@@ -51,6 +51,10 @@ func (v *Anchor) Handle(ctx context.Context, req admission.Request) admission.Re
 		v.Log.Error(err, "Couldn't decode request")
 		return deny(metav1.StatusReasonBadRequest, err.Error())
 	}
+	if decoded == nil {
+		// https://github.com/kubernetes-sigs/multi-tenancy/issues/688
+		return allow("")
+	}
 
 	resp := v.handle(decoded)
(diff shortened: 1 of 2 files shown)


adrianludwin:
/hold cancel

@rjbez17 if you have any concerns with this just LMK and I'll go make the fixes. But we're off tomorrow and you're off on Monday so I'd kind of like to get this in :)

PR #760, opened 2020-05-24
Add best practices for HNC health

adrianludwin:


adrianludwin:
/cc @rjbez17 

Ryan, we'll probably approve this while you're OOO but please feel free to take a look and send me any comments, and I can update the docs.

PR #762, opened 2020-05-25
Don't delete subnses with the Anchor CRD

adrianludwin:
A reasonable admin action is to call `kubectl delete -f manifests.yaml`,
but this will delete the CRDs and hence the subnamespace anchors. If any
namespace have cascading deletion enabled, this will delete the
namespaces themselves, which is surprising and destructive. This change
disables cascading deletion if the Anchor CRD itself is being deleted,
as opposed to any one anchor CR.

Tested: Added hack/test-delete-anchor-crd.sh and confirmed that it fails
on HEAD and passes with this fix. Also verified that log messages looked
good. Tried deleting some subnamespaces with and without setting
allowCascadingDelete and everything worked properly. Also reran
test-issue-501.sh, which _mostly_ worked although it looks like the
webhooks now prevent us from getting a SubnamespaceAnchorMissing
condition (which is good).

### DIFF
--- a/incubator/hnc/internal/reconcilers/anchor.go
+++ b/incubator/hnc/internal/reconcilers/anchor.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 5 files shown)


incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -135,17 +143,36 @@ func (r *AnchorReconciler) onDeleting(ctx context.Context, log logr.Logger, inst
...
 	case snsInst.Name == "":
-		log.Info("The subnamespace is already purged.")
+		log.Info("The subnamespace is already purged; allowing the anchor to be deleted")
 	case sOf != pnm:
-		log.Info("The subnamespace believes it is the subnamespace of another namespace.", "annotation", sOf)
-	case snsInst.DeletionTimestamp.IsZero() && !r.allowsCascadingDelete(cnm):
-		log.Info("The subnamespace is not being deleted and it doesn't allow cascading deletion.")
+		log.Info("The subnamespace believes it is the subnamespace of another namespace; allowing the anchor to be deleted", "annotation", sOf)
+	case snsInst.DeletionTimestamp.IsZero():
+		// onDelete has decided not to try to delete this namespace. Either cascading deletion isn't
```
yiqigao217, line 172: onDeleting

adrianludwin:
> Could you please also test the cascading delete? It took me a while to understand the `removeFinalizers()` logic and I think it's correct, but it may be safer to retest it after this change.

What should I test other than the unit tests? You just mean by hand?

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -135,17 +143,36 @@ func (r *AnchorReconciler) onDeleting(ctx context.Context, log logr.Logger, inst
...
 }
 
+// isDeletingCRD returns true if the Anchor CRD is being deleted. This can happen if HNC is being
+// uninstalled. In such cases, we shouldn't perform a cascading deletion.
+func (r *AnchorReconciler) isDeletingCRD(ctx context.Context) (bool, error) {
+	crd := &apiextensions.CustomResourceDefinition{}
+	nsn := types.NamespacedName{Name: api.Anchors + "." + api.MetaGroup}
+	if err := r.Get(ctx, nsn, crd); err != nil {
+		// Either the CRD wasn't found, in which case, HNC can't operate propertly and is hopefully
+		// being deleted itself. Otherwise, there was some transient error and we should just retry.
```
yiqigao217, line 153: For the first sentence, does it mean the CRD is already removed? What does "is hopefully being deleted itself" mean?

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: Done

yiqigao217:
> > Could you please also test the cascading delete? It took me a while to understand the `removeFinalizers()` logic and I think it's correct, but it may be safer to retest it after this change.
> 
> What should I test other than the unit tests? You just mean by hand?

We don't have unit test for cascading delete because the namespace cannot be deleted in test. Try `hack/test-issue-501.sh`? That's what we used before to test.

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: Updated comment

adrianludwin:
> We don't have unit test for cascading delete because the namespace cannot be deleted in test. Try `hack/test-issue-501.sh`? That's what we used before to test.

Ah, right. Done, see updating commit comment (also updated the PR description).

yiqigao217:
Thanks for retesting to confirm!
/lgtm
/approve
/hold
/assign @rjbez17

adrianludwin:
@rjbez17 lmk if you have any concerns and I'll address them in a followup PR.
/hold cancel

Issue #763, opened 2020-05-25
HNC: Implement HNC External Hierarchy


yiqigao217:
See design doc at https://bit.ly/hnc-external-hierarchy.

As stated in the doc, the implementation will include:
- Changes in the forest and HCR;
- Handling race conditions;
- Add new admission controllers or just additional rules on the existing ones.

yiqigao217:
The handling race conditions are already done with the current HNC and no changes are needed for it.

PR #764, opened 2020-05-25
Reconcile external namespaces and update tree labels

yiqigao217:
Part of #763 

Add `forest.Namespace.ExternalTreeLabels` field to store external
hierarchy in memory. Add `syncExternalNamespace()` func in
`HierarchyConfigReconciler` to update the external tree labels in the
forest. Enqueue descedants if the namespace is converted from external
to internal or vice versa so that the external tree labels can be
removed or propagated to the descendant namespaces.

Skip `syncSubnamespaceParent()`, `syncParent()`, `syncLabel()` for external
namespaces. In `syncLabel()`, for an internal namespace, if the root
ancestor is an external namespace, propagate all the external tree
labels too.

Tested by "make test" with three new integration test cases.

### DIFF
--- a/incubator/hnc/internal/forest/namespace.go
+++ b/incubator/hnc/internal/forest/namespace.go
@@ -41,6 +41,12 @@ type Namespace struct {
 
 	// Anchors store a list of anchors in the namespace.
 	Anchors []string
+
+	// ExternalTreeLabels stores external tree labels if this namespace is an external namespace.
+	// It will be empty if the namespace is not external. External namespace will at least have one
+	// tree label of itself. The key is the tree label without ".tree.hnc.x-k8s.io/depth" suffix.
+	// The value is the depth.
+	ExternalTreeLabels map[string]int
 }
 
 // Name returns the name of the namespace, of "<none>" if the namespace is nil.
@@ -135,4 +141,9 @@ func (ns *Namespace) SetAnchors(anchors []string) (diff []string) {
 
 	ns.Anchors = anchors
 	return
+}
+
+// IsExternal returns true if the namespace is not managed by HNC.
+func (ns *Namespace) IsExternal() bool {
+	return len(ns.ExternalTreeLabels) > 0
 }
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -117,6 +117,8 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 	origHC := inst.DeepCopy()
 	origNS := nsInst.DeepCopy()
 
+	// We will put finalizers on namespaces (the Hierarchy singleton actually)
+	// if there are anchors no matter the namespace is internal or external.
 	r.updateFinalizers(ctx, log, inst, nsInst, anms)
 
 	// Sync the Hierarchy singleton with the in-memory forest.
@@ -193,6 +195,9 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	hadCrit := ns.HasLocalCritCondition()
 	ns.ClearLocalConditions()
 
+	// Set external tree labels in the forest if this is an external namespace.
+	r.syncExternalNamespace(log, nsInst, ns)
+
 	// If this is a subnamespace, make sure .spec.parent is set correctly. Then sync the parent to the
 	// forest, and finally notify any relatives (including the parent) that might have been waiting
 	// for this namespace to be synced.
@@ -212,11 +217,49 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	r.syncLabel(log, nsInst, ns)
 }
 
+// syncExternalNamespace sets external tree labels to the namespace in the forest
+// if the namespace is an external namespace not managed by HNC.
+func (r *HierarchyConfigReconciler) syncExternalNamespace(log logr.Logger, nsInst *corev1.Namespace, ns *forest.Namespace) {
+	manager := nsInst.Annotations[api.AnnotationManagedBy]
+	if manager == "" || manager == api.MetaGroup {
+		// If the internal namespace is just converted from an external namespace,
+		// enqueue the descendants to remove the propagated external tree labels.
+		if ns.IsExternal() {
+			r.enqueueAffected(log, "subtree root converts from external to internal", ns.DescendantNames()...)
+		}
+		ns.ExternalTreeLabels = nil
+		return
+	}
+
+	// If the external namespace is just converted from an internal namespace,
+	// enqueue the descendants to propagate the external tree labels.
+	if !ns.IsExternal() {
+		r.enqueueAffected(log, "subtree root converts from internal to external", ns.DescendantNames()...)
+	}
+
+	// Get tree labels and set them in the forest. If there's no tree labels on the
+	// namespace, set only one tree label of itself.
+	etls := make(map[string]int)
+	for tl, d := range nsInst.Labels {
+		enm := strings.TrimSuffix(tl, api.LabelTreeDepthSuffix)
+		if enm != tl {
+			etls[enm], _ = strconv.Atoi(d)
+		}
+	}
+	etls[ns.Name()] = 0
+	ns.ExternalTreeLabels = etls
+}
+
 // syncSubnamespaceParent sets the parent to the owner and updates the SubnamespaceAnchorMissing
 // condition if the anchor is missing in the parent namespace according to the forest. The
 // subnamespaceOf annotation is the source of truth of the ownership (e.g. being a subnamespace),
 // since modifying a namespace has higher privilege than what HNC users can do.
 func (r *HierarchyConfigReconciler) syncSubnamespaceParent(log logr.Logger, inst *api.HierarchyConfiguration, nsInst *corev1.Namespace, ns *forest.Namespace) {
+	if ns.IsExternal() {
+		ns.IsSub = false
+		return
+	}
+
 	pnm := nsInst.Annotations[api.SubnamespaceOf]
 	if pnm == "" {
 		ns.IsSub = false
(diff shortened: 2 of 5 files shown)


incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -212,11 +218,36 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	r.syncLabel(log, nsInst, ns)
 }
 
+// syncExternalNamespace sets external tree labels to the namespace in the forest
+// if the namespace is an external namespace not managed by HNC.
+func (r *HierarchyConfigReconciler) syncExternalNamespace(log logr.Logger, nsInst *corev1.Namespace, ns *forest.Namespace) {
+	manager := nsInst.Annotations[api.AnnotationManagedBy]
+	if manager == "" || manager == api.MetaGroup {
+		return
```
adrianludwin, line 226: You should explicitly set ns.ExternalTreeLabels to `nil` here in case this _used_ to be an external namespace but isn't anymore (and add a unit test for this).

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -316,11 +352,22 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
...
+		l := anc.Name() + api.LabelTreeDepthSuffix
 		metadata.SetLabel(nsInst, l, strconv.Itoa(depth))
 		if anc.HasLocalCritCondition() {
 			break
 		}
+
+		// If the root is an external namespace, add all its external tree labels too.
+		// Note it's impossible to have an external namespace as a non-root, which is
+		// enforced by both admission controllers and the reconciler here.
+		if anc.IsExternal() && anc.Parent() == nil {
```
adrianludwin, line 364: So why are you checking for `anc.Parent()`? Just checking for `anc.IsExternal()` should be enough, and then you can `break` at the end of it.

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Good catch! Yes, I should set `ns.ExternalTreeLabels` to `nil` here. Additionally, I added if the namespace is just converted from external to internal, enqueue all the descendants to update the tree labels (remove propagated external tree labels). I added the unit test first and caught the error of not updating the descendants.

After this change, the new test case is passed. PTAL at the test cases.

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Ah right, done!

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Added "enqueue descendants" for the other case too (converted from internal to external) and the new test case.

incubator/hnc/internal/forest/namespace.go
```diff
@@ -136,3 +142,11 @@ func (ns *Namespace) SetAnchors(anchors []string) (diff []string) {
 	ns.Anchors = anchors
 	return
 }
+
+// IsExternal returns true if the namespace is not managed by HNC.
+func (ns *Namespace) IsExternal() bool {
+	if len(ns.ExternalTreeLabels) > 0 {
+		return true
+	}
+	return false
```
adrianludwin, line 151: `return len(ns.ExternalTreeLabels) > 0` ?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -193,6 +195,10 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	hadCrit := ns.HasLocalCritCondition()
 	ns.ClearLocalConditions()
 
+	// Set external tree labels in the forest if this is an external namespace. It skips some of the
+	// syncs below accordingly.
```
adrianludwin, line 199: Maybe take out the second sentence? At the top level of this function, nothing gets skipped, and I was a bit confused at first.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -212,11 +218,48 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
...
+			r.enqueueAffected(log, "subtree parent converts from external to internal", ns.DescendantNames()...)
+		}
+		ns.ExternalTreeLabels = nil
+		return
+	}
+
+	// If the external namespace is just converted from an internal namespace,
+	// enqueue the descendants to propagate the external tree labels.
+	if !ns.IsExternal() {
+		r.enqueueAffected(log, "subtree parent converts from internal to external", ns.DescendantNames()...)
```
adrianludwin, line 238: Nit: we usually use the term "subtree root" (here and on line 229)

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -257,6 +300,10 @@ func (r *HierarchyConfigReconciler) markExisting(log logr.Logger, ns *forest.Nam
 }
 
 func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.HierarchyConfiguration, ns *forest.Namespace) {
+	if ns.IsExternal() {
+		return
+	}
```
adrianludwin, line 305: Shouldn't we call `ns.SetParent(nil)` in case there was a parent in the HC?

incubator/hnc/internal/reconcilers/hierarchy_config_test.go
```diff
@@ -135,9 +131,105 @@ var _ = Describe("Hierarchy", func() {
...
+		Eventually(getLabel(ctx, barName, "ext2"+api.LabelTreeDepthSuffix)).Should(Equal("3"))
+		Eventually(getLabel(ctx, barName, "ext1"+api.LabelTreeDepthSuffix)).Should(Equal("2"))
+		// Even if baz doesn't have the tree label of itself with depth 0 when created,
+		// the tree label of itself is added to the forest automatically.
+		Eventually(getLabel(ctx, barName, bazName+api.LabelTreeDepthSuffix)).Should(Equal("1"))
+		Eventually(getLabel(ctx, barName, barName+api.LabelTreeDepthSuffix)).Should(Equal("0"))
+
+		Eventually(getLabel(ctx, bazName, "ext2"+api.LabelTreeDepthSuffix)).Should(Equal("2"))
+		Eventually(getLabel(ctx, bazName, "ext1"+api.LabelTreeDepthSuffix)).Should(Equal("1"))
+		Eventually(getLabel(ctx, bazName, bazName+api.LabelTreeDepthSuffix)).Should(Equal(""))
```
adrianludwin, line 161: Why is this the empty string and not "0"?

incubator/hnc/internal/reconcilers/hierarchy_config_test.go
```diff
@@ -135,9 +131,105 @@ var _ = Describe("Hierarchy", func() {
...
+			"ext2" + api.LabelTreeDepthSuffix: "2",
+			"ext1" + api.LabelTreeDepthSuffix: "1",
+		}
+		ns.SetLabels(l)
+		updateNamespace(ctx, ns)
+
+		// Verify all the labels
+		Eventually(getLabel(ctx, barName, "ext2"+api.LabelTreeDepthSuffix)).Should(Equal("3"))
+		Eventually(getLabel(ctx, barName, "ext1"+api.LabelTreeDepthSuffix)).Should(Equal("2"))
+		// Even if baz doesn't have the tree label of itself with depth 0 when created,
```
adrianludwin, line 187: There's no `baz` in this test

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: sg

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Good catch, yes, I also added `ns.IsSub = false` in the early exit in `syncSubnamespaceParent()`.

incubator/hnc/internal/reconcilers/hierarchy_config_test.go
yiqigao217: Updated!

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -300,12 +348,14 @@ func (r *HierarchyConfigReconciler) syncAnchors(log logr.Logger, ns *forest.Name
 }
 
 func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Namespace, ns *forest.Namespace) {
-	// Pre-define label depth suffix
-	labelDepthSuffix := fmt.Sprintf(".tree.%s/depth", api.MetaGroup)
+	if ns.IsExternal() {
+		metadata.SetLabel(nsInst, nsInst.Name+api.LabelTreeDepthSuffix, "0")
+		return
+	}
```
yiqigao217, line 354: Updated here to always set tree label of itself on external namespaces and updated the test cases. (Previously, HNC didn't change/set any labels on external namespaces and only automatically add the tree label of itself with `depth=0` in the forest).

incubator/hnc/internal/reconcilers/hierarchy_config_test.go
yiqigao217: Updated the logic and thus having "0" here now.

yiqigao217:
> I'm wondering if we should put `foo.tree.hnx.x-k8s.io/depth=0` on externally-managed namespace `foo` if it's missing? It's a bit weird in the tests for the namespace to be missing its own label.

Agreed, since we still need the `depth=0` label to select the subtree of an external namespace. Update the early exit in `syncLabels` to add that label for external namespace and updated the test cases.

incubator/hnc/internal/reconcilers/hierarchy_config.go
adrianludwin: Can you add this to the design doc?

yiqigao217:
> Can you add this to the design doc?

Done!

PR #766, opened 2020-05-26
Allow recreation of missing subns anchors

adrianludwin:
Without this change, if a subnamespace is missing its anchor, the
condition can never be resolved. This change allows an admin to fix the
problem without disabling the validating webhook.

Tested: new unit test; verified by hand that I can recreate missing
anchors but only with the correct parent.

/assign @yiqigao217


adrianludwin:
> lgtm, just a typo in commit message and PR message. "if".

Fixed, thanks!

PR #767, opened 2020-05-26
Document uninstallation, backup and restore

adrianludwin:
Also update some unclear Gitops directions.

### DIFF
--- a/incubator/hnc/docs/user-guide/best-practices.md
+++ b/incubator/hnc/docs/user-guide/best-practices.md
@@ -75,20 +75,30 @@ kubectl hns describe foo
 
 Since HNC is controlled by regular Kubernetes objects, you can check your YAML
 files into source control and apply them to your cluster(s) via `kubectl apply
--f`. However, HNC does impose two restrictions on the order in which changes can
-be applied:
+-f`. You generally will only want to do this for your cluster-wide configuration
+(the `HNCConfiguration` object) and your full namespace configurations (the
+`HierarchyConfiguration` objects in each namespace), and not your subnamespaces
+(the `SubnamespaceAnchor` objects) since subnamespaces are mainly for
+unprivileged users, not Gitops flows.
+
+When applying your `HierarchyConfiguration` objects, HNC imposes two
+restrictions on the order in which changes can be applied:
 
 1. A namespace must exist before it can be referenced as the parent of another
-   namespace. However, the would-be parent’s `HierarchicalConfiguration` does
-   _not_ need to exist.
+   namespace.
 1. You may not create any cycles between namespaces. For example, assume that
    Namespace A is the parent of B, and you wish to reverse this relationship so
    that B becomes the parent of A. If A’s config is applied before B’s, this
    will result in a cycle.
 
-In both cases, HNC’s validating admission controllers will reject the change.
-However, in many cases, simply re-running the `apply` operation will resolve the
-issue:
+If either condition is violated, HNC’s validating admission controllers will
+reject the change. However, in Gitops flows, it is possible to transiently
+violate both conditions. For example, a namespace may not be fully created
+before it is referenced as a parent, or you might change the parents of multiple
+namespaces simultaneously, resulting in a transient cycle between them.
+
+Fortunately, in most cases, simply re-running the `apply` operation will resolve
+any issues:
 
 1. If all namespaces are created during the first application, the second
    application will successfully allow them to be referenced as parents.
--- a/incubator/hnc/docs/user-guide/how-to.md
+++ b/incubator/hnc/docs/user-guide/how-to.md
@@ -16,6 +16,7 @@ This document describes common tasks you might want to accomplish using HNC.
 * [Administer HNC](#admin)
   * [Install or upgrade HNC on a cluster](#admin-install)
   * [Uninstall HNC from a cluster](#admin-uninstall)
+  * [Backing up and restoring HNC data](#admin-backup-restore)
   * [Administer who has access to HNC properties](#admin-access)
   * [Modify the object types propagated by HNC](#admin-types)
   * [Gather metrics](#admin-metrics)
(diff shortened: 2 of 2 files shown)


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 

This is a bit subtle; does this look good enough for now?

cc @sophieliu15 , @adrianchung

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -388,19 +389,70 @@ make deploy
...
-kubectl delete -f https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/hnc-manager.yaml
+kubectl -n hnc-system delete deployment hnc-controller-manager
+kubectl delete validatingwebhookconfiguration.admissionregistration.k8s.io hnc-validating-webhook-configuration
+```
+
+You may also completely delete HNC, including its CRDs and namespaces. However,
+**this is a destructive process that results in data loss.** In particular, you
+will lose any cluster-wide configuration in your `HNCConfig` object, as well as
+any hierarchical relationships between different namespaces, _excluding_
+subnamespaces (subnamespace relationships are saved as annotations on the
```
adrianludwin, line 403: Note that this will only be true after #762 is merged.

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -388,19 +389,70 @@ make deploy
...
+to delete the `.metadata.uid` and `metadata.resourceVersion` fields from each
+object **prior** to applying the file.  It may also be convenient to delete the
+`.metadata.selfLink` field, which is alphabetically between the other two
+fields; this is safe.
+
+Once you are ready, first reapply the cluster-wide configuration via `kubectl
+apply -f hncconfig.yaml`, and then the structural relationships via `kubectl
+apply -f structure.yaml`.
+
+Finally, resolve any `SubnamespaceAnchorMissing` conditions. Type `kubectl hns
```
adrianludwin, line 451: This requires #766 to be merged.

incubator/hnc/docs/user-guide/best-practices.md
```diff
@@ -75,20 +75,30 @@ kubectl hns describe foo
...
+`HierarchyConfiguration` objects in each namespace), and not your subnamespaces
+(the `SubnamespaceAnchor` objects) since subnamespaces are mainly for
+unprivileged users, not Gitops flows.
+
+When applying your `HierarchyConfiguration` objects, HNC imposes two
+restrictions on the order in which changes can be applied:
 
 1. A namespace must exist before it can be referenced as the parent of another
-   namespace. However, the would-be parent’s `HierarchicalConfiguration` does
-   _not_ need to exist.
```
yiqigao217, line 83: Why is this removed?

adrianludwin:
> Also do we need to mention that not only the hierarchy will be gone but also the propagated objects?

That's a good point. I'm wondering if we should prevent HNC from removing the propagated objects if the CRDs get deleted. After all, if the pod dies first, everything will stay in place; if the CRDs _start_ getting deleted, _then_ the pod dies, then an unknown number of objects will be left around. Do you agree?

incubator/hnc/docs/user-guide/best-practices.md
adrianludwin: It just seemed redundant to me - I didn't think it would be a question that people would ask.

adrianludwin: wdyt?

yiqigao217: sg

yiqigao217:
> I'm wondering if we should prevent HNC from removing the propagated objects if the CRDs get deleted. After all, if the pod dies first, everything will stay in place; if the CRDs _start_ getting deleted, _then_ the pod dies, then an unknown number of objects will be left around. Do you agree?

Agreed. Just confirming we don't have this prevention right now, right?


> @yiqigao217 We don't put finalizers on anything other than `SubnamespaceAnchors` for cascading deletion, right? So if all anchors are deleted (because the CRD is removed), there should be nothing stopping users from deleting their namespaces?

We also put finalizers on `HierarchyConfiguration` objects if the namespace has any subnamespace. Therefore, subnamespace can be deleted by users, but any namespace (including subnamespaces) with subnamespaces cannot be deleted.

adrianludwin:
> > I'm wondering if we should prevent HNC from removing the propagated objects if the CRDs get deleted. After all, if the pod dies first, everything will stay in place; if the CRDs _start_ getting deleted, _then_ the pod dies, then an unknown number of objects will be left around. Do you agree?
> 
> Agreed. Just confirming we don't have this prevention right now, right?

That's right - we don't.

> We also put finalizers on `HierarchyConfiguration` objects if the namespace has any subnamespace. Therefore, subnamespace can be deleted by users, but any namespace (including subnamespaces) with subnamespaces cannot be deleted. But if the `HierarchyConfiguration` CRD is also removed, there should be nothing stopping users from deleting any namespaces.

Ok, I just tried this and it works. Thanks! Fortunately, it looks like deleting a CRD isn't blocked by finalizers.

yiqigao217:
> > We also put finalizers on `HierarchyConfiguration` objects if the namespace has any subnamespace. Therefore, subnamespace can be deleted by users, but any namespace (including subnamespaces) with subnamespaces cannot be deleted. But if the `HierarchyConfiguration` CRD is also removed, there should be nothing stopping users from deleting any namespaces.
> 
> Ok, I just tried this and it works. Thanks! Fortunately, it looks like deleting a CRD isn't blocked by finalizers.

Nice!

/lgtm
/approve
/hold
/assign @rjbez17

adrianludwin:
@rjbez17 lmk if you have any concerns and I'll address them in a followup PR.

PR #769, opened 2020-05-26
Update admission controllers for external hierarchy

yiqigao217:
Part of #763 

Add additional restrictions in admission controllers to deny:
1. Create namespace with existing name in external hierarchy;
2. Update namespace to add "managedBy: other" annotation on namespace
with parent;
3. Make any changes to HierarchyConfiguration in external namespace.

Tested manually on a GKE cluster and by "make test" with new test cases.

### DIFF
--- a/incubator/hnc/internal/forest/namespace.go
+++ b/incubator/hnc/internal/forest/namespace.go
@@ -42,6 +42,10 @@ type Namespace struct {
 	// Anchors store a list of anchors in the namespace.
 	Anchors []string
 
+	// Manager stores the manager of the namespace. The default value
+	// "hnc.x-k8s.io" means it's managed by HNC.
+	Manager string
+
 	// ExternalTreeLabels stores external tree labels if this namespace is an external namespace.
 	// It will be empty if the namespace is not external. External namespace will at least have one
 	// tree label of itself. The key is the tree label without ".tree.hnc.x-k8s.io/depth" suffix.
--- a/incubator/hnc/internal/validators/hierarchy.go
+++ b/incubator/hnc/internal/validators/hierarchy.go
@@ -176,6 +176,11 @@ func (v *Hierarchy) checkNS(ns *forest.Namespace) admission.Response {
 
 // checkParent validates if the parent is legal based on the current in-memory state of the forest.
 func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admission.Response {
+	if ns.IsExternal() && newParent != nil {
+		msg := fmt.Sprintf("Namespace %q is managed by %q, not HNC, so it cannot have a parent in HNC.", ns.Name(), ns.Manager)
+		return deny(metav1.StatusReasonForbidden, msg)
+	}
+
 	if curParent == newParent {
 		return allow("parent unchanged")
 	}
--- a/incubator/hnc/internal/validators/hierarchy_test.go
+++ b/incubator/hnc/internal/validators/hierarchy_test.go
@@ -52,6 +52,50 @@ func TestStructure(t *testing.T) {
 	}
 }
 
+func TestChangeParentOnManagedBy(t *testing.T) {
+	f := foresttest.Create("-a-c") // a <- b; c <- d
+	h := &Hierarchy{Forest: f}
+	l := zap.Logger(false)
+
+	// Make c and d external namespaces
+	f.Get("c").ExternalTreeLabels = map[string]int{"c" + api.LabelTreeDepthSuffix: 0}
+	f.Get("d").ExternalTreeLabels = map[string]int{"d" + api.LabelTreeDepthSuffix: 0}
+
+	// These cases test changing parent for internal or external namespaces, described
+	// in the table at https://bit.ly/hnc-external-hierarchy#heading=h.z9mkbslfq41g
+	// with other cases covered in the namespace_test.go.
+	tests := []struct {
+		name string
+		nnm  string
+		pnm  string
+		fail bool
+	}{
+		{name: "ok: change internal namespace parent from none to existing", nnm: "a", pnm: "c"},
+		{name: "ok: change internal namespace existing parent", nnm: "b", pnm: "c"},
+		{name: "ok: change internal namespace parent from existing to none", nnm: "b", pnm: ""},
+		{name: "not ok: change external namespace parent from none to existing", nnm: "c", pnm: "a", fail: true},
+		{name: "not ok: change external namespace existing parent", nnm: "d", pnm: "a", fail: true},
+		{name: "ok: change external namespace parent from existing to none", nnm: "d", pnm: ""},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// Setup
+			g := NewGomegaWithT(t)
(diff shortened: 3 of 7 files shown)


adrianludwin:
I'm going to hold off on reviewing this until the earlier changes go in.
But as one quick question, are all the transitions from the table in the
design doc covered here?

yiqigao217:
/hold until #764 is merged first.
/assign @adrianludwin 
/assign @rjbez17 

Hi Adrian and Ryan , in #764 , I have the HierarchyConfiguration reconciler to always set the external namespace parent to `nil` in the forest. Therefore, I have the changing parent rules in the [design doc table](https://docs.google.com/document/d/1DSygeGUhjhl6d-ynS7PfhCAI_zqeMAkVizSay_CkmuU/edit#heading=h.z9mkbslfq41g) all combined into one "denying any HC request on external namespaces" here.

Let me rethink it through and please lmk what you think.

yiqigao217:
> Hi Adrian and Ryan , in #764 , I have the HierarchyConfiguration reconciler to always set the external namespace parent to `nil` in the forest. Therefore, I have the changing parent rules in the [design doc table](https://docs.google.com/document/d/1DSygeGUhjhl6d-ynS7PfhCAI_zqeMAkVizSay_CkmuU/edit#heading=h.z9mkbslfq41g) all combined into one "denying any HC request on external namespaces" here.
> 
> Let me rethink it through and please lmk what you think.

FYI, I will update the HC validator here back to more granular rules as described in the rule table in the design doc.

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -176,6 +176,11 @@ func (v *Hierarchy) checkNS(ns *forest.Namespace) admission.Response {
 
 // checkParent validates if the parent is legal based on the current in-memory state of the forest.
 func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admission.Response {
+	if ns.IsExternal() && newParent != nil {
+		msg := fmt.Sprintf("Namespace %q is an external namespace that cannot have parent. Please unset the parent if it exists.", ns.Name())
```
adrianludwin, line 180: Maybe replace "is an external namespace" to "is managed by %q"?

adrianludwin: I'd also just leave out the "Please unset the parent if it exists" part, it's fine to just say "it can't have a parent."

incubator/hnc/internal/validators/hierarchy_test.go
```diff
@@ -52,6 +52,49 @@ func TestStructure(t *testing.T) {
...
+func TestChangeParentOnManagedBy(t *testing.T) {
+	f := foresttest.Create("-a-c") // a <- b; c <- d
+	h := &Hierarchy{Forest: f}
+	l := zap.Logger(false)
+
+	// Make c and d external namespaces
+	f.Get("c").ExternalTreeLabels = map[string]int{"c" + api.LabelTreeDepthSuffix: 0}
+	f.Get("d").ExternalTreeLabels = map[string]int{"d" + api.LabelTreeDepthSuffix: 0}
+
+	// The following 6 test cases matches the last 6 cases in the table at
```
adrianludwin, line 64: You should say which test does the other four. Also, maybe there's a better description than "the last 6 cases" - what makes those cases similar to each other so they can be tested together?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -79,7 +79,26 @@ func (v *Namespace) checkForest(req *nsRequest) admission.Response {
 
 	ns := v.Forest.Get(req.ns.Name)
 
-	// Early exit to allow non-delete requests or if the namespace allows cascading deletion.
+	// Allow all create namespace requests except if the namespace name already exists in
+	// the external hierarchy (the default webhook already denies existing namespace names).
```
adrianludwin, line 83: What's "the default webhook"?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -79,7 +79,26 @@ func (v *Namespace) checkForest(req *nsRequest) admission.Response {
...
+	if req.op == v1beta1.Create {
+		for _, nm := range v.Forest.GetNamespaceNames() {
+			if _, ok := v.Forest.Get(nm).ExternalTreeLabels[req.ns.Name]; ok {
+				return deny(metav1.StatusReasonAlreadyExists, "The name is already used. Please us a different name.")
```
adrianludwin, line 87: Instead of "The name is already used," I think we need a more specific error message. For example, "The namespace name %q is reserved by the external hierarchy manager %q"?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -79,7 +79,26 @@ func (v *Namespace) checkForest(req *nsRequest) admission.Response {
...
+			}
+		}
+	}
+
+	// Allow all update namespace requests except if the namespace has a parent and wants
+	// to add "hnc.x-k8s.io/managedBy" annotation other than "hnc.x-k8s.io".
+	if req.op == v1beta1.Update && req.ns.Annotations[api.AnnotationManagedBy] != "" {
+		if req.ns.Annotations[api.AnnotationManagedBy] != api.MetaGroup && ns.Parent() != nil {
+			msg := fmt.Sprintf("Namespace %s is a child of %s. Namespaces with parent cannot be converted to external namespaces. Please unset the parent first.", ns.Name(), ns.Parent().Name())
```
adrianludwin, line 96: Use %q instead of %s (I've found it reads much more easily). Also how about the following change?

> Namespace %q is a child of %q. Namespaces with parents defined by HNC cannot also be managed externally. To manage this namespace with %q, first make it a root in HNC.

adrianludwin:
#764 is merged, so removing hold
/hold cancel

incubator/hnc/internal/validators/namespace.go
yiqigao217: Updated to apiserver instead of default webhook. ( I thought the request is denied by default webhook but it turned out to be apiserver itself).

incubator/hnc/internal/forest/namespace.go
```diff
@@ -27,6 +27,10 @@ type Namespace struct {
 	exists               bool
 	allowCascadingDelete bool
 
+	// manager stores the manager of the namespace. The default value
+	// "hnc.x-k8s.io" means it's managed by HNC.
+	manager string
```
adrianludwin, line 32: Maybe just make this public and get rid of the accessors? This isn't Java 😁

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -176,6 +176,11 @@ func (v *Hierarchy) checkNS(ns *forest.Namespace) admission.Response {
 
 // checkParent validates if the parent is legal based on the current in-memory state of the forest.
 func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admission.Response {
+	if ns.IsExternal() && newParent != nil {
+		msg := fmt.Sprintf("Namespace %q is managed by %q that cannot have parent.", ns.Name(), ns.Manager())
```
adrianludwin, line 180: Maybe: "Namespace %q is managed by %q, not HNC, so it cannot have a parent in HNC."

PR #774, opened 2020-05-27
Removes CannotUpdate when source object deleted

### LINKED ISSUE #771: HNC: CannotUpdate conditions are not cleared if the source object is deleted
To reproduce:

```
# Set up a simple tree
k create ns a
k create ns b
k hns set b --parent a

# Create a rolebinding HNC can't propagate
k create rolebinding --clusterrole=cluster-admin --serviceaccount=default:default -n a foo
k hns tree a # observe conditions

k delete rolebinding -n a foo
k hns tree a # conditions not removed from "b", only "a"
```

adrianludwin:
See issue #771. The CannotUpdate condition wasn't being cleared, since
conditions are indexed on the source object, but if the source object is
deleted and the propagated object (and its inheritedFrom label) was
never saved to the apiserver, we have no way to look up the condition by
the source object. This change simply looks through *all* conditions on
the same namespace as the missing object and removes the appropriate
ones; see the comments for more details.

This change also improves the logging for objects so that every action
(create, update or delete) is logged, as are the reasons for enqueuing
other objects or HCs for reconciliation. This should make postmortems
easier.

Tested: new testing script (hack/test-issue-771.sh) fails without this
fix and passes with it. Existing test scripts all pass as expected.

Fixed #771

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -193,6 +193,15 @@ type Condition struct {
 	Affects []AffectedObject `json:"affects,omitempty"`
 }
 
+func (c Condition) String() string {
+	affects := fmt.Sprint(c.Affects)
+	msg := c.Msg
+	if len(msg) > 100 {
+		msg = msg[:100] + "..."
+	}
+	return fmt.Sprintf("%s: %s (affects %s)", c.Code, msg, affects)
+}
+
 // AffectedObject defines uniquely identifiable objects.
 type AffectedObject struct {
 	Group     string `json:"group,omitempty"`
@@ -230,8 +239,8 @@ func (a AffectedObject) String() string {
 	}
 
 	// No namespace -> it *is* a namespace
-	if a.Namespace != "" {
-		return a.Namespace
+	if a.Namespace == "" {
+		return a.Name
 	}
 
 	// Generic object (note that Group may be empty for core objects, don't worry about it)
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -51,6 +51,8 @@ const (
 	actionRemove  syncAction = "remove"
 	actionWrite   syncAction = "write"
 	actionNop     syncAction = "no-op"
+
+	unknownSourceNamespace = "<unknown-source-namespace>"
 )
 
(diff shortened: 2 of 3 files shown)


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 
cc @sophieliu15, @adrianchung

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -601,6 +615,59 @@ func hasPropagatedLabel(inst *unstructured.Unstructured) bool {
...
+	if sao.Namespace != unknownSourceNamespace {
+		return false
+	}
+
+	cleared := false
+	for _, cond := range ns.Conditions() {
+		for _, ao := range cond.Affects {
+			aoCmp := ao
+			aoCmp.Namespace = unknownSourceNamespace // to make comparison easier
+			if aoCmp != sao {
```
yiqigao217, line 659: why not just compare `if ao.Namespace != unknownSourceNamespace`?

incubator/hnc/api/v1alpha1/hierarchy_types.go
```diff
@@ -230,8 +239,8 @@ func (a AffectedObject) String() string {
 	}
 
 	// No namespace -> it *is* a namespace
-	if a.Namespace != "" {
-		return a.Namespace
+	if a.Namespace == "" {
+		return a.Name
```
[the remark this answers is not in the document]
adrianludwin: It's used for logging, I think. I haven't looked at logs for the object reconciler for a while so probably just didn't notice.

incubator/hnc/internal/reconcilers/object.go
adrianludwin: The actual comparison we want is:

```
if sao.Group == ao.Group &&
  sao.Version == ao.Version &&
  sao.Kind == ao.Kind &&
  sao.Name == ao.Name {
  ...
}
```
But it's easier to make a copy of `ao`, overwrite the field we know we don't care about, and compare the rest of the struct in one line.

yiqigao217: Ah ok got you. I was confused by the new `unknownSourceNamespace` since it replaced the `source-namespace-placeholder`, which was initially used to mark propagated object at the beginning. Now this value is used in `onMissing()`.

adrianludwin:
/hold cancel
As discussed, @rjbez17 lmk if you have any concerns

PR #775, opened 2020-05-27
Don't unpropagate objects after CRDs are deleted

adrianludwin:
This change ensures that if the HierarchyConfig CRD is deleted before
the HNC pod, it will not delete the propagated objects. This reduces the
surprises when uninstalling HNC.

Tested: verified by hand that I can still delete the HC and this has the
effect of resetting the parent, and that cascading deletion still works.
Wrote a new test that verifies that the objects are not deleted after
five seconds, and verified that this test fails on HEAD but passes with
this patch.

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -17,6 +17,7 @@ package reconcilers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -25,7 +26,8 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 

cc @sophieliu15 @adrianchung

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -93,32 +95,33 @@ func (r *HierarchyConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
 		}
 		return err
 	}
+	origNS := nsInst.DeepCopy()
+
 	// Get singleton from apiserver. If it doesn't exist, initialize one.
 	inst, err := r.getSingleton(ctx, nm)
 	if err != nil {
 		return err
 	}
```
yiqigao217, line 115: Do we need to have the CRD being deleted case to return nil to not trigger a retry here?

adrianludwin: What did you have in mind if the CRD's been fully deleted?

yiqigao217: It seemed to me that when the CRD's been fully deleted, it will keep retrying even if the object is gone and reconciler watch won't be triggered. It will keep retrying until the reconciler is torn down, right?

Do you mean it's okay to have it keep retrying until the reconciler is down?

adrianludwin: I think so? I mean, what's the alternative? We could `os.Exit(0)` but then the pod will just start getting restarted, which isn't really much better.

yiqigao217: Oh I mentioned returning `nil` in the first comment.

yiqigao217: What about returning `nil, nil` in `getSingleton()` if the CRD is deleted and here if getting `nil, nil`, return `nil` instead of error to avoid retry? However, if doing so, and if the getting CRD just got a transient error, it won't retry while it should. Hmm, it seems too compllicated.

My only question to keep retrying is if the CRD is removed while HNC is actually not going to tear down, then the retry will never ends? BTW do we need to add webhook for users not to delete CRDs when HNC is supposed to be running?

adrianludwin: Ah, I see what you're saying. Yes, technically the reconciler could just early-exit if the CRD is not-found or is being deleted, but I don't mind if it spins for a while either and emits a lot of error messages (you _should_ see a lot of error messages if the pod isn't destroyed right away).

The retries have an exponential backoff so they'll get less and less frequent. I think that's good behaviour.

No, we shouldn't add a webhook to prevent this IMO. Deleting CRDs is a _very_ highly privileged operation, and it makes HNC _much_ harder to uninstall. I think we're fine for now, we can always come back to this later.

yiqigao217: Thanks for the explanation, sgtm!

adrianludwin:
/hold cancel
As discussed, @rjbez17 lmk if you have any concerns

PR #779, opened 2020-06-01
[HNC v0.4] Make Stackdriver optional and fix startup logs

adrianludwin:
The Stackdriver exporter shouldn't block HNC from starting; log any
error but ignore it and keep going. Also, ensure that we can display
startup logs that are generated before the manager starts.

Tested: HNC now works in Kind again, but is still exporting metrics to
SD on GKE. Startup logs are now showing up throughout the startup
process.

Fixes #770 
Fixes #777

PR #780, opened 2020-06-01
Make Stackdriver optional and fix startup logs

adrianludwin:
The Stackdriver exporter shouldn't block HNC from starting; log any
error but ignore it and keep going. Also, ensure that we can display
startup logs that are generated before the manager starts.

Tested: HNC now works in Kind again, but is still exporting metrics to
SD on GKE. Startup logs are now showing up throughout the startup
process.


yiqigao217:
Is this a dup of #779 ?

adrianludwin:
#779 was for 0.4.0, this is for the master branch.

PR #788, opened 2020-06-04
Make leaf subnamespace deletion easier

### LINKED ISSUE #716: HNC: consider making leaf subnamespace deletion easier
Right now, to delete a subnamespace, you first have to set `allowCascadingDelete` on the subnamespace itself (or its parent). This makes subnamespaces _harder_ to delete than regular namespaces, which have no such protection.

The goal of `allowCascadingDelete` was really to prevent _cascading_ deletions - that is, unintentionally deleting _multiple_ namespaces. I wonder if we could skip this check unless the namespace being deleted actually has subnamespaces itself. So for example, I could say:

```
k hns create sub -n parent
k delete subns sub -n parent
```

But not:
```
k hns create sub -n parent
k hns create subsub -n sub
k delete subns sub -n parent # fails because of subsub
```

yiqigao217:
Update anchor reconciler and webhook to allow deleting leaf subnamespace
even when allowCascadingDelete is not set.

Tested by three test-issue-716.sh files and new test cases in anchor
validator that the test only passed after the change.

Fixes #716

### DIFF
--- a/incubator/hnc/hack/test-issue-716-full.sh
+++ b/incubator/hnc/hack/test-issue-716-full.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+bold=$(tput bold)
+normal=$(tput sgr0)
+
+# Set up namespace structure
+echo "----------------------------------------------------"
+echo "${bold}Setting up a 2-level tree with 'a' as the root and 'b' as a child of 'a'${normal}"
+echo "Creating the root 'a'"
+kubectl create ns a
+sleep 1
+echo "Creating the child namespace 'b' for 'a'"
+kubectl create ns b
+kubectl hns set b --parent a
+
+echo "${bold}Here's the outcome of the tree hierarhy:${normal}"
+kubectl hns tree a
+
+echo "----------------------------------------------------"
+echo "${bold}Test-1:${normal} Remove parent namespace 'a'."
+echo "${bold}Operation:${normal} delete 'a' - kubectl delete ns a"
+kubectl delete ns a
+echo "${bold}Expected:${normal} Delete successfully."
+echo "${bold}Result:${normal} b should have 'CritParentMissing' condition:"
+kubectl hns tree b
+
+echo "----------------------------------------------------"
+echo "${bold}Test-2:${normal} Remove the orphan 'b'."
+echo "${bold}Operation:${normal} delete 'b' - kubectl delete ns b"
+kubectl delete ns b
+echo "${bold}Expected:${normal} Delete successfully."
+
+echo "----------------------------------------------------"
+echo "${bold}Cleaning up${normal}"
+kubectl delete ns a
+kubectl delete ns b
--- a/incubator/hnc/internal/reconcilers/anchor.go
+++ b/incubator/hnc/internal/reconcilers/anchor.go
@@ -126,11 +126,11 @@ func (r *AnchorReconciler) onDeleting(ctx context.Context, log logr.Logger, inst
 		// We've finished processing this, nothing to do.
 		log.Info("Do nothing since the finalizers are already gone.")
 		return true, nil
-	case !deletingCRD && snsInst.DeletionTimestamp.IsZero() && r.allowsCascadingDelete(cnm):
-		// The subnamespace is not already being deleted but it allows cascadingDelete.
+	case r.shouldDeleteSubns(cnm, snsInst, deletingCRD):
+		// The subnamespace is not already being deleted but it allows cascadingDelete or it's a leaf.
 		// Delete the subnamespace, unless the CRD is being deleted, in which case, we want to leave the
 		// namespaces alone.
-		log.Info("The subnamespace is not being deleted but it allows cascading deletion.")
+		log.Info("The subnamespace is not being deleted but it allows cascading deletion or it's a leaf.")
 		return true, r.deleteNamespace(ctx, log, snsInst)
 	case r.removeFinalizers(log, inst, snsInst):
 		// We've determined that this anchor is ready to be deleted.
@@ -156,6 +156,29 @@ func (r *AnchorReconciler) isDeletingCRD(ctx context.Context) (bool, error) {
 	}
 
 	return !crd.DeletionTimestamp.IsZero(), nil
+}
+
+// shouldDeleteSubns returns true if the namespace still exists and it is a leaf
+// subnamespace or it allows cascading delete unless the CRD is being deleted.
+func (r *AnchorReconciler) shouldDeleteSubns(nm string, inst *corev1.Namespace, deletingCRD bool) bool {
+	r.forest.Lock()
+	defer r.forest.Unlock()
+
+	// If the CRD is being deleted, we want to leave the subnamespace alone.
+	if deletingCRD {
+		return false
+	}
+
+	// If the subnamespace is already being deleted, or has already been deleted,
+	// then there's no need to delete it again.
+	ns := r.forest.Get(nm)
+	if !inst.DeletionTimestamp.IsZero() || !ns.Exists() {
+		return false
+	}
+
+	// The subnamespace exists and isn't being deleted. We should delete it if it
+	// doesn't have any children itself, or if cascading deletion is enabled.
+	return ns.ChildNames() == nil || ns.AllowsCascadingDelete()
 }
 
 func (r *AnchorReconciler) removeFinalizers(log logr.Logger, inst *api.SubnamespaceAnchor, snsInst *corev1.Namespace) bool {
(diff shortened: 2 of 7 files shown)


yiqigao217:
Updated the test cases in validators/anchor_test.go to resolve merge-conflict.

incubator/hnc/internal/validators/anchor.go
```diff
@@ -90,7 +90,18 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
...
-		if cns.Exists() && !cns.AllowsCascadingDelete() {
+		// If the subnamespace has full namespaces in the descendants, deny the request and ask users
+		// to remove the full descendant namespaces first.
+		fd := cns.FullDescendantNames()
+		if fd != nil {
+			msg := fmt.Sprintf("The subnamespace %s has full namespaces in the descendants. Please remove them first. "+
+				"Full descendant namespaces: %q.", cnm, fd)
+			return deny(metav1.StatusReasonForbidden, msg)
+		}
+
```
adrianludwin, line 101: This seems to belong in a different CL? This is about preventing orphans, not making it easier to delete leaves, right?

incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -12,8 +12,8 @@ import (
 
 func TestSubnamespaces(t *testing.T) {
 	// Two roots `a` and `b`, with `c` a subnamespace of `a`, `e` a regular child of `c`,
-	// `f` a subnamespace of `e` and `d` a regular child of `b`
-	f := foresttest.Create("--AbcE")
+	// `f` a subnamespace of `e`, `g` a subnamespace of `a` and `d` a regular child of `b`
+	f := foresttest.Create("--AbcEA")
 	h := &Anchor{Forest: f}
 	f.Get("c").UpdateAllowCascadingDelete(true)
```
adrianludwin, line 18: This test is getting so complicated, that I think each subtest should have its own forest at this point. That means that any one test can have a simpler forest.

Also, maybe split the overall test into TestSubnamespaceCreate and TestSubnamespaceDelete? That lets you take one parameter out of the test, so that the Delete test can look like this:

```
tests := []struct{
  name string
  forest string
  acd string // eg if this is "ab", then namespaces a and b will have allowCascadingDelete set to true
  nm string // no need for cnm and pnm
}{{...}}
```

incubator/hnc/hack/test-issue-716.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
adrianludwin, line 1: This test seems to do a _lot_ of stuff. Can we break it up into multiple smaller and simpler tests? E.g.:

1. Create a->b where b is a subns. Try to delete a (fail). Try to delete b (succeed). Try to delete a (succeed).
2. As in #1 but b is a full namespace.
3. Create a->b->c where b and c are subnamespaces. Try to delete b; fail. Try to delete c; succeed. Try to delete b: succeed.

incubator/hnc/internal/validators/anchor.go
yiqigao217: Right, I removed this part and will do it in a separate PR.

incubator/hnc/internal/validators/anchor_test.go
yiqigao217: SG. Updated.

incubator/hnc/hack/test-issue-716.sh
yiqigao217: SG. Split into 3 bash files.

incubator/hnc/hack/test-issue-716-full.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
adrianludwin, line 1: I'm not sure what this test is doing? This PR isn't about deleting orphans? Or do we not have other tests for this?

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -279,12 +279,17 @@ func (r *AnchorReconciler) deleteNamespace(ctx context.Context, log logr.Logger,
...
-func (r *AnchorReconciler) allowsCascadingDelete(nm string) bool {
+// canDelete returns true if the namespace still exists and it is a leaf
+// subnamespace or it allows cascading delete.
+func (r *AnchorReconciler) canDelete(nm string) bool {
 	r.forest.Lock()
 	defer r.forest.Unlock()
 
 	ns := r.forest.Get(nm)
-	return ns.AllowsCascadingDelete()
+	if !ns.Exists() {
```
adrianludwin, line 289: This seems like an odd place for this line. Why didn't we need it before, and why do we need it now?

incubator/hnc/internal/validators/anchor.go
```diff
@@ -90,7 +90,7 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
 	}
 
 	if req.op == v1beta1.Delete {
-		if cns.Exists() && !cns.AllowsCascadingDelete() {
+		if cns.Exists() && cns.ChildNames() != nil && !cns.AllowsCascadingDelete() {
 			msg := fmt.Sprintf("The subnamespace %s doesn't allow cascading deletion. Please set allowCascadingDelete flag first.", cnm)
```
adrianludwin, line 94: Maybe update the message?

incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -54,3 +49,53 @@ func TestSubnamespaces(t *testing.T) {
...
+	// "b", "d" and "e" are subnamespaces. This is set up in a long chain to test that
+	// subnamespaces will look all the way up to get the 'allowCascadingDelete` value
+	// and won't stop looking when the first full namespace ancestor is met.
+	f := foresttest.Create("-AbCD")
+	h := &Anchor{Forest: f}
+
+	tests := []struct {
+		name string
+		op   v1beta1.Operation
+		acd  []string
```
adrianludwin, line 64: Since each namespace name is one character long, this could just be a single string

incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -24,14 +22,11 @@ func TestSubnamespaces(t *testing.T) {
...
 	}{
-		{name: "ok-create", op: v1beta1.Create, pnm: "a", cnm: "brumpf"},
-		{name: "ok-delete", op: v1beta1.Delete, pnm: "a", cnm: "c"},
-		{name: "ok-delete with allowCascadingDelete set in ancestor", op: v1beta1.Delete, pnm: "e", cnm: "f"},
-		{name: "create anchor in excluded ns", op: v1beta1.Create, pnm: "kube-system", cnm: "brumpf", fail: true},
-		{name: "create anchor with existing ns name", op: v1beta1.Create, pnm: "a", cnm: "b", fail: true},
-		{name: "create anchor for existing non-subns child", op: v1beta1.Create, pnm: "b", cnm: "d", fail: true},
-		{name: "create anchor for existing subns", op: v1beta1.Create, pnm: "a", cnm: "c"},
-		{name: "delete anchor when allowCascadingDelete is false", op: v1beta1.Delete, pnm: "a", cnm: "b", fail: true},
+		{name: "ok - create a brand new anchor with a non-existing name", op: v1beta1.Create, pnm: "a", cnm: "brumpf"},
```
adrianludwin, line 25: You can take out `op` since it's now always `Create`

incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -54,3 +49,53 @@ func TestSubnamespaces(t *testing.T) {
...
+	// Create a chain of namespaces from "a" to "e", with "a" as the root. Among them,
+	// "b", "d" and "e" are subnamespaces. This is set up in a long chain to test that
+	// subnamespaces will look all the way up to get the 'allowCascadingDelete` value
+	// and won't stop looking when the first full namespace ancestor is met.
+	f := foresttest.Create("-AbCD")
+	h := &Anchor{Forest: f}
+
+	tests := []struct {
+		name string
+		op   v1beta1.Operation
```
adrianludwin, line 63: `op` is always `Delete` so you can take it out

incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -54,3 +49,53 @@ func TestSubnamespaces(t *testing.T) {
...
+		acd  []string
+		pnm  string
+		cnm  string
+		fail bool
+	}{
+		{name: "ok - delete subnamespace with allowCascadingDelete set in the parent namespace", op: v1beta1.Delete, acd: []string{"c"}, pnm: "c", cnm: "d"},
```
adrianludwin, line 69: Nit - I'd take out the `[not] ok - delete subnamespace with` prefix. Whether it's ok or not is encoded in the `fail` and the expected value will be printed out if there's a problem, and the test is called `TestAllowCascadingDeleteSubnamespaces` so we're just repeating that.

I'd suggest calling these something like:
`set in parent`
`set in non-leaf`
`set in ancestor that is not the first full namepsace`
`unset in leaf`
`unset in non-leaf`

PR #790, opened 2020-06-05
Enable object webhooks dynamically based on hncconfig

### LINKED ISSUE #761: HNC: webhooks are not created/deleted as watched types are modified
/assign @adrianludwin 

I just noticed that while the object _reconcilers_ are created dynamically as a result of the HNC config, the _admission controllers_ are not - there's a hardcoded list left over from v0.2 (and one of them is spelled incorrectly). We should fix this before v0.4.

yiqigao217:
Remove object webhooks of different types in the hardcoded list.
Instead, have one objects webhook for all types and in the Handle()
function, dynamically look up the current type configuration in the
forest and allows changes on all other types that are not in propagate
mode in the type configuration. As long as the request object has the
matching GroupKind in propagate mode in type configuration, we don't
allow changes to it.

Tested by newly added unit test cases and manually configuring types and
modifying objects.

Fixes #761


incubator/hnc/internal/validators/object.go
```diff
@@ -77,6 +71,12 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, inst *unstructured
 		return allow("")
 	}
 
+	// Allow changes to the types that are not in propagate mode.
```
adrianludwin, line 74: We can use `request.Kind` [link](https://godoc.org/k8s.io/api/admission/v1beta1#AdmissionRequest) and have this check much earlier in the flow, even before calling Decode(). In fact, I'd make this the very first check we do, even before calling `isHNCServiceAccount()`.

incubator/hnc/internal/validators/object.go
```diff
@@ -77,6 +71,12 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, inst *unstructured
...
+	if !o.isPropagateType(inst) {
+		log.V(1).Info("Allowed change to non-propagate-mode types.")
```
adrianludwin, line 76: I'd skip this myself, or make it a V(2). I want this to be super-fast and super-safe (e.g. the fewer functions we call, the better).

incubator/hnc/internal/validators/object.go
```diff
@@ -104,6 +104,14 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, inst *unstructured
 	return allow("")
 }
 
+func (o *Object) isPropagateType(inst *unstructured.Unstructured) bool {
+	o.Forest.Lock()
+	defer o.Forest.Unlock()
+
+	ts := o.Forest.GetTypeSyncer(inst.GroupVersionKind())
```
adrianludwin, line 111: Hmm. This means that if we're propagating CRD v1, and someone tries to make a change to v1beta1, we won't run the webhook since we won't have a reconciler for v1beta1. So I think we need to change this lookup to be based on GroupKind, not GroupVersionKind. sgty?

yiqigao217: Hmm, if we don't have a reconciler for v1beta1, then objects of v1beta1 shouldn't be propagated, right? Since we only deny changing the labels in any type that is in propagate mode. Even if for example, v1.Secret is removed from the type configuration and the previous propagated objects are left and we allow changes to those obsolete propagated objects, right?

incubator/hnc/internal/validators/object.go
yiqigao217: sg. removed.

incubator/hnc/internal/validators/object.go
yiqigao217: sg. Will move it to the top.

incubator/hnc/internal/validators/object.go
adrianludwin: No, that's not how it works. v1beta1 and v1 are two different views _onto the same object_. So when we say "propagate example.com/Foobar v1", what that actually means is "please use a v1 client to propagate objects of type Foobar." But _other_ clients could be reading and writing _the same objects_ usina a v1beta1 or even v2 API.

So imagine that there's a "Secret v2" that's added to K8s, which is a superset of the features available in Secret v1. The HNC admin would have to change the config to "Secret v2" to make sure the new fields get copied - they can't have _both_ v1 and v2 at the same time. But that means that if another client tries to modify a secret using the v1 interface, the webhook won't be applied (we'll early exit).

Does this make sense?

yiqigao217: Ah right got you, yes we just discussed. Thanks, will update it too!

incubator/hnc/internal/validators/object_test.go
```diff
@@ -5,14 +5,65 @@ import (
...
+	tests := []struct {
+		name    string
+		version string
+		kind    string
+		fail    bool
+	}{{
+		name:    "Deny GroupVersionKind in the propagate mode",
+		version: "v1",
+		kind:    "Secret",
+		fail:    true,
```
adrianludwin, line 40: It's unclear to me _why_ the admission controller is returning a "denied" here (maybe `allow` or `deny` would be a better term here). Is it because we're not actually including a `Secret` object, so the call to `Decode` will fail? So basically, we're passing in a request that's _guaranteed_ to fail _unless_ we skip it because the type isn't registered?

If so, I think that deserves a lot of commenting somehwere, maybe on line 53 or so.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -5,14 +5,65 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	"k8s.io/api/admission/v1beta1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
yiqigao217: sg updated.

incubator/hnc/internal/validators/object_test.go
yiqigao217: Added comments in validators/object.go#41 and object_test.go#21. 

TestType only tests the types from the admission requests. There is no object in the admission requests for all the test cases. The requests are supposed to fail with "BadRequest" since the object cannot be decoded. However, with early exit for non-propagate-mode types, they should return "allow" before the object gets decoded.

PR #799, opened 2020-06-11
Enable removing bad subns anchors safely 

### LINKED ISSUE #797: Deleting an anchor in the "conflict" state can also delete the namespace from another parent
This could be a race condition when two parents both create a subnamespace anchor with the same name simultaneously. The webhook could be bypassed if the second anchor is created before the forest is updated.

Test and see what would happen if we delete the bad anchor in one of the parent. The good anchor is the one in the namespace that the subnamespace points to as the parent.

yiqigao217:
Allow removing bad subns anchor if the subnamespace already exists and
the parent doesn't match. The race condition could happen when two
parents both create the same subns the same time and the webhook is
bypassed because the forest is not updated after the first change. We
would always allow resolving a bad state.

Most importantly, make sure the subnamespace is not deleted with the
deletion of a bad anchor. Before this change, the subnamespace created
from a good anchor will also be deleted when deleting a bad anchor.

Tested by test-issue-797-leaf.sh and test-issue-797-non-leaf.sh, where
they disable the webhook first to generate the race condition in both a
leaf and non-leaf subnamespace and enable the webhook to test the
change. One unit test case is added to the anchor validator too.

Fixes #797

### DIFF
--- a/incubator/hnc/hack/test-issue-797-non-leaf.sh
+++ b/incubator/hnc/hack/test-issue-797-non-leaf.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+bold=$(tput bold)
+normal=$(tput sgr0)
+
+# Set up namespace structure
+echo "----------------------------------------------------"
+echo "${bold}Creating a race condition of two parents with the same non-leaf subns (anchor)"
+echo "Creating two root namespaces 'p1' and 'p2'...${normal}"
+kubectl create ns p1
+kubectl create ns p2
+sleep 1
+
+echo "${bold}Disabling webhook to generate the race condition...${normal}"
+kubectl delete validatingwebhookconfigurations.admissionregistration.k8s.io hnc-validating-webhook-configuration
+sleep 1
+
+echo "${bold}Creating subns (anchor) 'sub' in both 'p1' and 'p2'...${normal}"
+kubectl hns create sub -n p1
+kubectl hns create sub -n p2
+sleep 1
+
+echo "${bold}Subnamespace 'sub' should be created and have 'p1' as the 'subnamespaceOf' annoation value:${normal}"
+kubectl get ns sub -o yaml
+
+echo "${bold}Creating a 'test-secret' in the subnamespace 'sub'${normal}"
+kubectl create secret generic test-secret --from-literal=key=value -n sub
+
+echo "${bold}subns (anchor) 'sub' in 'p2' should have 'status: conflict' because it's a bad anchor:${normal}"
+kubectl get subns sub -n p2 -o yaml
+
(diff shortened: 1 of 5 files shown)


yiqigao217:
Please hold review on this, I will let you know once it's ready.
/hold

yiqigao217:
Hi @adrianludwin and @rjbez17, it's ready for review now.

incubator/hnc/hack/test-issue-797-non-leaf.sh
```diff
@@ -0,0 +1,58 @@
...
+echo "${bold}subns (anchor) 'sub' in 'p2' should have 'status: conflict' because it's a bad anchor:${normal}"
+kubectl get subns sub -n p2 -o yaml
+
+echo "${bold}Creating subns (anchor) 'sub-sub' in 'sub' to make 'sub' non-leaf...${normal}"
+kubectl hns create sub-sub -n sub
+sleep 1
+
+echo "${bold}Setting allowCascadingDelete on 'sub'...${normal}"
+kubectl hns set sub -a
+
```
adrianludwin, line 35: Perhaps you should do this _after_ trying to delete the bad anchor? The bad anchor should be deleted even if ACD _isn't_ set, right?

incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -69,6 +69,7 @@ func TestAllowCascadingDeleteSubnamespaces(t *testing.T) {
 		{name: "set in ancestor that is not the first full namespace", acd: "a", pnm: "c", cnm: "d"},
 		{name: "unset in leaf", pnm: "d", cnm: "e"},
 		{name: "unset in non-leaf", pnm: "c", cnm: "d", fail: true},
+		{name: "unset in non-leaf but bad anchor", pnm: "bad", cnm: "d"},
```
adrianludwin, line 72: Probably best for `bad` to actually be a namespace in the tree, otherwise there's a chance that you could hit some other random error. Maybe use `b`, since `d` is actually a child of `c`?

incubator/hnc/hack/test-issue-797-non-leaf.sh
yiqigao217: Right. SG, updated.

incubator/hnc/internal/validators/anchor_test.go
yiqigao217: SG, updated.

yiqigao217:
Hi @adrianludwin , I also changed the ensuring of the subnamespace is not deleted by seeing the created secret is still there instead of matching the creatingTimestamp.

PR #801, opened 2020-06-12
[VC] Back populate super master clusterIP to tenant service object

Fei-Guo:
In vc, the tenant service and super master service have different cluster ips since the cluster ip allocations are done by separate service controllers in tenant master and super master individually. The Pods can use super master cluster ip to access service or other Pod without any problem. But tenant dns cannot translate service dnsname to super master cluster ip since it can only see  the tenant master cluster ip. 

This change is the first step towards resolving the problem. The syncer now will populate the super master service cluster ip in the annotation of the tenant service object. Other components now can get the super master cluster IP from tenant perspective if it is needed.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/service/uws.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/service/uws.go
@@ -56,7 +56,7 @@ func (c *controller) BackPopulate(key string) error {
 	}
 
 	// Make sure the super cluster IP is added to the annotation so that it can be back populated to the tenant object
-	if pService.Spec.ClusterIP != "" && (pService.Annotations == nil || pService.Annotations[constants.LabelSuperClusterIP] != pService.Spec.ClusterIP) {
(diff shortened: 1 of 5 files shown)


incubator/virtualcluster/pkg/syncer/resources/service/uws.go
```diff
@@ -54,6 +55,20 @@ func (c *controller) BackPopulate(key string) error {
 		return err
 	}
 
+	// Make sure the super cluster IP is added to the annotation so that it can be back populated to the tenant object
+	if pService.Spec.ClusterIP != "" && (pService.Annotations == nil || pService.Annotations[constants.LabelSuperClusterIP] != pService.Spec.ClusterIP) {
```
zhuangqh, line 59: just `pService.Annotations[constants.LabelSuperClusterIP] != pService.Spec.ClusterIP`

it won't panic even if pServer.Annotations is nil

Fei-Guo: You are right. Done

PR #804, opened 2020-06-13
[VC] Add document for tenant dns support

Fei-Guo:
This change adds document to explain how to customize coredns to provide super master cluster IP translation for tenant service domain name.

### DIFF
--- a/incubator/virtualcluster/README.md
+++ b/incubator/virtualcluster/README.md
@@ -60,9 +60,7 @@ hence the default grace period is too small.
 The DNS service should be created in `kube-system` namespace using name `kube-dns`. The syncer controller can then
 recognize the DNS service cluster IP in super master and inject it into Pod spec `dnsConfig`.
 
-- Since coredns is installed in tenant master, the service cluster IPs have to be identical in tenant
-master and super master respectively so that coredns can provide correct cluster IP translation for service cname.
-This requires both clusters to have the same CIDR.
+- VirtualCluster supports tenant DNS service using a customized coredns build. See this [document](./doc/tenant-dns.md) for details.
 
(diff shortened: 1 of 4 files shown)


ShadowJonathan:
Would it be possible for coreDNS to eventually automatically detect if it should operate in a "virtualized" fashion, or a "normal" fashion?

I'm just musing, as this means that tooling could detect something "wrong" with the coredns deployment and alter it so it has the "latest" image, and thereby breaking the dns resolver in a tenant namespace. If the dns image could be one and the same for normal k8s, this could be avoided.

(I know this is an unlikely hypothetical, i'm just asking if its possible for this functionality to be implemented transparently in the normal dns image, and switch on detection)

Fei-Guo:
> (I know this is an unlikely hypothetical, i'm just asking if its possible for this functionality to be implemented transparently in the normal dns image, and switch on detection)

Good suggestion! I have pushed a new change for this. Please review.
The virtualcluster/coredns:v1.6.8 image is also updated.

ShadowJonathan:
This looks good, but my suggestion was more one to actually consider PRing this into mainstream coreDNS to allow for that image to transparently work in both scenarios. 

To make a PR into mainstream coreDNS when virtualcluster has had it's first release.

Fei-Guo:
> This looks good, but my suggestion was more one to actually consider PRing this into mainstream coreDNS to allow for that image to transparently work in both scenarios.
> 
> To make a PR into mainstream coreDNS when virtualcluster has had it's first release.

This one is a hacky workaround. But I agree in general, we should think about how to make coredns tenant-aware. In my opinion, all tenants should be able to share a dns server if it has isolation mechanisms.

PR #807, opened 2020-06-16
[VC] update tenant master component to v1.16.x

zhuangqh:

### DIFF
--- a/incubator/virtualcluster/config/sampleswithspec/clusterversion_v1_loadbalancer.yaml
+++ b/incubator/virtualcluster/config/sampleswithspec/clusterversion_v1_loadbalancer.yaml
@@ -30,7 +30,7 @@ spec:
             subdomain: etcd
             containers:
             - name: etcd
-              image: registry.cn-hangzhou.aliyuncs.com/virtualcluster/etcd-v3.4.0
+              image: virtualcluster/etcd-v3.4.0
               imagePullPolicy: Always
               command: 
               - etcd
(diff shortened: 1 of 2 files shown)


christopherhein:
How do the image registries get updated? 

I'm seeing both `registry.cn-hangzhou.aliyuncs.com/virtualcluster/apiserver-v1.16.2` and `virtualcluster/apiserver-v1.15.4`.

ShadowJonathan:
> How do the image registries get updated?
> 
> I'm seeing both `registry.cn-hangzhou.aliyuncs.com/virtualcluster/...` and `virtualcluster/...`.

This, i also cannot access `registry.cn-hangzhou.aliyuncs.com`, it times out for me.

christopherhein:
I wonder if we should have the same registry domain for all images that come from the `virtualcluster` project. I know for myself I usually build all the images and push them to my own registry then have to change all the paths manually, at least until we have #738 when we can use `kustomize set image name=foobar`.

Fei-Guo:
> I wonder if we should have the same registry domain for all images that come from the `virtualcluster` project. I know for myself I usually build all the images and push them to my own registry then have to change all the paths manually, at least until we have #738 when we can use `kustomize set image name=foobar`.

Yes, we should. @zhuangqh, we'd better change images to docker hub ones for loadbalancer type of clusterversion.

zhuangqh:
fixed. good catch @Fei-Guo @christopherhein

PR #815, opened 2020-06-21
[MTB] auto register benchmark

phoenixking25:
/assign @JimBugwadia


phoenixking25:
@JimBugwadia added mtb-builder 
 `make builder`
A binary file will be generated at the root directory of the project.
Usage :- `./mtb-builder create block host IPC --profile-level 1` 
At present just taking benchmark name and profile level, leaving the rest of fields to be filled by user in config.yaml
Reason: for example `MTB-PL1-BC-CPI-5` in this we can generate most of the ID but can't serial order as builder is separate from the project and dont know about any previous added benchmarks.

PR #819, opened 2020-06-23
[VC] fix: respect vc status when reconciling

zhuangqh:

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/syncer.go
+++ b/incubator/virtualcluster/pkg/syncer/syncer.go
@@ -241,7 +241,16 @@ func (s *Syncer) syncVirtualCluster(key string) error {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


incubator/virtualcluster/pkg/syncer/syncer.go
```diff
@@ -298,17 +307,12 @@ func (s *Syncer) addCluster(key string, vc *v1alpha1.VirtualCluster) error {
 		return fmt.Errorf("failed to new tenant cluster %s/%s: %v", vc.Namespace, vc.Name, err)
 	}
 
-	s.mu.Lock()
-	if _, exist := s.clusterSet[key]; exist {
```
zhuangqh, line 302: workqueue promise to us that we never concurrently reconcile the same vc

PR #820, opened 2020-06-23
Update HC finalizer from hasOwnedNamespace to hasSubnamespace

yiqigao217:
Tested by make test and verified the updated finalizer name in GKE
cluster.

Part of #667

### DIFF
--- a/incubator/hnc/api/v1alpha1/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha1/hierarchy_types.go
@@ -31,11 +31,11 @@ const (
 
 // Constants for labels and annotations
 const (
-	MetaGroup                  = "hnc.x-k8s.io"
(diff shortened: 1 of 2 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

This is a minor change to the finalizer name not API.

yiqigao217:
> What happens if you upgrade HNC but objects with the old finalizer already exist? Do we need an upgrade path?

No, it updates the old finalizer since the finalizer on HC is reset every time reconciling HC. I tested on an existing "owner" namespace in the existing cluster and after `make deploy`, the finalizer is updated.

adrianludwin:
Oh, cool! Thanks :)

/approve
/lgtm
/hold
/assign @rjbez17

PR #822, opened 2020-06-24
Set apiVersion and kind as required fields in hncconfig.spec.types

yiqigao217:
apiVersion and kind fields in hncconfig.spec.types should never be
empty. It was a mistake to leave them omitempty. The current validating
webhook ensures the type generated from these two fields are valid,
therefore, empty fields are not allowed. Still we need to fix it in the
CRD.

Tested with the webhook disabled and it was still not allowed.

### DIFF
--- a/incubator/hnc/api/v1alpha1/hnc_config.go
+++ b/incubator/hnc/api/v1alpha1/hnc_config.go
@@ -54,9 +54,9 @@ const (
 // TypeSynchronizationSpec defines the desired synchronization state of a specific kind.
 type TypeSynchronizationSpec struct {
 	// API version of the kind defined below. This is used to unambiguously identifies the kind.
-	APIVersion string `json:"apiVersion,omitempty"`
+	APIVersion string `json:"apiVersion"`
 	// Kind to be configured.
-	Kind string `json:"kind,omitempty"`
+	Kind string `json:"kind"`
 	// Synchronization mode of the kind. If the field is empty, it will be treated
 	// as "propagate". An unsupported mode will be treated as "ignore".
 	// +optional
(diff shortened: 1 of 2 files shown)


yiqigao217:
I investigated the previous failed test (it was never flaky when I ran `make test` and it passed the retest). The failed test was on https://github.com/kubernetes-sigs/multi-tenancy/blob/hnc-v0.4/incubator/hnc/internal/reconcilers/hnc_config_test.go#L361 where it returns 2 instead of 1. I think it's irrelevant and is just an error in the prow job. Please let me know if you think differently.

adrianludwin:
> I investigated the previous failed test (it was never flaky when I ran `make test` and it passed the retest). The failed test was on https://github.com/kubernetes-sigs/multi-tenancy/blob/hnc-v0.4/incubator/hnc/internal/reconcilers/hnc_config_test.go#L361 where it returns 2 instead of 1. I think it's irrelevant and is just an error in the prow job. Please let me know if you think differently.

Can you please file a bug to look into the flake (and paste in the part of the log with the failure in case Prow deletes it)? Even rare flakes are annoying because they can mask real rare issues :) But I agree it's probably not relevant to this bug.

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -54,9 +54,9 @@ const (
 // TypeSynchronizationSpec defines the desired synchronization state of a specific kind.
 type TypeSynchronizationSpec struct {
 	// API version of the kind defined below. This is used to unambiguously identifies the kind.
-	APIVersion string `json:"apiVersion,omitempty"`
+	APIVersion string `json:"apiVersion"`
```
adrianludwin, line 57: Can you make the same change in `TypeSynchronizationStatus`? Technically it's not necessary since HNC will always fill them in, but we may as well make the code self-documenting.

yiqigao217: SG, updated! Question: by "self-documenting", did you mean that the it shows the fields are "required" when people read the code?

yiqigao217:
> > I investigated the previous failed test (it was never flaky when I ran `make test` and it passed the retest). The failed test was on https://github.com/kubernetes-sigs/multi-tenancy/blob/hnc-v0.4/incubator/hnc/internal/reconcilers/hnc_config_test.go#L361 where it returns 2 instead of 1. I think it's irrelevant and is just an error in the prow job. Please let me know if you think differently.
> 
> Can you please file a bug to look into the flake (and paste in the part of the log with the failure in case Prow deletes it)? Even rare flakes are annoying because they can mask real rare issues :) But I agree it's probably not relevant to this bug.

Good suggestion! Filed #823

incubator/hnc/api/v1alpha1/hnc_config.go
adrianludwin: Yup, exactly. It also shows up as "required" in the CRD.

PR #827, opened 2020-06-27
[VC] Remove tenant-controller from the demo, and use default NS

christopherhein:
While the tenant controller is really useful it's an additional component that isn't technically necessary for using `virtualcluster`. This PR removes the steps to setup the `tenant1admin` namespace from the demo doc making it easier to get up and running.


christopherhein:
Thanks @Fei-Guo

Issue #785, opened 2020-06-02
HNC: run all e2e tests automatically as part of a test suite, especially in Kind


adrianludwin:
We should be able to run all the tests in `hack/` in a single command, either on a "real" K8s cluster (e.g. GKE, EKS, etc) or a fake one (e.g. Kind). We should be able to make this part of our release suite.

This has several components:

1. Ensure that each existing test returns `0` on success and `1` on failure
2. Ensure that no two tests can affect each other (e.g. they all use different namespace names; HNC is repaired between each test)
3. Write a wrapper script to invoke all the tests.

/good-first-issue

adrianludwin:
This work is now complete. Thanks @GinnyJI !

/close

adrianludwin:
Note: we didn't _actually_ automate this, but it does all run now via `go test ./test/e2e/...` and I think that's good enough. Further automation (e.g. presubmit/postsubmit) can wait for another day.

PR #832, opened 2020-06-30
Fix the error in e2e test due to the prefix change

GinnyJI:
The prefix on test namespace was missing so the test was failing.

Tested by running:
     make kind-reset
     make kind-deploy
     k delete validatingwebhookconfigurations.admissionregistration.k8s.io hnc-validating-webhook-configuration
     go test ./e2e/...
All the e2e test passed now. 
Part of #785


GinnyJI:
I signed it

PR #833, opened 2020-07-01
Allow finalizers to be removed when deleting CRDs

### LINKED ISSUE #824: HNC: ensure there are no finalizers on any CRD if the CRD is being deleted
I noticed that we had a `hasSubnamespace` finalizer on an HC as we were trying to delete the CRD to uninstall HNC. This prevents uninstallation needlessly. We should remove all finalizers if the CRD is being deleted.

adrianludwin:
See #824. When the CRDs are being deleted, we need to continue
reconciling the CRs until they're gone in order to ensure that any
finalizers on the CRs are removed.

Tested: new test script usually fails without this fix, but reliably
passes with it. Observed the new log messages working as expected while
running the script. For subnamespace anchors, tried removing the
CRD-checking code entirely and verified that
hack/test-delete-anchor-crd.sh failed, and then replaced it with the
current code (the shared isDeletingCRD function) and verified that it
passed again.

Fixes #824

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -17,7 +17,6 @@ package reconcilers
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -26,7 +25,6 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -109,9 +107,15 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 	origNS := nsInst.DeepCopy()
 
 	// Get singleton from apiserver. If it doesn't exist, initialize one.
-	inst, err := r.getSingleton(ctx, nm)
+	inst, deletingCRD, err := r.getSingleton(ctx, nm)
 	if err != nil {
 		return err
+	}
+	// Don't _create_ the singleton if its CRD is being deleted. But if the singleton is already
+	// _present_, we may need to update it to remove finalizers (see #824).
+	if deletingCRD && inst.CreationTimestamp.IsZero() {
+		log.Info("Skipping reconcile due to CRD deletion")
+		return nil
 	}
 	origHC := inst.DeepCopy()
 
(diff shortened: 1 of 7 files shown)


adrianludwin:
/assign @yiqigao217 
/cc @GinnyJI @adrianchung 

Hey Yiqi, can you please review this one as well? Thanks!

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -109,10 +107,17 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
...
 	// Get singleton from apiserver. If it doesn't exist, initialize one.
-	inst, err := r.getSingleton(ctx, nm)
+	inst, deletingCRD, err := r.getSingleton(ctx, nm)
 	if err != nil {
 		return err
 	}
+	// If the CRD is being deleted, then there's no need to sync a missing
+	// CR. However, if the CR is _present_, we may need to update it to
+	// remove finalizers (see #824).
+	if deletingCRD && inst.CreationTimestamp.IsZero() {
```
yiqigao217, line 117: What does `CreationTimestamp.IsZero()` mean here?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -411,13 +416,17 @@ func (r *HierarchyConfigReconciler) syncConditions(log logr.Logger, inst *api.Hi
...
-func (r *HierarchyConfigReconciler) syncCritConditions(log logr.Logger, ns *forest.Namespace, hadCrit bool) {
+func (r *HierarchyConfigReconciler) syncCritConditions(log logr.Logger, ns *forest.Namespace, deletingCRD, hadCrit bool) {
 	// If we're in a cycle, determine that now
 	if cycle := ns.CycleNames(); cycle != nil {
 		msg := fmt.Sprintf("Namespace is a member of the cycle: %s", strings.Join(cycle, " <- "))
 		ns.SetLocalCondition(api.CritCycle, msg)
 	}
 
+	if deletingCRD {
+		ns.SetLocalCondition(api.CritDeletingCRD, "The HierarchyConfiguration CRD is being deleted; all syncing is disabled.")
```
yiqigao217, line 427: Is `CritDeletingCRD` condition very short-lived? Considering it will be gone when the CRD and CR are deleted.

adrianludwin: Yup. Its _main_ purpose it to stop the propagated objects from being deleted.

incubator/hnc/internal/reconcilers/hierarchy_config.go
adrianludwin: Exactly. Updated the comment.

adrianludwin:
> Why SubnamespaceAnchor CRD is not included? It has finalizers too. Is it that once the CRD is gone, all subnamespaces are cascading deleted? Maybe add that in the test script too?

SubnamespaceAnchors are already working - see https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/anchor.go#L116. There's a separate test script for them too - https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/hack/test-delete-anchor-crd.sh

We already _were_ trying to handle these issues properly, but there was a bug specifically for HierarchyController.

yiqigao217:
Thanks for the clarification! Not sure if we want to use the same `func isDeletingCRD(ctx context.Context, c crdClient, nm string) (bool, error)` for all 3 CRDs. Feel free to remove the hold if you think it's ok as is.

/lgtm
/hold

adrianludwin:
> Thanks for the clarification! Not sure if we want to use the same `func isDeletingCRD(ctx context.Context, c crdClient, nm string) (bool, error)` for all 3 CRDs. Feel free to remove the hold if you think it's ok as is.

Good call! Fixed and retested (see PR description or commit message).

yiqigao217:
Thanks looks nice!

/hold cancel

PR #834, opened 2020-07-01
Add controller-tools to /vendor

### LINKED ISSUE #778: HNC: include controller-gen in the vendor directory
The makefile relies on `controller-gen` being in the path, otherwise it downloads it. Now that we've enabled vendoring, if `controller-gen` _isn't_ in the path, this a) causes an error and b) isn't necessary; we should always be able to use the vendored version.

adrianludwin:
Adds a fake file (hack/tools.go) to fool `go mod tidy` into believing
there's a dependency on sigs.k8s.io/controller-tools, and modifies the
makefile to a) use the vendored version of this file and b) avoid ever
trying to compile it (you can't import a non-package).

Tested: regenerated manifests and autogenerate Go code

Fixes #778


adrianludwin:
/assign @rjbez17 
/cc @yiqigao217 @adrianchung @GinnyJI 

Ryan - I got tired of people accidentally having the wrong version of controller-gen in their paths so figured that we could just include it as a vendored dependency. The way to do this isn't _great_ but it does appear to work.

wdyt about this?

rjbez17:
It seems hacky but I could not find/think of any better solution at the moment. One thing to note was this answer in your linked thread: https://stackoverflow.com/a/55939956. TL;DR: As we think about including client libraries, the `tools.go` could cause us some pain.

/lgtm
/approve

adrianludwin:
> It seems hacky but I could not find/think of any better solution at the moment. One thing to note was this answer in your linked thread: https://stackoverflow.com/a/55939956. TL;DR: As we think about including client libraries, the `tools.go` could cause us some pain.

That's a good point, but I _think_ we can use submodules to get around that. The client library should have vastly fewer imports than the service anyway (e.g. no Prometheus).

PR #836, opened 2020-07-01
Restart the pod if certs change

### LINKED ISSUE #765: HNC: very slow startup after first installation
The builtin cert manager seems to stall the first time we install HNC. This is a poor user experience.

adrianludwin:
See #765. If a mounted secret changes _after_ a pod is started, it can
take a fairly long time (~60s) for the kubelet to notice the change and
project the new secret to the pod. Since our internal cert manager
writes a secret but then needs to wait for it to become available as a
file, this leads to a poor onboarding experience with HNC.

This change introduces a flag that exits the process as soon as the
internal cert manager changes a secret, which should only occur on
initial installation of HNC or every ten years (!). The restart time
takes <5s so this is overall a much better experience.

Tested: without changing the flags in the default manifest, observed no
change when HNC is installed for the first time (i.e. from the first log
message to when the HNCConfiguration is first reconciled takes 103s, and
there are no restarts). When the flag is added, the startup time
decreases to 10s with the one expected restart. Further restarts of HNC
(e.g. deleting and recreating the deployment but not the secret) does
not result in a restart and completes in 4s.

Fixes #765

### DIFF
--- a/incubator/hnc/config/manager/manager.yaml
+++ b/incubator/hnc/config/manager/manager.yaml
@@ -42,7 +42,8 @@ spec:
         - "--metrics-addr=127.0.0.1:8080"
         - "--max-reconciles=10"
         - "--apiserver-qps-throttle=50"
-        - "--enable-internal-cert-management=true"
+        - "--enable-internal-cert-management"
+        - "--cert-restart-on-secret-refresh"
         image: controller:latest
         name: manager
         resources:
--- a/incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
+++ b/incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
@@ -10,6 +10,7 @@ import (
 	"crypto/x509/pkix"
 	"encoding/base64"
 	"encoding/pem"
+	"flag"
 	"fmt"
 	"math/big"
 	"os"
@@ -46,6 +47,12 @@ var crLog = logf.Log.WithName("cert-rotation")
 var vwhGVK = schema.GroupVersionKind{Group: "admissionregistration.k8s.io", Version: "v1beta1", Kind: "ValidatingWebhookConfiguration"}
 
 var _ manager.Runnable = &CertRotator{}
+
+var restartOnSecretRefresh = false
+
+func init() {
+	flag.BoolVar(&restartOnSecretRefresh, "cert-restart-on-secret-refresh", false, "Kills the process when secrets are refreshed so that the pod can be restarted (secrets take up to 60s to be updated by running pods)")
+}
 
 // AddRotator adds the CertRotator and ReconcileVWH to the manager.
 func AddRotator(mgr manager.Manager, cr *CertRotator, vwhName string) error {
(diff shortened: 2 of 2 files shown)


adrianludwin:
/assign @rjbez17 

Hi Ryan, can you please review this? I'll remove the hold once #835 is merged.

/hold

rjbez17:
This should only ever cause us issues on initial startup correct? I'm not sure on the internal rotators exact implementation, but typically you rotate pretty early (like days/weeks/months) before a cert is set to expire, meaning the old cert will still be valid over the ~60 seconds it takes to re-mount. 

With that, can't we just add to our startup to wait for the `CertsMounted` channel to close? It would prevent HNC from fully starting until the certs are mounted. Since we call `CreateCertsIfNeeded` from the managers main.go, this should work no? Sorry, I'm not too familiar with this code so I tried to follow along the best I could, I may be wrong.

I generally don't like the idea of exiting to force a restart/picking up new data, but I think we should be avoiding the internal cert rotator in production workloads anyway (meaning this wouldn't come into play) so I'm not terribly concerned.

Feel free to remove the hold if your happy with it. I guess you can consider me neutral on which way it goes.

/lgtm
/approve
/hold

adrianludwin:
> This should only ever cause us issues on initial startup correct? I'm not sure on the internal rotators exact implementation, but typically you rotate pretty early (like days/weeks/months) before a cert is set to expire, meaning the old cert will still be valid over the ~60 seconds it takes to re-mount.

Pretty much, I believe we have a 10yr rotation period. These keys are all internal to the cluster so the risk of them being exfiltrated is fairly low.

> With that, can't we just add to our startup to wait for the `CertsMounted` channel to close? It would prevent HNC from fully starting until the certs are mounted. Since we call `CreateCertsIfNeeded` from the managers main.go, this should work no? Sorry, I'm not too familiar with this code so I tried to follow along the best I could, I may be wrong.

That's what we're currently doing. The problem is that it takes over a minute for a secret to be updated on a pod that's already running, whereas it's basically instant if the pod restarts.

> I generally don't like the idea of exiting to force a restart/picking up new data, but I think we should be avoiding the internal cert rotator in production workloads anyway (meaning this wouldn't come into play) so I'm not terribly concerned.

Agreed, I actually asked Yiqi to put in the current code (the original from Gatekeeper _did_ do a restart). But an almost two minute delay from when you install it to when you can use it is particularly user-hostile.

Thanks!

adrianludwin:
/hold cancel
/assign @yiqigao217 
/cc @GinnyJI @adrianchung 

Hey Yiqi, can you please review this? Ryan's away now.

yiqigao217:
> That's what we're currently doing. The problem is that it takes over a minute for a secret to be updated on a pod that's already running, whereas it's basically instant if the pod restarts.

For over a minute wait for the secret to be updated and it's instant if the pod restarts, is it because that in the internal cert, it checks the existence of certs with exponential backoff?

adrianludwin:
> For over a minute wait for the secret to be updated and it's instant if the pod restarts, is it because that in the internal cert, it checks the existence of certs with exponential backoff?

No, the problem is in the kubelet. AIUI, it only checks for changes to _existing_ secrets every 60s or so, so if you create a pod and then immediately change the secret that's mounted in it (which is what we do), then it will take almost 60s for the kubelet to notice that the secret's changed and modify what the pod can see.

Does that make sense?

Issue #718, opened 2020-05-13
HNC: allow us to uninstall easily without removing CRDs


adrianludwin:
Right now, if you say `kubectl delete -f <hnc-version>/manifest.yaml`, _everything_ gets deleted, including the CRDs. It would be nice if we could delete the HNC namespace, deployment, webhooks etc but not the user data.

/good-first-issue

Divya063:
@adrianludwin I would like to work on this.

adrianludwin:
Hi @Divya063, thanks for the offer! Do you have experience with kustomize, and some idea about how we'd do this? I think we somehow want to have an "uninstall" manifest, so you could say `kubectl delete -f <hnc-version>/safe-uninstall.yaml`, where `safe-uninstall.yaml` has everything that `manifest.yaml` does _except_ the CRDs. The build process will also need to be updated.

Do you know how to do all this? Thanks!

adrianludwin:
See #824 for an interesting inversion of this bug.

adrianludwin:
This is somewhat addressed by #839 so I think we can push this to the backlog.

adrianludwin:
Replaced by #1320

PR #839, opened 2020-07-02
Document changes in HNC v0.5

adrianludwin:
Documented the new behaviour of subnamespace deletion, and added docs
about how to safely uninstall HNC starting with the CRDs.

### DIFF
--- a/incubator/hnc/docs/user-guide/how-to.md
+++ b/incubator/hnc/docs/user-guide/how-to.md
@@ -166,7 +166,7 @@ This section is under construction (as of Apr 2020). For now, please see
 
 ### Delete a subnamespace
 
-In order to delete a subnamespace, you must have permissions to delete its
+In order to delete a subnamespace, you must first have permissions to delete its
 anchor in its parent namespace. Ask your cluster administrator to give you this
 permission if you do not have it.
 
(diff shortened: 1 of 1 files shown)


adrianludwin:
/assign @yiqigao217 
cc @GinnyJI @adrianchung

adrianludwin:
This addressed #718 as much as we're going to in HNC v0.5, so pushing this to the backlog.

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -176,78 +176,80 @@ Kubernetes 1.15 and higher installed. See [issue
...
 These errors are there for your protection. Deleting namespaces is very
 dangerous, and deleting _subnamespaces_ can result in entire subtrees of
-namespaces being deleted as well. Therefore, you may set the
-`allowCascadingDelete` field either on the child namespace, on its parent, or (if
-the parent is a subnamespace itself) on its parent, and so on. For example, set
-the field on `child` if you’re only trying to delete the child, or on `parent` if
-you’re trying to delete the parent.
+namespaces being deleted as well. Therefore, if deleting a namespace (or
+subnamespace) would result in the deletion of any namespace _other than the one
+explicitly being deleted_, HNC requires that you must specify the
```
yiqigao217, line 211: This is documented well but I am just thinking about the easy deletion of leaf subns. Is it possible that users may delete a leaf subns' parent (full ns) without knowing the subns will be deleted too?

adrianludwin: That's not allowed:

```
aludwin@aludwin0:hnc$ k create ns parent
namespace/parent created

aludwin@aludwin0:hnc$ k hns create child -n parent
Successfully created "child" subnamespace anchor in "parent" namespace

aludwin@aludwin0:hnc$ k delete ns parent
Error from server (Forbidden): admission webhook "namespaces.hnc.x-k8s.io" denied the request: Please set allowCascadingDelete first either in the parent namespace or in all the subnamespaces.
  Subnamespace(s) without allowCascadingDelete set: [child].
```

yiqigao217: Ah right, thank you for clarification. Yes, I tested it and got the same. Only deleting the leaf subns anchor doesn't need allowCascadingDelete set.

PR #841, opened 2020-07-03
Use CRD validation to enforce singletons

### LINKED ISSUE #831: HNC: CRD metadata.name validation for singletons
Add kustomization patch to add metadata.name validation for singletons. With this, we can remove the validating webhooks on the singleton names and remove the `CritSingletonNameInvalid`  HNCConfiguration condition.

HNCConfiguration CRD:
```
CRD.spec.validation.openAPIV3Schema.properties.
metadata:
  type: object
  properties:
    name:
      type: string
      pattern: "^config$"
```

HierarchyConfiguration CRD:
```
CRD.spec.validation.openAPIV3Schema.properties.
metadata:
  type: object
  properties:
    name:
      type: string
      pattern: "^hierarchy$"
```

/assign @adrianludwin

yiqigao217:
Add CRD validation kustomization patchs to HierarchyConfiguration and
HNCConfiguration CRDs to validate the metadata.name to be the singleton.

Remove the "CritSingletonNameInvalid" condition and remove the
validating and condition setting code in the HNCConfig reconciler.
HNCConfig reconiler only reconciles the 'config' singleton. The
validation here is only to set the condition so it's safe to remove.

Remove the validating webhook on HNCConfig resource name too. There's no
need to keep the validating webhook as a second layer of protection
because the server-side validation is more reliable than webhooks since
webhooks can be disabled or bypassed if not running.

This change is also backward compatible since the old bad HNCConfig
resource will still have the obsolete condition on it. HNCConfig
reconciler won't touch it.

Tested on kind and GKE 1.14 cluster. New HC or HNCConfig with bad
names cannot be created with the server-side validation. An old
HNCConfig with bad name created before introducing the CRD validation
when the webhook was also disabled can still be deleted now. There was
still an obsolete "CritSingletonNameInvalid" condition on it.

Fixes #831

### DIFF
--- a/incubator/hnc/api/v1alpha1/hnc_config.go
+++ b/incubator/hnc/api/v1alpha1/hnc_config.go
@@ -46,7 +46,6 @@ const (
 // HNCConfigurationCondition codes. *All* codes must also be documented in the
 // comment to HNCConfigurationCondition.Code.
 const (
-	CritSingletonNameInvalid         HNCConfigurationCode = "CritSingletonNameInvalid"
 	ObjectReconcilerCreationFailed   HNCConfigurationCode = "ObjectReconcilerCreationFailed"
 	MultipleConfigurationsForOneType HNCConfigurationCode = "MultipleConfigurationsForOneType"
 )
@@ -148,16 +147,10 @@ type HNCConfigurationCondition struct {
 	// shown below, but new values may be added over time. This field is always present in a
 	// condition.
 	//
-	// All codes that begin with the prefix `crit` indicate that reconciliation has
-	// been paused for this configuration. Future changes of the configuration will be
-	// ignored by HNC until the condition has been resolved. Non-critical conditions
-	// typically indicate some kinds of error that HNC itself can ignore. However,
+	// Conditions typically indicate some kinds of error that HNC itself can ignore. However,
 	// the behaviors of some types might be out-of-sync with the users' expectations.
 	//
 	// Currently, the supported values are:
-	//
-	// - "critSingletonNameInvalid": the specified singleton name is invalid. The name should be the
-	// same as HNCConfigSingleton.
 	//
 	// - "objectReconcilerCreationFailed": an error exists when creating the object
 	// reconciler for the type specified in Msg.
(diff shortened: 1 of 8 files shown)


incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -46,7 +46,6 @@ const (
 // HNCConfigurationCondition codes. *All* codes must also be documented in the
 // comment to HNCConfigurationCondition.Code.
 const (
-	CritSingletonNameInvalid         HNCConfigurationCode = "CritSingletonNameInvalid"
```
adrianludwin, line 49: Did you remove the documentation as well? See line 46, above.

yiqigao217: Ah good catch, will update it!

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -156,8 +155,7 @@ type HNCConfigurationCondition struct {
 	//
 	// Currently, the supported values are:
 	//
-	// - "critSingletonNameInvalid": the specified singleton name is invalid. The name should be the
-	// same as HNCConfigSingleton.
+	// (We don't have any critical conditions for now.)
```
adrianludwin, line 158: We don't need this comment IMO, also it would look weird as the beginning of a list. Maybe just delete the paragraph on critical conditions completely?

yiqigao217, line 158: I still kept the description (line 150-154) about critical or non-critical condition. Please lmk if you would prefer that to be removed entirely.

yiqigao217: ah sg!

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -148,17 +147,11 @@ type HNCConfigurationCondition struct {
 	// shown below, but new values may be added over time. This field is always present in a
 	// condition.
 	//
-	// All codes that begin with the prefix `crit` indicate that reconciliation has
-	// been paused for this configuration. Future changes of the configuration will be
-	// ignored by HNC until the condition has been resolved. Non-critical conditions
-	// typically indicate some kinds of error that HNC itself can ignore. However,
+	// Conditions typically indicate some kinds of error that HNC itself can ignore. However,
```
yiqigao217, line 150: PTAL here.

adrianludwin: lg

yiqigao217:
Part of #868

PR #842, opened 2020-07-03
Add timer for webhook server to get ready before test

### LINKED ISSUE #825: HNC: Failed e2e tests right after `make deploy`
Since it takes about 30s for the webhook server to be available after deployment, running e2e tests right after `make deploy` will fail (won't be able to create any namespaces because the webhook is fail close). e.g. 
```
Error from server (InternalError): Internal error occurred: failed calling webhook "namespaces.hnc.x-k8s.io": Post https://hnc-webhook-service.hnc-system.svc:443/validate-v1-namespace?timeout=30s: dial tcp 10.110.188.85:443: connect: connection refused
```

It took me a while to figure this out. Here are the steps:
1. I found creating namespaces outside the e2e tests using `exec.Command` won't fail (because the webhooks are already up then) and creating other resources in the e2e tests also worked. 
2. Therefore, I just tested deleting the `validatingwebhookconfiguration` to disable webhooks and run e2e tests by `go test ./e2e/...` instead of `make local-e2e-test` (with `make kind-deploy` in it before running e2e tests). The namespaces can be created! 

We will need to run the e2e tests after making sure the webhooks are up. I don't have a good idea of how to automate it for now. Will update in the comments.

GinnyJI:
Add a timer between deploy and go test to give webhook server enough
time to get ready. This change also invalidates the cached test results
before running the e2e test. Without 'go clean -testcache' it will always 
output the cached test result instead of the real one.

Tested:
On a real cluster, run:
    kubectl delete -f manifests/hnc-manager.yaml
    make e2e-test
On kind, run:
    make kind-reset
    make local-e2e-test

Fix #825

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -182,13 +182,20 @@ kind-deploy:
 	CONFIG=kind $(MAKE) deploy
 
 ###################### E2E TEST #########################
-
+# Sleep time is added for the webhook to get ready before test. The time 
+# comes from watching the log as well as multiple experiments. 
 .PHONY: e2e-test
 e2e-test: deploy
+	go clean -testcache
+	sleep 15
 	go test ./e2e/...
 
+# The sleep time needed for kind cluster is longer than that for a real 
+# cluster, from the experiments. I don't know why - @GinnyJI, July 2020
 .PHONY: local-e2e-test
 local-e2e-test: kind-deploy
+	go clean -testcache
+	sleep 20
 	go test ./e2e/...
 
 ###################### RELEASE ACTIONS #########################


adrianludwin:
Can you please take "Fix #825" out of the commit message? You can manually add that to the description of the PR, but it shouldn't be in the commit message.

incubator/hnc/Makefile
```diff
@@ -185,10 +185,14 @@ kind-deploy:
 
 .PHONY: e2e-test
 e2e-test: deploy
+	go clean -testcache
+	sleep 15
```
adrianludwin, line 189: Document where these numbers came from, and (more importantly) why they're different between the local and remote versions.

rjbez17:
lgtm after @adrianludwin comment is handled (and you sign the CLA).

GinnyJI:
I signed it

incubator/hnc/Makefile
GinnyJI: The numbers come from multiple experiments. I'm not sure why webhook server on kind takes more time.

PR #853, opened 2020-07-06
Prevent deletion of propagated objects

### LINKED ISSUE #845: HNC: don't allow deletion of propagated objects
In v0.5, the object controller was fixed to apply to all objects, but it only protected the "create" and "update" operations. Users can still _delete_ objects, and while HNC will quickly recreate them, this does allow a very short gap during which the policies won't be applied.

I have a working patch already but I don't want to hold up v0.5.0 as we should probably test this for at least a week or so. We may fix this in v0.5.1.

/assign @adrianludwin

adrianludwin:
Restructures the object validation controller to prevent deletions. Also
simplifiest the logic by explicitly checking the operation
(create/update/delete) that we're trying to validated.

Tested: new unit tests; tried a variety of legal and illegal
modifications on GKE 1.16 and confirmed that everything passed or failed
as expected in included namespaces, and that every change (including
illegal ones) pass in the excluded namespace `kube-system`. Confirmed
log messages are always printed out on failure while successes are never
printed.

Fixes #845

### DIFF
--- a/incubator/hnc/config/webhook/manifests.yaml
+++ b/incubator/hnc/config/webhook/manifests.yaml
@@ -96,5 +96,6 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
+    - DELETE
     resources:
     - '*'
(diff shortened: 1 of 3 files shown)


adrianludwin:
/assign @rjbez17 
/assign @yiqigao217 
cc @GinnyJI

PR #855, opened 2020-07-06
Prevent deletion of propagated objects

### LINKED ISSUE #854: HNC: don't allow deletion of propagated objects (v0.5)
This is to backport the fix to #845 into 0.5.1.

adrianludwin:
Restructures the object validation controller to prevent deletions. Also
simplifiest the logic by explicitly checking the operation
(create/update/delete) that we're trying to validated.

Tested: new unit tests; tried a variety of legal and illegal
modifications on GKE 1.16 and confirmed that everything passed or failed
as expected in included namespaces, and that every change (including
illegal ones) pass in the excluded namespace `kube-system`. Confirmed
log messages are always printed out on failure while successes are never
printed.

Fixes #854

### DIFF
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -5,6 +5,7 @@ import (
 	"reflect"
 
 	"github.com/go-logr/logr"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -27,7 +28,7 @@ const (
 // Note: the validating webhook FAILS OPEN. This means that if the webhook goes down, all further
 // changes to the objects are allowed.
 //
-// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="*",resources="*",verbs=create;update,versions="*",name=objects.hnc.x-k8s.io
+// +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="*",resources="*",verbs=create;update;delete,versions="*",name=objects.hnc.x-k8s.io
 
 type Object struct {
 	Log     logr.Logger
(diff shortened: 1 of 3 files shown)


adrianludwin:
This is the cherrypick for 0.5.1.

/assign @rjbez17 
/assign @yiqigao217

incubator/hnc/internal/validators/object.go
```diff
@@ -37,36 +38,52 @@ type Object struct {
...
 	if !o.isPropagateType(req.Kind) {
 		return allow("Non-propagate-mode types")
 	}
+	// Finally, let the HNC SA do whatever it wants.
 	if isHNCServiceAccount(&req.AdmissionRequest.UserInfo) {
 		log.V(1).Info("Allowed change by HNC SA")
 		return allow("HNC SA")
 	}
 
+	// Decode the old and new object. We validate both creations and deletions, so either of these can
```
yiqigao217, line 62: 'We validate both creations and deletions' seems confusing here. Maybe 'Either of these can be missing for creations or deletions'?

incubator/hnc/internal/validators/object.go
```diff
@@ -80,37 +97,44 @@ func (o *Object) isPropagateType(gvk metav1.GroupVersionKind) bool {
...
-	if newExists && newValue != oldValue {
-		return deny(metav1.StatusReasonForbidden, "Users should not change the value of label "+api.LabelInheritedFrom)
-	}
 
-	// If the existing object has an inheritedFrom label, it's a propagated object.
-	// Any user changes should be rejected.
-	if newExists && !reflect.DeepEqual(object.Canonical(inst), object.Canonical(oldInst)) {
-		return deny(metav1.StatusReasonForbidden,
-			"Illegal modification to an object propagated by the Hierarchical Namespace Controller")
+		return allow("all modifications allowed")
```
yiqigao217, line 132: What does the modifications imply here?

yiqigao217: Oh it's still in the case of `update`, but the only possibility is `DeepEqual` here, right?

incubator/hnc/internal/validators/object_test.go
```diff
@@ -102,13 +116,7 @@ func TestInheritedFromLabel(t *testing.T) {
 		name:     "Label is added",
 		newLabel: api.LabelInheritedFrom, newValue: "foo",
 		fail: true,
-	}, {
```
yiqigao217, line 105: This case is removed because it's handled in the `Handle()` func but not `handle()` func now?

adrianludwin: Yup, exactly.

incubator/hnc/internal/validators/object.go
adrianludwin: Done

incubator/hnc/internal/validators/object.go
adrianludwin: Yup, updated comment, PTAL

adrianludwin:
/hold

Should check in the master branch first

adrianludwin:
/hold cancel

Actually nvm, doesn't really matter what order they go in

PR #858, opened 2020-07-07
Remove deploy before E2E test

GinnyJI:
The user should be responsible to have the test cluster ready before
running the E2E test. It could be either a kind cluster or a real k8s
cluster.

Tested:
    on a real cluster: make e2e-test
    on a kind cluster: make local-e2e-test

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -182,20 +182,12 @@ kind-deploy:
 	CONFIG=kind $(MAKE) deploy
 
 ###################### E2E TEST #########################
-# Sleep time is added for the webhook to get ready before test. The time 
-# comes from watching the log as well as multiple experiments. 
+# This test will run on the current cluster the user deployed (either kind or 
+# a kubernetes cluster).
 .PHONY: e2e-test
-e2e-test: deploy
+e2e-test: 
+	@echo "If these tests fail due to the webhook not being ready, wait 30s and try again. Note that webhooks can take up to 30s to become ready after HNC is first deployed in a cluster."
 	go clean -testcache
-	sleep 15
-	go test ./e2e/...
-
-# The sleep time needed for kind cluster is longer than that for a real 
-# cluster, from the experiments. I don't know why - @GinnyJI, July 2020
-.PHONY: local-e2e-test
-local-e2e-test: kind-deploy
-	go clean -testcache
-	sleep 20
 	go test ./e2e/...
 
 ###################### RELEASE ACTIONS #########################


incubator/hnc/Makefile
```diff
@@ -182,18 +182,22 @@ kind-deploy:
...
+# a kubernetes cluster). Please make sure to have the test cluster ready 
+# before running the test.
+
 # Sleep time is added for the webhook to get ready before test. The time 
 # comes from watching the log as well as multiple experiments. 
 .PHONY: e2e-test
-e2e-test: deploy
+e2e-test: 
 	go clean -testcache
 	sleep 15
```
adrianludwin, line 194: Do we need the "sleep" now that we're asking the user to set up the cluster ahead of time?

incubator/hnc/Makefile
```diff
@@ -182,18 +182,22 @@ kind-deploy:
...
 	go test ./e2e/...
 
 # The sleep time needed for kind cluster is longer than that for a real 
 # cluster, from the experiments. I don't know why - @GinnyJI, July 2020
 .PHONY: local-e2e-test
-local-e2e-test: kind-deploy
+local-e2e-test:
```
adrianludwin, line 200: I'm not sure if we need this separate target now? Just have `e2e-test`?

incubator/hnc/Makefile
yiqigao217: Ah right. And agreed that we don't need 2 targets now.

GinnyJI: Done

incubator/hnc/Makefile
GinnyJI: Done

incubator/hnc/Makefile
```diff
@@ -182,21 +182,17 @@ kind-deploy:
 	CONFIG=kind $(MAKE) deploy
 
 ###################### E2E TEST #########################
-# Sleep time is added for the webhook to get ready before test. The time 
-# comes from watching the log as well as multiple experiments. 
+# This test will run on the current cluster the user deployed (either kind or 
+# a kubernetes cluster). Please make sure to have the test cluster ready
```
adrianludwin, line 186: Don't need to have the "please make sure" in the comment since it's displayed to the user, which is a much more visible spot.

Also, don't put a blank line between the comment and the rule.

incubator/hnc/Makefile
```diff
@@ -182,21 +182,17 @@ kind-deploy:
...
-# comes from watching the log as well as multiple experiments. 
+# This test will run on the current cluster the user deployed (either kind or 
+# a kubernetes cluster). Please make sure to have the test cluster ready 
+# before running the test.
+
 .PHONY: e2e-test
-e2e-test: deploy
+e2e-test: 
+	@echo "Please make sure that the webhook server is ready."
+	@echo "It takes around 20 seconds for the webhook server to get ready in a fresh cluster."
```
adrianludwin, line 192: This could be confusing, because it seems to ask users to do something but usually, nothing will be required.

How about something like: "If these tests fail due to the webhook not being ready, wait 30s and try again. Note that webhooks can take up to 30s to become ready after HNC is first deployed in a cluster."

GinnyJI: Done

incubator/hnc/Makefile
GinnyJI: Done

adrianludwin:
Whoops forgot to remove the hold

/hold cancel

PR #860, opened 2020-07-07
convert test-delete-anchor-crd to go test file

GinnyJI:
convert hack/test-delete-anchor-crd.sh to
e2e/tests/delete-anchor-crd_test.go

Tested: make e2e-test

### DIFF
--- a/incubator/hnc/e2e/tests/delete_crd_test.go
+++ b/incubator/hnc/e2e/tests/delete_crd_test.go
@@ -1,0 +1,41 @@
+package test
+
+import (
+	"os"
+
+	. "github.com/onsi/ginkgo"
+)
+
+var _ = Describe("Delete-anchor-crd", func() {
+	hncRecoverPath := os.Getenv("HNC_REPAIR")
+	AfterEach(func() {
+		// clean up
+		nses := []string{
+			"delete-crd-parent",
+			"delete-crd-child",
+		}
+		// Remove all possible objections HNC might have to deleting a network. Make sure it 
+		// has cascading deletion so we can delete any of its subnamespace descendants, and 
+		// make sure that it's not a subnamespace itself so we can delete it directly.
+		for _, ns := range nses {
+			tryRun("kubectl hns set", ns, "-a")
+			tryRun("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
+			tryRun("kubectl delete ns", ns)
+		}
+		mustRun("kubectl apply -f", hncRecoverPath)
+	})
+
+	It("should create parent and deletable child, and delete the CRD", func() {
+		// we don't want to destroy the HNC without being able to repair it, so skip this test if recovery path not set
+		if hncRecoverPath == ""{
+			Skip("Environment variable HNC_REPAIR not set. Skipping reocovering HNC.")
+		}
+		// set up
+		mustRun("kubectl create ns delete-crd-parent")
+		mustRun("kubectl hns create delete-crd-child -n delete-crd-parent")
+		// test
+		mustRun("kubectl delete customresourcedefinition/subnamespaceanchors.hnc.x-k8s.io")
+		// verify
+		mustRun("kubectl get ns delete-crd-child")
+	})
+})
--- a/incubator/hnc/e2e/tests/e2e_test.go
+++ b/incubator/hnc/e2e/tests/e2e_test.go
@@ -3,6 +3,8 @@ package test
 import (
 	"testing"
 	"time"
+	"os/exec"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -18,4 +20,21 @@ func TestE2e(t *testing.T) {
 	RunSpecsWithDefaultAndCustomReporters(t,
 		"HNC Suite",
 		[]Reporter{printer.NewlineReporter{}})
+}
+
+func mustRun(cmdln ...string) {
+	err := tryRun(cmdln...)
+	Expect(err).Should(BeNil())
+}
+
+func tryRun(cmdln ...string) error {
+	var args []string
+	for _, subcmdln := range cmdln {
+		args = append(args, strings.Split(subcmdln, " ")...)
+	}
+	GinkgoT().Log("Running: ", args)
+	cmd := exec.Command(args[0], args[1:]...)
+	stdout, err := cmd.CombinedOutput()
+	GinkgoT().Log("Output: ", string(stdout))
+	return err
 }


incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,35 @@
...
+		// create a namespace delete-crd-child and set its parent as delete-crdparent
+		cmd = exec.Command("kubectl", "hns", "create", "delete-crd-child", "-n", "delete-crd-parent")
+		Expect(cmd.Run()).Should(BeNil())
+
+		// allow cascading delete for delete-crd-child
+		cmd = exec.Command("kubectl", "hns", "set", "delete-crd-child", "--allowCascadingDelete")
+		Expect(cmd.Run()).Should(BeNil())
+	})
+
+	It("should delete the CRD", func() {
```
adrianludwin, line 25: Each "It()" block should be a complete test, and also there's no guarantee that they're run in order AFAIK. So this should all be one single "It()" block.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Delete-anchor-crd", func() {
+	It("should create parent and deletable child", func() {
+		// create a namespace delete-crd-parent
+		cmd := exec.Command("kubectl", "create", "ns", "delete-crd-parent")
+		Expect(cmd.Run()).Should(BeNil())
```
adrianludwin, line 14: Feel free to start building utility functions if you like. Eg this could be something like:

`RunSuccessfully("kubectl create ns delete-crd-parent")`

where `RunSuccessfully` creates the command, runs it and validates that it succeeds.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+		// allow cascading delete for delete-crd-child
+		cmd = exec.Command("kubectl", "hns", "set", "delete-crd-child", "--allowCascadingDelete")
+		Expect(cmd.Run()).Should(BeNil())
+	})
+
+	It("should delete the CRD", func() {
+		// delete the crd
+		cmd := exec.Command("kubectl", "delete", "ns", "customresourcedefinition.apiextensions.k8s.io/subnamespaceanchors.hnc.x-k8s.io")
```
adrianludwin, line 27: Instead of `customresourcedefinition.apiextensions.k8s.io`, you can say `customresourcedefinition` or `crd`, which are both synonyms.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,35 @@
...
+		Expect(cmd.Run()).Should()
+		// TODO: @ginnyji - output should be "customresourcedefinition.apiextensions.k8s.io "subnamespaceanchors.hnc.x-k8s.io" deleted"
```
adrianludwin, line 29: Do you _need_ to check that text output? Or will it return a non-nil error if it fails?

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
adrianludwin: Or you could build something higher-level like `CreateNamespace("parent")`

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
yiqigao217: Do you need "ns" here to delete crd?

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
yiqigao217: Or just have a expected failed `kubectl get crd subnamespaceanchor` after the deletion?

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+		// verify that the namespace still exists
+		cmd = exec.Command("kubectl", "get", "ns", "delete-crd-child")
+		Expect(cmd.Run()).Should(BeNil())
+	})
+})
```
yiqigao217, line 35: Add one more empty line here at the end.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
adrianludwin: Ah, I missed that :) No, you shouldn't need that ("ns" is the short synonym for "namespace").

adrianludwin:
Also - I know the original test script didn't try to clean up after itself, but can you please try to delete all the namespaces you create in the test? Use the [AfterEach()](https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach) function to do cleanup, since the test will immediately stop running if it hits a problem, but `AfterEach` will always be called to do the cleanup.

In a later PR, we can worry about undeleting that CRD :)

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Done

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Done. 
Removed ns and changed to customresourcedefinition

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Done

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,55 @@
...
+		Expect(err).Should(BeNil())
+
+		// cleaning up: fix hnc
+		cmd = exec.Command("kubectl", "apply", "-f", "manifests/hnc-manager.yaml")
+		stdout, err = cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
+
+		// cleaning up: delete namespace
+		// Note: the namespace delete-crd-child could not be deleted due to a known issue #847
```
adrianludwin, line 49: `delete-crd-child` is a [subnamespace](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/concepts.md#basic-subns), which means it has the `hnc.x-k8s.io/subnamespaceOf` [annotation](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/). You should be able to remove this as follows:

`kubectl annotate namespace delete-crd-child hnc.x-k8s.io/subnamespaceOf-`

Note the "minus" at the end of that line - that means "remove the annotation." Then you should be able to delete the namespace normally.

adrianludwin:
There should be a way to make sure that the tests don't run in parallel -
check the Gingko documentation. I *think* that by default, they don't run
in parallel, but they do run in a random order.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: The problem was because of the extra 'ns'. We should be expecting a Nil

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Done

GinnyJI: Done

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,68 @@
...
+		Expect(err).Should(BeNil())
+
+		// create a namespace delete-crd-child and set its parent as delete-crdparent
+		cmd = exec.Command("kubectl", "hns", "create", "delete-crd-child", "-n", "delete-crd-parent")
+		stdout, err = cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
+
+		// allow cascading delete for delete-crd-child
+		cmd = exec.Command("kubectl", "hns", "set", "delete-crd-child", "-a")
```
yiqigao217, line 26: Why do we need to set `allowCascadingDelete` here? Seems it's irrelevant here?

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,68 @@
...
+		stdout, err = cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
+
+		cmd = exec.Command("kubectl", "delete", "ns", "delete-crd-child")
+		stdout, err = cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
+
+
```
yiqigao217, line 66: Remove extra empty lines.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
adrianludwin: I put that in the original test. Hmm. I didn't comment _why_ I did that (d'oh!) but if I had to guess, I wanted to make it as easy as possible for HNC to delete it, and then make sure that it didn't.

But now that you don't need `allowCascadingDelete` to delete a leaf subnamespace anyway, I think we can delete this line too.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
adrianludwin: Couple of things here:

* Probably better to delete the child first, then the parent. In HNC, having a child can sometimes prevent the parent from being deleted.
* You should document _why_ you made the change that we recommended. You added the code, but will the next person to read the test understand why it's there?

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,74 @@
...
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Delete-anchor-crd", func() {
+	It("should create parent and deletable child, and delete the CRD", func() {
+		GinkgoT().Log(time.Now())
+
+		// create a namespace delete-crd-parent
+		err := runcmd("kubectl create ns delete-crd-parent")
+		Expect(err).Should(BeNil())
```
adrianludwin, line 19: Much better! But the line `Expect(err).Should(BeNil())` is repeated every time - why not include that in `runcmd` as well?

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Delete-anchor-crd", func() {
+	It("should create parent and deletable child, and delete the CRD", func() {
+		GinkgoT().Log(time.Now())
+
+		// create a namespace delete-crd-parent
```
adrianludwin, line 17: You don't need to comment every line, especially if the code is self-explaining. For example, if the code is:

`runcmd("kubectl create ns delete-crd-parent")`

Then the comment

`// Create the namespace delete-crd-parent`

basically just repeats what the code says, and doesn't add any new information. What might be more useful is:

```
// Setup
runcmd("kubectl create ns delete-crd-parent")
runcmd("kubectl hns create delete-crd-child -n delete-crd-parent")

// Test
runcmd("kubectl hns set delete-crd-child -a")

// Verify
...

// Cleanup
...
```

Because now the reader can very quickly look at the part of the test they care about. Then you only need to add _additional_ comments if it's unclear _why_ the code is doing what it's doing.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Removed

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: I was thinking what if we are actually expecting an error some time

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,55 @@
...
+	AfterEach(func() {
+		// clean up
+		nses := []string{
+			"delete-crd-parent",
+			"delete-crd-child",
+		}
+
+		for _, ns := range nses {
+			runSuccess("kubectl hns set", ns, "-a")
+			//remove subname annotation
```
adrianludwin, line 24: Comment: "Remove all possible objections HNC might have to deleting a network. Make sure it has cascading deletion so we can delete any of its subnamespace descendants, and make sure that it's not a subnamespace itself so we can delete it directly."

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,55 @@
...
+			runSuccess("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
```
adrianludwin, line 25: Consider adding a `tryRun` function that doesn't insist on succeeding (e.g. in case the namespace doesn't exist). Maybe rename `runSuccess` to `mustRun`, and `mustRun` can call `tryRun`.

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,55 @@
...
+			runSuccess("kubectl delete ns", ns)
+		}
+
+		runSuccess("kubectl apply -f", os.Getenv("HNC_REPAIR"))
+		GinkgoT().Log(time.Now())
```
adrianludwin, line 30: Either take this out if it was just for short-term debugging, or add some text like `Log("Time finished:", time.Now())`. Ditto on line 34

incubator/hnc/e2e/tests/delete_crd_test.go
adrianludwin: Note that the call to `kubectl apply` should always succeed.

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Added function runSuccess()

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,55 @@
...
+		cmd := exec.Command("kubectl", "create", "ns", "delete-crd-parent")
+		stdout, err := cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
+
+		// create a namespace delete-crd-child and set its parent as delete-crdparent
+		cmd = exec.Command("kubectl", "hns", "create", "delete-crd-child", "-n", "delete-crd-parent")
+		stdout, err = cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
GinnyJI: Done

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,43 @@
...
+	})
+
+	It("should create parent and deletable child, and delete the CRD", func() {
+		GinkgoT().Log(time.Now())
+		// set up
```
adrianludwin, line 35: Don't forget to `Skip()` if `HNC_REPAIR` isn't set.

Also, since the `"HNC_REPAIR"` string would now be used twice (if you follow this advice), it's a good idea to replace it with a constant so if you mistype it somewhere, the compiler will complain. Maybe at the top of this file (around line 12):

```
const (
  hncYamlEnvVar = "HNC_REPAIR"
)
```

incubator/hnc/e2e/tests/e2e_test.go
```diff
@@ -19,3 +19,15 @@ func TestE2e(t *testing.T) {
...
+
+func  runSuccess(cmdln ...string) {
+	var args []string
+	for _, subcmdln := range cmdln {
+		args = append(args, strings.Split(subcmdln, " ")...)
+	}
+	GinkgoT().Log(args)
+	cmd := exec.Command(args[0], args[1:]...)
+	stdout, err := cmd.CombinedOutput()
+	GinkgoT().Log(string(stdout))
```
adrianludwin, line 31: Consider differentiating somehow between the command (line 28) and the output (here).

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: Removed

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: done

incubator/hnc/e2e/tests/delete-anchor-crd_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+		// verify that the namespace still exists
+		cmd = exec.Command("kubectl", "get", "ns", "delete-crd-child")
+		stdout, err = cmd.CombinedOutput()
+		GinkgoT().Log(string(stdout))
+		Expect(err).Should(BeNil())
+
+		// cleaning up: fix hnc
+		cmd = exec.Command("kubectl", "apply", "-f", "manifests/hnc-manager.yaml")
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: done

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: done

incubator/hnc/e2e/tests/e2e_test.go
GinnyJI: added prefix messages

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,44 @@
...
+			"delete-crd-parent",
+			"delete-crd-child",
+		}
+		// Remove all possible objections HNC might have to deleting a network. Make sure it 
+		// has cascading deletion so we can delete any of its subnamespace descendants, and 
+		// make sure that it's not a subnamespace itself so we can delete it directly.
+		for _, ns := range nses {
+			mustRun("kubectl hns set", ns, "-a")
+			tryRun("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
+			mustRun("kubectl delete ns", ns)
```
adrianludwin, line 26: I think these should all by tryRun - what happens if the namespace wasn't successfully created? Then they'll all fail.

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		hncRecoverPath := os.Getenv(hncYamlEnvVar)
+		if hncRecoverPath == ""{
+			Skip("Environment variable HNC_REPAIR not set. Skipping reocovering HNC.")
+		}
```
adrianludwin, line 31: Did you try this out? I would have thought that the `Skip()` function had to be part of the _test_ (the `It()` function), not the cleanup.

adrianludwin: When you've tested this, can you please update the commit comment as well to say that you tested it with and without HNC_REPAIR set?

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: done

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: Yes tested it and in the log it says "[SKIPPING] in Spec Teardown (AfterEach)".
Added to the commit message

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -0,0 +1,41 @@
...
+		// Remove all possible objections HNC might have to deleting a network. Make sure it 
+		// has cascading deletion so we can delete any of its subnamespace descendants, and 
+		// make sure that it's not a subnamespace itself so we can delete it directly.
+		for _, ns := range nses {
+			tryRun("kubectl hns set", ns, "-a")
+			tryRun("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
+			tryRun("kubectl delete ns", ns)
+		}
+		mustRun("kubectl apply -f", hncRecoverPath)
```
adrianludwin, line 25: @GinnyJI Can you please file an issue to make this more robust? I think I know what you mean - that the recovery "has" to work - but mustRun will only stop the current _test_ from running if it fails, not the current _test suite_. We might want to do something more like:

```
if err := tryRun( ... ); err != nil {
  GinkoT().Logf("Couldn't repair HNC: %v", err)
  os.Exit(1)
}
```

That will stop all _other_ tests from running, which is probably what we want here.

GinnyJI: Done at issue #881

PR #866, opened 2020-07-09
Allow subnamespace deletion if parent/anchor is missing

### LINKED ISSUE #847: HNC: allow direct deletion of a subnamespace if its parent doesn't exist
/assign @yiqigao217 

This comes up in some of the e2e tests - a namespace with the subnamespaceOf annotation exists, but its parent doesn't. We should allow it to be deleted.

yiqigao217:
Since we allow changes to the subnamespaceOf annotation on subnamespace
as an escape hatch and there's a scenario of recovering from CRD
deletion, this commit makes it easier to delete a subnamespace, but only
when the parent or the anchor is missing. If the anchor is ok (state),
we still don't allow direct deletion of a subnamespace and only allows
deletion from the anchor.

Tested with new unit test cases and manually before and after.

Fixes #847

### DIFF
--- a/incubator/hnc/internal/validators/namespace.go
+++ b/incubator/hnc/internal/validators/namespace.go
@@ -117,7 +117,15 @@ func (v *Namespace) conflictBetweenParentAndExternalManager(req *nsRequest, ns *
 
 func (v *Namespace) cannotDeleteSubnamespace(req *nsRequest) admission.Response {
 	parent := req.ns.Annotations[api.SubnamespaceOf]
-	if parent != "" {
+	// Early exit if the namespace is not a subnamespace.
+	if parent == "" {
+		return allow("")
+	}
+
+	// If the anchor doesn't exist, we want to allow it to be deleted anyway.
+	// See issue https://github.com/kubernetes-sigs/multi-tenancy/issues/847.
+	anchorExists := v.Forest.Get(parent).HasAnchor(req.ns.Name)
+	if anchorExists {
 		msg := fmt.Sprintf("The namespace %s is a subnamespace. Please delete the anchor from the parent namespace %s to delete the subnamespace.", req.ns.Name, parent)
 		return deny(metav1.StatusReasonForbidden, msg)
 	}
--- a/incubator/hnc/internal/validators/namespace_test.go
+++ b/incubator/hnc/internal/validators/namespace_test.go
@@ -12,29 +12,58 @@ import (
 )
 
 func TestDeleteSubNamespace(t *testing.T) {
-	// Create a namespace with owner annotation.
-	sub := &corev1.Namespace{}
-	sub.Name = "a"
-	setSubAnnotation(sub, "b")
 
-	// a (subnamespace of b) -> b
-	f := foresttest.Create("B-")
-	vns := &Namespace{Forest: f}
+	// We will put an anchor a in the namespace c if it exists.
+	tests := []struct {
+		name string
+		// We only use the forest here to create full namespace hierarchy.
+		forest string
+		// The subnamespaces are created by manually adding the annotation.
+		subnamespaceOf string
+		fail           bool
+	}{
+		// There's no test case for when parent and annotation don't match, since the
+		// reconciler always sets the parent to match the subnamespaceOf annotation.
+		// - a (subnamespaceOf b)
+		{name: "when parent is missing", forest: "-", subnamespaceOf: "b"},
+		// - a (subnamespaceOf b) -> b (no anchor)
+		{name: "when anchor is missing", forest: "b-", subnamespaceOf: "b"},
+		// - a (subnamespaceOf c) -> c (has anchor a), b
+		{name: "when annotation and anchor match", forest: "c--", subnamespaceOf: "c", fail: true},
+		// - a (subnamespaceOf b) -> b, c(has anchor a)
+		{name: "when annotation and anchor don't match", forest: "b--", subnamespaceOf: "b"},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
 
-	t.Run("Delete namespace with owner annotation", func(t *testing.T) {
-		g := NewGomegaWithT(t)
-		req := &nsRequest{
-			ns: sub,
-			op: v1beta1.Delete,
-		}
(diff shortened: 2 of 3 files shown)


incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,19 +64,15 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
 // handle implements the non-boilerplate logic of this validator, allowing it to be more easily unit
 // tested (ie without constructing a full admission.Request).
 func (v *Namespace) handle(req *nsRequest) admission.Response {
+	v.Forest.Lock()
+	defer v.Forest.Unlock()
+
 	parent := req.ns.Annotations[api.SubnamespaceOf]
-	if parent != "" && req.op == v1beta1.Delete {
+	if parent != "" && v.Forest.Get(parent).HasAnchor(req.ns.Name) && req.op == v1beta1.Delete {
```
adrianludwin, line 71: This needs a lot more commenting IMO plus maybe a bit of refactoring. E.g. something like:

```
// If the anchor doesn't exist, we want to allow it to be deleted anyway because ...
// See issue #xxx
anchorExists := v.Forest.Get(parent).HasAnchor(req.ns.Name) <-- should be false if this isn't a subns, I think
if req.op == v1beta1.Delete && anchorExists {
   msg := ...
   return deny(forbidden, msg)
}
```

incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,19 +64,15 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
...
 		return deny(metav1.StatusReasonForbidden, msg)
 	}
 
-	return v.checkForest(req)
-}
-
-func (v *Namespace) checkForest(req *nsRequest) admission.Response {
-	v.Forest.Lock()
-	defer v.Forest.Unlock()
-
```
adrianludwin, line 79: This function is already much too long IMO, I don't think we should be making it longer. What's the motivation behind merging them?

If anything, maybe hold off on this change, submit another PR to restructure this webhook so each function is much smaller (maybe max 40 lines) and then reapply this change?

yiqigao217: Ah the only reason I did so is because we need to check (lock) the forest for that part too now before `checkForest()` func.
sgtm, I will work on the restructure first then come back to this PR.

incubator/hnc/internal/validators/namespace.go
yiqigao217: sg. updated!

incubator/hnc/internal/validators/namespace.go
yiqigao217: Done in #870

incubator/hnc/internal/validators/namespace_test.go
```diff
@@ -12,29 +12,53 @@ import (
 )
 
 func TestDeleteSubNamespace(t *testing.T) {
-	// Create a namespace with owner annotation.
+	// Create a namespace instance a.
 	sub := &corev1.Namespace{}
 	sub.Name = "a"
```
adrianludwin, line 17: Maybe move this into the `t.Run` portion of the test? That's where we actually use it, plus it will ensure we get a fresh copy for each subtest.

incubator/hnc/internal/validators/namespace_test.go
```diff
@@ -12,29 +12,53 @@ import (
...
-	f := foresttest.Create("B-")
-	vns := &Namespace{Forest: f}
+	// We will put an anchor a in the namespace c if it exists.
+	tests := []struct {
+		name           string
+		forest         string
+		subnamespaceOf string
+		fail           bool
+	}{
+		// There's no test case for when parent and anchor don't match, because the
```
adrianludwin, line 26: Do you mean: "when parent and annotation don't match"?

incubator/hnc/internal/validators/namespace_test.go
```diff
@@ -12,29 +12,53 @@ import (
...
-	setSubAnnotation(sub, "b")
 
-	// a (subnamespace of b) -> b
-	f := foresttest.Create("B-")
-	vns := &Namespace{Forest: f}
+	// We will put an anchor a in the namespace c if it exists.
+	tests := []struct {
+		name           string
+		forest         string
+		subnamespaceOf string
```
adrianludwin, line 23: Maybe comment this a bit more? E.g., "we'll use the forest to create only full namespaces, and then manually turn them into subnamespaces by adding the annotations" or something like that?

incubator/hnc/internal/validators/namespace_test.go
```diff
@@ -12,29 +12,53 @@ import (
...
 
-	t.Run("Delete namespace with owner annotation", func(t *testing.T) {
-		g := NewGomegaWithT(t)
-		req := &nsRequest{
-			ns: sub,
-			op: v1beta1.Delete,
-		}
+			// Construct the forest
+			f := foresttest.Create(tc.forest)
+			// Add anchor a to namespace c if it exists.
```
adrianludwin, line 48: This needs more comments somewhere, either here or maybe on lines 32 and 34. I can see that we _are_ adding the anchor if `c` exists, but _why_ are we doing this? (I figured it out, I'd just like it to be a bit better documented in the code).

rjbez17:
lgtm after @adrianludwin comments are resolved.

incubator/hnc/internal/validators/namespace_test.go
yiqigao217: No, since parent and annotation always match.  There is a mistake here in the comment in line 34, the test case is subnamespaceOf b, but I commented "subnamespaceOf c". The comment is wrong. I updated it to b in the comment.

incubator/hnc/internal/validators/namespace_test.go
yiqigao217: I added the reason here in the same line.

incubator/hnc/internal/validators/namespace_test.go
yiqigao217: oh sorry somehow read it in the other way. will update..

yiqigao217:
/retest

Similar to another flaky test #871 . I added the failed test logs to the issue.

PR #867, opened 2020-07-09
Restore webhook in e2e test from the current manifest

### LINKED ISSUE #846: HNC: test-issue-797-* shouldn't rely on `manifests/hnc-manager.yaml`
/assign @yiqigao217 

These two scripts aren't working for me because:
1. I run them from the `hack/` directory, so `/manifests/` wasn't available, but more importantly...
2. I was using a released version of HNC, so `manifests/hnc-manager.yaml` might not be available and if it is, it refers to the _local_ version of HNC, not the deployed version.

Ideally, this test shouldn't rely on any pre-existing files on disk.

yiqigao217:
Remove the `kubectl apply -f manifests/hnc-manager.yaml` in the e2e
test as a way to restore webhook. Instead, output the webhook manifest
into a file and later apply this file to restore the webhook. This makes
sure the webhook is the one used in the current deployment when the e2e
test is run.

Tested by running the script and the generated yaml file is deleted
after each test.

Fixes #846

### DIFF
--- a/incubator/hnc/hack/test-issue-797-leaf.sh
+++ b/incubator/hnc/hack/test-issue-797-leaf.sh
@@ -11,6 +11,8 @@ kubectl create ns p1
 kubectl create ns p2
 sleep 1
 
+echo "${bold}Creating a copy of the webhook manifest at hnc_e2e_797_webhook.yaml...${normal}"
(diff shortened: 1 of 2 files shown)


yiqigao217:
Please let me know if there's a concern of writing/deleting a file in the test.
/assign @adrianludwin 
/assign @rjbez17 

/cc @GinnyJI , @adrianchung

Issue #868, opened 2020-07-09
HNC: Implement API v1alpha2


yiqigao217:
Useful links:
* Design doc - http://bit.ly/hnc_v1alpha2
* [v1alpha1](https://github.com/kubernetes-sigs/multi-tenancy/tree/master/incubator/hnc/api/v1alpha1) API review - #748

We will switch the served version and storage version from v1alpha1 to v1alpha2 in 0.6. Thus it's a one-way conversion and we don't support downgrade from 0.6 to 0.5.

We will start the development of v1a2 by simply copying v1a1, switching all our internal code to v1a2, and adding no-op conversion webhooks to make sure the framework is working. Then we will implement v1a2 directly.

/assign @yiqigao217

yiqigao217:
Closing this since the implementation is done and all the open "sub" issues are updating docs or not related to API conversion (#989)

PR #869, opened 2020-07-09
Add fake v1alpha2 and keep v1alpha1 the only served version

yiqigao217:
Make a copy of the v1alpha1 api folder and change it into v1alpha2 with
the exact same schema. Make v1alpha1 the only served version and the
storage version. Make v1alpha2 unserved.

The `kubebuilder:unservedversion` marker seems not working, so this
commit used `kubebuilder:skipversion` on v1alpha2 CRDs. Without this
marker, v1alpha2 had `served: true` and `storage: false`. When I looked
at the CRs, I saw v1alpha2 version is served.

We don't want to serve v1alpha2 before it's ready, so we will only
manually change it to be servedversion for testing during development.

Part of #868 
The issue with `unservedversion` marker is filed [here](https://github.com/kubernetes-sigs/controller-tools/issues/462).


incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -0,0 +1,298 @@
...
+)
+
+// ClearConditionCriteria is initialized in init(). See ClearConditionCriterion for more details.
+var ClearConditionCriteria map[Code]ClearConditionCriterion
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// +kubebuilder:object:root=true
+// +kubebuilder:skipversion
```
yiqigao217, line 95: This is the `skipversion` marker added to the v1alpha2 copy from v1alpha1. The `unservedversion` seemed not working, see [issue](https://github.com/kubernetes-sigs/controller-tools/issues/462).

yiqigao217:
> Why do you want to start with a full copy of _all_ the files, even the ones that will change over time? Did you consider only adding an empty skeleton for, say, HC but not adding anything else yet?

Because the changes are minor based on the v1alpha1 and I felt it's easier to review for later PR to make a diff on top of that, so reviewer can just see the changes instead of to review the newly-added code (which is changes upon v1alpha1).

adrianludwin:
Makes sense

/lgtm
/approve

yiqigao217:
/retest

The previous failed test logs are put in a gist and #871 is filed to investigate this failed test since it doesn't seem relevant.

PR #870, opened 2020-07-09
Restructure handle() in the namespace validator

yiqigao217:
Split handle() and checkForest() into smaller functions.

Tested manually and by `make test`.

Part of #847

### DIFF
--- a/incubator/hnc/internal/validators/namespace.go
+++ b/incubator/hnc/internal/validators/namespace.go
@@ -64,48 +64,73 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
 // handle implements the non-boilerplate logic of this validator, allowing it to be more easily unit
 // tested (ie without constructing a full admission.Request).
 func (v *Namespace) handle(req *nsRequest) admission.Response {
-	parent := req.ns.Annotations[api.SubnamespaceOf]
-	if parent != "" && req.op == v1beta1.Delete {
-		msg := fmt.Sprintf("The namespace %s is a subnamespace. Please delete the anchor from the parent namespace %s to delete the subnamespace.", req.ns.Name, parent)
-		return deny(metav1.StatusReasonForbidden, msg)
-	}
-
-	return v.checkForest(req)
-}
-
-func (v *Namespace) checkForest(req *nsRequest) admission.Response {
 	v.Forest.Lock()
 	defer v.Forest.Unlock()
 
 	ns := v.Forest.Get(req.ns.Name)
 
-	// Allow all create namespace requests except if the namespace name already exists in
-	// the external hierarchy (the apiserver already denies existing namespace names).
-	if req.op == v1beta1.Create {
-		for _, nm := range v.Forest.GetNamespaceNames() {
-			if _, ok := v.Forest.Get(nm).ExternalTreeLabels[req.ns.Name]; ok {
-				msg := fmt.Sprintf("The namespace name %q is reserved by the external hierarchy manager %q.", req.ns.Name, v.Forest.Get(nm).Manager)
-				return deny(metav1.StatusReasonAlreadyExists, msg)
-			}
+	switch req.op {
+	case v1beta1.Create:
+		// This check only applies to the Create operation since namespace name
+		// cannot be updated.
+		if rsp := v.nameExistsInExternalHierarchy(req); !rsp.Allowed {
+			return rsp
+		}
+	case v1beta1.Update:
+		// This check only applies to the Update operation. Creating a namespace
+		// with external manager is allowed and we will prevent this conflict by not
+		// allowing setting a parent when validating the HierarchyConfiguration.
+		if rsp := v.conflictBetweenParentAndExternalManager(req, ns); !rsp.Allowed {
+			return rsp
+		}
+	case v1beta1.Delete:
+		if rsp := v.cannotDeleteSubnamespace(req); !rsp.Allowed {
+			return rsp
+		}
+		if rsp := v.illegalCascadingDeletion(ns); !rsp.Allowed {
+			return rsp
 		}
 	}
 
-	// Allow all update namespace requests except if the namespace has a parent and wants
-	// to add "hnc.x-k8s.io/managedBy" annotation other than "hnc.x-k8s.io".
-	if req.op == v1beta1.Update && req.ns.Annotations[api.AnnotationManagedBy] != "" {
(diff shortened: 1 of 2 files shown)


yiqigao217:
Thanks for the suggestion I used the 1st one you suggested. As for the 2nd, they do always use the same parameters, but some of the checkers don't need both. They may only use either `req` or `ns`. @adrianludwin

incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,48 +64,70 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
...
-	if req.op == v1beta1.Update && req.ns.Annotations[api.AnnotationManagedBy] != "" {
-		if req.ns.Annotations[api.AnnotationManagedBy] != api.MetaGroup && ns.Parent() != nil {
-			msg := fmt.Sprintf("Namespace %q is a child of %q. Namespaces with parents defined by HNC cannot also be managed externally. "+
-				"To manage this namespace with %q, first make it a root in HNC.", ns.Name(), ns.Parent().Name(), req.ns.Annotations[api.AnnotationManagedBy])
-			return deny(metav1.StatusReasonForbidden, msg)
+	return allow("")
+}
+
+// nameExistsInExternalHierarchy only applies to the Create operation since
+// namespace name cannot be updated.
```
adrianludwin, line 92: I'd actually put this comment on line 74, since that's where you'll be looking when you say "I wonder why this isn't called as part of the update..."

incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,48 +64,70 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
...
-	if req.op == v1beta1.Create {
-		for _, nm := range v.Forest.GetNamespaceNames() {
-			if _, ok := v.Forest.Get(nm).ExternalTreeLabels[req.ns.Name]; ok {
-				msg := fmt.Sprintf("The namespace name %q is reserved by the external hierarchy manager %q.", req.ns.Name, v.Forest.Get(nm).Manager)
-				return deny(metav1.StatusReasonAlreadyExists, msg)
-			}
+	switch req.op {
+	case v1beta1.Create:
+		// Allow all create namespace requests except if the namespace name already exists in
+		// the external hierarchy (the apiserver already denies existing namespace names).
```
adrianludwin, line 75: This comment could get out of date ("allow all except in this one case") as soon as we add a second case. I'm not actually sure that we really need this comment at all, or the one on line 78 - it's pretty self-explanatory.

incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,48 +64,70 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
...
+		return v.nameExistsInExternalHierarchy(req)
```
adrianludwin, line 76: Nit: TBH, even though it's a _bit_ more verbose, I'd actually suggest using the same pattern as on line 82 for _all_ these test, and then have a single `allow("")` at the end of the function. That makes it easier to add new tests in the future and also, I think, reads a bit better as English.

For example, it looks to me like `nameExistsInExternalHierarchy()` should return `true` if the name already exists, which means... that the request should be allowed? That's not right. But maybe if we call it:

```
if rsp := v.conflictsWithExternalHierarchy(req); !rsp.Allowed {
   return rsp
}
```

to me, that reads more clearly even though it's more verbose (which is usually the opposite of what I like).

Anyway, I don't actually feel too strongly about this, so up to you if you want to make this change or not.

incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,48 +64,70 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
...
+			return deny(metav1.StatusReasonAlreadyExists, msg)
 		}
 	}
+	return allow("")
+}
 
-	// Early exit to allow other non-delete requests or if the namespace allows cascading deletion.
-	if req.op != v1beta1.Delete || ns.AllowsCascadingDelete() {
+// conflictBetweenParentAndExternalManager only applies to the Update operation
+// to check if changing namespace manager is allowed.
```
adrianludwin, line 104: Ditto (this comment belongs in the switch statement, IMO). Maybe explain _why_ this can't be a problem with Create.

incubator/hnc/internal/validators/namespace.go
yiqigao217: Agreed on both new tests in the future and easy to read!

incubator/hnc/internal/validators/namespace.go
```diff
@@ -64,48 +64,73 @@ func (v *Namespace) Handle(ctx context.Context, req admission.Request) admission
...
-	// Early exit to allow other non-delete requests or if the namespace allows cascading deletion.
-	if req.op != v1beta1.Delete || ns.AllowsCascadingDelete() {
+func (v *Namespace) illegalCascadingDeletion(ns *forest.Namespace) admission.Response {
+	// Early exit if the namespace allows cascading deletion.
+	if ns.AllowsCascadingDelete() {
 		return allow("")
 	}
 
-	// Check if the deleting namespace has subnamespaces that can't be deleted.
+	// Check if all children allow cascading deletion.
```
adrianludwin, line 133: Maybe file an issue to consider removing this exception? It made sense back when we required allowCascadingDelete to delete a leaf namespace as well, but the new interpretation of aCD now seems to be "will this operation delete _any_ namespace other than the one explicitly named?"

yiqigao217: sg will test first and then file an issue.

yiqigao217: So i think we still need to keep it and it's a right behavior. Here's the test I did.
```
k create ns a
k hns create b -n a
k delete ns a
```
It shouldn't allow it because the user may not aware it would also delete ns b. Currently, with this rule, the webhook will deny the deletion. WDYT?

yiqigao217: Basically, we should allow `k delete subns b -n a` or ( `k hns set a -a` + `k delete ns a` ), but not `k delete ns a` directly. WDYT?

adrianludwin: Yeah I was talking about making the webhook _more_ strict, not _less_. We can discuss it in another issue, but for now the behaviour of the webhook is correct.

yiqigao217: Thanks for the clarification, filed #876 for this.

PR #875, opened 2020-07-10
Support using ClusterIP for the API Server service

coltonstapper:
Use ClusterIP in the kubeconfig if it exists
Allow ClusterIP service type in vcctl
Create the apiserver-svc ahead of time to add ClusterIP to PKI.

### DIFF
--- a/incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
+++ b/incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
@@ -76,26 +76,40 @@ func (mpn *MasterProvisionerNative) CreateVirtualCluster(vc *tenancyv1alpha1.Vir
 	if err != nil {
 		return err
 	}
+	isClusterIP := cv.Spec.APIServer.Service != nil && cv.Spec.APIServer.Service.Spec.Type == v1.ServiceTypeClusterIP
+	// if ClusterIP, have to create API Server ahead of time to lay it down in the PKI
+	if isClusterIP {
+		log.Info("deploying ClusterIP Service for API component", "component", cv.Spec.APIServer.Name)
+		complementAPIServerTemplate(conversion.ToClusterKey(vc), cv.Spec.APIServer)
+		err = mpn.Create(context.TODO(), cv.Spec.APIServer.Service)
+		if err != nil {
+			if !apierrors.IsAlreadyExists(err) {
+				return err
+			}
+			log.Info("service already exist",
+				"service", cv.Spec.APIServer.Service.GetName())
+		}
+	}
 
-	// 3. create PKI
-	err = mpn.createPKI(vc, cv)
+	// 2. create PKI
+	err = mpn.createPKI(vc, cv, isClusterIP)
 	if err != nil {
 		return err
 	}
 
-	// 4. deploy etcd
+	// 3. deploy etcd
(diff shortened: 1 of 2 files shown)


incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
```diff
@@ -272,25 +285,42 @@ func (mpn *MasterProvisionerNative) createPKI(vc *tenancyv1alpha1.VirtualCluster
...
+		var err error
+		clusterIP, err = kubeutil.GetSvcClusterIP(mpn, conversion.ToClusterKey(vc), cv.Spec.APIServer.Service.GetName())
+		if err != nil {
+			log.Info("Warning: failed to get API Service", "service", cv.Spec.APIServer.Service.GetName(), "err", err)
+		}
+	}
+
 	apiserverDomain := cv.GetAPIServerDomain(ns)
-	apiserverCAPair, err := vcpki.NewAPIServerCrtAndKey(rootCAPair, vc, apiserverDomain)
+	apiserverCAPair, err := vcpki.NewAPIServerCrtAndKey(rootCAPair, vc, apiserverDomain, clusterIP)
```
coltonstapper, line 299: (the PKI function makes sure not to add empty strings)

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
```diff
@@ -272,25 +285,42 @@ func (mpn *MasterProvisionerNative) createPKI(vc *tenancyv1alpha1.VirtualCluster
 	}
 	caGroup.ETCD = etcdCAPair
 
+	clusterIP := ""
+	isClusterIP := cv.Spec.APIServer.Service != nil && cv.Spec.APIServer.Service.Spec.Type == v1.ServiceTypeClusterIP
```
Fei-Guo, line 289: I suggest to move isClusterIP definition before line 80 so that the if statement in line 80 can be `if isClusterIP`.

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
```diff
@@ -272,25 +285,42 @@ func (mpn *MasterProvisionerNative) createPKI(vc *tenancyv1alpha1.VirtualCluster
...
 	}
 	caGroup.APIServer = apiserverCAPair
 
+	var ip string
+	if clusterIP != "" {
+		ip = clusterIP
+	} else {
+		ip = apiserverDomain
+	}
+
```
charleszheng44, line 311:
```golang
ip := apiserverDomain
if clusterIP != "" {
    ip = clusterIP
}
```

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
coltonstapper: good catch, fixing

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
coltonstapper: Thanks :)

coltonstapper:
Thanks @Fei-Guo @charleszheng44, addressed your comments

PR #877, opened 2020-07-10
Add nodes/proxy to vn-agent RBAC

coltonstapper:
Without adding nodes/proxy, vn-agent cannot get/create proxy requets to the kubelet:
Error from server (Forbidden): Forbidden (user=vn-agent, verb=get, resource=nodes, subresource=proxy) ( pods/log net-tools)

### DIFF
--- a/incubator/virtualcluster/config/setup/all_in_one.yaml
+++ b/incubator/virtualcluster/config/setup/all_in_one.yaml
@@ -350,6 +350,7 @@ rules:
   - pods/log
   - pods/exec
   - pods/portforward
+  - nodes/proxy
   verbs:
   - get
   - list


charleszheng44:
Hi @coltonstapper,
what does "vn-agent cannot get/create proxy requets to the kubelet" mean? Could you give an example?

christopherhein:
With the all_in_one_setup.yaml, when trying to use requests like `kubectl exec` or anything else that needs to proxy with a 1.17 cluster it needs to have node/proxy permissions to allow the requests to get to the kubelet through the van-agent.

charleszheng44:
> With the all_in_one_setup.yaml, when trying to use requests like `kubectl exec` or anything else that needs to proxy with a 1.17 cluster it needs to have node/proxy permissions to allow the requests to get to the kubelet through the van-agent.

Thanks for explanation

/lgtm

PR #884, opened 2020-07-14
Warn the testers if HNC cannot be repaired

### LINKED ISSUE #881: Recover HNC in e2e test
In some of the e2e tests, for example in incubator/hnc/e2e/tests/delete_crd_test.go, we are breaking HNC. Currently how we repair HNC is by "mustRun("kubectl apply -f", hncRecoverPath)", but mustRun will only stop the current test from running if it fails, not the current test suite.

GinnyJI:
If HNC is broken and cannot be repaired, log loud warnings to the user that any test after this could fail because of broken hnc.

Tested: 'make e2e-test' with invalid HNC_REPAIR path

Fix #881


adrianludwin:
/cc @rjbez17 

Ryan, I'm assuming you're not too interested in all these e2e tests but lmk if you want to be more in the loop.

/lgtm
/approve

Issue #885, opened 2020-07-14
HNC: CRD conversion for HNCConfig SynchronizationMode


yiqigao217:
Part of #868 

In v1alpha2, we will change the modes from `propagate`, `remove`, `ignore` to `Propagate`, `Remove`, `Ignore` (camel case). A straightforward conversion is to convert the valid modes to the other version and keep invalid values as is. 

However, there are two scenarios that the simple mapping doesn't work when:
* A type has `Propagate` mode in v1alpha1 HNCConfig. Since the value is not recognized, the mode in status will be `ignore`. However, if you get the resource in v1alpha2, it will look confusing with `Propagate` in spec and `Ignore` in status.
* A type has `propagate` mode in v1alhpa2 HNCConfig, which shouldn't be recognized and should be treated as `Ignore` in status but now converted to a valid mode in v1alpha1, so the status for it is `propagate` (since v1alpha1 is the storage version). 

A simple solution could be that for the valid modes, no matter in lower case or camel case, they will all be converted to the other version. Then the above scenarios won't happen.

adrianludwin:
Is it possible to have `propagate` in v1a2 if we set up the schemas correctly? I'm not sure if this is worth worrying about _too_ much.

I think it's fine to say that if our storage version is v1a1 and it contains the text `Propagate` (which is invalid in v1a1), when we convert that to v1a2 we could display it as something like `InvalidConvertedValuePropagate`.

yiqigao217:
Update: We will not convert any unknown modes since we introduced the enum schema validation first #902 .

adrianludwin:
@yiqigao217 is this complete?

yiqigao217:
Fixed by #1026

PR #886, opened 2020-07-14
Add conversion webhook to HNCConfig for sync mode

yiqigao217:
This is the initial commit to introduce CRD conversion webhook in HNC.
The conversion of sync mode in HNCConfig is minor so it's a good
starting point. v1alpha2 is now set as a served version in HNCConfig.

Set v1alpha1 as the conversion hub and v1alpha2 as a spoke. Add
ConvertTo() and ConvertFrom() funcs with conversion details.

Add conversion strategy and webhookClientConfig in a kustomize patch.
Update the internal cert reconciler to inject cert to the conversion
webhook too. (The cer-manager option is not available for this current
conversion webhook and the caninjectcert patch is not included either)

Tested by applying sample v1alpha1 and v1alpha2 configs with one invalid
mode for "Secret" type. Got expected mode "Ignore" for v1alpha2 and
"ignore" for v1alpha1 in the status. Fixing it by either applying a
correct v1alpha1 config or v1alpha2 config worked.

Part of #885 , #868


incubator/hnc/cmd/manager/main.go
```diff
@@ -36,6 +36,7 @@ import (
 	// +kubebuilder:scaffold:imports
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha1"
+	api2 "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
```
adrianludwin, line 39: Maybe call this `v1a2` and rename the other one to `v1a1` to keep them parallel? (Here and elsewhere)

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import ctrl "sigs.k8s.io/controller-runtime"
+
+// Hub marks this type as a conversion hub.
```
adrianludwin, line 20: Can you add a link to the documentation of the term "conversion hub"?

incubator/hnc/api/v1alpha2/hnc_config.go
```diff
@@ -95,7 +95,6 @@ type CodeAndAffectedNamespaces struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:path=hncconfigurations,scope=Cluster
-// +kubebuilder:skipversion
```
adrianludwin, line 98: Does this need to be removed from v1a1 as well?

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,190 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+
+	"sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha1"
```
adrianludwin, line 21: Suggestion: import as v1a1 to parallel what this will look like in other places (consistency can be important IMO)

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+// ConvertTo converts this hncconfig to the Hub version (v1alpha1).
+func (src *HNCConfiguration) ConvertTo(dstRaw conversion.Hub) error {
+	dst := dstRaw.(*v1alpha1.HNCConfiguration)
+
+	// Spec
+	sspts := src.Spec.Types
```
adrianludwin, line 29: sspts = "Source Spec Propagation Types"?

Maybe srcTypes - just a tiny bit longer but quite a bit more clear? The loop variable can just be `st` instead.

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,190 @@
...
+		// The following are the invalid modes in v1alpha2, so they should remain
+		// invalid in v1alpha1. See https://github.com/kubernetes-sigs/multi-tenancy/issues/885
+		case SynchronizationMode(v1alpha1.Propagate):
+			dspt.Mode = v1alpha1.SynchronizationMode(Propagate)
+		case SynchronizationMode(v1alpha1.Ignore):
+			dspt.Mode = v1alpha1.SynchronizationMode(Ignore)
+		case SynchronizationMode(v1alpha1.Remove):
+			dspt.Mode = v1alpha1.SynchronizationMode(Remove)
+		default:
+			dspt.Mode = v1alpha1.SynchronizationMode(sspt.Mode)
```
adrianludwin, line 51: This seems a bit odd to me? If someone writes `propagate` in v1a2, it will turn into `Propagate` in v1a1, correct?

I feel like I'd rather see something like `UnknownV1Alpha2Value-propagate` which should make it a lot more clear what's going on?

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,190 @@
...
+		dspt := v1alpha1.TypeSynchronizationSpec{}
+		dspt.APIVersion = sspt.APIVersion
+		dspt.Kind = sspt.Kind
+		switch sspt.Mode {
+		case Propagate:
+			dspt.Mode = v1alpha1.Propagate
+		case Ignore:
+			dspt.Mode = v1alpha1.Ignore
+		case Remove:
+			dspt.Mode = v1alpha1.Remove
```
adrianludwin, line 41: This might be more clearly expressed as a map, and lets you skip the second switch statements.

```
modeMap := map[SynchronizationMode]v1a1.SynchronizationMode{
  Propagate: v1a1.Propagate,
  Ignore: v1a1.Ignore,
  Remove: v1a1.Remove,
}
...
mapped, ok := modeMap[sspt.Mode]
if !ok {
  mapped = v1a1.SynchronizationMode(unknownModePrefix+string(sspt.Mode))
}
```

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,190 @@
...
+	dstcs := []v1alpha1.HNCConfigurationCondition{}
+	for _, sstc := range sstcs {
+		dstc := v1alpha1.HNCConfigurationCondition{}
+		dstc.Code = v1alpha1.HNCConfigurationCode(sstc.Code)
+		dstc.Msg = sstc.Msg
+		dstcs = append(dstcs, dstc)
+	}
+	dst.Status.Conditions = dstcs
+
+	// Status.NamespaceConditions
```
adrianludwin, line 91: Are we keeping NamespaceConditions? I have a funny memory that we weren't?

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -54,23 +56,24 @@ func init() {
...
+	reconciler := &ReconcileWH{
 		client:    mgr.GetClient(),
 		scheme:    mgr.GetScheme(),
 		ctx:       context.Background(),
 		secretKey: cr.SecretKey,
 		vwhKey:    vwhKey,
+		crdNames:  crdNames,
 	}
-	if err := addController(mgr, reconciler, cr.SecretKey, vwhKey); err != nil {
+	if err := addController(mgr, reconciler, cr.SecretKey, vwhKey, crdNames); err != nil {
```
adrianludwin, line 76: Suggestion: change `addController` to accept the `ReconcileWH` type as its second parameter, instead of the `reconcile.Reconciler` interface. Using an interface doesn't buy us anything (the type is local; we don't mock it in the unit tests) and it would let us avoid passing the other three parameters since they're all in `ReconcileWH` anyway.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -503,20 +538,50 @@ func (r *ReconcileVWH) Reconcile(request reconcile.Request) (reconcile.Result, e
 		return reconcile.Result{Requeue: true}, err
 	}
 
+	crds := []*unstructured.Unstructured{}
```
adrianludwin, line 541: What do we gain from using `Unstructured` here as opposed to the real CRD type? We know exactly what type it is and we manipulate its fields directly. (I have the same question for VWH but that's out of scope for this change)

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -503,20 +538,50 @@ func (r *ReconcileVWH) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+	crd := &unstructured.Unstructured{}
+	crd.SetGroupVersionKind(crdGVK)
```
adrianludwin, line 543: I think this will fail as soon as there's more than one CRD type - on line 555, you add this _pointer_ to an array, but you keep reading into the same pointer. So if there are three CRDs, I _think_ that `crds` will land up with three copies of whatever the last CRD was.

As a general rule, I like to define my variables at the lowest scope possible, which in this case is inside the loop, not outside.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -503,20 +538,50 @@ func (r *ReconcileVWH) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+				// For additional cleanup logic use finalizers.
+				return reconcile.Result{}, nil
+			}
+			// Error reading the object - requeue the request.
+			return reconcile.Result{Requeue: true}, err
+		}
+		crds = append(crds, crd)
+	}
+
 	if secret.GetDeletionTimestamp().IsZero() {
```
adrianludwin, line 558: There's no "else" here, so maybe turn this into an early exit, and then collect all the new CRD stuff (loading the existing CRDs, modifying them and writing them back) into a new function that gets called after line 571?

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
yiqigao217: Agreed, will update.

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
yiqigao217: We didn't have a conclusion on whether to keep or delete namespace conditions in HNCConfig, but only that if we have this condition, make sure we keep it rate-limited and length-limited. But someone did mention that users are more likely to find namespace conditions from namespaces (not somewhere else like in this config resource) to fix them.

So it's more like our call if we want to keep it or not. The purpose is to have all the conditions in one place for user to look. It's generated from forest, right? Do you think we can remove it since we already have metrics showing the same data now? (The metrics are generated from forest not from here, right?)

adrianludwin: I forget where the metrics come from but they don't identify _which_ namespaces have the problem. Hmm. If an admin has 1,000 namespaces, one of which has a condition, how would they find it if we don't have this?

adrianludwin: But you can ignore that question for the purpose of this PR :)

yiqigao217: sg, I will have it in its own PR if we decide a change there.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -54,23 +56,24 @@ func init() {
 	flag.BoolVar(&restartOnSecretRefresh, "cert-restart-on-secret-refresh", false, "Kills the process when secrets are refreshed so that the pod can be restarted (secrets take up to 60s to be updated by running pods)")
 }
 
-// AddRotator adds the CertRotator and ReconcileVWH to the manager.
-func AddRotator(mgr manager.Manager, cr *CertRotator, vwhName string) error {
+// AddRotator adds the CertRotator and ReconcileWH to the manager.
+func AddRotator(mgr manager.Manager, cr *CertRotator, vwhName string, crdNames []string) error {
```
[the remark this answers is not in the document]
yiqigao217: Seems it's better to add the vwh and crds to the `ReconcileWH`. I made the `VWHKey` and `CRDNames` into public fields so that here we can pass a `ReconcileWH` object with these two values set.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
yiqigao217: Maybe it's easier to make changes to the version that we just need to update it in one place in the global variable `vwhGVK` instead of updating all of them in the code?

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
yiqigao217: I tested this pattern in the go playground with a fake struct to make sure it worked, but anyway I cannot remember now why I preferred putting this block outside the loop then. Agreed that it's better to place it inside the loop.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
adrianludwin: I'm not sure I follow? I'd suggest trying to use the real types for your new code and seeing if you like it. We can worry about the existing code later.

incubator/hnc/api/v1alpha2/hnc_config.go
yiqigao217: This is not added in any v1a1. The other two places that it still remains are v1a2/ subnamespaceanchor.go and hierarchy_type.go. For these two other CRDs, we will still skip the v1a2 before there's any conversion webhook is there.

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
yiqigao217: sspts = "src spec types". Updated to `srcSpecTypes`.

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
yiqigao217: Good suggestion updated!

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
yiqigao217: Good sugguestion, updated!

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
yiqigao217: Further split the logic in `Reconcile()` into 3 parts:
1. Get secret and generate certs
2. Ensure certs on validatingwebhookconfiguration
3. Ensure certs on CRDs

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
yiqigao217: Oh even in the new code, I refactored the cert injection to take the current `injectCert` functions as a parameter, and update the unstructed object accordingly (either validatingwebhookconfiguration or CRD).

incubator/hnc/internal/validators/setup.go
```diff
@@ -42,7 +50,7 @@ func CreateCertsIfNeeded(mgr ctrl.Manager, novalidation, internalCert bool) (cha
 		CaOrganization: caOrganization,
 		DNSName:        dnsName,
 		CertsMounted:   setupFinished,
-	}, vwhName)
+	}, rwh)
```
adrianludwin, line 53: This feels very strange to me. We're already dealing with one object - `CertRotator` - so why are we now dealing with _another_ object (`ReconcileWH`) that's really just an implementation detail of the rotator? I'd far rather only deal with one object here, and I think that should be the rotator, not the reconciler.

I'd go so far as to say that `ReconcileWH` should be private so it's not even accessible here.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -54,23 +56,23 @@ func init() {
...
-	vwhKey := types.NamespacedName{Name: vwhName}
-	reconciler := &ReconcileVWH{
+	reconciler := &ReconcileWH{
 		client:    mgr.GetClient(),
 		scheme:    mgr.GetScheme(),
 		ctx:       context.Background(),
 		secretKey: cr.SecretKey,
-		vwhKey:    vwhKey,
+		VWHKey:    rwh.VWHKey,
+		CRDNames:  rwh.CRDNames,
```
adrianludwin, line 73: I definitely find it strange that this function accepts one object of type `ReconcileWH` and then creates another completely separate object of the same type to use "for real." I think this is good evidence that `ReconcileWH` should be private and not used outside this package.

Maybe `reconcileWH` could have a pointer to the `CertRotator` so it doesn't have to copy all its fields?

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -435,91 +452,169 @@ func (m *mapper) Map(object handler.MapObject) []reconcile.Request {
...
-// object and makes sure the most recent CA cert is included
-func (r *ReconcileVWH) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+// Reconcile reads the state of the cluster for webhooks in validatingwebhookconfiguration
+// and CRD to make sure the most recent CA cert is included
+func (r *ReconcileWH) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	if request.NamespacedName != r.secretKey {
 		return reconcile.Result{}, nil
 	}
+
+	// Get secret and generate certs.
```
adrianludwin, line 522: This is a much bigger change than it used to be, right? It would be better if you could minimize the size of the change so it's very clear what's being added, since that will make it possible to upstream this code.

I think what I originally had in mind was that just underneath the call to `injectCertToWebhook` and updating the VWH, you'd just add a call to a function called something like `handlCRDWebhooks` and then _everything_ would go into that one function. That makes it much, much easier for an outside to review this diff in a few weeks.

adrianludwin:
Perhaps it makes sense to split this PR into two? One for the cert rotator changes, another for everything else?

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+const (
+	unknownv1a1Prefix = "UnknownV1Alpha1Value-"
+	unknownv1a2Prefix = "UnknownV1Alpha2Value-"
+)
+
+var (
+	// Maps to and from the Hub version (v1alpha1) for all valid values.
+	toMap = map[SynchronizationMode]v1a1.SynchronizationMode{
```
adrianludwin, line 33: Nit: the "to" and "from" here are ambiguous, maybe something like `toV1A1` and `toV1A2` would be more clear?

Including the name of the type ("map") is sometimes a code smell - describe what the variable is _for_, not how it's implemented.

yiqigao217:
> Perhaps it makes sense to split this PR into two? One for the cert rotator changes, another for everything else?

Makes sense, I will add another PR for cert changes and hold this until that one is merged.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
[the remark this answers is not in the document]
adrianludwin: Sorry I didn't really understand what this comment meant before. I don't think this is the best way to go for the reasons I explained in other comments.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
adrianludwin: Ok why don't you leave this as-is.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
adrianludwin: As I mentioned in my other comments, we should have as little refactoring as possible since that makes it very difficult to upstream this code. My suggestion (collect all the new stuff into one function and call it on line 571) would make it easier to merge these changes in the future.

yiqigao217: Got you yes, that makes sense. Do you think I can make minimum change in certs.go and still have it with this PR?

adrianludwin: Hmm, maybe try it out and see?

rjbez17:
/hold 

until split prs

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
yiqigao217: sgtm. updated!

adrianludwin:
> hold until split prs

@rjbez17 we're thinking the split PRs may no longer be necessary as this PR seems to be converging. The main motivation to split was to make it easier to review.

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,191 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/conversion"
+
+	v1a1 "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha1"
+)
+
+const (
+	unknownv1a1Prefix = "UnknownV1Alpha1Value-"
+	unknownv1a2Prefix = "UnknownV1Alpha2Value-"
+)
+
```
adrianludwin, line 30: Based on our conversations with Jordan (offline) to use a schema to limit the allowed values in v1a2, this is no longer necessary, right?

* We don't need the unknownv1a1 because the schema in v1a2 will mean that we _have_ to call it "Ignore."
  * Perhaps we can add a condition in v1a2 that there was a bad value in v1a1? That's not needed in this PR though, we can file an issue to follow up.
* We don't need the unknownv1a2 because the schema will prevent that from ever happening

Is that correct?

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,191 @@
...
+			dtm = v1a1.SynchronizationMode(unknownv1a2Prefix + st.Mode)
+		}
+		dt.Mode = dtm
+		dstStatusTypes = append(dstStatusTypes, dt)
+	}
+	dst.Status.Types = dstStatusTypes
+
+	// Status.Conditions
+	srcConditions := src.Status.Conditions
+	distConditions := []v1a1.HNCConfigurationCondition{}
```
adrianludwin, line 90: Maybe add a TODO here to switch to real conditions in v1a2 in a followup PR

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
```diff
@@ -0,0 +1,191 @@
...
+			dtm = SynchronizationMode(unknownv1a1Prefix + st.Mode)
+		}
+		dt.Mode = dtm
+		dstStatusTypes = append(dstStatusTypes, dt)
+	}
+	dst.Status.Types = dstStatusTypes
+
+	// Status.Conditions
+	srcConditions := src.Status.Conditions
+	dstConditions := []HNCConfigurationCondition{}
```
adrianludwin, line 167: TODO: real conditions

incubator/hnc/api/v1alpha2/hnc_config.go
```diff
@@ -32,15 +32,15 @@ type SynchronizationMode string
 
 const (
 	// Propagate objects from ancestors to descendants and deletes obsolete descendants.
-	Propagate SynchronizationMode = "propagate"
+	Propagate SynchronizationMode = "Propagate"
```
adrianludwin, line 35: Per offline discussion with Jordan, use kubebuilder:validation:Enum?

https://book.kubebuilder.io/reference/markers/crd-validation.html

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -517,6 +554,42 @@ func (r *ReconcileVWH) Reconcile(request reconcile.Request) (reconcile.Result, e
...
+		}
+
+		if !crd.GetDeletionTimestamp().IsZero() {
+			return reconcile.Result{}, nil
+		}
+
+		crLog.Info("ensuring CA cert on CRD conversion webhook")
+		if err := injectCertToConversionWebhook(crd, certPem); err != nil {
+			crLog.Error(err, "unable to inject cert to CRD conversion webhook")
+			return reconcile.Result{}, err
```
yiqigao217, line 587: @adrianludwin since here we don't want to enqueue the request again if there is an error in the injection (e.g. cannot find the `caBundle` field), I still keep this `ensureCRDCerts` func to return both result and error.

adrianludwin: I'm not sure I understand? Returning a non-nil error guarantees that we'll re-enqueue the request.

adrianludwin: And I think that's fine. Maybe the `caBundle` field was missing due to an operator error but it gets added later, in which case this is self-repaired (good). Otherwise, we continue trying this every few seconds and putting the error in the log each time, which means that when someone looks at the log, the error message will always be fresh (also good). So re-enqueuing is usually a very good idea :)

incubator/hnc/api/v1alpha2/hnc_config_conversion.go
yiqigao217: Yes, but I need to test if `unknownv1a1` prefix can still be taken with enum schema validation now.

yiqigao217: Removed since we use one-way conversion and the enum schema validation is added in #902

yiqigao217:
Closing this because as updated in #868 , we will make a no-op CRD conversion first (see #953 ) and then implement v1a2.

PR #888, opened 2020-07-14
Add delete-hc-crd test to e2e test suite

GinnyJI:
Translate test-delete-hc-crd.sh to delete_crd_test.go. Move hnc path
check to BeforeEach() to avoid repeated checks in each It block.

Tested: 'make e2e-test' with and without HNC_REPAIR


adrianludwin:
/approve
/lgtm

Nice work!

PR #891, opened 2020-07-15
Upgrade to controller-runtime v0.6.1

adrianludwin:
In preparation for importing Gatekeeper's cert-controller, upgrade to
the latest version of controller-runtime, as well as update all K8s
libraries from 0.17.x to 0.18.x.

Tested: ran e2e tests on GKE clusters with versions v1.14 and 1.16 (for
1.14, I patched in #890 first). Everything looks fine.


adrianludwin:
/assign @rjbez17 
/assign @yiqigao217 
cc @GinnyJI

incubator/hnc/internal/kubectl/root.go
```diff
@@ -114,14 +116,17 @@ func homeDir() string {
 }
 
 func (cl *realClient) getHierarchy(nnm string) *api.HierarchyConfiguration {
-	if _, err := k8sClient.CoreV1().Namespaces().Get(nnm, metav1.GetOptions{}); err != nil {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
yiqigao217, line 119: Question: what does this 5s timeout do?

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -20,23 +20,18 @@ spec:
       description: Hierarchy is the Schema for the hierarchies API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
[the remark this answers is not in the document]
adrianludwin: As far as I can tell, yes. I didn't make them manually, and TBH I don't much like them either (they make it impossible to review in PRs) but not much I can do about that!

incubator/hnc/internal/kubectl/root.go
adrianludwin: It means that if the total length of that function (eg all calls to the apiserver within `getHierarchy`) takes longer than 5s, the call will fail. Does that make sense?

yiqigao217: I see thanks for the explanation!

adrianludwin:
I don't think @rjbez17 will care too much about this one :)

/hold cancel

PR #894, opened 2020-07-15
Add delete-hc-and-anchor-crd test

GinnyJI:
Convert test-delete-hc-and-anchor-crd-test.sh to delete_crd_test.go. Add
helper function runCommand() to return the stdout and err from running
the input command.

Tested: make e2e-test

### DIFF
--- a/incubator/hnc/e2e/tests/delete_crd_test.go
+++ b/incubator/hnc/e2e/tests/delete_crd_test.go
@@ -7,7 +7,7 @@ import (
 	. "github.com/onsi/ginkgo"
 )
 
-var _ = Describe("Delete-anchor-crd", func() {
+var _ = Describe("When deleting CRDs", func() {
 
 	hncRecoverPath := os.Getenv("HNC_REPAIR")
 
@@ -33,9 +33,11 @@ var _ = Describe("Delete-anchor-crd", func() {
 			GinkgoT().Log("------------------------------------------------------------------")
 			GinkgoT().FailNow()
 		}
+		// give HNC enough time to repair
+		time.Sleep(5 * time.Second)
 	})
 
-	It("should create parent and deletable child, and delete the CRD", func() {
+	It("should not delete subnamespaces", func() {
 		// set up
 		mustRun("kubectl create ns", nsParent)
 		mustRun("kubectl hns create", nsChild, "-n", nsParent)
@@ -61,5 +63,19 @@ var _ = Describe("Delete-anchor-crd", func() {
 		time.Sleep(5 * time.Second)
 		// verify
 		mustRun("kubectl get -oyaml rolebinding foo -n", nsChild)
+	})
+
+	It("should fully delete all CRDs", func() {
+		// set up
+		mustRun("kubectl create ns", nsParent)
+		mustRun("kubectl hns create", nsChild, "-n", nsParent)
+		// test
+		mustRun("kubectl delete crd hierarchyconfigurations.hnc.x-k8s.io")
+		time.Sleep(1 * time.Second)
+		mustRun("kubectl delete crd subnamespaceanchors.hnc.x-k8s.io")
+		mustRun("kubectl delete crd hncconfigurations.hnc.x-k8s.io")
+		// Give HNC 10s to have the chance to fully delete everything (5s wasn't enough).
+		// Verify that the HNC CRDs are gone (if nothing's printed, then they are).
+		runShouldNotContain("hnc", "10s", "kubectl get crd")
 	})
 })
--- a/incubator/hnc/e2e/tests/e2e_test.go
+++ b/incubator/hnc/e2e/tests/e2e_test.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"errors"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Delete-anchor-crd", func() {
```
adrianludwin, line 11: I think `Delete-anchor-crd` isn't really a good description anymore since we're not _only_ deleting the anchor CRD.

How about `"When deleting CRDs"`, and then the `It()` blocks can have names like `"should not delete subnamespaces"`. This means that when it runs, it will print out something close to "When deleting CRDs, it should not delete subnamespaces" which is exactly what we're testing.

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -62,4 +63,21 @@ var _ = Describe("Delete-anchor-crd", func() {
 		// verify
 		mustRun("kubectl get -oyaml rolebinding foo -n", nsChild)
 	})
+
+	It("should delete the HC CRD, anchor CRD, and verify that the HNC CRDs are gone", func() {
```
adrianludwin, line 67: I think this description should be `should fully delete all CRDs`

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -62,4 +63,21 @@ var _ = Describe("Delete-anchor-crd", func() {
...
+		mustRun("kubectl delete crd hierarchyconfigurations.hnc.x-k8s.io &")
+		time.Sleep(1 * time.Second)
+		mustRun("kubectl delete crd subnamespaceanchors.hnc.x-k8s.io &")
+		mustRun("kubectl delete crd hncconfigurations.hnc.x-k8s.io &")
+		// Sleeping for 10s to give HNC the chance to fully delete everything (5s wasn't enough).
+		time.Sleep(10 * time.Second)
+		// Verify that the HNC CRDs are gone (if nothing's printed, then they are).
+		stdout, err := runCommand("kubectl get crd | grep hnc")
+		Expect(err).Should(BeNil())
+		Expect(stdout).Should(BeNil())
```
adrianludwin, line 81: Here's a good chance to use `Eventually`:

```
Eventually(func() error{
  stdout, err := runCommand("kubectl get crd | grep hnc")
  if err != nil {
    return err
  }
  if stdout != nil {
    return error.New(string(stdout))
  }
}).Should(Succeed())
```

lmk if you need any clarity on how that works.

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: done

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: done

incubator/hnc/e2e/tests/delete_crd_test.go
GinnyJI: Done

incubator/hnc/e2e/tests/delete_crd_test.go
```diff
@@ -62,4 +64,18 @@ var _ = Describe("Delete-anchor-crd", func() {
...
 		mustRun("kubectl get -oyaml rolebinding foo -n", nsChild)
 	})
+
+	It("should fully delete all CRDs", func() {
+		// set up
+		mustRun("kubectl create ns", nsParent)
+		mustRun("kubectl hns create", nsChild, "-n", nsParent)
+		// test
+		mustRun("kubectl delete crd hierarchyconfigurations.hnc.x-k8s.io")
+		time.Sleep(1 * time.Second)
```
adrianludwin, line 74: What's this one for?

GinnyJI: It's from the shell script, the original comment is "Delete the HC CRD, then wait 1s, then delete the anchor CRD." My guess is that `delete crd hierarchyconfigurations.hnc.x-k8s.io` takes some time

adrianludwin: Maybe try removing it in a followup CL. Also if you see stuff like this that you don't understand, it's good to comment it or see if you can delete it.

incubator/hnc/e2e/tests/e2e_test.go
```diff
@@ -33,15 +34,33 @@ func mustNotRun(cmdln ...string) {
 }
 
 func tryRun(cmdln ...string) error {
+	stdout, err := runCommand(cmdln...)
+	GinkgoT().Log("Output: ", stdout)
+	return err
+}
+
+func runShouldNotContain(substr string, timeDuration string, cmdln ...string) {
```
yiqigao217, line 42: Nice function with `timeDuration`!

PR #898, opened 2020-07-16
Add e2e test for issue 328

GinnyJI:
Add issue_test.go that test github issue #328. Add helper function
runShouldContain() to run the command and check if the output contains
the give slice of strings.

Tested: make e2e-test

### DIFF
--- a/incubator/hnc/e2e/tests/e2e_test.go
+++ b/incubator/hnc/e2e/tests/e2e_test.go
@@ -39,7 +39,34 @@ func tryRun(cmdln ...string) error {
 	return err
 }
 
-func runShouldNotContain(substr string, timeDuration string, cmdln ...string) {
+func runShouldContain(substr string, seconds float64, cmdln ...string) {
+	runShouldContainMultiple([]string{substr}, seconds, cmdln...)
+}
+
+func runShouldContainMultiple(substrs []string, seconds float64, cmdln ...string) {
+	Eventually(func() error {
+		stdout, err := runCommand(cmdln...)
+		if err != nil {
+			return err
+		}
+
+		allContained := true
+		for _, substr := range substrs {
+			if strings.Contains(stdout, substr) == false {
+				allContained = false
+				break
+			}
+		}
+
+		if allContained == false {
+			return errors.New("Expecting: " + strings.Join(substrs, ", ") + " but get: " + stdout)
+		}
+
+		return nil
+	}, seconds).Should(Succeed())
+}
+
+func runShouldNotContain(substr string, seconds float64, cmdln ...string) {
 	Eventually(func() error {
 		stdout, err := runCommand(cmdln...)
 		if err != nil {
@@ -49,7 +76,7 @@ func runShouldNotContain(substr string, timeDuration string, cmdln ...string) {
 			return errors.New("Not expecting: " + substr + " but get: " + stdout)
 		}
 		return nil
-	}, timeDuration).Should(Succeed())
+	}, seconds).Should(Succeed())
 }
 
 func runCommand(cmdln ...string) (string, error) {
--- a/incubator/hnc/e2e/tests/issues_test.go
+++ b/incubator/hnc/e2e/tests/issues_test.go
@@ -1,0 +1,39 @@
+package test
+
+import (
+	. "github.com/onsi/ginkgo"
+)
+
+var _ = Describe("Issues", func() {
+
+	const (
+		nsParent = "issues-parent"
+		nsChild = "issues-child"
+	)
+
+	BeforeEach(func() {
+		cleanupNamespaces(nsParent, nsChild)
+	})
+
+	AfterEach(func() {
+		cleanupNamespaces(nsParent, nsChild)
+	}) 
+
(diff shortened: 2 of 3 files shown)


incubator/hnc/e2e/tests/e2e_test.go
```diff
@@ -39,6 +39,30 @@ func tryRun(cmdln ...string) error {
 	return err
 }
 
+func runShouldContain(substrs []string, timeDuration string, cmdln ...string) {
```
adrianludwin, line 42: Might be less ambiguous for `timeDuration string` to be `seconds float64` (see http://onsi.github.io/gomega/#eventually).

adrianludwin: Also, why do we need multiple substrings? That adds a lot of visual complexity to the caller (`[]string{X}` where you really only needed `X`) in the vast majority of cases. Maybe just accept a single string, and if we _also_ need a "multi" version, add a `runShouldContainMultiple` function?

incubator/hnc/e2e/tests/issues_test.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+var _ = Describe("Issues", func() {
+
+	const (
+		nsParent = "issues-parent"
+		nsChild = "issues-child"
+	)
+
+	AfterEach(func() {
```
adrianludwin, line 16: _Also_ do this in `BeforeEach` in case the last run was interrupted? E.g. I see a failure, I hit "ctrl-c" but the next test is already running and has created namespaces. But because I hit ctrl-c, `AfterEach` never gets run, and the test fails the next time I run it.

This is a general comment for all e2e tests.

incubator/hnc/e2e/tests/issues_test.go
```diff
@@ -0,0 +1,39 @@
...
+		// Setting up hierarchy with rolebinding that HNC doesn't have permission to copy.
+		mustRun("kubectl create ns", nsParent)
+		mustRun("kubectl create ns", nsChild)
+		mustRun("kubectl hns set", nsChild, "--parent", nsParent)
+		// cluster-admin is the highest-powered ClusterRole and HNC is missing some of
+		// its permissions, so it cannot propagate it.
+		mustRun("kubectl create rolebinding cluster-admin-rb -n", nsParent, 
+			"--clusterrole='cluster-admin' --serviceaccount=issue-parent:default")
+		// wait 2s
+		time.Sleep(2 * time.Second)
```
adrianludwin, line 30: This is no longer needed given that `runShouldContain` has an `Eventually` clause (which the bash script didn't). You should _never_ need an explicit sleep unless you're waiting to prove that something _hasn't_ happened.

incubator/hnc/e2e/tests/issues_test.go
GinnyJI: Done. will put up another PR to apply this to other tests

incubator/hnc/e2e/tests/issues_test.go
GinnyJI: done

incubator/hnc/e2e/tests/e2e_test.go
GinnyJI: done

GinnyJI: we need the multiple version for 
<pre><code>
// Tree should show CannotPropagateObject in nsParent and CannotUpdateObject in nsChild
runShouldContainMultiple([]string{"1) CannotPropagateObject", "2) CannotUpdateObject"}, 1, "kubectl hns tree", nsParent)
</code></pre>

incubator/hnc/e2e/tests/e2e_test.go
```diff
@@ -39,7 +39,44 @@ func tryRun(cmdln ...string) error {
 	return err
 }
 
-func runShouldNotContain(substr string, timeDuration string, cmdln ...string) {
+func runShouldContain(substr string, seconds float64, cmdln ...string) {
```
adrianludwin, line 42: No copy-paste :D There's a much shorter way to implement `runShouldContain` now that `runShouldContainMultiple` exists!

incubator/hnc/e2e/tests/issues_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	BeforeEach(func() {
+		cleanupNamespaces(nsParent, nsChild)
+	})
+
+	AfterEach(func() {
+		cleanupNamespaces(nsParent, nsChild)
+	}) 
+
+	It("Should test issue #328", func() {
```
yiqigao217, line 22: Probably better to change the it block here to say "it should remove obsolete conditions - see issue #328"?

incubator/hnc/e2e/tests/e2e_test.go
GinnyJI: ah good catch. Done

incubator/hnc/e2e/tests/issues_test.go
GinnyJI: done

PR #902, opened 2020-07-16
Use enum schema validation for HNCConfig spec modes

### LINKED ISSUE #899: HNC: Use enum schema validation for modes in v1alpha1 HNCConfig spec
Part of #868 

Currently in v1alpha1, the validation for propagation modes in HNCConfig spec is done by validating webhooks. The webhook doesn't allow any values other than `propagate`, `ignore` and `remove`. If somehow the webhook is bypassed and an invalid value is set in the spec, the unknown mode will be treated as `ignore`.

Schema validation won't have issues like being bypassed and it's much simpler in code. To replace the webhook, the only thing we need to take care of is the unknown modes in the existing HNCConfig resources. Currently, they will remain as is in the spec but will be in `ignore` mode in the status.

To switch to schema validation, we need to add a function in the HNCConfig reconciler to change unknown modes to `ignore` in the spec. With that change, users won't have any issue updating the existing HNCConfig with old unknown modes.

yiqigao217:
Add schema validation and remove the mode validation in HNCConfig
validator.

Tested with an obsolete existing HNCConfig resource with unknown modes
in the spec. Changes to the config is not allowed until the unknown
modes are fixed.

The validator test was failed before removing the test case.

Fix #899 
Part of #868

### DIFF
--- a/incubator/hnc/api/v1alpha1/hnc_config.go
+++ b/incubator/hnc/api/v1alpha1/hnc_config.go
@@ -57,8 +57,9 @@ type TypeSynchronizationSpec struct {
 	// Kind to be configured.
 	Kind string `json:"kind"`
 	// Synchronization mode of the kind. If the field is empty, it will be treated
-	// as "propagate". An unsupported mode will be treated as "ignore".
+	// as "propagate".
 	// +optional
+	// +kubebuilder:validation:Enum=propagate;ignore;remove
 	Mode SynchronizationMode `json:"mode,omitempty"`
 }
 
--- a/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hncconfigurations.yaml
+++ b/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hncconfigurations.yaml
@@ -42,7 +42,11 @@ spec:
                     description: Kind to be configured.
                     type: string
                   mode:
-                    description: Synchronization mode of the kind. If the field is empty, it will be treated as "propagate". An unsupported mode will be treated as "ignore".
+                    description: Synchronization mode of the kind. If the field is empty, it will be treated as "propagate".
+                    enum:
+                    - propagate
+                    - ignore
+                    - remove
                     type: string
                 required:
                 - apiVersion
--- a/incubator/hnc/internal/validators/hncconfig.go
+++ b/incubator/hnc/internal/validators/hncconfig.go
@@ -139,13 +139,7 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
 			fmt.Sprintf("Cannot find the %s in the apiserver with error: %s", gvk, err.Error()))
 	}
 
-	// The mode of a type should be either unset or set to one of the supported modes.
-	switch t.Mode {
-	case api.Propagate, api.Ignore, api.Remove, "":
-		return allow("")
-	default:
-		return deny(metav1.StatusReasonInvalid, fmt.Sprintf("Unrecognized mode '%s' for %s", t.Mode, gvk))
-	}
+	return allow("")
 }
 
 // realGVKValidator implements gvkValidator, and is not used during unit tests.
--- a/incubator/hnc/internal/validators/hncconfig_test.go
+++ b/incubator/hnc/internal/validators/hncconfig_test.go
@@ -160,17 +160,6 @@ func TestNonRBACTypes(t *testing.T) {
 			},
 			validator: f,
 			allow:     false,
-		},
-		{
-			name: "Unrecognized mode",
-			configs: []api.TypeSynchronizationSpec{
-				{APIVersion: "rbac.authorization.k8s.io/v1", Kind: "Role", Mode: "propagate"},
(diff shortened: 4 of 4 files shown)


incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -162,6 +163,22 @@ func (r *ConfigReconciler) validateRBACTypes(inst *api.HNCConfiguration) {
 	}
 }
 
+// validateModes set modes to ignore if they are invalid in the Spec. This func
+// can only be tested manually since the schema validation is now in place.
```
adrianludwin, line 167: s/manually/in unit tests/

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -162,6 +163,22 @@ func (r *ConfigReconciler) validateRBACTypes(inst *api.HNCConfiguration) {
...
+// There's no way to add an "existing" config with unknown modes in the test.
+// TODO delete this function when v1alpha1 is no longer the storage version of
+//  HNCConfig CRD - https://github.com/kubernetes-sigs/multi-tenancy/issues/901
+func (r *ConfigReconciler) validateModes(inst *api.HNCConfiguration) {
+	for i := 0; i < len(inst.Spec.Types); i++ {
+		t := &inst.Spec.Types[i]
+		switch t.Mode {
+		case api.Propagate, api.Ignore, api.Remove, "":
```
adrianludwin, line 175: You don't want to change "" to Propagate? Or do we even need this function anymore?

incubator/hnc/api/v1alpha1/hnc_config.go
```diff
@@ -59,6 +59,7 @@ type TypeSynchronizationSpec struct {
 	// Synchronization mode of the kind. If the field is empty, it will be treated
 	// as "propagate". An unsupported mode will be treated as "ignore".
```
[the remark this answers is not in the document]
yiqigao217: sg updated. The empty fields will still remain empty (so still work).

incubator/hnc/internal/reconcilers/hnc_config.go
yiqigao217: updated.

incubator/hnc/internal/reconcilers/hnc_config.go
yiqigao217: This field is still remain optional, so we can keep it as is for now?

adrianludwin: I'm not sure I understand. With the schema in place, what does this function do? The comment to the function says it can't be tested anymore, so why keep it around?

yiqigao217: hmm, this is for all other non-invalid mode to be set to `ignore` in line 176. If you are asking what `validateModes` func for, as stated in #899 , this is to fix any unknown modes in the already existing config resource so that the resource can be updated (otherwise with schema validation and existing invalid values, the resource cannot be updated).

adrianludwin: I wouldn't worry about fixing up existing specs as long as HNC doesn't do anything too terrible if bad values are already there - and even if they do, we can actually just ask people to fix it in the release notes.

yiqigao217: sgtm will remove this func in the reconciler.

yiqigao217: It actually affects user experience. Say I have a Secret with unknown mode and it is treated as `ignore` now. After using schema validation without this `validateModes` func, if users want to set type for ConfigMap with: 
```
k hns config set-type --apiVersion v1 --kind ConfigMap remove
```
They are not able to and will get this message that is not relevant to their changes:
```
Could not update the HNC Configuration: HNCConfiguration.hnc.x-k8s.io "config" is invalid: spec.types.mode: Unsupported value: "ignsssore": supported values: "propagate", "ignore", "remove"
```

adrianludwin: Yes, but I think that's fine given the circumstances (not too many users yet, beta software, etc). We can just put it in the release notes that the admin will need to fix any existing problems before they make any unrelated changes.

yiqigao217: ok sg makes sense let me remove it.

rjbez17:
> /approve
> /lgtm
> /hold
> /assign @rjbez17
> 
> Hey Ryan, any concerns about how we treat unsupported values here? The impact to anyone using this should be extremely low, IMO.

I think it is acceptable, the fix/workaround isn't too terrible. Do we have a ticket to add it to the docs/release notes?

rjbez17:
/lgtm
/approve

can release hold when we have answer about follow up docs ticket.

yiqigao217:
Thanks @rjbez17 ! The issue is just filed at #907 

/hold cancel

PR #903, opened 2020-07-17
Replace pattern with enum to enforce singleton

yiqigao217:
As suggested by API reviewers, the enum validation produces a nicer
error message than patterns.

Tested by applying wrong names and got nicer message:
Unsupported value: "wconfig": supported values: "hierarchy"

Part of #868

### DIFF
--- a/incubator/hnc/config/crd/patches/singleton_validation_hierarchyconfiguration.yaml
+++ b/incubator/hnc/config/crd/patches/singleton_validation_hierarchyconfiguration.yaml
@@ -12,4 +12,5 @@ spec:
           properties:
             name:
               type: string
-              pattern: '^hierarchy$'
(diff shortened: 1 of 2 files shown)


adrianludwin:
/lgtm
/approve

Out of interest, what was the error message before?

yiqigao217:
> Out of interest, what was the error message before?

@adrianludwin Good question! Attaching both here:
```
The HierarchyConfiguration "wconfig" is invalid: metadata.name: Invalid value: "": metadata.name in body should match '^hierarchy$'
```

```
The HierarchyConfiguration "wconfig" is invalid: metadata.name: Unsupported value: "wconfig": supported values: "hierarchy"
```

adrianludwin:
Nice :)

PR #906, opened 2020-07-20
Use discovery to validate types in hncconfig

### LINKED ISSUE #904: HNC: Updating mode in HNCConfig is broken at HEAD
Currently at HEAD, updating mode in HNCConfig is broken that the webhook always denies any change. E.g:
```
$ kubectl hns config set-type --apiVersion v1 --kind Secret propagate

Could not update the HNC Configuration: admission webhook "hncconfigurations.hnc.x-k8s.io" denied the request: Cannot find the /v1, Kind=Secret in the apiserver with error: an empty namespace may not be set when a resource name is provided
```
The issue is that we rely on the error message returned from controller-runtime client get - `Get(ctx, types.NamespacedName{Name: "nm"}, inst)` - to determine if the type exists. It was not broken in HNC v0.5 or before we upgraded controller-runtime to v0.6.1. Also, if I specify namespace in the `Get()` now, it is no longer broken.

However, we shouldn't rely on controller-runtime to see if a type exists and should use [discovery](https://godoc.org/k8s.io/client-go/discovery) instead. Discovery will also be used in the v1alpha1 to v1alpha2 conversion (see [details](https://docs.google.com/document/d/1xoi0v_b__A6mMHYUAI640RCPhufSHyBhmqD84vU2TWM/edit#bookmark=id.kg6fialuqx9p)).

yiqigao217:
Currently, the hncconfig validator validate types by making a Get() call
using the controller-runtime client to see if the type exists from the
error types returned. This is indirect and error-prone when the error
types returned fron controller-runtime change.

This commit uses the client-go discovery client to look for a matching
type.

Tested manually that the cannot set-type issue is fixed and requests
with incorrect GVK are still denied.

Fixes #904

### DIFF
--- a/incubator/hnc/internal/validators/hncconfig.go
+++ b/incubator/hnc/internal/validators/hncconfig.go
@@ -8,10 +8,9 @@ import (
 	"k8s.io/api/admission/v1beta1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -150,25 +149,38 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+
+	// The GV exists. Look for the matching kind now.
+	for _, resource := range resources.APIResources {
+		if resource.Kind == gvk.Kind {
+			return nil
+		}
+	}
+
+	// No matching kind. Use the same error message when the GV is not found above.
+	return errors.NewBadRequest("the server could not find the requested resource")
```
adrianludwin, line 179: Specify which resource?

yiqigao217: The caller has the info already in the `deny()` message. The current output looks like this:
```
Could not update the HNC Configuration: admission webhook "hncconfigurations.hnc.x-k8s.io" denied the request: Cannot find the /v2, Kind=Secret in the apiserver with error: the server could not find the requested resource
```

adrianludwin: Great!

PR #908, opened 2020-07-21
Updated expected return message from test.

macleanj:
It seems that for test "sigs.k8s.io/multi-tenancy/benchmarks/e2e/tests/block_bind_mounts" in K8S 1.17 on Azure AKS the return message is different.

Was:
	expectedVal = "Host path volumes are not allowed"

Observed:
	expectedVal = "hostPath volumes are not allowed"

After this correction the test passed.

### DIFF
--- a/benchmarks/e2e/tests/block_bind_mounts/block_bind_mounts.go
+++ b/benchmarks/e2e/tests/block_bind_mounts/block_bind_mounts.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	expectedVal = "Host path volumes are not allowed"
+	// expectedVal = "Host path volumes are not allowed"
+	expectedVal = "hostPath volumes are not allowed"
 )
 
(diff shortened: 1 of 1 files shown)


adrianludwin:
Is this PR still active?

macleanj:
Yes it is. It seems that the expected value is no longer valid.

JimBugwadia:
Hi @macleanj thanks for fixing this! Happy to accept / merge, but do want to mention that we are in the process of migrating all tests to https://github.com/kubernetes-sigs/multi-tenancy/tree/master/benchmarks/kubectl-mtb and will be cleaning up the old tests and directory soon.

Would be great if you can try kubectl-mtb and let us know if you see any issues there. Thanks!

macleanj:
Hi @JimBugwadia , thanks for your verification. I'll deviate to the new repo for multi-tenant testing and will take this for a spin. I've scheduled this for around next week.

PR #909, opened 2020-07-21
Add e2e test for clearing CannotUpdate conditions in descendants when a

GinnyJI:
hierachy changes

Convert test for issue #605 to issue_test.go

Tested: make e2e-test

### DIFF
--- a/incubator/hnc/e2e/tests/issues_test.go
+++ b/incubator/hnc/e2e/tests/issues_test.go
@@ -1,6 +1,8 @@
 package test
 
 import (
+	"time"
+
 	. "github.com/onsi/ginkgo"
 )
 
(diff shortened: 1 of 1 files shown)


incubator/hnc/e2e/tests/issues_test.go
```diff
@@ -112,4 +117,40 @@ var _ = Describe("Issues", func() {
...
+		// Tree should show CannotPropagateObject in parent and CannotUpdateObject in child and subchild
+		runShouldContain("CannotPropagateObject", 1, "kubectl hns describe", nsParent)
+		runShouldContain("CannotUpdateObject", 1, "kubectl hns describe", nsChild)
+		runShouldContain("CannotUpdateObject", 1, "kubectl hns describe", nsSubChild)
+		// Remove the grandchild to avoid removing CannotPropagate condition in parent or CannotUpdate condition in child.
+		// Verify that the conditions in grandchild and greatgrandchild are gone.
+		mustRun("kubectl hns set", nsSubChild, "--root")
+		// We use 10 because it takes around 9s here before the controller-runtime retry from an exponential backoff
+		// after the 30s sleep earlier. We should see 'greatgrandchild-605' has conditions for the first 9 attempts before
+		// #605 is fixed, and should see 'greatgrandchid-605' has no conditions for all 10 attemps after the fix.
```
yiqigao217, line 149: The original test has a loop here with N=10 and sleep 1 second each round and the e2e test should have a loop of 10 and  make sure there's no condition in each round.

adrianludwin:
/approve

For @yiqigao217 to lgtm

incubator/hnc/e2e/tests/issues_test.go
yiqigao217: Discussed offline that we can actually omit the loop since after the fix the obsolete condition should be gone immediately (before the issue is fixed, it takes about 8s for the condition to be gone).

So we just need to update the comment to mention why we have `sleep 30` before and the reason is that it should trigger that the obsolete condition still stays for 8s if the issue is NOT fixed.

GinnyJI: done

incubator/hnc/e2e/tests/issues_test.go
```diff
@@ -112,4 +117,34 @@ var _ = Describe("Issues", func() {
...
+		// immediately 
+		time.Sleep(30 * time.Second)
+		// Tree should show CannotPropagateObject in parent and CannotUpdateObject in child and subchild
+		runShouldContain("CannotPropagateObject", 1, "kubectl hns describe", nsParent)
+		runShouldContain("CannotUpdateObject", 1, "kubectl hns describe", nsChild)
+		runShouldContain("CannotUpdateObject", 1, "kubectl hns describe", nsSubChild)
+		// Remove the grandchild to avoid removing CannotPropagate condition in parent or CannotUpdate condition in child.
+		// Verify that the conditions in grandchild and greatgrandchild are gone.
+		mustRun("kubectl hns set", nsSubChild, "--root")
+		// We should see sub-child has conditions, and should see sub-sub-child has no conditions for all 10 attemps after the fix.
```
yiqigao217, line 143: Nit:  r/for all 10 attemps/immediately

GinnyJI: done

PR #913, opened 2020-07-22
 Add test for CritParentMissing condition

GinnyJI:
CritParentMissing condition should appear when the parent namespace is
deleted, and we should be able to delete a namespace with CritParent
Missing condition.

Tested: make e2e-test

### DIFF
--- a/incubator/hnc/e2e/tests/issues_test.go
+++ b/incubator/hnc/e2e/tests/issues_test.go
@@ -147,4 +147,27 @@ var _ = Describe("Issues", func() {
 		runShouldContain("CannotPropagateObject", 1, "kubectl hns describe", nsParent)
 		runShouldContain("CannotUpdateObject", 1, "kubectl hns describe", nsChild)
 	})
+
+	It("Should have CritParentMissing condition when parent namespace is deleted - issue #716", func(){
+		// Setting up a 2-level tree with 'a' as the root and 'b' as a child of 'a'"
+		mustRun("kubectl create ns", nsParent)
+		mustRun("kubectl create ns", nsChild)
+		mustRun("kubectl hns set", nsChild, "--parent", nsParent)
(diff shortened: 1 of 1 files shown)


adrianludwin:
Can you please rebase this change to make sure it only has one commit? Thanks!

incubator/hnc/e2e/tests/issues_test.go
```diff
@@ -18,16 +18,18 @@ var _ = Describe("Issues", func() {
...
 	BeforeEach(func() {
 		cleanupNamespaces(nsParent, nsChild, nsSub1, nsSub2, nsSub1Sub1, nsSub2Sub1, 
-			nsSubSub2, nsSubChild, nsSubSubChild)
+			nsSubSub2, nsSubChild, nsSubSubChild, nsA, nsB)
 	})
 
 	AfterEach(func() {
 		cleanupNamespaces(nsParent, nsChild, nsSub1, nsSub2, nsSub1Sub1, nsSub2Sub1, 
-			nsSubSub2, nsSubChild, nsSubSubChild)
+			nsSubSub2, nsSubChild, nsSubSubChild, nsA, nsB)
```
adrianludwin, line 32: Is there any way to get rid of this duplication?

adrianludwin:
/approve

@yiqigao217 to lgtm

incubator/hnc/e2e/tests/issues_test.go
GinnyJI: I did a quick search and don't see a good way to iterate through constants with values. I was thinking of having a cleanup() function instead and have BeforeEach() and AfterEach() calling it, but then seems like I cannot create a function inside a var.

How about change the namespaces into a struct. Instead of using `nsParent` we use `ns.parent`. Then it will be easier to iterate through when cleaning up

PR #926, opened 2020-07-23
Add test for creating and clearing unpropagatable object

GinnyJI:
Convert test-issue-771.sh to issue_test.go

Tested: make e2e-test


incubator/hnc/test/e2e/issues_test.go
```diff
@@ -209,4 +209,18 @@ var _ = Describe("Issues", func() {
 		// Expected: delete successfully
 		mustRun("kubectl delete subns", nsChild, "-n", nsParent)
 	})
+
+	It("Should create and delete an unpropagatable object - issue 771", func(){
```
yiqigao217, line 213: Oh sorry that I missed this in the first review. It should say "Should **set and unset CannotPropagateObject condition** - issue 771".

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -209,4 +209,18 @@ var _ = Describe("Issues", func() {
...
+		// set up
+		mustRun("kubectl create ns", nsParent)
+		mustRun("kubectl create ns", nsChild)
+		mustRun("kubectl hns set", nsChild, "--parent", nsParent)
+		// Creating unpropagatable object; both namespaces should have a condition
+		mustRun("kubectl create rolebinding --clusterrole=cluster-admin --serviceaccount=default:default -n", nsParent, "foo")
+		// verify
+		runShouldContain("CannotPropagateObject", 1, "kubectl hns describe", nsParent)
```
adrianludwin, line 221: The comment says "both namespaces should have a condition" but you only appear to be testing one namespace for the condition? Ditto on line 224, you only appear to test that one namespace does *not* have a condition.

yiqigao217: Ah right, you should also add `runShouldContain` `CannotUpdateObject` condition in the child namespace.

incubator/hnc/test/e2e/issues_test.go
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
GinnyJI: added test for child

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -209,4 +209,20 @@ var _ = Describe("Issues", func() {
 		// Expected: delete successfully
 		mustRun("kubectl delete subns", nsChild, "-n", nsParent)
 	})
+
+	It("Should set and unset CannotPropagateObject condition - issue 771", func(){
```
yiqigao217, line 213: one last nit "CannotPropagateObject/CannotUpdateObject" here

GinnyJI: done

PR #927, opened 2020-07-24
Add test for propagating admin rolebindings

GinnyJI:
Tested: make e2e-test


incubator/hnc/test/e2e/issues_test.go
```diff
@@ -225,4 +225,15 @@ var _ = Describe("Issues", func() {
 		runShouldNotContain("CannotPropagateObject", 1, "kubectl hns describe", nsParent)
 		runShouldNotContain("CannotUpdateObject", 1, "kubectl hns describe", nsChild)
 	})
+
+	It("Should propogate admin rolebindings - issue 772", func(){
+		// set up
+		mustRun("kubectl create ns", nsParent)
+		mustRun("kubectl create ns", nsChild)
+		mustRun("kubectl hns set", nsChild, "--parent", nsParent)
+		// Creating admin rolebinding object and waiting 2s
```
yiqigao217, line 234: Remove wait 2s in the comment. Does `mustRun` have eventually?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -225,4 +225,15 @@ var _ = Describe("Issues", func() {
...
+		mustRun("kubectl create rolebinding --clusterrole=admin --serviceaccount=default:default -n", nsParent, "foo")
+		// Object should exist in the child, and there should be no conditions
+		mustRun("kubectl get rolebinding foo -n", nsChild, "-oyaml")
```
yiqigao217, line 237: Just double checking, if you remove line 235, can this line pass?

What I want to see is that if the rolebinding resource is not there and the `k get rolebinding` will pass or fail the `mustRun()`.

adrianludwin:
/approve

@yiqigao217 to lgtm

incubator/hnc/test/e2e/issues_test.go
GinnyJI: Yes. When I had a typo in line 235 before and this test kept failing until I corrected it.

incubator/hnc/test/e2e/issues_test.go
GinnyJI: Yes. Done

incubator/hnc/test/e2e/issues_test.go
yiqigao217: Ah what was the typo? Just wanted to see if the test failed on line 235 because of the typo or line 237 because the resource is not there.

yiqigao217:
Left a last question here https://github.com/kubernetes-sigs/multi-tenancy/pull/927#discussion_r460102587

incubator/hnc/test/e2e/issues_test.go
GinnyJI: I wrote it as `--clusterrole=cluster-admin`

yiqigao217: Ah it means it will return an error when `mustRun(k get)` if the resource doesn't exist. Thanks for confirming!

PR #931, opened 2020-07-24
Add test for deleting good and bad parent namespaces under race condition

GinnyJI:
Tested: make e2e-test


incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
...
+		err := tryRun("kubectl apply -f", hncRecoverPath)
+		if err != nil {
+			GinkgoT().Log("-----------------------------WARNING------------------------------")
+			GinkgoT().Logf("WARNING: COULDN'T REPAIR HNC: %v", err)
+			GinkgoT().Log("ANY TEST AFTER THIS COULD FAIL BECAUSE WE COULDN'T REPAIR HNC HERE")
+			GinkgoT().Log("------------------------------------------------------------------")
+			GinkgoT().FailNow()
+		}
+		// give HNC enough time to repair
+		time.Sleep(5 * time.Second)
```
adrianludwin, line 269: Eliminate cut-and-paste?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
 		mustRun("kubectl get rolebinding foo -n", nsChild, "-oyaml")
 	})
 })
+
+var _ = Describe("Issues that require reparing HNC", func() {
+
+	hncRecoverPath := os.Getenv("HNC_REPAIR")
```
adrianludwin, line 244: Turn `"HNC_REPAIR"` into a constant that's shared?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
...
+	const (
+		nsParent = "parent"
+		nsParent2 = "parent-2"
+		nsChild = "child"
+	)
+
+	BeforeEach(func() {// we don't want to destroy the HNC without being able to repair it, so skip this test if recovery path not set
+		if hncRecoverPath == ""{
+			Skip("Environment variable HNC_REPAIR not set. Skipping reocovering HNC.")
+		}
```
adrianludwin, line 254: I know it's only three lines, but maybe eliminate the cut-and-paste?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
...
+			GinkgoT().Logf("WARNING: COULDN'T REPAIR HNC: %v", err)
+			GinkgoT().Log("ANY TEST AFTER THIS COULD FAIL BECAUSE WE COULDN'T REPAIR HNC HERE")
+			GinkgoT().Log("------------------------------------------------------------------")
+			GinkgoT().FailNow()
+		}
+		// give HNC enough time to repair
+		time.Sleep(5 * time.Second)
+	})
+
+	It("Should not delete child namespace when deleting a bad parent namespace - issue #797", func(){
```
adrianludwin, line 272: Can we be more specific than "bad parent namespace"? I _think_ this is "namespace with bad anchor"?

incubator/hnc/test/e2e/issues_test.go
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
 		mustRun("kubectl get rolebinding foo -n", nsChild, "-oyaml")
 	})
 })
+
+var _ = Describe("Issues that require reparing HNC", func() {
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
...
+
+	hncRecoverPath := os.Getenv("HNC_REPAIR")
+	const (
+		nsParent = "parent"
+		nsParent2 = "parent-2"
+		nsChild = "child"
+	)
+
+	BeforeEach(func() {// we don't want to destroy the HNC without being able to repair it, so skip this test if recovery path not set
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -237,3 +238,80 @@ var _ = Describe("Issues", func() {
...
+		if hncRecoverPath == ""{
+			Skip("Environment variable HNC_REPAIR not set. Skipping reocovering HNC.")
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/test/e2e/issues_test.go
GinnyJI: done

GinnyJI:
> I wonder if we will add more `it` blocks in this `describe` block for other tests that require repairing. If so, maybe move the `create ns` out of `beforeEach` and put them in each `it` block? If not, the `k apply -f recover` can also be moved from `it` blocks into `beforeEach`.
> 
> WDYT @GinnyJI @adrianludwin

I looked at the two other tests left and they both requires the same set up. I'm not sure in the future if we would add any other new test under this `Describe` block. For now it seems cleaner to put the set up under `BeforeEach` than copying them four times

yiqigao217:
> I looked at the two other tests left and they both require the same set up. I'm not sure in the future if we would add any other new test under this `Describe` block. For now it seems cleaner to put the set up under `BeforeEach` than copying them four times

sgtm. Also we can change the `describe` message in the future if there's any change. For now, do you want to move the `k apply -f recover` to `beforeEach` as well? Or you can leave it as if in the other two tests, it is not needed. LMK and I will lgtm it.

GinnyJI:
> > I looked at the two other tests left and they both require the same set up. I'm not sure in the future if we would add any other new test under this `Describe` block. For now it seems cleaner to put the set up under `BeforeEach` than copying them four times
> 
> sgtm. Also we can change the `describe` message in the future if there's any change. For now, do you want to move the `k apply -f recover` to `beforeEach` as well? Or you can leave it as if in the other two tests, it is not needed. LMK and I will lgtm it.

Right, I was thinking of that as well. The other two tests require setting up a sub-child namespace under child. Do you know if we can set it up after we repair the HNC, since we have a bad anchor in this tree? If not, we can just leave the `k apply -f recover` as it is here.

yiqigao217:
> Right, I was thinking of that as well. The other two tests require setting up a sub-child namespace under child. Do you know if we can set it up after we repair the HNC, since we have a bad anchor in this tree? If not, we can just leave the `k apply -f recover` as it is here.

Yes, we can create subnamespace in this child (which is a subnamespace of the good anchor) without having to bypass the webhook. So we can recover the webhook before that.

GinnyJI:
> > Right, I was thinking of that as well. The other two tests require setting up a sub-child namespace under child. Do you know if we can set it up after we repair the HNC, since we have a bad anchor in this tree? If not, we can just leave the `k apply -f recover` as it is here.
> 
> Yes, we can create subnamespace in this child (which is a subnamespace of the good anchor) without having to bypass the webhook. So we can recover the webhook before that.

Great, updated

PR #932, opened 2020-07-24
Add test for deleting a non-leaf child namespace under race condition

GinnyJI:
Tested: make e2e-test


incubator/hnc/test/e2e/issues_test.go
```diff
@@ -286,4 +287,21 @@ var _ = Describe("Issues that require repairing HNC", func() {
...
+		mustRun("kubectl get secret -n", nsChild)
+	})
+
+	It("Should delete a non-leaf child namespace when deleting a parent namespace with good anchor under race condition - issue #797", func(){
+		mustRun("kubectl hns create", nsSubChild, "-n", nsChild)
+		// Test: Remove subns (anchor) in the good parent 'parent'
+		// Expected: The subns (anchor) is deleted successfully and the 'child' is also deleted
+		mustRun("kubectl delete subns", nsChild, "-n", nsParent)
+		mustNotRun("kubectl get ns", nsChild, "-o yaml")
+		mustNotRun("kubectl get ns", nsSubChild, "-o yaml")
```
yiqigao217, line 305: Hmm, it seems not right here. The original test sets the `allowCascadingDelete` first. It shouldn't delete a non-leaf subnamespace if `allowCascadingDelete` is not set.

GinnyJI: Thats's right.. I think I accidentally removed it when I was removing the `k apply -f`. Added back now

adrianludwin:
/approve

For @yiqigao217 to lgtm

GinnyJI:
> The updated code lgtm. Just confirming that the deleted old bash test is not included in this PR?

Removed the bash test

PR #951, opened 2020-07-28
add vcname to metadata

thousandhu:
Nowadays, syncer will add tenancy.x-k8s.io/cluster to object annotations to distinguish which vc this object belongs to. But the annotation is not convenient in real product because the cluster id is generated by vc-manager. The cluster id's format is ${namespace}-${uuid}-${vc-name}, where user can't get uuid without access k8s super master. 

In real product system, there are always many modules and not all modules need to access k8s. For example, module A deals with users order and pass a vcname to the k8s; module B collects pods information and module C get order information from module A and pods information from module B. Without vcname in pod metadata, module C can't match module A's information with module B unless module C get the vcname and cluster id mapping from super k8s.

In this issue, I changes the build metadata method to add tenancy.x-k8s.io/vcname to object annotation and label. In this way, module C can match module A and B's information by vc name. I aims to let all module's to use vcname as unified id in the whole system and do not need to convert vcname to cluster id.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/helper.go
@@ -105,7 +105,7 @@ func GetKubeConfigOfVC(c v1core.SecretsGetter, vc *v1alpha1.VirtualCluster) ([]b
 	return adminKubeConfigSecret.Data[constants.KubeconfigAdminSecretName], nil
 }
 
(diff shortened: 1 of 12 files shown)


Fei-Guo:
Thanks for working on this.
Mostly LG.  
+1 to zhuangqh's comment, then we are good to go.

thousandhu:
@Fei-Guo @zhuangqh 
Thanks for the comments.
I've fixed the typo in the newest commit. Please help to review it again.

zhuangqh:
please squash your commits

PR #957, opened 2020-07-29
Backport all e2e tests to HNC v0.5

adrianludwin:
Right now, the e2e tests at HEAD all work on HNC v0.5.1, but this won't
be true for long. This change mass-imports all tests expected to work on
v0.5.1 so they can live together on the same branch.

Tested: all tests pass on HNC v0.5.1-rc1 on GKE 1.16.


adrianludwin:
Test changes like this don't need multiple signoffs IMO :)

/hold cancel

PR #964, opened 2020-07-29
Add e2e test for 'Propagating different types' in demo

GinnyJI:
Tested: make e2e-test

### DIFF
--- a/incubator/hnc/test/e2e/demo_test.go
+++ b/incubator/hnc/test/e2e/demo_test.go
@@ -1,6 +1,8 @@
 package e2e
 
 import (
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "sigs.k8s.io/multi-tenancy/incubator/hnc/pkg"
 )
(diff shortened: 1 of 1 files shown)


incubator/hnc/test/e2e/demo_test.go
```diff
@@ -61,4 +61,23 @@ var _ = Describe("Demo", func() {
...
+
+	It("Should propagate different types", func(){
+		MustRun("kubectl create ns", nsOrg)
+		MustRun("kubectl hns create", nsTeamA, "-n", nsOrg)
+		MustRun("kubectl hns create", nsTeamB, "-n", nsOrg)
+		MustRun("kubectl create ns", nsService1)
+		MustRun("kubectl hns set", nsService1, "--parent", nsTeamB)
+		MustRun("kubectl -n", nsTeamB, "create secret generic my-creds --from-literal=password=iamteamb")
+		// secret does not show up in service-1 because we haven’t configured HNC to propagate secrets in HNCConfiguration.
+		RunShouldNotContain("my-creds", 5, "kubectl -n", nsService1, "get secrets")
```
adrianludwin, line 73: Because we're waiting for something to _not_ happen, it actually does make sense to do an explicit sleep here (2s should be enough).

Why do you have a 5s timeout here? Was this needed to make the test pass?

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -61,4 +61,23 @@ var _ = Describe("Demo", func() {
...
+		MustRun("kubectl hns config set-type --apiVersion v1 --kind Secret propagate")
+		MustRun("kubectl get hncconfiguration config -oyaml")
+		RunShouldContain("my-creds", 5, "kubectl -n", nsService1, "get secrets")
```
adrianludwin, line 76: I'd have thought that a 2s timeout would be enough here, is that not true?

incubator/hnc/test/e2e/demo_test.go
GinnyJI: No I was trying to keep it consistent with the 5s we used in Eventually(). It's basically a random number, and if anything succeeds before 5s, it will return only using the time it need

GinnyJI: Actually after sleeping 2s, the secrete does get propagated to `service-1`

incubator/hnc/test/e2e/demo_test.go
GinnyJI: 2s is enough, I changed them all to 2s. I thought 5s might make it more stable, and for any command that takes less than 5s they will just return early

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -61,4 +63,25 @@ var _ = Describe("Demo", func() {
...
+		MustRun("kubectl hns create", nsTeamB, "-n", nsOrg)
+		MustRun("kubectl create ns", nsService1)
+		MustRun("kubectl hns set", nsService1, "--parent", nsTeamB)
+		MustRun("kubectl -n", nsTeamB, "create secret generic my-creds --from-literal=password=iamteamb")
+		// wait 2 seconds to give time for secret to propogate if it was to
+		time.Sleep(2*time.Second)
+		// secret does not show up in service-1 because we haven’t configured HNC to propagate secrets in HNCConfiguration.
+		RunShouldNotContain("my-creds", 2, "kubectl -n", nsService1, "get secrets")
+		MustRun("kubectl hns config set-type --apiVersion v1 --kind Secret propagate")
+		MustRun("kubectl get hncconfiguration config -oyaml")
```
yiqigao217, line 79: Seems we don't need this in the e2e test. This is for human to read and see the setting. Or you can have `RunShouldNotContain` secret before `k hns config set-type` line 78 and have `RunShouldContain` secret here.

adrianludwin: If a human was supposed to see it, then we should check that it's there in the demo. Just be careful because the spec and status look very similar, and we should definitely be looking for the status here. I _think_ there's a way to just get the status from an object using kubectl, maybe look into that.

adrianludwin:
/approve

To @yiqigao217 to lgtm

incubator/hnc/test/e2e/demo_test.go
GinnyJI: No we don't need to have this cmd here. I put it here because this is part of the demo, and I want to confirm that the user can run this cmd without any error. I'm not sure if it's necessary tho. Should I remove it instead?

incubator/hnc/test/e2e/demo_test.go
adrianludwin: Maybe turn "5" into `testutils.Timeout`?

wdyt "the secret does get propagated?" The code currently says the opposite?

```
		// wait 2 seconds to give time for secret to propogate if it was to
		time.Sleep(2*time.Second)
		// secret does not show up in service-1 because we haven’t configured HNC to propagate secrets in HNCConfiguration.
```

GinnyJI: Resolved it by setting to ignore the Secret at the beginning of the test

PR #986, opened 2020-08-04
E2e test for hierarchical network policy

GinnyJI:
Tested: make e2e-test

### DIFF
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -2,6 +2,7 @@ package testutils
 
 import (
 	"errors"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
@@ -46,25 +47,48 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
 
 func RunShouldContainMultiple(substrs []string, seconds float64, cmdln ...string) {
 	Eventually(func() error {
-		stdout, err := RunCommand(cmdln...)
+		missing, err := runShouldContainMultiple(substrs, cmdln...)
 		if err != nil {
 			return err
 		}
-
-		allContained := true
-		for _, substr := range substrs {
-			if strings.Contains(stdout, substr) == false {
-				allContained = false
-				break
-			}
+		if len(missing) > 0 {
+			return errors.New("Missing the expected strings: " + strings.Join(missing, ", "))
 		}
-
-		if allContained == false {
-			return errors.New("Expecting: " + strings.Join(substrs, ", ") + " but get: " + stdout)
-		}
-
 		return nil
 	}, seconds).Should(Succeed())
+}
+
+func RunErrorShouldContain(substr string, seconds float64, cmdln ...string) {
+	RunErrorShouldContainMultiple([]string{substr}, seconds, cmdln...)
+}
+
+func RunErrorShouldContainMultiple(substrs []string, seconds float64, cmdln ...string) {
+	Eventually(func() error {
+		missing, err := runShouldContainMultiple(substrs, cmdln...)
+		if len(missing) > 0 {
+			return errors.New("Missing the expected strings: " + strings.Join(missing, ", "))
+		}
+		if err == nil {
+			return errors.New("Expecting command to fail but get succeed.")
+		}
+		return nil
+	}, seconds).Should(Succeed())
+}
+
+func runShouldContainMultiple(substrs []string, cmdln ...string) ([]string, error) {
+		stdout, err := RunCommand(cmdln...)	
+		GinkgoT().Log("Output: ", stdout)
+		return missAny(substrs, stdout), err
+}
+
+func missAny(substrs []string, teststring string) []string {
+	var missing []string 
+	for _, substr := range substrs {
+		if strings.Contains(teststring, substr) == false {
+			missing = append(missing, substr)
+		}
+	}
+	return missing
 }
 
 func RunShouldNotContain(substr string, seconds float64, cmdln ...string) {
@@ -144,4 +168,16 @@ func RecoverHNC() {
 	}
 	// give HNC enough time to repair
 	time.Sleep(5 * time.Second)
+}
+
+func WriteTempFile(cxt string) string {
+	f, err := ioutil.TempFile(os.TempDir(), "e2e-test-*.yaml")
+	Expect(err).Should(BeNil())
+	defer f.Close()
+	f.WriteString(cxt)
+	return f.Name()
+}
+
+func RemoveFile(path string) {
+	Expect(os.Remove(path)).Should(BeNil())
 }
--- a/incubator/hnc/test/e2e/demo_test.go
+++ b/incubator/hnc/test/e2e/demo_test.go
@@ -93,7 +93,88 @@ var _ = Describe("Demo", func() {
 		RunShouldNotContain("my-creds", 2, "kubectl -n", nsService1, "get secrets")
 	})
 
-	It("Should create and delete subnamespaces", func() {
+	FIt("Should intergrate hierarchical network policy", func(){
+		MustRun("kubectl create ns", nsOrg)
+		MustRun("kubectl hns create", nsTeamA, "-n", nsOrg)
+		MustRun("kubectl hns create", nsTeamB, "-n", nsOrg)
+		MustRun("kubectl hns create", nsService1, "-n", nsTeamA)
+		MustRun("kubectl hns create", nsService2, "-n", nsTeamA)
+		// create a web service s2 in namespace service-2, and a client pod client-s1 in namespace service-1 that can access this web service
+		MustRun("kubectl run s2 -n", nsService2, "--image=nginx --restart=Never --expose --port 80")
+		clientArgs := "-i --image=alpine --restart=Never --rm -- sh -c"
+		cmdln := "\"wget -qO- --timeout 2 http://s2.service-2\""
+		// at least 20 seconds is needed here from experiments 
+		RunShouldContain("Welcome to nginx!", 20, 
+			"kubectl run client -n", nsService1, clientArgs, cmdln)
+		RunShouldContain("Welcome to nginx!", 10, 
+			"kubectl run client -n", nsTeamA, clientArgs, cmdln)
+		RunShouldContain("Welcome to nginx!", 10, 
+			"kubectl run client -n", nsTeamB, clientArgs, cmdln)
+
+		// create a default network policy that blocks any ingress from other namespaces 
+		policy := `# temp file created by demo_test.go
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: deny-from-other-namespaces
+  namespace: acme-org
+spec:
+  podSelector:
+    matchLabels:
+  ingress:
+  - from:
+    - podSelector: {}`
+		
+		filename := WriteTempFile(policy)
+		defer RemoveFile(filename)
+		MustRun("kubectl apply -f", filename)
+		// ensure this policy can be propagated to its descendants
+		MustRun("kubectl hns config set-type --apiVersion networking.k8s.io/v1 --kind NetworkPolicy propagate")
+		expected := "deny-from-other-namespaces"
+		RunShouldContain(expected, 2, "kubectl get netpol -n", nsOrg)
+		RunShouldContain(expected, 2, "kubectl get netpol -n", nsTeamA)
+		RunShouldContain(expected, 2, "kubectl get netpol -n", nsTeamB)
+		RunShouldContain(expected, 2, "kubectl get netpol -n", nsService1)
+		RunShouldContain(expected, 2, "kubectl get netpol -n", nsService2)
+
+		// Now we’ll see that we can no longer access service-2 from the client in service-1:
+		RunErrorShouldContain("wget: download timed out", 10, 
+			"kubectl run client -n", nsService1, clientArgs, cmdln)
+		
+		// create a second network policy that will allow all namespaces within team-a to be able to communicate with each other
+		policy = `# temp file created by demo_test.go
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: allow-team-a
+  namespace: team-a
+spec:
+  podSelector:
+    matchLabels:
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchExpressions:
+          - key: 'team-a.tree.hnc.x-k8s.io/depth'
+            operator: Exists`
+
+		filename2 := WriteTempFile(policy)
+		defer RemoveFile(filename2)
+		MustRun("kubectl apply -f", filename2)
+
+		expected = "allow-team-a"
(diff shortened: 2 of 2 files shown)


incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -46,7 +46,23 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
...
 	Eventually(func() error {
-		stdout, err := RunCommand(cmdln...)
+		return runShouldContainMultiple(substrs, cmdln...)
+	}, seconds).Should(Succeed())
+}
+
+func RunShouldContainAndExitWithError(substr string, seconds float64, cmdln ...string) {
+	RunShouldContainMultipleAndExitWithError([]string{substr}, seconds, cmdln...)
+}
+
```
adrianludwin, line 56: Nit: consider `RunErrorShouldContain`?

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+			"apiVersion: networking.k8s.io/v1" + "\n" +
+			"metadata:" + "\n" +
+			"  name: deny-from-other-namespaces" + "\n" +
+			"  namespace: acme-org" + "\n" +
+			"spec:" + "\n" +
+			"  podSelector:" + "\n" +
+			"    matchLabels:" + "\n" +
+			"  ingress:" + "\n" +
+			"  - from:" + "\n" +
+			"    - podSelector: {}"
```
adrianludwin, line 124: IMO a Go multiline string would be easier to read here, despite the fact that it would be indented strangely.

```
      policy := `# temp file created by demo_test.go
kind: NetworkPolicy
apiVersion: networking.k8s.io
...
`
```

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+		filename := "temp.yaml"
+		WriteToFile(policy, filename)
```
adrianludwin, line 126: Suggestion: this should be something like:

```
tmp := WriteTempFile(policy)
defer RemoveTempFile(tmp)
MustRun("kubectl apply -f", tmp)
```

The file should be created in the `/tmp` directory

adrianludwin: Also _consider_ https://golang.org/pkg/io/ioutil/#TempFile but don't feel obligated to use.

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+		// ensure this policy can be propagated to its descendants
+		MustRun("kubectl hns config set-type --apiVersion networking.k8s.io/v1 --kind NetworkPolicy propagate")
+		expected := []string {
+			nsOrg 	+ "    deny-from-other-namespaces",
+			nsService1 + "   deny-from-other-namespaces",
+			nsService2 + "   deny-from-other-namespaces",
+			nsTeamA + "      deny-from-other-namespaces",
+			nsTeamB + "      deny-from-other-namespaces",
+		}
+		RunShouldContainMultiple(expected, 2, "kubectl get netpol --all-namespaces")
```
adrianludwin, line 138: The constant number of strings seems quite brittle. Maybe better to iterate through the namespaces one-by-one and call `RunShouldContain` in each namespace individually.

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+		// create a second network policy that will allow all namespaces within team-a to be able to communicate with each other
+		policy = "" +
+			"kind: NetworkPolicy" + "\n" +
+			"apiVersion: networking.k8s.io/v1" + "\n" +
+			"metadata:" + "\n" +
+			"  name: allow-team-a" + "\n" +
+			"  namespace: team-a" + "\n" +
+			"spec:" + "\n" +
+			"  podSelector:" + "\n" +
+			"    matchLabels:" + "\n" +
```
adrianludwin, line 153: I have a funny feeling these lines aren't needed

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+		expected = []string {
+			nsService1 + "   allow-team-a",
+			nsService2 + "   allow-team-a",
+			nsTeamA + "      allow-team-a",
+		}
+		RunShouldContainMultiple(expected, 2, "kubectl get netpol --all-namespaces")
+
+		// Now, we can access the service from other namespaces in team-a, but not outside of it:
+		RunShouldContain("Welcome to nginx!", 10, 
+			"kubectl run client -n", nsService1, "-i --image=alpine --restart=Never --rm -- sh -c", "\"wget -qO- --timeout 2 http://s2.service-2\"")
```
adrianludwin, line 171: This would be more more clear if it looked like this:

`RunShouldContain("Welcome to nginx!", 10, "kubectl run client -n", nsService1, clientArgs)`

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+		expected := []string {
+			nsOrg 	+ "    deny-from-other-namespaces",
+			nsService1 + "   deny-from-other-namespaces",
+			nsService2 + "   deny-from-other-namespaces",
+			nsTeamA + "      deny-from-other-namespaces",
+			nsTeamB + "      deny-from-other-namespaces",
+		}
+		RunShouldContainMultiple(expected, 2, "kubectl get netpol --all-namespaces")
+
+		// Now we’ll see that we can no longer access service-2 from client a client in service-1:
```
yiqigao217, line 140: from a client?  Remove the extra "client".

incubator/hnc/test/e2e/demo_test.go
GinnyJI: done

incubator/hnc/test/e2e/demo_test.go
GinnyJI: done

incubator/hnc/pkg/testutils/testutils.go
GinnyJI: done

incubator/hnc/test/e2e/demo_test.go
GinnyJI: done

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -93,8 +93,91 @@ var _ = Describe("Demo", func() {
...
+			"kubectl run client -n", nsService1, "-i --image=alpine --restart=Never --rm -- sh -c", "\"wget -qO- --timeout 2 http://s2.service-2\"")
+		RunShouldContainAndExitWithError("wget: download timed out", 10, 
+			"kubectl run client -n", nsTeamB, "-i --image=alpine --restart=Never --rm -- sh -c", "\"wget -qO- --timeout 2 http://s2.service-2\"")
+
+
+		RemoveFile(filename)
+	})
+
+	It("Should create and delete subnamespaces", func(){
+		// set up initial structure
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/test/e2e/demo_test.go
GinnyJI: done

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -73,7 +88,7 @@ func RunShouldNotContain(substr string, seconds float64, cmdln ...string) {
 
 func RunShouldNotContainMultiple(substrs []string, seconds float64, cmdln ...string) {
 	Eventually(func() error {
-		stdout, err := RunCommand(cmdln...)
+		stdout, err := RunCommand(cmdln...)
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -46,7 +46,23 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
...
+	RunShouldContainMultipleAndExitWithError([]string{substr}, seconds, cmdln...)
+}
+
+func RunShouldContainMultipleAndExitWithError(substrs []string, seconds float64, cmdln ...string) {
+	Eventually(func() error {
+		return runShouldContainMultiple(substrs, cmdln...)
+	}, seconds).Should(Not(BeNil()))
+}
+
+func runShouldContainMultiple(substrs []string, cmdln ...string) error {
```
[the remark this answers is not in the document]
GinnyJI: RunErrorShouldContainMultiple() and RunShouldContainMultiple() both pass in a slice of the strings to runShouldContainMultiple(). Is this what you are asking for?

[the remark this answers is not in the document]
yiqigao217: oh my bad, you are right.

incubator/hnc/test/e2e/demo_test.go
GinnyJI: I tried this approach and kept getting 
```  Expected
      <*os.PathError | 0xc0002ad3b0>: {
          Op: "remove",
          Path: "tempfile.yaml",
          Err: 0x2,
      }
  to be nil```

Comment (adrianludwin) [reply incubator/hnc/test/e2e/demo_test.go:129]: Did you create two `tempfile.yaml` copies and try to delete them twice?

**Comment by yiqigao217** [comment]
FYI I will wait for the write-temp-file to give another review.

File: incubator/hnc/pkg/testutils/testutils.go [hunk 11]
```diff
@@ -46,7 +46,23 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
...
-		stdout, err := RunCommand(cmdln...)
+		return runShouldContainMultiple(substrs, cmdln...)
+	}, seconds).Should(Succeed())
+}
+
+func RunErrorShouldContain(substr string, seconds float64, cmdln ...string) {
+	RunErrorShouldContainMultiple([]string{substr}, seconds, cmdln...)
+}
+
+func RunErrorShouldContainMultiple(substrs []string, seconds float64, cmdln ...string) {
```
adrianludwin, line 57: This doesn't look right to me? `runShouldContainMultiple` returns an error _both_ if `RunCommand` fails, and _also_ if the text cannot be found. So this function would "pass" if the command _succeeded_ but then _didn't_ find the correct string, which is not what you intend.

GinnyJI: Thanks for catching it! Fixed the logic

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -46,25 +46,50 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
...
+func runShouldContainMultiple(substrs []string, cmdln ...string) error {
+		stdout, err := RunCommand(cmdln...)	
+		GinkgoT().Log("Output: ", stdout)
+		if err != nil {
+			return err
 		}
 
+		allContained := containAll(substrs, stdout)
+
 		if allContained == false {
```
yiqigao217, line 77: These two lines can now be combined after the change (into ` if !containAll(substrs, stdout){}`)

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -46,25 +46,50 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
...
+
+func RunErrorShouldContain(substr string, seconds float64, cmdln ...string) {
+	RunErrorShouldContainMultiple([]string{substr}, seconds, cmdln...)
+}
+
+func RunErrorShouldContainMultiple(substrs []string, seconds float64, cmdln ...string) {
+	Eventually(func() error {
+		err := runShouldContainMultiple(substrs, cmdln...)
+		GinkgoT().Log("Error from: RunErrorShouldContainMultiple ", err)
+		if containAll([]string {"Expecting: ", " but get: "}, err.Error()) {
```
yiqigao217, line 61: Just confirming, the `substrs []string` are the ones should appear in the error message, right? Seems the `err` on line 59 should come from just running the command instead of caring about whether `substrs` are there or not. Then line 61 should see if the error msg contains all `substrs`. CMIIW.

incubator/hnc/test/e2e/demo_test.go
GinnyJI: done

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -46,25 +47,49 @@ func RunShouldContain(substr string, seconds float64, cmdln ...string) {
...
+	RunErrorShouldContainMultiple([]string{substr}, seconds, cmdln...)
+}
+
+func RunErrorShouldContainMultiple(substrs []string, seconds float64, cmdln ...string) {
+	Eventually(func() error {
+		err := runShouldContainMultiple(substrs, cmdln...)
+		if containAll([]string {"Expecting: ", " but get: "}, err.Error()) {
 			return err
 		}
+		return nil
```
adrianludwin, line 64: It's generally considered to be bad practice to encode functional information in strings without a very well-defined schema (e.g., unless the string is JSON or something). One reason it that it's quite brittle - what happens if you change "but get" to "but got"? (The latter is actually more standard in Go). Another is that it's just confusing to the reader.

Instead, I'd suggest using multiple returned values in `runShouldContainMultiple`:

`found, err := runShouldContainMultiple(substrs, cmdln...)`

where `found` is true if everything was found, false otherwise, and `err` is only set if the command couldn't be run or returned a nonzero value. Or maybe:

`missing, err := runShouldContainMultiple(substrs, cmdln...)`

where `missing` contains any element from substrs that _wasn't_ found, so you can print them in a nice error message.

incubator/hnc/test/e2e/demo_test.go
adrianludwin: Ping. Can you try without these lines?

incubator/hnc/pkg/testutils/testutils.go
GinnyJI: `substrs []string` are the ones should appear in the `stdout` from running the cmd, which is checked in `runShouldContainMultiple`. Here we just want to distinguish if the `err` comes from 
    1. `substrs` not found in `stdout`
    2. running `cmd` returns an error

If the `err` contains `"Expecting: ", " but get: "`, then it means `stdout` does not contain the `substrs` we want, so it returns an err. Otherwise, the `err` comes from running the cmd, which is what we are expecting

incubator/hnc/pkg/testutils/testutils.go
GinnyJI: done

incubator/hnc/pkg/testutils/testutils.go
yiqigao217: Thanks for the explanation! The new change LGTM!

PR #988, opened 2020-08-04
Remove v1alpha1 from CRD status.storedVersions

### LINKED ISSUE #954: HNC: Remove v1alpha1 from CRD status.storedVersions
Part of #868 

After all v1alpha1 CRs are gone in the cluster, we can remove v1alpha1 from CRD `status.storedVersions` so that we can eventually remove v1alpha1 from the CRD `spec.versions`.

I tested manually after the conversion #953 by `kubectl edit crd XXX` and deleting the v1alpha1 listed under `status.storedVersions`. The v1alpha1 was added back immediately. It shouldn't be the case if the reconcilers only read and write in v1alpha2 that should overwrite the v1alpha1 resources. After investigation, it seems that we cannot remove it manually by `kubectl` - see issue https://github.com/kubernetes/website/issues/18248 . We will have to test it by golang script.

Anyway, this change is needed. We can have it as the 2nd PR for the API transition.

yiqigao217:
Add a reconciler to watch for HNC CRDs to remove v1alpha1 from CRD
status.storedVersions. Start this reconciler with other reconcilers in
main.go but exclude it in integration tests.

Tested manually by installing HNC v0.5, applying v1alpha1 resources in
config/sample/v1alpha1_anchor.yaml and then deploy the change. The
v1alpah1 was removed from the CRD status.storedVersions.

Next PR will automate this into an e2e test for the API conversion.

Fixes #954 
Part of #868

### DIFF
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -174,7 +174,8 @@ func startControllers(mgr ctrl.Manager, setupFinished chan struct{}) {
 	// Create all reconciling controllers
 	f := forest.NewForest()
 	setupLog.Info("Creating controllers", "maxReconciles", maxReconciles)
-	if err := reconcilers.Create(mgr, f, maxReconciles); err != nil {
+	removeOldCRDVersion := true
+	if err := reconcilers.Create(mgr, f, maxReconciles, removeOldCRDVersion); err != nil {
 		setupLog.Error(err, "cannot create controllers")
 		os.Exit(1)
 	}
--- a/incubator/hnc/internal/reconcilers/crd_removeversion.go
+++ b/incubator/hnc/internal/reconcilers/crd_removeversion.go
@@ -1,0 +1,104 @@
+package reconcilers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextension "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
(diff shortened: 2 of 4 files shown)


incubator/hnc/internal/reconcilers/crd_removeversion.go
```diff
@@ -0,0 +1,106 @@
...
+func (r *RemoveObsoleteCRDVersionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+
+	// Early exit if the CRD is not the ones we want to remove the version.
+	if !r.CRDNames[req.Name] {
+		return ctrl.Result{}, nil
+	}
+
+	nm := req.Name
+	msg := fmt.Sprintf("Reconciling CRD %s to remove %s from status.storedVersions", nm, r.ObsoleteVersion)
```
adrianludwin, line 41: Better to use structured logging (here and elsewhere): `r.Log.Info("Ensuring obsolete storedVersions are removed", "name", nm, "oldVersion", r.ObsoleteVersion)`

Or even better:
```
log := r.Log.WithValues("name", nm, "oldVersion", r.ObsoleteVersion)
...
log.Info("Reconciling")
```

This ensures that the name is included if you call `log.Error`, which it's currently missing.

incubator/hnc/cmd/manager/main.go
```diff
@@ -174,7 +174,8 @@ func startControllers(mgr ctrl.Manager, setupFinished chan struct{}) {
 	// Create all reconciling controllers
 	f := forest.NewForest()
 	setupLog.Info("Creating controllers", "maxReconciles", maxReconciles)
-	if err := reconcilers.Create(mgr, f, maxReconciles); err != nil {
+	removeCRDVersion := true
```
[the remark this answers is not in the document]
yiqigao217: updated.

incubator/hnc/internal/reconcilers/crd_removeversion.go
yiqigao217: Thanks for the suggestion! Yes, `.WithValues()` is much better, updated!

yiqigao217:
/hold cancel

@rjbez17 lmk if you have any concerns.

PR #990, opened 2020-08-04
Try injecting certs to all webhooks before early exit

yiqigao217:
Update the iteration to make sure no early return before attempting to
inject certs to all webhooks. This is to address comment in
https://github.com/open-policy-agent/cert-controller/pull/7/commits/1a1e6f93f227bffb771588dd7cf9eb255336200d#r463349960

Tested manually by deploying v0.5 and config/sample/v1alpha1_anchor.yaml
first and then deploying the change. Worked as expected.

### DIFF
--- a/incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
+++ b/incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
@@ -231,19 +231,19 @@ func injectCertToWebhook(vwh *unstructured.Unstructured, certPem []byte) error {
 	return nil
 }
 
-func injectCertToConversionWebhook(crd *unstructured.Unstructured, certPem []byte) (notFound bool, err error) {
+func injectCertToConversionWebhook(crd *unstructured.Unstructured, certPem []byte) error {
 	_, found, err := unstructured.NestedMap(crd.Object, "spec", "conversion", "webhookClientConfig")
 	if err != nil {
-		return false, err
+		return err
 	}
 	if !found {
-		return true, errors.New("`webhookClientConfig` field not found in CustomResourceDefinition")
+		return errors.New("`webhookClientConfig` field not found in CustomResourceDefinition")
 	}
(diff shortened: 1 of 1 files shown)


yiqigao217:
/hold
Until it's good to go in https://github.com/open-policy-agent/cert-controller/pull/7

/assign @adrianludwin 
/assign @rjbez17 
LMK if you have any concerns or any review comments.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -580,18 +570,45 @@ func (r *ReconcileWH) Reconcile(request reconcile.Request) (reconcile.Result, er
...
+	}
+	if err := r.client.Update(r.ctx, vwh); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileWH) ensureCRDConvWHCerts(certPem []byte) error {
+	var err error = nil
```
adrianludwin, line 598: Suggestion: call this `anyError` and then you can use `err` in the rest of the function. That makes more sense to me than `tmpErr` vs `err`.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
```diff
@@ -580,18 +570,45 @@ func (r *ReconcileWH) Reconcile(request reconcile.Request) (reconcile.Result, er
...
+		return err
+	}
+	if err := r.client.Update(r.ctx, vwh); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *ReconcileWH) ensureCRDConvWHCerts(certPem []byte) error {
```
adrianludwin, line 597: Document that this function essentially returns an arbitrary error if multiple errors are encountered; if it encounters multiple errors, these are logged but not returned.

incubator/hnc/third_party/open-policy-agent/gatekeeper/pkg/webhook/certs.go
yiqigao217: sgtm!

yiqigao217:
/hold cancel

@rjbez17 lmk if you have any concerns.

PR #996, opened 2020-08-06
Bump Kubernetes dependencies to 1.18

munnerz:
As part of general repository maintenance, and to help get us over the 'context hump', bump dependency versions for `k8s.io/*` and `controller-runtime` to use the latest 1.18 patch release.

This also involves running clientgofix, as well as making some manual edits throughout the codebase too, to have it continue to compile.

I _also_ added some more information around *why* `replace` directives are required in the `go.mod` file. In future, it'd be great if we can remove the dependency on `k8s.io/kubernetes` by either:

1) working upstream to move the required packages into `staging/` repositories in k/k
2) copying in/making our own version of the packages we depend on to avoid depending on the main repo.

It's really not advised to depend on it like this, and it causes a number of issues (such as requiring replace directives even for repositories we don't directly depend upon, such as sample-apiserver, sample-controller etc).

/cc @christopher-hein


christopherhein:
/lgtm

Nice work @munnerz !

Fei-Guo:
Nice work!  I agree we should avoid importing kubernetes directly. This can be our code refactor opportunity. 

/approve

PR #999, opened 2020-08-06
Add e2e test suite for API conversion

### LINKED ISSUE #992: HNC: Create e2e test suite for API conversion from v1alpha1 to v1alpha2
Part of #868 

Currently, the tests for API conversion are done manually with these steps:
1. Deploy HNC v0.5 with `kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-v0.5.0/hnc-manager.yaml`;
2. Apply `config/samples/hnc_v1alpha1_anchor.yaml` so that there are v1alpha1 resources for all three CRDs;
3. `make deploy` at the HEAD and see if CRD changes and HNC demo operations still work as expected.

We can automate this process and at the end, run the current HNC e2e test suite using v1alpha2 HNC plugin after the conversion. We can make the test more thorough by applying more v1alpha1 resources before the conversion.

We will clear all the test resources after each run to clean up the cluster (also because we cannot downgrade from v1alpah2 to v1alpha1 anyway).

yiqigao217:
The test installs HNC v1alpha1 first and creates v1alpha1 resources
both directly and by kubectl-hns plugin. Verify the CRD versions,
kubectl plugin and HNC behaviors are as expected in v1alpha1.

Then upgrade HNC and kubectl plugin to v1alpha2. Verify everything
mentioned above also works fine in v1alpha2.

Before and after the test, the namespaces and HNC deployment or
resources are all cleaned up.

Change some `TryRunQuietly()` function in testutils package to
`TryRun()` so that if there's a failed test, we can see the output.

Tested by `make conversion-test`. The verification of v1alpha1 to be
gone in the CRD status.storedVersions failed in the current HEAD but
passed on top of the change to remove v1alpha1 from storedVersions at
https://github.com/kubernetes-sigs/multi-tenancy/pull/988.

Fixes #992 
Part of #868

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -205,8 +205,8 @@ kind-deploy:
 # Note the `-timeout 0` that's passed to the `go test` command - by default, a
 # Go test suite has a 10m timeout, and the flag disables that timeout (as of
 # July 2020, these tests take ~15m and that number is expected to grow).
-.PHONY: e2e-test
-e2e-test:
+.PHONY: test-e2e
+test-e2e:
 	@echo
 	@echo "If these tests fail due to the webhook not being ready, wait 30s and try again. Note that webhooks can take up to 30s to become ready"
 	@echo "after HNC is first deployed in a cluster."
@@ -220,7 +220,20 @@ ifndef HNC_REPAIR
 	@sleep 5
 endif
 	go clean -testcache
-	go test -v -timeout 0 ./test/...
+	go test -v -timeout 0 ./test/e2e/...
+
+# This test is for conversion from v1alpha1 to v1alpha2. It will run on any
+# cluster w/o HNC. It will tear down any existing HNC deployment before and
+# after the test.
+#
+# Note the `-timeout 0` that's passed to the `go test` command - by default, a
+# Go test suite has a 10m timeout, and the flag disables that timeout.
+.PHONY: test-conversion
+# Have the manifests and kubectl plugin ready for the current change but do not
+# apply it. It will be applied in the test.
+test-conversion: docker-push kubectl manifests
+	go clean -testcache
+	go test -v -timeout 0 ./test/conversion/...
 
 ###################### RELEASE ACTIONS #########################
 # Build the container image by Cloud Build and build YAMLs locally
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -18,10 +18,54 @@ const eventuallyTimeout = 5
 
 var hncRecoverPath = os.Getenv("HNC_REPAIR")
 
+func FieldShouldContain(resource, ns, nm, field, want string){
+	FieldShouldContainMultiple(resource, ns, nm, field, []string{want})
+}
+
+func FieldShouldContainMultiple(resource, ns, nm, field string, want []string){
+	FieldShouldContainMultipleWithTimeout(resource, ns, nm, field, want, eventuallyTimeout)
+}
+
+func FieldShouldContainWithTimeout(resource, ns, nm, field, want string, timeout float64){
+	FieldShouldContainMultipleWithTimeout(resource, ns, nm, field, []string{want}, timeout)
+}
+
+func FieldShouldContainMultipleWithTimeout(resource, ns, nm, field string, want []string, timeout float64){
+	if ns != "" {
+		RunShouldContainMultiple(want, timeout, "kubectl get", resource, nm, "-n", ns, "-o template --template={{"+field+"}}")
+	} else {
+		RunShouldContainMultiple(want, timeout, "kubectl get", resource, nm, "-o template --template={{"+field+"}}")
+	}
+}
+
+func FieldShouldNotContain(resource, ns, nm, field, want string){
+	FieldShouldNotContainMultiple(resource, ns, nm, field, []string{want})
+}
+
+func FieldShouldNotContainMultiple(resource, ns, nm, field string, want []string){
+	FieldShouldNotContainMultipleWithTimeout(resource, ns, nm, field, want, eventuallyTimeout)
+}
+
+func FieldShouldNotContainWithTimeout(resource, ns, nm, field, want string, timeout float64){
+	FieldShouldNotContainMultipleWithTimeout(resource, ns, nm, field, []string{want}, timeout)
+}
+
+func FieldShouldNotContainMultipleWithTimeout(resource, ns, nm, field string, want []string, timeout float64){
+	if ns != "" {
+		RunShouldNotContainMultiple(want, timeout, "kubectl get", resource, nm, "-n", ns, "-o template --template={{"+field+"}}")
+	} else {
+		RunShouldNotContainMultiple(want, timeout, "kubectl get", resource, nm, "-o template --template={{"+field+"}}")
+	}
+}
+
 func MustRun(cmdln ...string) {
+	MustRunWithTimeout(eventuallyTimeout, cmdln...)
+}
+
+func MustRunWithTimeout(timeout float64, cmdln ...string) {
 	Eventually(func() error {
 		return TryRun(cmdln...)
-	}, eventuallyTimeout).Should(BeNil())
+	}, timeout).Should(BeNil())
 }
 
 func MustNotRun(cmdln ...string) {
@@ -117,6 +161,12 @@ func RunShouldNotContainMultiple(substrs []string, seconds float64, cmdln ...str
 	}, seconds).Should(Succeed())
 }
 
+func MustApplyYAML(s string){
+	filename := WriteTempFile(s)
+	defer RemoveFile(filename)
+	MustRun("kubectl apply -f", filename)
+}
+
 // RunCommand passes all arguments to the OS to execute, and returns the combined stdout/stderr and
 // and error object. By default, each arg to this function may contain strings (e.g. "echo hello
 // world"), in which case we split the strings on the spaces (so this would be equivalent to calling
@@ -148,6 +198,25 @@ func CleanupNamespaces(nses ...string) {
 		TryRunQuietly("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
 		TryRunQuietly("kubectl delete ns", ns)
 	}
+}
+
+// TearDownHNC removes CRDs first and then the entire manifest from the current
+// change. If a specific HNC version is provided, its manifest will also be
+// deleted. It will ensure HNC is cleared at the end.
+func TearDownHNC(hncVersion string) {
+	// Delete all CRDs first to ensure all finalizers are removed. Since we don't
+	// know the version of the current HNC in the cluster so we will try tearing
+	// it down twice with the specified version and what's in the HEAD.
+	TryRunQuietly("k delete crd subnamespaceanchors.hnc.x-k8s.io")
+	TryRunQuietly("k delete crd hierarchyconfigurations.hnc.x-k8s.io")
+	TryRunQuietly("k delete crd hncconfigurations.hnc.x-k8s.io")
+	TryRunQuietly("kubectl delete -f ../../manifests/hnc-manager.yaml")
+	if hncVersion != ""{
+		TryRunQuietly("kubectl delete -f https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-"+hncVersion+"/hnc-manager.yaml")
+	}
+	// Wait for HNC to be fully torn down (the namespace and the CRDs are gone).
+	RunShouldNotContain("hnc-system", 10, "kubectl get ns")
+	RunShouldNotContain(".hnc.x-k8s.io", 10, "kubectl get crd")
 }
 
 func CheckHNCPath() {
--- a/incubator/hnc/test/conversion/conversion_test.go
+++ b/incubator/hnc/test/conversion/conversion_test.go
@@ -1,0 +1,222 @@
+package conversion
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+
+	. "sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils"
+)
+
+var (
+	crds      = []string{anchorCRD, hierCRD, configCRD}
+)
+
+const (
+	certsReadyTime = 20
+	// Some reconciliation may take longer so we have it as 7 seconds, e.g. removing
+	// v1alpha1 from CRD status.storedVersions after CRD conversion because it can
+	// be removed only if all the v1alpha1 CRs are reconciled and converted to v1alpha2.
+	crdConversionTime = 7
+
+	anchorCRD       = "subnamespaceanchors.hnc.x-k8s.io"
+	hierCRD         = "hierarchyconfigurations.hnc.x-k8s.io"
+	configCRD       = "hncconfigurations.hnc.x-k8s.io"
+	hierSingleton   = "hierarchy"
+	configSingleton = "config"
+)
+
+func TestConversion(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"HNC v1alpha2 Conversion Test Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
+	const (
+		// hncFromVersion has to be a version using v1alpha1.
+		hncFromVersion = "v0.5.0"
+		namspacePrefix = "e2e-conversion-test-"
+		nsA            = namspacePrefix + "a"
+		nsB            = namspacePrefix + "b"
+	)
+
+	BeforeEach(func() {
+		CleanupNamespaces(nsA, nsB)
+		// Tear down HNC for both the specified version and at HEAD.
+		TearDownHNC(hncFromVersion)
+		setupV1alpha1(hncFromVersion)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsA, nsB)
+		// Only tear down HNC at HEAD, since that's what we just deployed.
+		TearDownHNC("")
+	})
+
+	It("should convert subnamespace anchors", func() {
+		// Before conversion, create namespace A and a subnamespace B.
+		MustRun("kubectl create ns", nsA)
+		subnsB := `# temp file created by conversion_test.go
+apiVersion: hnc.x-k8s.io/v1alpha1
+kind: SubnamespaceAnchor
+metadata:
+  name: e2e-conversion-test-b
+  namespace: e2e-conversion-test-a`
+		MustApplyYAML(subnsB)
+		FieldShouldContain(anchorCRD, nsA, nsB, ".status.status", "ok")
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion.
+		verifyCRDConversion()
+		// Verify subnamespace anchor status in the new version.
+		FieldShouldContainWithTimeout(anchorCRD, nsA, nsB, ".apiVersion", "v1alpha2", crdConversionTime)
+		FieldShouldContain(anchorCRD, nsA, nsB, ".status.status", "ok")
+	})
+
+	It("should convert HCs with parent", func() {
+		// Before conversion, create namespace A and B and set A as the parent of B.
+		MustRun("kubectl create ns", nsA)
+		MustRun("kubectl create ns", nsB)
+		hierA := `# temp file created by conversion_test.go
+apiVersion: hnc.x-k8s.io/v1alpha1
+kind: HierarchyConfiguration
+metadata:
+  name: hierarchy
+  namespace: e2e-conversion-test-b
+spec:
+  parent: e2e-conversion-test-a`
+		MustApplyYAML(hierA)
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion.
+		verifyCRDConversion()
+		// Verify the parent spec and the children status in the new version.
+		FieldShouldContainWithTimeout(hierCRD, nsA, hierSingleton, ".apiVersion", "v1alpha2", crdConversionTime)
+		FieldShouldContain(hierCRD, nsA, hierSingleton, ".status.children", nsB)
+		FieldShouldContainWithTimeout(hierCRD, nsB, hierSingleton, ".apiVersion", "v1alpha2", crdConversionTime)
+		FieldShouldContain(hierCRD, nsB, hierSingleton, ".spec.parent", nsA)
+	})
+
+	It("should convert HCs with allowCascadingDelete", func() {
+		// Before conversion, create namespace A with allowCascadingDelete.
+		MustRun("kubectl create ns", nsA)
+		hierA := `# temp file created by conversion_test.go
+apiVersion: hnc.x-k8s.io/v1alpha1
+kind: HierarchyConfiguration
+metadata:
+  name: hierarchy
+  namespace: e2e-conversion-test-a
+spec:
+  allowCascadingDelete: true`
+		MustApplyYAML(hierA)
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion.
+		verifyCRDConversion()
+		// Verify allowCascadingDelete in the new version.
+		FieldShouldContainWithTimeout(hierCRD, nsA, hierSingleton, ".apiVersion", "v1alpha2", crdConversionTime)
+		FieldShouldContain(hierCRD, nsA, hierSingleton, ".spec", "allowCascadingDelete:true")
+	})
+
+	It("should still have HC condition if it exists in v1alpha1", func() {
+		// Before conversion, create namespace B with a missing parent A (have to
+		// create A first and then delete it because otherwise the webhook will deny
+		// setting a non-existing namespace as parent).
+		MustRun("kubectl create ns", nsA)
+		MustRun("kubectl create ns", nsB)
+		hierB := `# temp file created by conversion_test.go
+apiVersion: hnc.x-k8s.io/v1alpha1
+kind: HierarchyConfiguration
+metadata:
+  name: hierarchy
+  namespace: e2e-conversion-test-b
+spec:
+  parent: e2e-conversion-test-a`
+		MustApplyYAML(hierB)
+		MustRun("kubectl delete ns", nsA)
+		FieldShouldContain(hierCRD, nsB, hierSingleton, ".status.conditions", "CritParentMissing")
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion.
+		verifyCRDConversion()
+		// Verify conditions in the new version.
+		FieldShouldContainWithTimeout(hierCRD, nsB, hierSingleton, ".apiVersion", "v1alpha2", crdConversionTime)
+		FieldShouldContain(hierCRD, nsB, hierSingleton, ".status.conditions", "CritParentMissing")
+	})
+
+	It("should convert HNCConfig", func() {
+		// Convert
+		setupV1alpha2()
+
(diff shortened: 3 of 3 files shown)


incubator/hnc/Makefile
```diff
@@ -220,7 +220,20 @@ ifndef HNC_REPAIR
...
+# This test is for conversion from v1alpha1 to v1alpha2. It will run on any
+# cluster w/o HNC. It will tear down any existing HNC deployment before and
+# after the test.
+#
+# Note the `-timeout 0` that's passed to the `go test` command - by default, a
+# Go test suite has a 10m timeout, and the flag disables that timeout.
+.PHONY: conversion-test
+# Have the manifests and kubectl plugin ready for the current change but do not
+# apply it or copy the plugin binary to the $PATH yet. It will be done in the test.
+conversion-test: docker-push build manifests
```
adrianludwin, line 234: Can we change this to `test-conversion`? Maybe update the existing one to `test-e2e` as well. That way, we can more easily use autocomplete if we forget what test we're trying to run.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -144,19 +145,38 @@ func CleanupNamespaces(nses ...string) {
 	// has cascading deletion so we can delete any of its subnamespace descendants, and
 	// make sure that it's not a subnamespace itself so we can delete it directly.
 	for _, ns := range nses {
-		TryRunQuietly("kubectl hns set", ns, "-a")
-		TryRunQuietly("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
-		TryRunQuietly("kubectl delete ns", ns)
+		TryRun("kubectl hns set", ns, "-a")
+		TryRun("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
+		TryRun("kubectl delete ns", ns)
```
adrianludwin, line 150: Why? This can produce a lot of spam.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -144,19 +145,38 @@ func CleanupNamespaces(nses ...string) {
...
 	}
 }
 
+// TearDownHNC removes CRDs first and then the entire manifest from the current
+// change. If a specific HNC version is provided, its manifest will also be deleted.
```
adrianludwin, line 155: I don't see the conditional logic here?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		RunShouldContain("CritParentMissing", 2, "kubectl hns describe", nsC)
+	})
+})
+
+// Install HNC and kubectl plugin with v1alpha1.
+func setupV1alpha1(hncVersion string){
+	goPath := os.Getenv("GOPATH")
+	MustRun("kubectl apply -f https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-"+hncVersion+"/hnc-manager.yaml")
+	MustRun("curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-"+hncVersion+"/kubectl-hns -o "+goPath+"/bin/kubectl-hns")
+	MustRun("chmod +x "+goPath+"/bin/kubectl-hns")
```
adrianludwin, line 103: A few things here:
* I don't think there's any need to install the plugin in the gopath. You could, say, install it in `/tmp` and just change any calls to `kubectl hns` to `/tmp/kubectl-hns` and I _think_ it would all work fine.
* More importantly: I'm not sure that this is getting us enough extra coverage to be worth the effort. These are conversion tests, not e2e tests - I _think_ it's sufficient to just use checked-in v1alpha1 yaml files, which are more explicit than using the plugin anyway. That would simplify a bunch of these tests, especially the setup.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		MustRun("kubectl hns set", nsA, "-a")
+		MustRun("kubectl delete ns", nsA)
+		// Verify conditions still work.
+		RunShouldContain("CritParentMissing", 2, "kubectl hns describe", nsC)
+	})
+})
+
+// Install HNC and kubectl plugin with v1alpha1.
+func setupV1alpha1(hncVersion string){
+	goPath := os.Getenv("GOPATH")
```
adrianludwin, line 100: Better: `goPath := testutils.GetGopath()`

And `GetGopath` calls `Skip()` if the env var isn't set. That way, you don't have the string `"GOPATH"` floating around, and no-one can forget to call `CheckGopath`.

But see the next comment about not needing Gopath at all.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+	}
+}
+
+// ensureHNCCleared ensures the namespace and the CRDs are gone.
+func ensureHNCCleared(crds []string){
+	RunShouldNotContain("hnc-system", 10, "kubectl get ns")
+	for _, crd := range crds {
+		RunShouldNotContain(crd, 10, "kubectl get crd")
+	}
```
adrianludwin, line 130: All the CRDs have the name "hnc" in them, so you could just say: `RunShouldNotContain("hnc", 10, "kubectl get crds")`

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// Just create a 'check-webhook' namespace to make sure it's not rejected. It
+// will be rejected if the validating webhook is not ready.
+func ensureVWHReady(){
+	RunShouldContain("namespace/check-webhook created", 20, "kubectl create ns check-webhook")
```
adrianludwin, line 136: Does `RunShouldContain` add anything here? I'd have thought that `MustRun` would be fine here IMO (it fails if the command doesn't run). I'd avoid using `RunShouldContain` unless the returned text is essential to the test.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+	SetDefaultEventuallyTimeout(time.Second * 2)
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"HNC v1alpha2 Conversion Test Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
+	const (
+		// hncVersion has to be a version using v1alpha1.
+		hncVersion = "v0.5.0"
```
adrianludwin, line 29: Maybe: `hncFromVersion`?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+
+	. "sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils"
+)
+
+const namspacePrefix = "e2e-conversion-test-"
```
adrianludwin, line 15: This can be part of the main `const` block on line 27, right?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		// Only tear down HNC at HEAD, since that's what we just deployed.
+		TearDownHNC("")
+	})
+
+	It("should convert CRD versions properly and HNC should still work", func() {
+		// Set up v1alpha1 and verify everything is fine before conversion.
+		GinkgoT().Log("Set up v1alpha1")
+		setupV1alpha1(hncVersion)
+		// Wait for the validating webhooks to be ready.
+		ensureVWHReady()
```
adrianludwin, line 60: Maybe combine these two? `setupV1alpha1` shouldn't return until it's ready to use.

You could also argue that we should _delete_ the VWHes. They're not needed since these are not e2e tests, and deleting them makes it easier/possible to create objects with conditions.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		ensureHNCCleared(crds)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsA, nsB, nsC)
+		// Only tear down HNC at HEAD, since that's what we just deployed.
+		TearDownHNC("")
+	})
+
+	It("should convert CRD versions properly and HNC should still work", func() {
```
adrianludwin, line 55: This is a very monolithic test. I'd prefer more targeted tests - e.g., what this test _actually_ seems to be doing is testing that subnamespace anchors work.

I'm also not sure I'd even try to test the full functionality of HNC. Simply testing the conversion process of the individual CRDs is much easier IMO.

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: Also, everything up to this line should probably be part of `BeforeEach`, right? They'll be the same for every test in this suite.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		// Set up v1alpha1 and verify everything is fine before conversion.
+		GinkgoT().Log("Set up v1alpha1")
+		setupV1alpha1(hncVersion)
+		// Wait for the validating webhooks to be ready.
+		ensureVWHReady()
+		// Create namespace 'a' and a subnamespace 'b'
+		GinkgoT().Log("Apply v1alpha1 resources")
+		MustRun("kubectl apply -f v1alpha1_resources.yaml")
```
adrianludwin, line 63: This makes the test much harder to understand, IMO. I'd rather see something like:

```
  MustRun("kubectl create ns", nsA)
  subns := `#
apiVersion: hnc.x-k8s.io/v1alpha1
kind: SubnamespaceAnchor
metadata:
  name: `+nsB+`
  namespace: `+nsA+`
`
  MustApplyYAML(subns) // new function - write temp file and call 'kubectl apply'
```

Now it's clear from reading the code what's happening - you don't need to skip around.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		// Create a namespace 'c' and use kubectl plugin to set its parent to 'a'
+		MustRun("kubectl create ns", nsC)
+		GinkgoT().Log("Use v1alpha1 kubectl-hns plugin")
+		MustRun("kubectl hns set", nsC, "--parent", nsA)
+
+		// Verify kubectl-hns plugin works and 'b' and 'c' are children of 'a'.
```
adrianludwin, line 69: This should be a different test.

`It("should convert subnamespace anchors", ...)`
`It("should convert HCs with parents", ...)`
`It("should convert HCs with allowCascadingDelete", ...)`

etc.

BTW are we changing `allowCascadingDelete` to `allowCascadingDeletion`?

The only time I'd try to do something more end-to-end like is to make sure that conditions look good before and after. But I wouldn't test conditions exhaustively - I'd just make sure that each of the three objects can be converted in _one_ bad state each.

adrianludwin: Plus, as mentioned above, I wouldn't use `kubectl-hns` - I'd rather see the API objects directly so I can see what's getting converted.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+		RunShouldContainMultiple([]string{nsB, nsC}, 2, "kubectl hns tree", nsA)
+		// Verify there's no 'v1alpha2' in all three CRDs for now.
+		checkCRDVersionInField("v1alpha2", ".spec", crds, false,2)
+		checkCRDVersionInField("v1alpha2", ".status", crds, false,2)
+
+		// Deploy the HEAD to include API conversion and install v1alpha2 kubectl plugin.
+		GinkgoT().Log("Set up v1alpha2")
+		setupV1alpha2()
+		// Wait for the cert rotator to write caBundles in CRD conversion webhooks.
+		ensureCRDConvWHReady(crds)
```
adrianludwin, line 79: This should all be one line IMO

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		// Verify the conversion works with 'v1alpha1' still in the spec.versions but not in status.storedVersions.
+		checkCRDVersionInField("v1alpha1", ".spec.versions", crds, true, 2)
+		checkCRDVersionInField("v1alpha2", ".spec.versions", crds, true, 2)
+		// Give this a longer timeout for the reconciler to remove v1alpha1 from storedVersions.
+		checkCRDVersionInField("v1alpha1", ".status.storedVersions", crds, false, 5)
+		checkCRDVersionInField("v1alpha2", ".status.storedVersions", crds, true,2)
```
adrianludwin, line 86: Maybe always use a 5s timeout and remove it from the list of parameters? It's a little confusing to read and waiting an additional 3s for a (rare) failure seems fine to me.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		// Verify v1alpha2 kubectl-hns plugin still works
+		RunShouldContainMultiple([]string{nsB, nsC}, 2, "kubectl hns tree", nsA)
+		// Verify cascading deletion still works.
+		MustRun("kubectl hns set", nsA, "-a")
+		MustRun("kubectl delete ns", nsA)
+		// Verify conditions still work.
+		RunShouldContain("CritParentMissing", 2, "kubectl hns describe", nsC)
```
adrianludwin, line 94: None of this tests that any information from before the conversion was preserved, so I'd take it out IMO.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		// Verify the conversion works with 'v1alpha1' still in the spec.versions but not in status.storedVersions.
+		checkCRDVersionInField("v1alpha1", ".spec.versions", crds, true, 2)
+		checkCRDVersionInField("v1alpha2", ".spec.versions", crds, true, 2)
+		// Give this a longer timeout for the reconciler to remove v1alpha1 from storedVersions.
+		checkCRDVersionInField("v1alpha1", ".status.storedVersions", crds, false, 5)
+		checkCRDVersionInField("v1alpha2", ".status.storedVersions", crds, true,2)
+
+		// Verify v1alpha2 kubectl-hns plugin still works
+		RunShouldContainMultiple([]string{nsB, nsC}, 2, "kubectl hns tree", nsA)
```
adrianludwin, line 89: This is _ok_ to test, but I'd prefer to see something like:

`HCFieldShouldContain(nsB, ".spec.parent", nsA)`

Look directly at the CRs. A general-purpose function might also make sense:

`func FieldShouldContain(resource, ns, nm, field, want string)`

Usage:

`FieldShouldContain("hierarchyconfiguration", nsB, "hierarchy", ".spec.parent", nsA)`

incubator/hnc/pkg/testutils/testutils.go
GinnyJI: +1

yiqigao217: I think it only outputs when the test failed and won't appear if the test passes. I mentioned it in the commit/PR message as "Change some `TryRunQuietly()` function in testutils package to
`TryRun()` so that if there's a failed test, we can see the output."

WDYT?

incubator/hnc/Makefile
yiqigao217: sg will do.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: sgtm, I will remove the verification for the v1alpha1 kubectl-hns plugin but only keep the v1alpha2 plugin test.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Thanks for the suggestion this simplifies it a lot!

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: sg.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Removing GOPATH as sugguested in the next comment.

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: I guess? `TryRunQuietly` has only ever been used for lines that we _expect_ to fail even in successful runs, so they could cause red herrings. Did you run into a case where this wasn't true? If not, I think I'd prefer to leave these alone.

yiqigao217: Oh I was trying to see what could be the issue of a namespace not being deleted. The issue is gone and I changed it back to quietly.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Oh this one waits for the VWH to be ready so it may fail before that.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Updated.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Broke it into 5 tests.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -150,6 +156,30 @@ func CleanupNamespaces(nses ...string) {
...
+	TryRunQuietly("kubectl delete -f ../../manifests/hnc-manager.yaml")
+	if hncVersion != ""{
+		TryRunQuietly("kubectl delete -f https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-"+hncVersion+"/hnc-manager.yaml")
+	}
+	// Wait for HNC to be fully torn down.
+	ensureHNCCleared()
+}
+
+// ensureHNCCleared ensures the namespace and the CRDs are gone.
+func ensureHNCCleared(){
```
adrianludwin, line 178: This is only called in one place, maybe inline?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,201 @@
...
+		RunShouldContain(nsB, 10, "kubectl get ns")
+		RunShouldContain("ok", 10, "kubectl get subnamespaceanchors.hnc.x-k8s.io", nsB, "-n", nsA, "-oyaml")
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify
+		verifyCRDConversion()
+		MustRun("kubectl -n", nsA, "delete subnamespaceanchors.hnc.x-k8s.io", nsB)
+		RunShouldNotContain(nsB, 4, "kubectl get ns")
```
adrianludwin, line 68: Why?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,201 @@
...
+spec:
+  parent: e2e-conversion-test-a`
+		MustApplyYAML(hc)
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion and kubectl plugin with v1alpha2.
+		verifyCRDConversion()
+		RunShouldContain(nsB, 5, "kubectl hns tree", nsA)
```
adrianludwin, line 90: I'd prefer to read the HCs of nsA and nsB directly and confirm:
1. Both have version v1alpha2
2. nsA has child nsB
3. nsB has parent nsA

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,201 @@
...
+spec:
+  allowCascadingDelete: true`
+		MustApplyYAML(anchor)
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify
+		verifyCRDConversion()
+		RunShouldContain("is already set to true; unchanged", 2, "kubectl hns set", nsA, "-a")
```
adrianludwin, line 111: Again, better to confirm the yaml directly (especially when we change the name of this field)

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,201 @@
...
+apiVersion: hnc.x-k8s.io/v1alpha1
+kind: HierarchyConfiguration
+metadata:
+  name: hierarchy
+  namespace: e2e-conversion-test-b
+spec:
+  parent: e2e-conversion-test-a`
+		MustApplyYAML(hc)
+		MustRun("kubectl delete ns", nsA)
+
```
adrianludwin, line 130: Maybe confirm condition before moving on?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,201 @@
...
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion and the condition.
+		verifyCRDConversion()
+		RunShouldContain("CritParentMissing: missing parent", 2, "kubectl hns describe", nsB)
```
adrianludwin, line 136: Again, use `get -oyaml` so we can directly observe the v1alpha2 conditions when we have them.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,201 @@
...
+		MustRun("kubectl create ns", nsA)
+		anchor := `# temp file created by conversion_test.go
+apiVersion: hnc.x-k8s.io/v1alpha1
+kind: SubnamespaceAnchor
+metadata:
+  name: e2e-conversion-test-b
+  namespace: e2e-conversion-test-a`
+		MustApplyYAML(anchor)
+		RunShouldContain(nsB, 10, "kubectl get ns")
+		RunShouldContain("ok", 10, "kubectl get subnamespaceanchors.hnc.x-k8s.io", nsB, "-n", nsA, "-oyaml")
```
adrianludwin, line 60: Replace all numeric constants with named constants and documentation for why the values were chosen

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: I'm not sure I understand? If it waits for the VWH to be ready but fails, it will still fail if you just call `MustRun`

incubator/hnc/test/conversion/v1alpha1_resources.yaml
```diff
@@ -0,0 +1,13 @@
+# This file can be applied to a cluster with HNC v0.5 deployment to create
```
adrianludwin, line 1: Remove this file?

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Oh `MustRun` func has a fixed timeout of 5 seconds. Let me create a new `MustRunEventually(seconds float64, cmdln ...string)` func for this? Since VWH takes longer.

adrianludwin: Ah, I see. I'm glad I asked :) Maybe consider `MustRunWithTimeout`?

yiqigao217: sgtm will use that for the new func name.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+		// Verify CRD conversion.
+		verifyCRDConversion()
+		// Verify default types in the new version.
+		//time.Sleep(20*time.Second)
+		// TEST PASSED:
+		RunShouldContainMultiple([]string{"v1alpha2", "Role", "RoleBinding"}, statusUpdateTime, "kubectl get", configCRD, "-oyaml")
+		// TEST FAILED: got "v1error: resource name may not be empty"
+		FieldShouldContain(configCRD, "", configSingleton, ".apiVersion", "v1alpha2")
+		FieldShouldContainMultiple(configCRD, "", configSingleton, ".spec.types", []string{"Role", "RoleBinding"})
```
yiqigao217, line 169: /hold

Until this issue is resolved. All other 4 tests PASSED.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,223 @@
...
+)
+
+var (
+	crds      = []string{anchorCRD, hierCRD, configCRD}
+)
+
+const (
+	certsReadyTime = 20
+	// Some reconciliation may take longer so we have it as 6 seconds.
+	statusUpdateTime = 6
```
adrianludwin, line 20: Can you say anything more about _why_ these reconciliations take longer?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,223 @@
...
+		FieldShouldContain(anchorCRD, nsA, nsB, ".status.status", "ok")
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify CRD conversion.
+		verifyCRDConversion()
+		// Verify subnamespace anchor status in the new version.
+		FieldShouldContain(anchorCRD, nsA, nsB, ".apiVersion", "v1alpha2")
+		FieldShouldContain(anchorCRD, nsA, nsB, ".status.status", "ok")
```
adrianludwin, line 78: Perfect :)

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,223 @@
...
+		setupV1alpha1(hncFromVersion)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsA, nsB)
+		// Only tear down HNC at HEAD, since that's what we just deployed.
+		TearDownHNC("")
+	})
+
+	FIt("should convert subnamespace anchors", func() {
```
adrianludwin, line 59: Make sure to turn these back into "It"

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -0,0 +1,223 @@
...
+// Check if a specific version exists in the field as expected for a list of CRDs.
+func checkCRDVersionInField(version, field string, expected bool) {
+	for _, crd := range crds {
+		if expected {
+			RunShouldContain(version, statusUpdateTime, "kubectl get -o template crd/"+crd+" --template={{"+field+"}}")
+		} else {
+			RunShouldNotContain(version, statusUpdateTime, "kubectl get -o template crd/"+crd+" --template={{"+field+"}}")
+		}
+	}
+}
```
adrianludwin, line 210: Use `FieldShouldContain` now that you have it?

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: FYI the failed line 168 translates to `kubectl get hncconfigurations.hnc.x-k8s.io config  -o template --template={{.apiVersion}}`.

I also tried `kubectl get hncconfigurations config  -o template --template={{.apiVersion}}` as the cmd for this line and got the same error `v1error: resource name may not be empty`.

While I manually try either cmd, it worked and I got the expected result `hnc.x-k8s.io/v1alpha2`.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Using `FieldShouldContain()` now.

incubator/hnc/test/conversion/conversion_test.go
liggitt: passing an empty ns translates to `kubectl get hncconfigurations.hnc.x-k8s.io config "" -o template --template={{.apiVersion}}`

yiqigao217: Ah right verified by removing the "" and the test passed. Thanks!

PR #1015, opened 2020-08-14
Add annotation for subnamespace when print tree

### LINKED ISSUE #844: HNC: `kubectl hns tree` should identify subnamespaces
It would be nice to easily see which namespaces are subnamespaces.

GinnyJI:
Tested: manually. The output looks like this:
  ```
  #setup
  $ k create ns root
  namespace/root created
  $ k create ns nonsub
  namespace/nonsub created
  $ k hns set nonsub --parent root
  Setting the parent of nonsub to root
  Succesfully updated 1 property of the hierarchical configuration of nonsub
  $ k hns create child -n root
  Successfully created "child" subnamespace anchor in "root" namespace

  #result
  $ k hns tree root
  root
  ├── [s] child
  └── nonsub
[s] indicates subnamespaces.
  ```

Fix #844

### DIFF
--- a/incubator/hnc/internal/kubectl/tree.go
+++ b/incubator/hnc/internal/kubectl/tree.go
@@ -51,6 +51,9 @@ var treeCmd = &cobra.Command{
 			fmt.Printf("Error: Must specify the root of the tree(s) to display or else specify --all-namespaces\n")
 			os.Exit(1)
 		}
+
+		hasSubnamespace := false
+
 		for _, nnm := range nsList {
 			hier := client.getHierarchy(nnm)
 			// If we're showing the default list, skip all non-root namespaces since they'll be displayed
@@ -62,8 +65,13 @@ var treeCmd = &cobra.Command{
 				continue
 			}
 			fmt.Println(txt)
-			printSubtree("", hier, cycle)
+			hasSubnamespace = hasSubnamespace || printSubtree("", hier, cycle)
 		}
+
+		if hasSubnamespace {
+			fmt.Printf("[s] indicates subnamespaces.\n")
+		}
+
 		if len(footnotes) > 0 {
 			fmt.Printf("\nConditions:\n")
 
@@ -74,21 +82,38 @@ var treeCmd = &cobra.Command{
 	},
 }
 
-func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool) {
+func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool) (hasSubnamespace bool) {
+	hasSubnamespace = false
 	for i, cn := range hier.Status.Children {
 		ch := client.getHierarchy(cn)
 		txt, cycle := nameAndFootnotes(ch)
 		if cycle && inCycle {
 			continue
 		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		ns, err := k8sClient.CoreV1().Namespaces().Get(ctx, txt, metav1.GetOptions{})
+		if err != nil {
+			fmt.Printf("Could not get namespaces: %s\n", err)
+			os.Exit(1)
+		}
+
+		if _, ok := ns.ObjectMeta.Annotations[api.SubnamespaceOf]; ok {
+			txt = "[s] " + txt
+			hasSubnamespace = true
+		}
+
 		if i < len(hier.Status.Children)-1 {
 			fmt.Printf("%s├── %s\n", prefix, txt)
-			printSubtree(prefix+"│   ", ch, cycle)
+			hasSubnamespace = hasSubnamespace || printSubtree(prefix+"│   ", ch, cycle)
 		} else {
 			fmt.Printf("%s└── %s\n", prefix, txt)
-			printSubtree(prefix+"    ", ch, cycle)
+			hasSubnamespace = hasSubnamespace || printSubtree(prefix+"    ", ch, cycle)
 		}
 	}
+	return
 }
 
 // nameAndFootnotes returns the text to print to describe the namespace, in the form of the


adrianludwin:
> Tested:
> make deploy
> k create ns root
> k create ns nonsub
> k hns create child -n root
> k hns set nonsub --parent root
> k hns tree root
> 
> Fix #844

Perhaps a better description would be something like:

```
Tested: manually, output is similar to:
<paste the output showing subns and full ns children>
```

Note that if you amend your existing commit and force-push, the _PR description_ will not be updated, so you'll have to copy the _commit_ description into the PR description.

incubator/hnc/internal/kubectl/tree.go
```diff
@@ -81,6 +81,21 @@ func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool)
 		if cycle && inCycle {
 			continue
 		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		ns, err := k8sClient.CoreV1().Namespaces().Get(ctx, txt, metav1.GetOptions{})
+		if err != nil {
+			fmt.Printf("Could not list namespaces: %s\n", err)
```
adrianludwin, line 90: Update error message

incubator/hnc/internal/kubectl/tree.go
```diff
@@ -81,6 +81,21 @@ func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool)
...
+			os.Exit(1)
+		}
+
+		_, hasSubnamespaceAnnotation := ns.ObjectMeta.Annotations["hnc.x-k8s.io/subnamespaceOf"]
```
adrianludwin, line 94: You should almost never have raw strings in your programs. Use `api.SubnamespaceOf`, defined here: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/api/v1alpha2/subnamespace_anchor.go#L27

incubator/hnc/internal/kubectl/tree.go
```diff
@@ -81,6 +81,21 @@ func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool)
...
+		if hasSubnamespaceAnnotation {
+			txt = "[sub]" + txt
```
adrianludwin, line 96: Looking it over now, "[sub]" seems unclear.

@rjbez17 , wdyt? I'm in favour of either making it "[s] name" and have a "footer" showing what "[s]" means, or "[subns]" which is self-explanatory but longer.

rjbez17: I think "[s]" + footnote may be slightly less clear but it will make viewing large trees much easier and frequent users will quickly understand what it means. But I could go either way really. I do agree "[sub]" is unclear.

adrianludwin:
Ok, let's go with [s] then. Thanks!

incubator/hnc/internal/kubectl/tree.go
GinnyJI: done

incubator/hnc/internal/kubectl/tree.go
GinnyJI: done

incubator/hnc/internal/kubectl/tree.go
GinnyJI: done

incubator/hnc/internal/kubectl/tree.go
```diff
@@ -81,6 +81,21 @@ func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool)
...
+
+		ns, err := k8sClient.CoreV1().Namespaces().Get(ctx, txt, metav1.GetOptions{})
+		if err != nil {
+			fmt.Printf("Could not get namespaces: %s\n", err)
+			os.Exit(1)
+		}
+
+		_, hasSubnamespaceAnnotation := ns.ObjectMeta.Annotations[api.SubnamespaceOf]
+		if hasSubnamespaceAnnotation {
+			txt = "[s]" + txt
```
adrianludwin, line 96: Need a space after the square bracket

`[s] foo`

Not

`[s]foo`

rjbez17:
> +1 to @yiqigao217's comment. We need a footnote, I think the text should be:
> 
> ```
> [s] indicates subnamespaces
> ```
> 
> @rjbez17 lgty?

lgtm

incubator/hnc/internal/kubectl/tree.go
GinnyJI: done

incubator/hnc/internal/kubectl/tree.go
```diff
@@ -64,6 +64,9 @@ var treeCmd = &cobra.Command{
 			fmt.Println(txt)
 			printSubtree("", hier, cycle)
 		}
+
+		fmt.Printf("[s] indicates subnamespaces.\n")
+
```
adrianludwin, line 69: We only want to print this if there _are_ subnamespaces. See the logic just below (`if len(footnotes) > 0`) as an example - but note that after your change, not all footnotes will be conditions anymore, so you might want to change some of the existing logic too.

incubator/hnc/internal/kubectl/tree.go
```diff
@@ -81,6 +84,21 @@ func printSubtree(prefix string, hier *api.HierarchyConfiguration, inCycle bool)
...
+		defer cancel()
+
+		ns, err := k8sClient.CoreV1().Namespaces().Get(ctx, txt, metav1.GetOptions{})
+		if err != nil {
+			fmt.Printf("Could not get namespaces: %s\n", err)
+			os.Exit(1)
+		}
+
+		_, hasSubnamespaceAnnotation := ns.ObjectMeta.Annotations[api.SubnamespaceOf]
+		if hasSubnamespaceAnnotation {
```
adrianludwin, line 98: Go prefers short variable names if the name's only used once or twice. So you could use the following here:

```
// Identify subnamespaces
if _, ok := ns.ObjectMeta.Annotations[api.SubnamespaceOf]; ok {
  txt = "[s] " + txt
}
``

GinnyJI: done

incubator/hnc/internal/kubectl/tree.go
GinnyJI: done

rjbez17:
/lgtm
/approve
/hold 

@adrianludwin or @yiqigao217 feel free to release hold

PR #1016, opened 2020-08-14
Soften pre-GA warnings

adrianludwin:
Change HNC from "alpha" to "pre-GA," clarify the roadmap, be a bit more
positive about getting involved and add vendor installation links.


adrianludwin:
/assign @rjbez17 
cc @tashimi 

Any concerns about these updates to the HNC landing page? 

Staged version: https://github.com/adrianludwin/multi-tenancy/blob/hnc-beta/incubator/hnc/README.md

PR #1020, opened 2020-08-17
Change `hasSubnamespace` to a global variable

GinnyJI:
As discussed with @adrianludwing offline, changing the `hasSubnamespace`
variable to a global variable for cleaner code struct.

Tested: make kubectl


incubator/hnc/internal/kubectl/tree.go
```diff
@@ -42,6 +43,7 @@ var treeCmd = &cobra.Command{
 		flags := cmd.Flags()
 		footnotesByMsg = map[string]int{}
 		footnotes = []string{}
+		hasSubnamespace = false
```
adrianludwin, line 46: This isn't needed - boolean variables default to `false`.

GinnyJI: done

PR #1021, opened 2020-08-17
Line up Cloud Build with makefile

adrianludwin:
The makefile recently changed the name of the released image, but we
didn't update Cloud Build to match; this commit fixes that. It also
allows CB to pull from a different repo to make testing release changes
easier.

Tested: created a fake release and built from
gihub.com/adrianludwin/multi-tenancy using the new override.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -27,6 +27,8 @@ else
 endif
 endif
 
+# Override this to release from a different repo
+HNC_REPO_USER ?= "kubernetes-sigs"
 HNC_RELEASED_IMG ?= "gcr.io/k8s-staging-multitenancy/${HNC_IMG_NAME}:${HNC_IMG_TAG}"
 
 CONTROLLER_GEN ?= "./bin/controller-gen"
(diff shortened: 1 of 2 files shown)


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 

I think the first person who reviews this can just send it in, no need for dual reviews
/cc @GinnyJI

incubator/hnc/Makefile
```diff
@@ -27,6 +27,8 @@ else
 endif
 endif
 
+# Override this to release from a different repo
+HNC_REPO_USER ?= "kubernetes-sigs"
```
rjbez17, line 31: nit: This is generally called the org/organization in docker world, not user.

rjbez17:
/lgtm
/approve
/hold

one nit, feel free to remove hold if the nit is fine with you.

incubator/hnc/cloudbuild.yaml
```diff
@@ -57,6 +57,6 @@ steps:
   - 'https://uploads.github.com/repos/kubernetes-sigs/multi-tenancy/releases/$_HNC_RELEASE_ID/assets?name=kubectl-hns'
 # Build Docker image
 - name: gcr.io/cloud-builders/docker
-  args: ['build', '-t', 'gcr.io/$PROJECT_ID/hnc/controller:$_HNC_IMG_TAG', 'multi-tenancy/incubator/hnc']
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/$_HNC_IMG_NAME:$_HNC_IMG_TAG', 'multi-tenancy/incubator/hnc']
```
yiqigao217, line 60: Question: the `HNC_IMG_NAME` defaults to `hnc-manager` which is different from `hnc/controller` here?

incubator/hnc/Makefile
adrianludwin: Good call. This is for a Github repo and apparently they're called [owners](https://github.com/gitbucket/gitbucket/wiki/List-of-URL-Formats#repositories) there, so I'll make that change. But I already have another dependent PR so I'll make it in the followup change rather than try to merge it in.

/hold cancel

incubator/hnc/cloudbuild.yaml
adrianludwin: Yes - we switched the makefile from `hnc/controller` to `hnc-manager` some time ago. Having multiple paths in a Docker registry isn't standard - the Docker Hub, for example, doesn't support it.

yiqigao217: Thanks for the explanation!

PR #1022, opened 2020-08-17
More hermetic, unified builds

adrianludwin:
This commit makes two significant changes:

* Includes the kustomize tool so that we don't need to either fetch it
from the internet or rely on one being installed locally.
* Replaces portions of the Cloud Build process with the equivalent steps
from the Makefile so they can't get out of sync.

As a side effect, this required a bit of refactoring in the makefile,
and I've moved some lines around to be more closely located with the
relevant recipes.

I tried removing `kubebuilder` (I thought we just needed controller-gen)
but it contains elements required by Ginkgo. Since it's rather large
(>50MB) and doesn't affect the correctness of the binaries as they're
built (only tested), I decided not to check it in. I did upgrade the
version we use from a 2.0 alpha to the latest release (as of Aug 2020).

Finally, this includes a small docs change that was mistakenly left out
of an earlier PR.

Tested: made all targets locally; confirmed `kubectl hns version` gave
the expected result; pushed a test tag to my repo and used the
HNC_RELEASE_REPO_OWNER flag to build a fake release from that tag.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -4,34 +4,53 @@
 # If CONFIG is `kind`, various defaults will be optimized for deploying locally to Kind
 CONFIG ?= "default"
 
-PROJECT_ID=$(shell gcloud config get-value project)
+# The GCP project ID useful to have when performing operations that require one
+# (e.g. release). If you don't have gcloud, all other operations in this
+# makefile should still succeed.
+#
+# We use simple expansion to prevent this from being called every time we need
+# the project ID. gcloud prints some noise to stderr, hence the redirect.
+PROJECT_ID := ${shell gcloud config get-value project 2>/dev/null}
 
-# Set default version tag for krew build and Docker image (unless version is set)
+# The registry is the location of the image to be built, if any:
+#
+# * If you are building locally (e.g. 'make deploy'), this is where the image
+#   will be pushed after it is built, and what the manifest files will use.
+# * If you are releasing HNC, this is where the *temporary* image will be
+#   built. The *final* image will be in HNC_RELEASE_REGISTRY.
+#
+# By default, we use the Container Registry in the current GCP project, but you
+# can set it to anything, UNLESS you are calling 'make release' since the image
+# needs to be built in the same project as the GCB instance that builds it.
+HNC_REGISTRY ?= gcr.io/${PROJECT_ID}
+
+# The image name is the *base* name - excluding the registry and the tag.
 HNC_IMG_NAME ?= hnc-manager
-HNC_IMG_TAG ?= latest
 
-# Image URL to use all building/pushing image targets
-ifeq ($(CONFIG),kind)
-# The tag is `kind-local` since K8s always attempst to re-pull an image with
-# the `latest` tag, and this doesn't work when we're testing locally (we rely
-# on the docker-push target, below, to push the image into Kind).
-HNC_IMG ?= controller:kind-local
+# By default, the image tag is "latest" but you should override this when
+# creating a release. If you're using Kind, the tag is 'kind-local' since K8s
+# always attempts to re-pull an image the 'latest' tag, and this doesn't work
+# when we're testing locally (we rely on the docker-push target, below, to push
+# the image into Kind).
+ifneq ($(CONFIG),kind)
+  HNC_IMG_TAG ?= latest
 else
-HNC_IMG ?= "gcr.io/${PROJECT_ID}/${HNC_IMG_NAME}:${HNC_IMG_TAG}"
-# We don't want to overwrite an existing docker image and tag so we query the repository
-# to see if the image already exists. This can be overridden by setting FORCE_RELEASE=true
-ifeq ($(FORCE_RELEASE), true)
-	IMG_READ_ERROR=1
-else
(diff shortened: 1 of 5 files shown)


incubator/hnc/Makefile
```diff
@@ -4,34 +4,49 @@
...
 CONFIG ?= "default"
 
-PROJECT_ID=$(shell gcloud config get-value project)
+# The GCP project ID useful to have when performing operations that require one
+# (e.g. release). If you don't have gcloud, all other operations in this
+# makefile should still succeed.
+#
+# We use simple expansion to prevent this from being called every time we need
+# the project ID. gcloud prints some noise to stderr, hence the redirect.
+PROJECT_ID := ${shell gcloud config get-value project 2>/dev/null}
```
yiqigao217, line 13: Question: the purpose of the redirect here is just to not include stderr in the `PROJECT_ID`, right?

incubator/hnc/Makefile
```diff
@@ -4,34 +4,49 @@
...
 endif
 
-# Override this to release from a different repo
-HNC_REPO_USER ?= "kubernetes-sigs"
-HNC_RELEASED_IMG ?= "gcr.io/k8s-staging-multitenancy/${HNC_IMG_NAME}:${HNC_IMG_TAG}"
+# HNC_IMG is the full image name. It can't be overridden in its entirety since
+# parts of the Makefile and Cloud Build target make certain assumptions about
+# its components.
+#
+# Note that if you're using Kind, this image will never actually be pushed to
```
yiqigao217, line 41: Does kind skip push and fetch, and use the image directly?

incubator/hnc/Makefile
```diff
@@ -239,39 +254,77 @@ test-conversion: docker-push kubectl manifests
...
 
+# Set up error checking variables
 ERR_MSG=Ensure that HNC_IMG_TAG (eg v0.1.0), HNC_USER (your Github username), HNC_PAT (Github personal access token) and HNC_RELEASE_ID (Github numeric ID) are set
+# During a release, We don't want to overwrite an existing docker image and tag
+# so we query the repository to see if the image already exists. This can be
+# overridden by setting HNC_FORCE_RELEASE=true
+ifeq ($(HNC_FORCE_RELEASE), true)
+	COULDNT_READ_RELEASE_IMG=1
+else
+	COULDNT_READ_RELEASE_IMG=$(shell gcloud container images describe $(HNC_RELEASE_IMG) > /dev/null 2>&1; echo $$?)
```
yiqigao217, line 311: Does this set it to 0 if there's no errors?

incubator/hnc/cloudbuild.yaml
```diff
@@ -39,7 +33,7 @@ steps:
   - '-H'
   - 'Content-Type: application/x-application'
   - '--data-binary'
-  - '@multi-tenancy/incubator/hnc/out/hnc-manager.yaml'
```
yiqigao217, line 42: Why was it using out here?

incubator/hnc/Makefile
adrianludwin: Exactly. Can I make the comment on line 12 more clear?

incubator/hnc/Makefile
adrianludwin: Sort of. Kind never pulls from a registry, so the `docker-push:` target pushes the image directly _to Kind_, and not to a registry.

incubator/hnc/Makefile
adrianludwin: Yes. By convention, all Linux (or Unix-style) processes exit with the code 0 if they run successfully, and we confirmed that `gcloud container images describe` follows this pattern.

incubator/hnc/cloudbuild.yaml
adrianludwin: It's replaced by the line below. The `out` directory was only used by the Cloud Build script; now that we're generating these files using `make`, the manifests are in the `manifests/` directory, just like when you're developing it yourself.

yiqigao217:
Thanks for the explanations! @adrianludwin can you help me resolve the comments (somehow I don't have that option).

/lgtm
/hold
/assign @rjbez17

adrianludwin:
/wip

This conflicts with #1032 so I need to revise

adrianludwin:
/hold cancel

Updated to include multiplatform kubectl plugins. Tested by releasing to https://github.com/adrianludwin/multi-tenancy/releases/tag/hnc-v0.6.0-test1.

adrianludwin:
@yiqigao217 Can you please re-review? Note that the hold is gone.
@rjbez17 Are you interested in reviewing this?

Thanks, A

incubator/hnc/hack/README.md
```diff
@@ -0,0 +1,31 @@
...
+`boilerplate.go.txt` includes the Apache header. Kubebuilder put it here and it
+seems like as good a place as any.
+
+`krew-hierarchical-namespaces.yaml` is a template for the Krew `kubectl-hns`
+plugin.
+
+## CI
+
+Other projects seem to put their presubmits, postsubmits etc here, so we did
+too. See ../README.md for where these are configured in Prow.
```
yiqigao217, line 31: Replace readme with more specific https://github.com/kubernetes-sigs/multi-tenancy/tree/master/incubator/hnc#test-infrastructure?

adrianludwin: Done

PR #1024, opened 2020-08-18
[VC] notify tenant about pod FailCreated event in super.

zhuangqh:


zhuangqh:
/hold 
no fully tested

thousandhu:
Very useful function. In our cluster, there's some webhook installed on super master. If a  pod rejected by the webhook, we can't get any information from tenant side. The only way to get the pod rejection reason is to check the log of webhooks or super apiserver. With this issue we can get the information from pod's event on tenant master and it's very helpful for debugging.

zhuangqh:
/hold cancel
ready to merge  @Fei-Guo

incubator/virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -45,55 +45,59 @@ func (c *controller) StartDWS(stopCh <-chan struct{}) error {
...
 			klog.Errorf("failed reconcile Pod %s/%s CREATE of cluster %s %v", request.Namespace, request.Name, request.ClusterName, err)
+			go func() {
+				if parentRef := getParentRefFromPod(vPod); parentRef != nil {
+					c.multiClusterPodController.Eventf(request.ClusterName, parentRef, v1.EventTypeWarning, "FailedCreate", "Error creating: %v", err)
+				}
+				c.multiClusterPodController.Eventf(request.ClusterName, &v1.ObjectReference{
+					Kind:      "Pod",
+					Namespace: vPod.Namespace,
+					UID:       vPod.UID,
+				}, v1.EventTypeWarning, "FailedCreate", "Error creating: %v", err)
```
zhuangqh, line 84: we filled the `controller` field in event, so unnecessary to contains context info in error message

incubator/virtualcluster/pkg/syncer/mccontroller/mccontroller.go
```diff
@@ -390,33 +431,43 @@ func (c *MultiClusterController) processNextWorkItem() bool {
...
+		// get queued again until another change happens.
+		c.Queue.Forget(obj)
 		return true
 	}
 
-	metrics.RecordDWSOperationStatus(c.objectKind, req.ClusterName, constants.StatusCodeOK)
+	// rejected by apiserver(maybe rejected by webhook or other admission plugins)
+	// we take a negative attitude on this situation and fail fast.
+	if apierr, ok := err.(apierrors.APIStatus); ok {
+		if code := apierr.Status().Code; code == http.StatusBadRequest || code == http.StatusForbidden {
```
zhuangqh, line 451: we cannot assume webhook error filled the `Reason` field. check code is the most reliable solution

PR #1033, opened 2020-08-20
Update v1alpah2 SubnamespaceAnchor status to camel case

yiqigao217:
Tested by 'make test-conversion'. It failed before and passed after
changing the v1alpha2 status check from 'ok' to 'Ok'.

Part of #868


adrianludwin:
/lgtm

I think this is small enough that a single review is fine, but:
/cc @rjbez17

PR #1038, opened 2020-08-21
Update guide to choose OS for the plugin

tcnksm:
[This](https://github.com/kubernetes-sigs/multi-tenancy/pull/1032) changes the plugin installation URL so update the user guide to support it.

### DIFF
--- a/incubator/hnc/docs/user-guide/how-to.md
+++ b/incubator/hnc/docs/user-guide/how-to.md
@@ -35,7 +35,7 @@ Kubernetes tools such as `kubectl`. However, the `kubectl-hns`
 greatly simplifies several tasks. This guide illustrates both methods, but we
 recommend installing the `kubectl-hns` plugin as well.
 
-To install the plugin (Linux-only), first switch to a directory on your `PATH`
+To install the plugin, first switch to a directory on your `PATH`
 (e.g. `~/bin`), then run the following commands:
 
 ```
@@ -44,8 +44,14 @@ To install the plugin (Linux-only), first switch to a directory on your `PATH`
 # shown below, as of August 2020.
 HNC_VERSION=v0.5.1
 
-# Download the plugin
+# HNC v0.5.1 or earlier: download the plugin (Linux only)
 curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/kubectl-hns -o ./kubectl-hns
+
+# HNC v0.5.2 or later: set the platform and download the plugin. The support platforms are:
+# * linux_amd64 (Linux on Intel/AMD)
+# * darwin_amd64 (MacOS on Intel)
+HNC_PLATFORM=linux_amd64
+curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/kubectl-hns_${HNC_PLATFORM} -o ./kubectl-hns
 
 # Make the plugin executable.
 chmod +x ./kubectl-hns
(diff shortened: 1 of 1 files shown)


incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -44,8 +44,11 @@ To install the plugin (Linux-only), first switch to a directory on your `PATH`
 # shown below, as of August 2020.
 HNC_VERSION=v0.5.1
 
+# Seletct the OS type where you run the plugin. Currently, you can choose "linux" or "darwin" (Mac).
+HNC_OS=linux
```
adrianludwin, line 48: I'd prefer to include the architecture here, especially since "darwin" will very shortly have multiple variants (amd64 vs arm). Maybe `HNC_PLATFORM=linux_amd64`?

Also, we should specify that this is only applies to versions _after_ 0.5.1. For 0.5.1, there's only the single release.

tcnksm: > Maybe HNC_PLATFORM=linux_amd64?

Ok. Let me change to this.

> Also, we should specify that this is only applies to versions after 0.5.1. For 0.5.1, there's only the single release.

Ok. So let me add the separated installation instruction for pre 0.5.1 and after 0.5.1

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -44,9 +44,16 @@ To install the plugin (Linux-only), first switch to a directory on your `PATH`
...
+# Seletct the OS and architecture type (from v0.5.2) where you run the plugin. 
+# Currently, you can choose "linux_amd64" or "darwin_amd64" (Mac).
+HNC_PLATFORM=linux_amd64
+
 # Download the plugin
 curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/kubectl-hns -o ./kubectl-hns
 
+# Download the plugin (from v0.5.2)
+curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/kubectl-hns_${HNC_PLATFORM} -o ./kubectl-hns
+
```
adrianludwin, line 56: Firstly, there's a typo ("Select", not "Seletct").

Secondly, this has a 0.5.2 instruction, then an instruction that doesn't appear to care which version it's for (but actually does), then another 0.5.2 instruction. How about grouping them:

```
# HNC v0.5.1 or earlier: download the plugin (Linux only)
curl ...

# HNC v0.5.2 or later: set the platform and download the plugin. The support platforms are:
# * linux_amd64 (Linux on Intel/AMD)
# * darwin_amd64 (MacOS on Intel)
HNC_PLATFORM=linux_amd64
curl ...
```

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -339,17 +346,25 @@ As of August 2020, the most recent official release is v0.5.1.
 # Set the desired release:
 HNC_VERSION=v0.5.1
 
+# Set the desired OS and architecture (from v0.5.2)
+HNC_PLATFORM=linux_amd64
+
```
adrianludwin, line 351: This isn't needed to install the Docker image

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -339,17 +346,25 @@ As of August 2020, the most recent official release is v0.5.1.
...
 
 ```bash
+# Download the plugin
 curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/kubectl-hns -o ./kubectl-hns
+
+# Download the plugin (from v0.5.2)
+curl -L https://github.com/kubernetes-sigs/multi-tenancy/releases/download/hnc-${HNC_VERSION}/kubectl-hns_$HNC_PLATFORM} -o ./kubectl-hns
+
 chmod +x ./kubectl-hns
 ```
```
adrianludwin, line 369: This section is now sufficiently complicated that I don't think it makes sense to repeat it twice. Instead of a full separate section, can you just add a link to `#use-prepare` to reference the earlier section?

tcnksm: Ok! Make sense. Let me change it.

tcnksm:
Thank you for the review. Updated and squashed.

adrianludwin:
/ok-to-test
/approve
/lgtm

Thanks @tcnksm for this documentation!

Issue #1045, opened 2020-08-22
Change the parent of a subnamespace


johnpemberton:
Hi,

I can envisage a scenario where changing the parent of a subnamespace would be really beneficial.   Is this something that HNC is likely to support?

Scenario:
 - Large organisation, lots of teams, who each look after a number of products
 - Sometimes the ownership of those products changes from one team to another 

Lets start out with a namespace tree like this:
```
- org
   - orders (fullnamespace)
      - order-management (subnamespace)
      - order-notifications (subnamespace)
   - payments (fullnamespace)
      - payment-services (subnamespace)
      - payment-notifications (subnamespace)
```

It's great that the orders and payments teams were in control of creating their own product subnamespaces without ClusterRoles, or putting in requests for sign-off for creating full namespaces from the platform team.   That's one of the things I really like about subnamespaces.

But then the org notices the duplication in that both orders and payments teams have notification services; they decide to create a notifications team, and want them to take ownership of the order-notifications and payment-notifications applications, with the end view of eliminating the duplication.  

The tree wants to look like this:

```
- org
   - orders (fullnamespace)
      - order-management (subnamespace)
   - payments (fullnamespace)
      - payment-services (subnamespace)
   - notifications (fullnamespace)
      - order-notifications (subnamespace)
      - payment-notifications (subnamespace)
```

They create a new notifications namespace, and try to move the order-notifications and payment-notifications under it as child namespaces. 

Unfortunately as the order-notifications and payment-notifications namespaces are subnamespaces, this isn't possible as you [can't change the parent of a subnamespace](https://github.com/kubernetes-sigs/multi-tenancy/blob/hnc-v0.5.1/incubator/hnc/docs/user-guide/concepts.md#subnamespaces-and-full-namespaces).   

If they were created as full namespaces, this would be possible, but that would negate the freedom that our orders and payments teams had in being able to create their own subnamespaces in the first place, and we'd be back to having the administrative burden and delay for approval/sign-off etc for full namespace creation.

I do believe this type of scenario is highly likely to happen, particularly in larger organisations, so I think the ability to reorganise the namespace hierarchy tree, whether involving full namespaces or subnamespaces would be really beneficial.

Also aware of Conway's Law :)

Thanks,
John

johnpemberton:
@adrianludwin  

Just been playing with this again, and I think the correct sequence for moving a subnamespace within the hierarchy tree is:

- Remove the annotation:
```
[...]

PR #1046, opened 2020-08-25
Document how to move a subnamespace

johnpemberton:
Relates to #1045 

This PR adds documentation in the howto on how to move a subnamespace within the hierarchy tree.

### DIFF
--- a/incubator/hnc/docs/user-guide/how-to.md
+++ b/incubator/hnc/docs/user-guide/how-to.md
@@ -13,6 +13,7 @@ This document describes common tasks you might want to accomplish using HNC.
   * [Select namespaces based on their hierarchies](#use-select)
   * [Delete a subnamespace](#use-subns-delete)
   * [Organize full namespaces into a hierarchy](#use-full)
+  * [Move a subnamespace](#move-subnamespace)
 * [Administer HNC](#admin)
   * [Install or upgrade HNC on a cluster](#admin-install)
   * [Uninstall HNC from a cluster](#admin-uninstall)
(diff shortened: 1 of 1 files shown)


rjbez17:
Thanks @johnpemberton! Can you please sign the CLA?

/assign @rjbez17 
/assign @adrianludwin 
/ok-to-test

Overall this looks good but some of the wording may need to change to keep with the same vocabulary we use elsewhere.

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -13,6 +13,7 @@ This document describes common tasks you might want to accomplish using HNC.
   * [Select namespaces based on their hierarchies](#use-select)
   * [Delete a subnamespace](#use-subns-delete)
   * [Organize full namespaces into a hierarchy](#use-full)
+  * [Move a subnamespace](#move-subnamespace)
```
adrianludwin, line 16: Let's move this to the "admin" section since this requires cluster-level privileges (editing a namespace). Also, can we change the name to "Convert subnamespace to full namespaces" and the anchor to `#admin-convert-subns` (all anchors either begin with `use-` or `admin-`).

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -315,6 +316,69 @@ Similarly, if you want to make `ns-bar` a root again, you must be an
 administrator of the root namespace that is an ancestor of `ns-bar`, since the
 admins of that namespace will lose access to `ns-bar` once it becomes a root.
 
+
+<a name="move-subnamespace"/>
+
+### Move a Subnamespace
+
+Changing the parent of a subnamespace is not yet permitted in HNC, though there may be times
+where you want to move a subnamespace from one parent to another.
```
adrianludwin, line 325: Let's use the language of "conversion" here - e.g., how to convert a subns to a full namespace and back.

incubator/hnc/docs/user-guide/how-to.md
```diff
@@ -315,6 +316,69 @@ Similarly, if you want to make `ns-bar` a root again, you must be an
...
+```
+org
+ - ns-team-1 (full namespace)
+   - ns-product-1 (subnamespace)
+   - ns-product-2 (subnamespace)
+ - ns-team-2 (full namespace)
+   - ns-product-3 (subnamespace)
+```
+
+We now want to move `ns-product-2` from `ns-team-1` to `ns-team-2`.
```
adrianludwin, line 338: Can we change these instructions as follows? To turn subns->full ns:

1. Prereqs: you must have "edit" permission on the subns and "delete" permission on anchors in the parent namespace.
2. Remove the `subnamespaceOf` annotation from the child.
3. Confirm that the anchor in the parent is now in the `Conflict` state.
4. Remove the anchor from the parent.

The namespace is now a full namespace. It remains a child of the same parent unless otherwise changed.

It's far more rare that you'll want to turn a full namespace into a subns - the only advantage of this is that users without cluster-level privileges will be able to _delete_ the namespace. But if you want to do this:

1. Ensure the full namespace has the correct parent.
2. Add the `subnamespaceOf` to the child, pointing to the current parent.
3. Create the anchor in the parent (`kubectl hns create child -n parent`). Confirm that the anchor has the `Ok` state.

The child is now a subnamespace, and will be deleted if the anchor is deleted.

You may want to double-check all these steps (and the exact format of the status messages) to ensure they're correct, I'm writing by memory.

incubator/hnc/docs/user-guide/how-to.md
yiqigao217: +1 to move this to `admin` section.

Since `move a subnamespace` contains two parts - `Convert subnamespace to full namespaces` and `Convert full namespaces to subnamespaces`, maybe consider having two separate ones instead of having 1?
(+1 to https://github.com/kubernetes-sigs/multi-tenancy/pull/1046/commits/1ebdd8b44f9172e28caa6bd28aaff77c184ede9d#r476477948)

johnpemberton:
I signed it

adrianludwin:
> I signed it

Hmm, the k8s-ci-robot still isn't seeing it - you followed all the instructions in the "Some checks were not successful" notice at the bottom of this PR?

adrianludwin:
Also after you've addressed the comments, can you please squash all your commits into one? (Or before you've addressed them - your choice). We try to have one commit per PR. Thanks!

johnpemberton:
Missed part of the signing steps.   Should all be good now.  Will hopefully get the changes made to the PR in the next few days.

adrianludwin:
sgtm!

rjbez17:
Any update on this @johnpemberton?

adrianludwin:
I believe this is now covered by #1302 - John, please re-open if you think that doesn't work for you!

Cheers, A

/close

PR #1039, opened 2020-08-21
Add timeoutSeconds patch for webhook configuration

tcnksm:
Context: https://github.com/kubernetes-sigs/multi-tenancy/issues/1023

Add `timeoutSeconds` patch for webhook configuration. 

Currently, `controller-tools` does not support `timetouSeconds` marker ([I'm sending patches for it](https://github.com/kubernetes-sigs/controller-tools/pull/473)). So just add kustomize patches for it. 

## Test

Executing the following command (which is executed in `cloudbuild.yaml`) on my local PC, 

```
cd multi-tenancy/incubator/hnc
mkdir out
cd out
touch kustomization.yaml
kustomize edit add resource ../config/default
kustomize build . -o ./hnc-manager.yaml
```

[This](https://gist.github.com/tcnksm/1bfb91d412c7635315559b3f318806fb) is the output of `hnc-manager.yaml`. You can see `timeoutSecond` is added.

Issue #1023, opened 2020-08-18
HNC: ValidatingWebhook should have a way to avoiding operation on specific namespaces and/or object kinds


tcnksm:
We've been using HNC for our experimental GKE environment. But because HNC validation webhook `objects.hnc.x-k8s.io` was often non-responsive, various system operations such as leader elections failing and this caused cluster itself unstable... 

I think non-responsive itself is an issue (in this time, since it's the experimental environment, we don't have enough monitoring and clue for the reason of non-responsive ... we would like to notify issue it faced when we found the root cause) but, at the same time, [as official documentation says](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#avoiding-operating-on-the-kube-system-namespace), we should avoid using validation webhook on `kube-system` related operation (or some other critical components). It caused the problems which we faced.

To skip the validation webhook, we can use `namespaceSelector`. But to use it we need the label. Since `ValidatingWebhookConfiguration` manifest itself is free to modify,  we can introduce our own defined label for this purpose but I think HNC itself can provide the label for it? For example, how about introducing `admission.hnc.x-k8s.io/ignore` like [gatekeeper project does](https://github.com/open-policy-agent/gatekeeper#exempting-namespaces-from-the-gatekeeper-admission-webhook-using---exempt-namespace-flag)?  How do you think?

adrianludwin:
Let me look into that ASAP. In the meantime, can you describe the
circumstances when the webhook became unresponsive? Was HNC under heavy
load at the time, or was it due to some kind of outage (e.g. the pod being
rescheduled onto another node)?

tcnksm:
> can you describe the circumstances when the webhook became unresponsive? Was HNC under heavy
load at the time, or was it due to some kind of outage

It was running on the experimental cluster which does not run many workloads (the number of namespaces is around 30). Since it's the experimental cluster, we don't receive any production traffic, and we don't often touch it (only when we try something new). I deployed it exactly the same way as the user guide described (no manifest modification). 

Actually, cluster instability was notified by GCP support (as I described above we don't have any monitoring on this cluster). To recover it, I just uninstalled it. So I don't have much information about what happened ... I tried to search log a bit but I cloud not find any clue about this non-responsive state. 

I will try to install it again later and see this happens again. Then I can tell you more information.

adrianludwin:
[...]

PR #1048, opened 2020-08-25
Add timeoutSeconds patch for webhook configuration

adrianludwin:
Cherrypick of #1039 . Tested by adding a 15s sleep to the object validator and confirmed that everything was ridiculously slow without this change and only annoyingly slow with it. I think this is good enough for v0.5.

Addresses #1023

### DIFF
--- a/incubator/hnc/config/webhook/kustomization.yaml
+++ b/incubator/hnc/config/webhook/kustomization.yaml
@@ -4,3 +4,6 @@ resources:
 
 configurations:
 - kustomizeconfig.yaml
+
+patchesStrategicMerge:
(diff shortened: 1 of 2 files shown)


adrianludwin:
No - I modified the webhook itself to say "time.Sleep(15 * time.Second)"
but verified that with this change, all the changes went in after 2s and
not 15s. I removed the Sleep before committing.

yiqigao217:
> No - I modified the webhook itself to say "time.Sleep(15 * time.Second)" but verified that with this change, all the changes went in after 2s and not 15s. I removed the Sleep before committing.

sgtm
/lgtm

Issue #1052, opened 2020-08-25
VC: make test return 0 even if there is FAIL test case


charleszheng44:
@zhuangqh the return code (#?) of the `make test` is 0, even though there exist failed test case (TestPriorityClassPatrol/pPriorityClass_not_found,_vPriorityClass_exists). Is this an expected behavior? Some CI relies on this return code.

PR #1053, opened 2020-08-26
[VC]: 🐜 Fix make test not catching and fix testsuite

christopherhein:
The way that `make test` was written it would not catch failures with the sweet, adding `|| exit 1` will cause failures to stop the make execution. This was uncovered by seeing that the priorityclass syncer was failing because of the group change and the apigroup being hardcoded.

This fixes https://github.com/kubernetes-sigs/multi-tenancy/issues/1052


Fei-Guo:
Nice change.

/lgtm
/approve

PR #1057, opened 2020-08-27
Improve e2e test pass rate and infra

adrianludwin:
I've noticed that a bunch of e2e tests are failing on master. This
change makes a few improvements:

* Fix a functional bug in the plugin that confused the text displayed
  _about_ a namespace with the actual namespace name.
* testutils.TryRun now prints out more informative log messages and
  returns a more informative error.
* Increased the default timeout from 1s to 2s in issues_test
* Update demo_test to include the now-expected output about
  subnamespaces in `kubectl hns tree`.

Tested: fewer tests fail now than before.

### DIFF
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -2,6 +2,7 @@ package testutils
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -76,7 +77,14 @@ func MustNotRun(cmdln ...string) {
 
 func TryRun(cmdln ...string) error {
 	stdout, err := RunCommand(cmdln...)
-	GinkgoT().Log("Output: ", stdout)
+	if err != nil {
+		// Add stdout to the error, since it's the error that gets displayed when a test fails and it
+		// can be very hard looking at the log to see which failures are intended and which are not.
+		err = fmt.Errorf("Error: %s\nOutput: %s", err, stdout)
+		GinkgoT().Log("Output (failed): ", err)
+	} else {
+		GinkgoT().Log("Output (passed): ", stdout)
+	}
 	return err
 }
 
@@ -95,8 +103,8 @@ func RunShouldContainMultiple(substrs []string, seconds float64, cmdln ...string
 		if err != nil {
 			return err
 		}
-		if len(missing) > 0 {
-			return errors.New("Missing the expected strings: " + strings.Join(missing, ", "))
+		if missing != "" {
+			return errors.New(missing)
 		}
 		return nil
 	}, seconds).Should(Succeed())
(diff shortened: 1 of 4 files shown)


incubator/hnc/test/e2e/demo_test.go
```diff
@@ -183,9 +183,9 @@ spec:
 		MustRun("kubectl hns create", nsService3, "-n", nsTeamA)
 		expected := "" + // empty string make go fmt happy
 			nsTeamA + "\n" +
-			"├── " + nsService1 + "\n" +
-			"├── " + nsService2 + "\n" +
-			"└── " + nsService3
+			"├── [s] " + nsService1 + "\n" +
+			"├── [s] " + nsService2 + "\n" +
+			"└── [s] " + nsService3
```
yiqigao217, line 188: Oh, this change should be merged in the same kubeclt-hns plugin change, right?

Was this the main cause of the failed e2e test at HEAD?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -269,11 +275,11 @@ var _ = Describe("Issues that require repairing HNC", func() {
...
 		MustRun("kubectl hns create", nsChild, "-n", nsParent2)
 		// Subnamespace child should be created and have parent as the 'subnamespaceOf' annoation value:
-		RunShouldContain("subnamespaceOf: "+nsParent, 1, "kubectl get ns", nsChild, "-o yaml")
+		RunShouldContain("subnamespaceOf: "+nsParent, defTimeout, "kubectl get ns", nsChild, "-o yaml")
 		// Creating a 'test-secret' in the subnamespace child
 		MustRun("kubectl create secret generic test-secret --from-literal=key=value -n", nsChild)
-		// subns (anchor) child in parent2 should have 'status: conflict' because it's a bad anchor:
-		RunShouldContain("status: conflict", 1, "kubectl get subns", nsChild, "-n", nsParent2, "-o yaml")
+		// subns (anchor) child in parent2 should have 'status: Conflict' because it's a bad anchor:
+		RunShouldContain("status: Conflict", defTimeout, "kubectl get subns", nsChild, "-n", nsParent2, "-o yaml")
```
yiqigao217, line 282: Similar to the `[s]` right? I will remember to update e2e test too when making future API conversion changes.

incubator/hnc/test/e2e/demo_test.go
adrianludwin: It's one of the causes but there seem to be many others too.

incubator/hnc/test/e2e/issues_test.go
adrianludwin: We should eventually make it part of our presubmits

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -119,20 +127,32 @@ func RunErrorShouldContainMultiple(substrs []string, seconds float64, cmdln ...s
...
 		return missAny(substrs, stdout), err
 }
 
-func missAny(substrs []string, teststring string) []string {
-	var missing []string 
+// If any of the substrs are missing from teststring, returns a string of the form:
+//   Missing: <string1>, <string2>, ...
+//   Got: teststring
+// Otherwise returns the empty string.
+func missAny(substrs []string, teststring string) string {
```
GinnyJI, line 140: Don't have an objection on it, but why do we prefer return a concatenated string than slice of strings?

adrianludwin: Good question. Usually, we should return structured data, but in this case, all the callers were doing exactly the same thing - displaying all the missing elements. I wanted to modify them all in the same way - not _just_ display what we _expected_, but also what we _got_. So it made the most sense to change the meaning of this function to return a single human-friendly _message_ to display, which now includes what we actually got.

If we were to find that in some cases we wanted the structured data (e.g. "ignore this missing string if that happens") then we'd have to break this function into two. But given how we were using it, returning a human-readable string made more sense.

PR #1065, opened 2020-08-31
Add KUBERNETES_SERVICE_HOST env variable to synced pods

salaxander:
Mutates synced pods with:

```
KUBERNETES_SERVICE_HOST=kubernetes
```

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/mutate.go
@@ -275,6 +275,7 @@ func getServiceEnvVarMap(ns, cluster string, enableServiceLinks *bool, services 
 	for _, e := range envvars.FromServices(mappedServices) {
 		m[e.Name] = e.Value
 	}
+	// ensure that tenant pods can reach the API server
 	m["KUBERNETES_SERVICE_HOST"] = "kubernetes"
 	return apiServerService, m
 }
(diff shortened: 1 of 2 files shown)


Fei-Guo:
I suggest you change here 
https://github.com/kubernetes-sigs/multi-tenancy/blob/23e6e2489656e0888094810314ae382c5fe0dfcd/incubator/virtualcluster/pkg/syncer/conversion/mutate.go#L275 , which may lead to the least code change. 

Also please add a comment to explain why this special conversion is needed in the code.

salaxander:
> I suggest you change here
> 
> https://github.com/kubernetes-sigs/multi-tenancy/blob/23e6e2489656e0888094810314ae382c5fe0dfcd/incubator/virtualcluster/pkg/syncer/conversion/mutate.go#L275
> 
> , which may lead to the least code change.
> Also please add a comment to explain why this special conversion is needed in the code.
> Btw, please change the copyright year if you touch any old files.

Is this loop just targeting pods that are exposed via services? I think we would want this change to apply to containers that are also not exposed. I'll add a comment as well :)

Fei-Guo:
> > I suggest you change here
> > https://github.com/kubernetes-sigs/multi-tenancy/blob/23e6e2489656e0888094810314ae382c5fe0dfcd/incubator/virtualcluster/pkg/syncer/conversion/mutate.go#L275
> > 
> > , which may lead to the least code change.
> > Also please add a comment to explain why this special conversion is needed in the code.
> > Btw, please change the copyright year if you touch any old files.
> 
> Is this loop just targeting pods that are exposed via services? I think we would want this change to apply to containers that are also not exposed. I'll add a comment as well :)

I believe it targets all Pods since every Pod needs to have Kubernetes service. Correct me if I am wrong.

Fei-Guo:
> @Fei-Guo pushed an update based on your comment :) Tests are still passing

Excellent, a comment would be preferred for the change.

salaxander:
@Fei-Guo added comment, tests still ok

PR #1068, opened 2020-09-01
Fix some flaky testes

GinnyJI:
Extend the timeout limit for some operations to make them more stable.
Checks that the previous operation finishes before proceeding to the
next one, otherwise some of them will be flaky. Warns the user that the
network policy has to be enabled for the test to pass.

Tested: make test-e2e

### DIFF
--- a/incubator/hnc/test/e2e/demo_test.go
+++ b/incubator/hnc/test/e2e/demo_test.go
@@ -38,7 +38,7 @@ var _ = Describe("Demo", func() {
 		MustRun("kubectl -n", nsOrg, "create role", nsOrg+"-sre", "--verb=update --resource=deployments")
 		MustRun("kubectl -n", nsOrg, "create rolebinding", nsOrg+"-sres", "--role", nsOrg+"-sre", "--serviceaccount="+nsOrg+":default")
 		// none of this affects service-1
-		RunShouldContain("No resources found in "+nsService1, 1, "kubectl -n", nsService1, "get rolebindings")
+		RunShouldContain("No resources found in "+nsService1, defTimeout, "kubectl -n", nsService1, "get rolebindings")
 
 		// make acme-org the parent of team-a, and team-a the parent of service-1.
 		MustRun("kubectl hns set", nsTeamA, "--parent", nsOrg)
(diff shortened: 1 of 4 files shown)


GinnyJI:
> I think the test fails only when networkPolicy is not enabled in the cluster. I tested late last week that it failed on my existing GKE cluster without networkPolicy enabled. After I upgraded the cluster to enable networkPolicy, this test passed as expected.

I think you are correct. Updated the pr

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -186,7 +188,8 @@ spec:
 			"├── [s] " + nsService1 + "\n" +
 			"├── [s] " + nsService2 + "\n" +
 			"└── [s] " + nsService3
-		RunShouldContain(expected, 2, "kubectl hns tree", nsTeamA)
+		// The subnamespaces takes a bit of time to show up
+		RunShouldContain(expected, 10, "kubectl hns tree", nsTeamA)
```
yiqigao217, line 192: Does this take as much as 10 seconds? I know it's in an `Eventually` that if it passes, it won't take 10 seconds, so basically having a larger number could improve the pass rate for our e2e tests. Just curious that this case could fail with a smaller number than 10?

Nit: can we make the 10 and the 5 seconds below into constants with easy to understand name for different sync time instead of having numbers all over? (A suggestion from Adrian on my conversion test too).

GinnyJI: Done. And 5s actually has a good pass rate too, I tried for 5 times and they all pass, so I changed the time limit to 5s

incubator/hnc/test/e2e/e2e_test.go
```diff
@@ -9,7 +9,15 @@ import (
...
 
-const namspacePrefix = "e2e-test-"
+const (
+	namspacePrefix = "e2e-test-"
+	// A 1s timeout was too short; 2s *seems* stable and also matches the Ginkgo default
+	defTimeout = 2
+	// For some operations that takes longer than 2s to finish, 5s seems to be a more stable time choice
+	longerTimeout = 5
+	// During experiments we find that there are a few cases where we need 10s timeout
+	tenSecTimeout = 10
```
yiqigao217, line 19: Actually I meant for something descriptive about the behavior or the reason why it takes longer for certain validation in the caller, e.g. `RunShouldContain()` with `cleanupTimeout`. I reviewed all the cases for `longerTimeout` and `tenSecTimeout`, seems they can map perfectly to `propagationTimeout = 5` and `cleanupTimeout = 10`. WDYT?

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -106,9 +108,9 @@ var _ = Describe("Demo", func() {
 		// at least 20 seconds is needed here from experiments 
 		RunShouldContain("Welcome to nginx!", 20, 
 			"kubectl run client -n", nsService1, clientArgs, cmdln)
-		RunShouldContain("Welcome to nginx!", 10, 
+		RunShouldContain("Welcome to nginx!", tenSecTimeout, 
 			"kubectl run client -n", nsTeamA, clientArgs, cmdln)
-		RunShouldContain("Welcome to nginx!", 10, 
+		RunShouldContain("Welcome to nginx!", tenSecTimeout,
```
yiqigao217, line 113: BTW here after the first 20s, maybe the other two can be just `defTimeout` instead of 10s?

incubator/hnc/test/e2e/e2e_test.go
GinnyJI: Agree. Changed the names and comments

incubator/hnc/test/e2e/demo_test.go
GinnyJI: With `defTimeout` this test fails half of the time actually

yiqigao217:
/lgtm

/assign @rjbez17 
/assign @adrianludwin 
To approve.

Issue #1035, opened 2020-08-20
HNC: 'allowCascadingDelete' to 'allowCascadingDeletion' field change in CRD conversion


yiqigao217:
Part of #868 

I found that if we replace `allowCascadingDelete` field with `allowCascadingDeletion` field directly in v1alpha2 schema, when passing the value of `v1alpha1.allowCascadingDelete` to `v1alpha2.allowCascadingDeletion` in the CRD conversion, the value is missing.

If I keep both `allowCascadingDelete` and `allowCascadingDeletion` fields in v1alpha2 schema, the value can be passed.

I suspect the CRD schema change (with v1alpha2 as storage version) drops the non-existing `allowCascadingDelete` field in etcd immediately before the CRD conversion. That's why only replacing the field directly failed the value passing.

WDYT @adrianludwin

adrianludwin:
As in, even _inside_ the conversion webhook, you can't see the value of `v1a1.aCD`? That's not how storage versions are supposed to work AIUI.

yiqigao217:
> As in, even _inside_ the conversion webhook, you can't see the value of `v1a1.aCD`? That's not how storage versions are supposed to work AIUI.

Yes, I manually tested replacing `dst.Spec.AllowCascadingDeletion = src.Spec.AllowCascadingDelete` with `dst.Spec.AllowCascadingDeletion = true` in the conversion webhook. After conversion the v1alpha2 HC has `spec.allowCascadingDeletion` in the `k get` yaml.

It means the new field can **get** value from conversion webhook. However, if passing `src.Spec.AllowCascadingDelete`, I got empty spec in v1alpha2 HC after conversion - `spec:{}`.

yiqigao217:
Update:
Setting `CRD_OPTIONS ?= "crd:trivialVersions=false"` does work that because it moves validations from `top-level` to `per-version` so that v1alpha1 will have `allowCascadingDelete` field and v1alpha2 has `allowCascadingDeleting` field in each validation.

The only issue is that I need to remove the [work-around patch](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/config/crd/patches/webhook_in_hierarchyconfigurations.yaml#L19-L27) for unserving v1alpha1, because otherwise it overwrites the validations per-version. See https://github.com/kubernetes-sigs/controller-tools/issues/462

I will leave v1alpha1 still serving for now to wait for `unservedversion` marker issue.

[...]

PR #1037, opened 2020-08-20
Update v1alpha2 HC schema with allowCascadingDeletion

yiqigao217:
Use `allowCascadingDeletion` instead of `allowCascadingDelete` in
v1alpha2 HC schema. Update conversion test, integration test and e2e
tests.

Add a skipped (`PIt`) test case for #1155 . This test failed on both
master branch and this change, so it's not caused by this change. Added
the test so that the later fix can use `allowCascadingDeletion` directly.

Tested by `make test-conversion`, `make test` and `make test-e2e` (with
`FIt` on related tests only).

Fixes #1035 
Part of #868

PR #1074, opened 2020-09-01
Set crd:trivialVersions=false and add patch per-version

yiqigao217:
Set `crd:trivialVersions=false` to replace top-level CRD schema with
per-version schema.

Use `kubebuilder:unservedversion` marker in v1alpha1 since it's working
after we switched to controller-gen v0.2.8 from v0.2.5.

There's no way to have metadata validation from kubebuilder marker so we
had to use kustomize. However, after switching to per-version schema,
there's no way to kustomize it per version. Thus, write script to
automatically patch the singleton validation per version.

With kubernetes-sigs/controller-runtime#1154
bug, integration test was broken with any versions unserved. Thus, write
another script to have a set of CRD manifests generated specifically for
the integration test. It still tests v1alpha2 since we still imported it.

Include the `hack/` scripts in `manifests` and `test` target in Makefile.

Tested with `make test` and `make test-conversion` and manual testing on
a GKE cluster.

Part of #1035


yiqigao217:
/assign @rjbez17 
/assign @adrianludwin 
/cc @GinnyJI 
Will hold #1037 until this is merged first.
See https://github.com/kubernetes-sigs/multi-tenancy/issues/1035#issuecomment-684879911 for why we make this change.

incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -112,7 +112,7 @@ type HierarchyConfigurationSpec struct {
 	Parent string `json:"parent,omitempty"`
 
 	// AllowCascadingDelete indicates if the subnamespaces of this namespace are allowed to cascading
-	// delete.
+	// delete. TODO update it to allowCascadingDeletion.
```
yiqigao217, line 115: FYI This is added just for controller-gen to generate a `per-version` schema for this HC CRD. Since all fields were the same, it still generated the `top-level` schema without this minor difference.

incubator/hnc/config/crd/patches/singleton-validation_unserve-v1alpha1_hierarchyconfiguration.yaml
```diff
@@ -0,0 +1,165 @@
...
+# that the entire "spec.versions" field can be patched. Here are the changes:
+# 1) Add singleton metadata validation to each version, e.g.
+#            metadata:
+#              properties:
+#                name:
+#                  type: string
+#                  enum:
+#                    - hierarchy
+#              type: object
+# 2) Set v1alpha1 as unserved, e.g. "served: false", in v1alpha1 schema.
```
yiqigao217, line 15: Mostly the reviewers just need to review this part and the manual update part below.

rjbez17:
This lgtm but I'd prefer @adrianludwin take a look.

incubator/hnc/config/crd/patches/singleton-validation_unserve-v1alpha1_hierarchyconfiguration.yaml
```diff
@@ -0,0 +1,165 @@
+# Please remember to re-copy CRD from bases/ if there's a CRD change. See
+# https://github.com/kubernetes-sigs/multi-tenancy/issues/1035 for detail.
```
adrianludwin, line 2: Automate this by adding a script in `hack/` and calling it as a part of `make generate`?

yiqigao217: Good idea, will work on that!

rjbez17:
Any update on this?

yiqigao217:
> Any update on this?

Ah yes, sorry for the delay. I will be finishing #1076 first and come back to this. Two more changes I will need to make here:
1) The `unservedVersion` marker should be working after #1093 so I  will update that. 
2) As for the singleton enum validation, it still needs manual patch, since kustomize cannot add patch to a specific item in the list, e.g. v1alpha1 or v1alpha2 in `versions` list after they are split with `crd:trivialVersions=false`. I will add a script to auto-generate it.

incubator/hnc/hack/ensure-all-served-for-envtest.sh
```diff
@@ -0,0 +1,21 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+# Generate CRD manifests in config/crd/envtest. Replace "served: false" with
+# "served: true" in the CRD manifests for envtest integration tests. See bug -
+# https://github.com/kubernetes-sigs/controller-runtime/issues/1154
+cp -R config/crd/bases/. config/crd/envtest/
+sed -i 's/served: false/served: true/' config/crd/envtest/hnc.x-k8s.io_hierarchyconfigurations.yaml
+sed -i 's/served: false/served: true/' config/crd/envtest/hnc.x-k8s.io_subnamespaceanchors.yaml
+sed -i 's/served: false/served: true/' config/crd/envtest/hnc.x-k8s.io_hncconfigurations.yaml
```
adrianludwin: Add newline

incubator/hnc/config/crd/patches/config_singleton_enum.yaml
```diff
@@ -0,0 +1,5 @@
+            properties:
```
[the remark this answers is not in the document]
adrianludwin: If these are only used by the `hack` scripts, I'd prefer to put this in `hack` as well, or maybe even as a heredoc in the script itself, rather than having a bunch of extra files floating around that don't make a lot of sense in isolation.

yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

This PR is ready for another review. The updates are listed in the commit message (basically two hack/ scripts to review).

incubator/hnc/hack/ensure-all-served-for-envtest.sh
yiqigao217: done.

incubator/hnc/config/crd/envtest/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -0,0 +1,169 @@
+
```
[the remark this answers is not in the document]
yiqigao217: sgtm, will do now.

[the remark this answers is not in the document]
yiqigao217: Done, moved to `manifests/envtest`.

incubator/hnc/config/crd/patches/config_singleton_enum.yaml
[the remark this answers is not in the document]
yiqigao217: Done, added a comment and moved everything into `hack/crd_patches`.

incubator/hnc/Makefile
```diff
@@ -61,7 +61,7 @@ KUSTOMIZE ?= ${CURDIR}/hack/kustomize-3.8.1
 CONTROLLER_GEN ?= ${CURDIR}/bin/controller-gen
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+CRD_OPTIONS ?= "crd:trivialVersions=false"
```
adrianludwin, line 64: Fix comment. Or maybe just delete these two lines completely - what's the default?

incubator/hnc/hack/crd_patches/ensure-all-served-for-envtest.sh
```diff
@@ -0,0 +1,21 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+# Generate CRD manifests in config/crd/envtest. Replace "served: false" with
+# "served: true" in the CRD manifests for envtest integration tests. See bug -
+# https://github.com/kubernetes-sigs/controller-runtime/issues/1154
```
adrianludwin, line 17: We should file our own issue for this to track when it's fixed upstream, and reference it here

incubator/hnc/hack/crd_patches/singleton-enum-patch.sh
```diff
@@ -0,0 +1,33 @@
...
+# 2) hncconfig singleton:
+#            metadata:
+#              properties:
+#                name:
+#                  type: string
+#                  enum:
+#                    - config
+#              type: object
+sed -i -e "/ metadata:/ r hack/crd_patches/hierarchy_singleton_enum.yaml" config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
+sed -i -e "/ metadata:/ r hack/crd_patches/config_singleton_enum.yaml" config/crd/bases/hnc.x-k8s.io_hncconfigurations.yaml
```
adrianludwin, line 33: For those of us who don't know `sed` well 🙋 can you comment exactly what this does? I take it we assume there's only one line called `' metadata:'` in each file? If so, that seems like a reasonable assumption to document here as well.

incubator/hnc/internal/reconcilers/suite_test.go
```diff
@@ -66,7 +66,7 @@ var _ = BeforeSuite(func(done Done) {
 
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
-		CRDDirectoryPaths: []string{filepath.Join("..", "..", "config", "crd", "bases")},
+		CRDDirectoryPaths: []string{filepath.Join("..", "..", "manifests", "envtest")},
```
adrianludwin, line 69: Maybe worth a comment here about what it _should_ be after we fix the bug and why it's set to the value that it is, a link to the bug, etc etc

adrianludwin (commented):
lgtm after some final nits

/assign @rjbez17

incubator/hnc/hack/crd_patches/singleton-enum-patch.sh
yiqigao217: There are 3 lines. Commented why now, PTAL.

incubator/hnc/hack/crd_patches/ensure-all-served-for-envtest.sh
yiqigao217: Done.

incubator/hnc/Makefile
yiqigao217: The default is only top-level schema no per-version schema. Per-version is introduced in k8s 1.13 so the default is to be compatible with older versions. Updated in the comment and changed the code to `CRD_OPTIONS ?= crd` according to https://book.kubebuilder.io/reference/generating-crd.html#multiple-versions

incubator/hnc/Makefile
```diff
@@ -60,8 +60,12 @@ CURDIR = $(shell pwd)
...
 CONTROLLER_GEN ?= ${CURDIR}/bin/controller-gen
 
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+# Enable CRD per-version validation. As of Kubernetes 1.13, we can have multiple
+# versions of the Kind defined in the CRD, and use a webhook to convert between
+# them. By default, KubeBuilder disables generating different validation for
+# different versions of the Kind in the CRD, to be compatible with older
+# Kubernetes versions.
+CRD_OPTIONS ?= crd
```
adrianludwin, line 68: If the default is trivial, don't we need "trivialVersions=false" here?

incubator/hnc/hack/crd_patches/singleton-enum-patch.sh
adrianludwin: Glad I asked :) Thanks!

adrianludwin (commented):
One last question but otherwise lgtm

incubator/hnc/Makefile
yiqigao217: According to https://book.kubebuilder.io/reference/generating-crd.html#multiple-versions, it should be changed to `CRD_OPTIONS ?= crd`. I tested both and `CRD_OPTIONS ?= crd` and `CRD_OPTIONS ?= "crd:trivialVersions=false` are the same.

adrianludwin: Hmm, I find the second version easier to read - it's unclear to a casual reader what "CRD options is 'crd'" means. Maybe use the more explicit version or explain that this funny syntax means "force all options to false" or something, and link to the book?

yiqigao217: SG, updated with `CRD_OPTIONS ?= "crd:trivialVersions=false` and tested.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

Issue #1076, opened 2020-09-02
HNC: Handle object propagation conflict


yiqigao217:
This issue is raised on Slack. See http://bit.ly/hnc-propagation-conflict for detailed design.

Consider an example tree structure with one source object demo role in namespace tenant1:
```
org1
├── tenant1 (with a demo role)
│   └── space1 (with a propagated demo role from tenant1) 
└── tenant2
   └── space2
```
Now let’s create a `demo` role in `org1` namespace: currently, HNC propagates `demo` from `org1` to `space1`, `tenant2` and `space2`, but it skips propagation to namespace `tenant1` since a conflicting source object already exists.

For now, we will fix this issue by adding new webhook rules to prevent it (creating conflicting source). If the webhook is bypassed, HNC will overwrite any conflicts in the descendants, e.g. overwriting `demo` in `tenant1` with the one in `org1`.

adrianludwin:
@yiqigao217 I think this is finished now, correct?

yiqigao217:
> @yiqigao217 I think this is finished now, correct?

I would say yes, since #1120 is a rare corner case that we thought it was not a rush and could be done later, so we may close this for now.

PR #1077, opened 2020-09-02
Update object reconciler to overwrite conflicting source

yiqigao217:
For source objects with a conflicting source in the ancestors, treat
them as propagated objects and overwrite them with the source in the
ancestor. See http://bit.ly/hnc-propagation-conflict

Tested by `make test` with a new integration test case.

Part of #1076


yiqigao217:
/hold

For the upcoming webhook change.

adrianludwin:
/lgtm
/approve

Why does this need to wait for the webhook?

yiqigao217:
> Why does this need to wait for the webhook?

Ah right, the change in the reconciler can be merged without webhooks are in place on master.
/hold cancel

PR #1080, opened 2020-09-03
Add post-submit github test file

GinnyJI:
Test:
```
docker run -it -v /var/run/docker.sock:/var/run/docker.sock --entrypoint /bin/bash gcr.io/k8s-testimages/kubekins-e2e:latest-master
git clone https://github.com/GinnyJI/multi-tenancy.git
cd multi-tenancy/
git checkout post-submit
incubator/hnc/hack/post-submit-test.sh
```
This test does not fully function, it still requires
dind(docker-in-docker) to be setup

### DIFF
--- a/incubator/hnc/hack/post-submit-test.sh
+++ b/incubator/hnc/hack/post-submit-test.sh
@@ -1,0 +1,23 @@
+# This file is run by Prow during all postsubmits
+export PATH=$(go env GOPATH)/bin:$PATH
+mkdir -p $(go env GOPATH)/bin
+
+echo "Installing 'kubebuilder' to include the Ginkgo test suite requirements"
+kb=2.3.1
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb}/kubebuilder_${kb}_linux_amd64.tar.gz
+tar -zxvf kubebuilder_${kb}_linux_amd64.tar.gz
+mv kubebuilder_${kb}_linux_amd64 /usr/local/kubebuilder
+
+# install and setup kind according to https://kind.sigs.k8s.io/docs/user/quick-start/
+echo "Setting up Kind"
+curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.8.1/kind-linux-amd64
+chmod +x ./kind
+mv ./kind /usr/local/bin/kind
+kind create cluster
+
+hack_dir=$(dirname ${BASH_SOURCE})
+
+echo "Running 'make test-e2e'"
+cd "$hack_dir/.."
+make kind-deploy
+make test-e2e


GinnyJI:
/assign @yiqigao217 
/assign @adrianchung 
Note that this has not been tested. I don't know how to trigger a post submit test manually

incubator/hnc/hack/post-submit-test.sh
```diff
@@ -0,0 +1,15 @@
...
+kb=2.3.1
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb}/kubebuilder_${kb}_linux_amd64.tar.gz
+tar -zxvf kubebuilder_${kb}_linux_amd64.tar.gz
+mv kubebuilder_${kb}_linux_amd64 /usr/local/kubebuilder
+
+hack_dir=$(dirname ${BASH_SOURCE})
+
+echo "Running 'make test-e2e'"
+cd "$hack_dir/.."
+make test-e2e
```
adrianludwin, line 15: Add newline

adrianludwin (commented):
> Note that this has not been tested. I don't know how to trigger a post submit test manually

It's probably a good idea to try to test this in the Go container referenced in https://github.com/kubernetes/test-infra/pull/19128/files. E.g.:

```
docker run -it golang:1.14 /bin/bash # or whatever the exact command is
# Inside the container:
git clone https://sigs.k8s.io/multi-tenancy # path_alias from the test-infra PR
./incubator/hnc/hack/ci-test.sh # command from the test-infra PR
```

Not sure if this is exactly right but it should be something like this.

incubator/hnc/hack/post-submit-test.sh
```diff
@@ -0,0 +1,15 @@
+# This file is run by Prow during all postsuubmits
```
yiqigao217, line 1: typo `uu` in postsubmits

incubator/hnc/hack/post-submit-test.sh
GinnyJI: done

incubator/hnc/hack/post-submit-test.sh
GinnyJI: done

yiqigao217 (commented):
Is it adding the docker and Kind installation in this new patch? Maybe add a commit message to give more info?

yiqigao217:
/retest

FYI I'm looking this flaky test.

GinnyJI:
> Is it adding the docker and Kind installation in this new patch? Maybe add a commit message to give more info?

We are not trying to install docker here at this stage. We assume that the test environment has docker installed. And yes we are installing Kind.

rjbez17:
lgtm after verifying works in the go container per @adrianludwin request. Has that been done?

adrianludwin:
We had a chat with the Kind developers and there's quite a bit more work here needed (e.g. to get Docker-in-Docker working). We should probably close this PR since whatever we do will be very different from what we originally thought we needed.

/close

PR #1082, opened 2020-09-05
add rest config into SyncerConfiguration

weiling61:
The purpose of this PR is to make the syncer framework more generic so it can be reused by users that need to support their own CRDs. This change passes rest config of super master in SyncerConfiguration struct. So users can construct/handle their CRD client to perform the DWS/UWS sync.

### DIFF
--- a/incubator/virtualcluster/cmd/syncer/app/options/options.go
+++ b/incubator/virtualcluster/cmd/syncer/app/options/options.go
@@ -142,7 +142,7 @@ func (o *ResourceSyncerOptions) Config() (*syncerappconfig.Config, error) {
 	c.ComponentConfig = o.ComponentConfig
 
 	// Prepare kube clients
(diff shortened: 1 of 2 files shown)


Fei-Guo:
@weiling61 Can you elaborate how this change helps to achieve your goal described in the change description?
"So users can construct/handle their CRD client to perform the DWS/UWS sync"  I don't quite understand this part.

weiling61:
@Fei-Guo thanks for the question. System level CRDs exist in super cluster. Tenant in Virtual cluster or Syncer inevitably needs to deal with these CRDs, for example, 1) syncer needs to check status of a security CRD in super cluster, or 2) some environment CRDs needs to be synced between tenant cluster and super cluster.  A separate customized syncer is required to handle these CRD cases. 
When I tried to build the customized syncer, I found that 80% of current syncer code can be reused, but still missing some pieces that cannot be directly imported. One critical piece is the super cluster restful config, which is used to construct CRD client for the customized syncer. If the rest config can be passed to /resources as this PR does, packages, such as 'app', 'options', and 'syncer' in current syncer can be directly imported in the customized syncer, which reduces quite a lot maintenance efforts.

Fei-Guo:
Thanks for the explanation.

/lgtm
/approve

PR #1085, opened 2020-09-08
Update object webhook to deny creating conflict object

### LINKED ISSUE #1079: HNC: Update object validator to avoid overwriting source
Part of #1076 , see http://bit.ly/hnc-propagation-conflict.

The current object validator already denies any requests on propagated objects, but it allows any requests on source objects. We will add new rules to restrict source objects to prevent this conflict.

  |  SyncMode |   _Operation*_| Condition |  Allow/Deny |
|---|---|---|---|
| `Propagate`  | Create |  Conflicting source in descendant | **Deny** |
| `Propagate`  | Create |  Conflicting source in ancestor |  Allow (apiserver will reject it since a propagated resource with that name already exist in the namespace) |
|  `Propagate` | Create | Any other | Allow  |
|  `Remove` | Create | Conflicting source in descendant | Allow (because `Remove` mode won't delete the conflicting **source** in the descendant) |
|  `Remove` | Create | Conflicting source in ancestor | Allow (because `Remove` mode won't delete this **source**)  |
|  `Remove` | Create | Any other | Allow  |
|  `Ignore` |  Create  | Any | Allow  |

_* We only discuss `Create` operation here because `Update` or `Delete` a source won't introduce this conflict._

yiqigao217:
Update object webhook to deny creating an object if an object with the
same name and type already exits in the descendants, and the type has
'Propagate' syncMode. Add new a new test case.

Tested manually on GKE cluster and by 'make test'.

The webhook output is like this:
```
Error from server (Conflict): admission webhook "objects.hnc.x-k8s.io" denied the request: Cannot create /v1, Kind=Secret 'my-creds' in namespace 'acme-org' because it would overwrite objects in the following descendant namespace(s): [team-a team-b]. Choose a different name for the object, or remove the conflicting objects from these namespaces.
```

Fixes #1079 
Part of #1076

### DIFF
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -2,6 +2,7 @@ package validators
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	"github.com/go-logr/logr"
@@ -106,8 +107,13 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	oldSource, oldInherited := metadata.GetLabel(oldInst, api.LabelInheritedFrom)
 	newSource, newInherited := metadata.GetLabel(inst, api.LabelInheritedFrom)
 
-	// If the object wasn't and isn't inherited, it's none of our business.
+	// If the object wasn't and isn't inherited, we will check to see if the
+	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
+		if yes, cnses := o.hasConflict(inst); yes {
+			msg := fmt.Sprintf("Cannot create %s '%s' in namespace '%s' because it would overwrite objects in the following descendant namespace(s): %v. Choose a different name for the object, or remove the conflicting objects from these namespaces.", inst.GroupVersionKind(), inst.GetName(), inst.GetNamespace(), cnses)
+			return deny(metav1.StatusReasonConflict, msg)
+		}
 		return allow("source object")
 	}
 
@@ -144,6 +150,33 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	// If you get here, it means the webhook config is misconfigured to include an operation that we
 	// actually don't support.
 	return deny(metav1.StatusReasonInternalError, "unknown operation: "+string(op))
+}
+
+// hasConflict checks if there's any conflicting objects in the descendants. Returns
+// true and a list of conflicting descendants, if yes.
+func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
+	o.Forest.Lock()
+	defer o.Forest.Unlock()
+
+	// If the instance is empty (for a delete operation) or it's not namespace-scoped,
+	// there must be no conflict.
+	if inst == nil || inst.GetNamespace() == "" {
+		return false, nil
+	}
+
+	nm := inst.GetName()
+	gvk := inst.GroupVersionKind()
+	descs := o.Forest.Get(inst.GetNamespace()).DescendantNames()
+	conflicts := []string{}
+
+	// Get a list of conflicting descendants if there's any.
+	for _, desc := range descs {
+		if o.Forest.Get(desc).HasOriginalObject(gvk, nm) {
+			conflicts = append(conflicts, desc)
+		}
+	}
+
+	return len(conflicts) != 0, conflicts
 }
 
 func (o *Object) InjectClient(c client.Client) error {
--- a/incubator/hnc/internal/validators/object_test.go
+++ b/incubator/hnc/internal/validators/object_test.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/foresttest"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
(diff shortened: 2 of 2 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

First one out of three webhooks.

adrianludwin:
Can you please update the commit comment with an example of what the message looks like if there are one or more conflicts?

incubator/hnc/internal/validators/object.go
```diff
@@ -108,6 +109,10 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 
 	// If the object wasn't and isn't inherited, it's none of our business.
```
adrianludwin, line 110: This comment is now out of date.

incubator/hnc/internal/validators/object.go
```diff
@@ -108,6 +109,10 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
 	if !oldInherited && !newInherited {
+		if yes, cnses := o.hasConflict(inst); yes {
+			msg := fmt.Sprintf("Cannot create and propagate this object because one or more objects with the same name already exist in the descendants: %v. Please resolve the conflict first.", cnses)
```
adrianludwin, line 113: "This object" might be ambiguous, e.g. if the kubectl command is running in a script and this log message is all the user has to go on, or if the user didn't specify the namespace and is using the one in their kubectx. Maybe include a bunch more information, e.g.:

```
Cannot create <gvk> <name> in namespace <ns> because it would overwrite objects in the following descendant namespace(s):
  <ns 1>
  <ns 2>
Choose a different name for the object, or remove the conflicting objects from these namespaces.
```

yiqigao217:
> Can you please update the commit comment with an example of what the message looks like if there are one or more conflicts?

Good point, done!

incubator/hnc/internal/validators/object_test.go
```diff
@@ -142,7 +143,14 @@ func TestInheritedFromLabel(t *testing.T) {
...
 func TestUserChanges(t *testing.T) {
-	f := forest.NewForest()
+	// Construct a forest with a <- b ('a' is the parent). Create a secret object
+	// called 'conflict' in namespace 'b'.
+	f := foresttest.Create("-a")
+	inst := &unstructured.Unstructured{}
+	inst.SetName("conflict")
+	inst.SetNamespace("b")
+	inst.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+	f.Get("b").SetOriginalObject(inst)
```
adrianludwin, line 153: Since this has a different setup for any other case in `TestUserChanges`, maybe make a new top-level function? That way, you can also test a few different topologies (e.g. have a conflict in a child and a grandchild).

adrianludwin (commented):
Mainly lgtm, just some nits about the message and testing.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -332,3 +333,63 @@ func TestUserChanges(t *testing.T) {
...
+	// a <- b <- c
+	//   |
+	//   <- d
+	f := foresttest.Create("-aba")
+
+	// Create secret objects in namespace 'c' and 'd'.
+	createSecret("conflict-c", "c", f)
+	createSecret("conflict-d", "d", f)
+
+	o := &Object{Forest: f}
```
adrianludwin, line 348: Better to create this inside `t.Run()` so you get a fresh forest for each test. Otherwise each test run can impact the next test run.

Also - _consider_ having a minimal forest for each test case. E.g. for the first one, the forest could be `-a`, for the second it can be `-ab`, with a way to communicate which namespace should have the secrets. This might be overkill, but the reason I wrote `foresttest` in the first place was so that we didn't have to create an "omni-forest" that could handle every possible test case.

Though in this case, it's still fairly simple so my proposed change might be overkill - your call.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 

lgtm with a testing nit. @rjbez17 does everything lgty?

incubator/hnc/internal/validators/object_test.go
yiqigao217: sgtm will update it

yiqigao217:
FYI will look into the flaky test on the new object reconciler change next.

incubator/hnc/internal/validators/object.go
```diff
@@ -146,6 +152,33 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
 
+// hasConflict checks if there's any conflicting objects in the descendants. Returns
+// true and a list of conflicting descendants, if yes.
+func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
+	o.Forest.Lock()
+	defer o.Forest.Unlock()
+
+	// If the instance is empty (for a delete operation) or it's not namespace-scoped,
+	// there must be no conflict.
+	if inst == nil || inst.GetNamespace() == "" {
```
rjbez17, line 163: If the instance is marked for deletion but still terminating wont we run into errors when we try to create a different resource of the same name?

adrianludwin: Isn't that the correct behaviour? We need to wait until the instance is _fully_ deleted?

rjbez17: Yes, so then wouldn't we return true (there is a conflict) until it is fully deleted? As is this will indicate there is no conflict (unless I'm reading it wrong).

yiqigao217: According to #1079 , `Delete` or `Update` an object won't introduce source object overwriting issue so we only discuss the the `Create` operation. This `inst` is the new instance to operate on, and it will only be `nil` for `Delete` operation. That's why it's allowed by the webhook. Does it sound right?

"Fully deleted" concern seems more like for the reconcilers not the webhooks, since if a resource is terminating, the API server won't allow creating a resource with the same name (will report alreadyExist), right?

rjbez17: 👍

incubator/hnc/internal/validators/object.go
```diff
@@ -106,8 +107,13 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	oldSource, oldInherited := metadata.GetLabel(oldInst, api.LabelInheritedFrom)
 	newSource, newInherited := metadata.GetLabel(inst, api.LabelInheritedFrom)
 
-	// If the object wasn't and isn't inherited, it's none of our business.
+	// If the object wasn't and isn't inherited, we will check to see if the
+	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
+		if yes, cnses := o.hasConflict(inst); yes {
+			msg := fmt.Sprintf("Cannot create %s '%s' in namespace '%s' because it would overwrite objects in the following descendant namespace(s): %v. Choose a different name for the object, or remove the conflicting objects from these namespaces.", inst.GroupVersionKind(), inst.GetName(), inst.GetNamespace(), cnses)
```
adrianludwin, line 114: Nit: use `%q` instead of `'%s'`

adrianludwin: Also, it might be nice to introduce newlines, e.g.:

```
Cannot create %s '%s' in namespace '%s' because it would overwrite objects in the following descendant namespace(s):
  * NS 1
  * NS 2
Choose a different name for the object, or remove the conflicting objects from these namespaces.
```

This will make the message _much_ easier to read, IMO, especially the last line.

incubator/hnc/internal/validators/object.go
adrianludwin: @yiqigao217 is right; I was confusing this with a reconciler. Yes, we _never_ want to block a "delete" operation, so this code is correct.

incubator/hnc/internal/validators/object.go
yiqigao217: sg, will update the fmt in #1092, sgty?

adrianludwin: sgtm

PR #1086, opened 2020-09-08
Remove resource creation time rest.

weiling61:
Resource creation time is used by performance monitoring module in k8s to present end to end resource creation latency. In virtual cluster case, resource creation time should be counted when tenant initiate the resource creation request in virtual cluster. Performing reset during super cluster syncing destroys useful metrics.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/helper.go
@@ -189,7 +189,6 @@ func ResetMetadata(obj metav1.Object) {
 	obj.SetUID("")
 	obj.SetResourceVersion("")
 	obj.SetGeneration(0)
-	obj.SetCreationTimestamp(metav1.Time{})
 	obj.SetDeletionTimestamp(nil)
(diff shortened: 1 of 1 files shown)


Fei-Guo:
It is ok to remove it from reset code but I doubt the creation time in tenant object will persist according to the API comment

```
        // CreationTimestamp is a timestamp representing the server time when this object was
	// created. It is not guaranteed to be set in happens-before order across separate operations.
	// Clients may not set this value. It is represented in RFC3339 form and is in UTC.
	//
	// Populated by the system.
```

Have you verified?

weiling61:
@Fei-Guo That is a good point. Based on further testing results, super master assigns a new timestamp for the resouce under creation regardless we do a reset or not. The mistake came from: CreationTimeStamp provides a 'fake' nanosecond precision, the timestamp set by super master falls into the same value  as timestamp created by virtual cluster. ------ seems this PR takes effect.

Fei-Guo:
Thanks for the verification. The change is still valid though - do less unnecessary work. 

/lgtm
/approve

PR #1087, opened 2020-09-09
Move webhook setup before manager start

yiqigao217:
According to
https://github.com/kubernetes-sigs/controller-runtime/issues/1148,
there's a race condition between manager start and webhook register.
This commit moved webhook setup before manager start.

Tested on GKE cluster. The nil pointer error was not detected for 20
times.

### DIFF
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -153,32 +153,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	go startControllers(mgr, setupFinished)
-
-	setupLog.Info("Starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
-		os.Exit(1)
-	}
-}
-
-func startControllers(mgr ctrl.Manager, setupFinished chan struct{}) {
-	setupLog.Info("Waiting for certificate generation to complete")
-	// Block until the setup finishes.
-	<-setupFinished
-
-	if testLog {
-		stats.StartLoggingActivity()
-	}
-
-	// Create all reconciling controllers
+	// Register webhooks before manager start to avoid potential race conditions.
+	// See https://github.com/kubernetes-sigs/controller-runtime/issues/1148.
 	f := forest.NewForest()
-	setupLog.Info("Creating controllers", "maxReconciles", maxReconciles)
-	removeOldCRDVersion := true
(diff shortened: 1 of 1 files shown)


incubator/hnc/cmd/manager/main.go
```diff
@@ -153,32 +153,8 @@ func main() {
...
-	setupLog.Info("Waiting for certificate generation to complete")
-	// Block until the setup finishes.
-	<-setupFinished
-
-	if testLog {
-		stats.StartLoggingActivity()
-	}
-
-	// Create all reconciling controllers
 	f := forest.NewForest()
```
adrianludwin, line 156: This is a very significant line so I might comment it :)

incubator/hnc/cmd/manager/main.go
```diff
@@ -153,32 +153,8 @@ func main() {
...
-	setupLog.Info("Creating controllers", "maxReconciles", maxReconciles)
-	removeOldCRDVersion := true
-	if err := reconcilers.Create(mgr, f, maxReconciles, removeOldCRDVersion); err != nil {
-		setupLog.Error(err, "cannot create controllers")
-		os.Exit(1)
-	}
+	go startControllers(mgr, f, setupFinished)
```
adrianludwin, line 157: While it _probably_ doesn't really matter, I might move this to be the last statement before starting the manager. This reduces the amount of possible parallelism.

incubator/hnc/cmd/manager/main.go
```diff
@@ -153,32 +153,8 @@ func main() {
...
 
 	// Create all validating admission controllers.
```
adrianludwin, line 159: Maybe add a comment here about the bug we're working around. I know it won't exist in the future, but it's nice to say why we're starting the webhooks before the reconcilers.

adrianludwin (commented):
lgtm with minor nits. You don't need to test another 20 times after making my suggested changes, once will be fine :)

incubator/hnc/cmd/manager/main.go
yiqigao217: sg

yiqigao217:
> lgtm with minor nits. You don't need to test another 20 times after making my suggested changes, once will be fine :)

Updated. BTW tested 20 times because last time it took about 20 times to reproduce the race condition.

adrianludwin (commented):
/lgtm
/approve

@rjbez17 this is a fairly straightforward fix, lmk if you have any concerns about it but I'm just approving for now.

yiqigao217:
/retest

looking the object reconciler issue now.

PR #1088, opened 2020-09-09
Fix a flaky test for Prow specifically

yiqigao217:
Update the 5-millisecond sleep to 500-millisecond before detecting
object overwriting for object reconciler test.

This should fix the flaky test on Prow presubmit.

Tested by 'make test'. It was not flaky before and after the change
locally but may help for Prow.


adrianludwin (commented):
/approve
/lgtm

PR #1091, opened 2020-09-10
Fix the Rolebinding update logic

### LINKED ISSUE #798: HNC: if a RoleBinding is quickly deleted and recreated with a new Role, HNC may fail to update it
I was recently reminded that RoleBindings can't have their Roles changed after they're created. This suggests that if we _quickly_ recreate an RB in an ancestor namespace - fast enough that by the time that HNC notices, the new RB exists - HNC could see it as an _update_ (not a delete + create) and attempt to update the RBs in all descendant namespaces. This will fail.

We should add a test that exhibits this issue and find a way to fix it.

I don't _think_ this needs to be backported to HNC v0.4. I'd _imagine_ that in most cases, HNC _will_ notice that the RB has been deleted before it's been recreated, even if the change is being made by some automated tool. But I'll update the release notes to indicate this issue.

GinnyJI:
RoleBindings can't have their Roles changed after they're created. If an RB was quickly delete and re-created in an ancestor namespace - fast enough that by the time that HNC notices, the new RB exists - HNC could see it as an update (not a delete + create) and attempt to update the RBs in all descendant namespaces, and this will fail. In order to handle this case, we try to delete and re-create the rolebinding instead.

Tested: make test-e2e

Fix #798

### DIFF
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -538,10 +538,38 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
 	var err error = nil
 	stats.WriteObject(inst.GroupVersionKind())
 	if exist {
-		log.Info("Updating")
+		log.Info("Updating object")
 		err = r.Update(ctx, inst)
+		// RoleBindings can't have their Roles changed after they're created 
+		// (see  https://github.com/kubernetes-sigs/multi-tenancy/issues/798).
+		// If an RB was quickly delete and re-created in an ancestor namespace
+		// - fast enough that by the time that HNC notices, the new RB exists; or 
+		// if there's a change to the RBs when HNC isn't running - HNC could see 
+		// it as an update (not a delete + create) and attempt to update the RBs in
+		// all descendant namespaces, and this will fail. In order to handle this 
+		// case, we try to delete and re-create the rolebinding here
+		
+		// We only found this issue with the RoleBinding object, but we *think* this 
+		// will also be helpful for other similar objects that end up with the same error
+		// type. If we find out later that this assumption is not true, we can update the
+		// logic here to only deal with RoleBinding.
+
+		// The error type is 'Invalid' after I tested it out with different error types 
+		// from https://godoc.org/k8s.io/apimachinery/pkg/api/errors
+		if err != nil && errors.IsInvalid(err) {
+			if err = r.Delete(ctx, inst); err == nil {
+				err = r.Create(ctx, inst)
+				if err != nil {
+					log.Info("Unable to create new object.") // error is handles below
+				} else {
+					log.Info("Couldn't update object but delete and re-create it.")
+				}
+			} else {
+				log.Info("Unable to delete the existing object.") // error is handles below
+			}
+		}
 	} else {
-		log.Info("Creating")
+		log.Info("Creating object")
 		err = r.Create(ctx, inst)
 	}
 	if err != nil {
--- a/incubator/hnc/test/e2e/rolebinding_test.go
+++ b/incubator/hnc/test/e2e/rolebinding_test.go
@@ -1,0 +1,46 @@
+package e2e
+
+import (
+	"time"
+	. "github.com/onsi/ginkgo"
+	. "sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils"
+)
+
+var _ = Describe("HNC should delete and create a new Rolebinding instead of updating it", func() {
+
+	const (
+		nsParent = "parent"
+		nsChild  = "child"
+	)
+
+	BeforeEach(func() {
+		CheckHNCPath()
+		CleanupNamespaces(nsParent, nsChild)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsParent, nsChild)
+		RecoverHNC()
+	})
+
+	FIt("Should delete and create a Rolebinding when HNC is undeployed - issue #798", func() {
+		// NOTE: THERE IS ONE CASE THAT THIS TEST WILL ALWAYS PASS EVEN IF CODE IS BROKEN:
+		// After recovering HNC, if nsChild gets reconciled first, the 'admin' rolebinding will
+		// be deleted, and the 'edit' rolebinding will be created when nsParent gets reconciled.
+		// In this case the rolebinding would not be considered as 'updated' and the test will pass
+		MustRun("kubectl create ns", nsParent)
+		MustRun("kubectl hns create", nsChild, "-n", nsParent)
+		MustRun("kubectl create rolebinding test --clusterrole=admin --serviceaccount=default:default -n", nsParent)
+		FieldShouldContain("rolebinding", nsChild, "test", ".roleRef.name", "admin")
+		MustRun("kubectl delete deployment --all -n hnc-system")
+		// The pod might take up to a minite to be deleted, we force the deletion here to save time
+		MustRun("kubectl delete pods --all -n hnc-system --grace-period=0 --force")
+		RunShouldContain("No resources found", 60, "kubectl get pods -n hnc-system")
+		MustRun("kubectl delete rolebinding test -n", nsParent)
+		MustNotRun("kubectl describe rolebinding test -n", nsParent)
+		MustRun("kubectl create rolebinding test --clusterrole=edit --serviceaccount=default:default -n", nsParent)
+		FieldShouldContain("rolebinding", nsParent, "test", ".roleRef.name", "edit")
+		RecoverHNC()
+		FieldShouldContain("rolebinding", nsChild, "test", ".roleRef.name", "edit")
+	})
+})


incubator/hnc/internal/reconcilers/object.go
```diff
@@ -540,6 +540,25 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
 	if exist {
 		log.Info("Updating")
 		err = r.Update(ctx, inst)
+		// RoleBindings can't have their Roles changed after they're created. 
+		// If an RB was quickly delete and re-created in an ancestor namespace 
+		// - fast enough that by the time that HNC notices, the new RB exists - 
+		// HNC could see it as an update (not a delete + create) and attempt to 
+		// update the RBs in all descendant namespaces, and this will fail. 
+		// In order to handle this case, we try to delete and re-create the 
+		// rolebinding here
```
adrianludwin, line 549: This is a good comment, but it only talks about RBs while the code is general. We should explain why that is.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -540,6 +540,25 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+		if err != nil && errors.IsInvalid(err) {
```
adrianludwin, line 550: Comment why you picked `IsInvalid`

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -540,6 +540,25 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+			if err = r.Delete(ctx, inst); err == nil {
+				err = r.Create(ctx, inst)
+				if err != nil {
+					log.Info("Unable to create new object.")
+				} else {
+					log.Info("Successfully delete and create object.")
+				}
+			} else {
+				log.Info("Unable to delete the existing object.")
```
adrianludwin, line 559: It's not clear how we're handling the error here. Maybe add a comment that `err` will be handled below.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -540,6 +540,25 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+		// In order to handle this case, we try to delete and re-create the 
+		// rolebinding here
+		if err != nil && errors.IsInvalid(err) {
+			if err = r.Delete(ctx, inst); err == nil {
+				err = r.Create(ctx, inst)
+				if err != nil {
+					log.Info("Unable to create new object.")
+				} else {
+					log.Info("Successfully delete and create object.")
+				}
```
adrianludwin, line 557: We usually don't log _successful_ operations, but in this case we might want to say something like "Couldn't update object but delete/create worked" to indicate that something unusual happened.

incubator/hnc/internal/reconcilers/object.go
adrianludwin: Also you talk about "quickly" deleting and re-creating, but this probably also comes up if there's a change while HNC isn't running (which is what you actually test). Maybe add that to the comment as well.

adrianludwin (commented):
Looks good overall, just a couple of comments.

/assign @rjbez17

rjbez17:
lgtm after the comments are updated.

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done. Added a link to why the RB is immutable

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done

incubator/hnc/test/e2e/rolebinding_test.go
```diff
@@ -0,0 +1,39 @@
...
+		MustRun("kubectl delete deployment --all -n hnc-system")
+		// The pod might take up to a minite to be deleted, we force the deletion here to save time
+		MustRun("kubectl delete pods --all -n hnc-system --grace-period=0 --force")
+		RunShouldContain("No resources found", 60, "kubectl get pods -n hnc-system")
+		MustRun("kubectl delete rolebinding test -n", nsParent)
+		MustRun("kubectl create rolebinding test --clusterrole=edit --serviceaccount=default:default -n", nsParent)
+		RecoverHNC()
+		FieldShouldContain("rolebinding", nsChild, "test", ".roleRef.name", "edit")
+	})
+})
```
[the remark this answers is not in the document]
GinnyJI: Done

GinnyJI:
> BTW I have a question: the `delete->create` logic here is basically to "update" RBs in the descendants, right?
Yes

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -538,12 +538,35 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+		log.Info("Updating object")
 		err = r.Update(ctx, inst)
+		// RoleBindings can't have their Roles changed after they're created 
+		// (see  https://kubernetes.io/docs/reference/access-authn-authz/rbac/#clusterrolebinding-example).
+		// If an RB was quickly delete and re-created in an ancestor namespace
+		// - fast enough that by the time that HNC notices, the new RB exists; or 
+		// if there's a change to the RBs when HNC isn't running - HNC could see 
+		// it as an update (not a delete + create) and attempt to update the RBs in
+		// all descendant namespaces, and this will fail. In order to handle this 
+		// case, we try to delete and re-create the rolebinding here
```
adrianludwin, line 550: But this code isn't active _only_ for RBs - it works on _every_ object. You should explain why we did that.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -538,12 +538,35 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+			} else {
+				log.Info("Unable to delete the existing object.") // error is handles below
+			}
+		}
 	} else {
-		log.Info("Creating")
+		log.Info("Creating object")
 		err = r.Create(ctx, inst)
 	}
+	log.Info("Ginny", "raw", err)
```
adrianludwin, line 569: Remove :)

incubator/hnc/test/e2e/rolebinding_test.go
```diff
@@ -0,0 +1,46 @@
...
+		CleanupNamespaces(nsParent, nsChild)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsParent, nsChild)
+		RecoverHNC()
+	})
+
+	FIt("Should delete and create a Rolebinding when HNC is undeployed - issue #798", func() {
+		// NOTE: THERE IS ONE CASE THAT THIS TEST WILL ALWAYS PASS:
```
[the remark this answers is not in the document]
GinnyJI: Done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Is it to improve the success rate of this operation?

GinnyJI: Updated after our offline discussion

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -538,10 +538,37 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
 	var err error = nil
 	stats.WriteObject(inst.GroupVersionKind())
 	if exist {
-		log.Info("Updating")
+		log.Info("Updating object")
 		err = r.Update(ctx, inst)
+		// RoleBindings can't have their Roles changed after they're created 
+		// (see  https://kubernetes.io/docs/reference/access-authn-authz/rbac/#clusterrolebinding-example).
```
adrianludwin, line 544: Change this link to a reference to the issue, and make sure the issue includes this link? The issue will have more context on this problem.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -538,10 +538,37 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
+		// it as an update (not a delete + create) and attempt to update the RBs in
+		// all descendant namespaces, and this will fail. In order to handle this 
+		// case, we try to delete and re-create the rolebinding here
+		
+		// We only found this issue with the RoleBinding object, but we *think* this 
+		// will also be helpful for other similar objects that end up with the same error
+		// type. If we find out later that this assumption is not true, we can update the
+		// logic here to only deal with RoleBinding.
+
+		// The error type is 'Invalid' after I tested it out with different error types from https://godoc.org/k8s.io/apimachinery/pkg/api/errors
```
adrianludwin, line 557: Wrap this line

adrianludwin (commented):
Two tiny nits and then lgtm. Thanks!

incubator/hnc/internal/reconcilers/object.go

incubator/hnc/internal/reconcilers/object.go

adrianludwin (commented):
/lgtm
/approve

PR #1092, opened 2020-09-10
Update hc validator to check object conflicts

yiqigao217:
Add an additional rule to make sure there's no potential object
overwriting in hierarchy changes. Format the webhook message to:
```
Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "team-b": my-creds (/v1, Kind=Secret)
  * Namespace "team-b": my-creds2 (/v1, Kind=Secret)
```

Format the object webhook message to:
```
Cannot create "my-creds2" (/v1, Kind=Secret) in namespace "acme-org" because it would overwrite objects in the following descendant namespace(s):
  * team-a
  * team-b
To fix this, choose a different name for the object, or remove the conflicting objects from the above namespaces.
```

Tested manually on GKE cluster and by 'make test' with new test
cases.

Part of #1076

### DIFF
--- a/incubator/hnc/internal/forest/namespaceobjects.go
+++ b/incubator/hnc/internal/forest/namespaceobjects.go
@@ -63,6 +63,14 @@ func (ns *Namespace) GetPropagatedObjects(gvk schema.GroupVersionKind) []*unstru
 	return o
 }
 
+// GetPropagatingObjects returns all the source objects to be propagated into the
+// descendants of this namespace.
+// TODO update this function to reflect the changes introduced by the future HNC
+//  exceptions implementation.
+func (ns *Namespace) GetPropagatingObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
+	return append(ns.GetPropagatedObjects(gvk), ns.GetOriginalObjects(gvk)...)
+}
+
 // GetSource returns the original copy in the ancestors if it exists.
 // Otherwise, return nil.
 func (ns *Namespace) GetSource(gvk schema.GroupVersionKind, name string) *unstructured.Unstructured {
--- a/incubator/hnc/internal/validators/hierarchy.go
+++ b/incubator/hnc/internal/validators/hierarchy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/go-logr/logr"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
@@ -12,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
@@ -210,7 +212,52 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
 		return deny(metav1.StatusReasonConflict, "Illegal parent: "+reason)
 	}
 
+	// Prevent overwriting source objects in the descendants after the hierarchy change.
+	if co := v.getConflictingObjects(newParent, ns); len(co) != 0 {
+		msg := "Cannot update hierarchy because it would overwrite the following object(s):\n"
+		msg += "  * " + strings.Join(co, "\n  * ") + "\n"
+		msg += "To fix this, please rename or remove the conflicting objects first."
+		return deny(metav1.StatusReasonConflict, msg)
+	}
+
 	return allow("")
+}
+
+// getConflictingObjects returns a list of namespaced objects if there's any conflict.
+func (v *Hierarchy) getConflictingObjects(newParent, ns *forest.Namespace) []string {
+	// Traverse all the types with 'Propagate' mode to find any conflicts.
+	conflicts := []string{}
+	for _, t := range v.Forest.GetTypeSyncers() {
+		if t.GetMode() == api.Propagate {
+			conflicts = append(conflicts, v.getConflictingObjectsOfType(t.GetGVK(), newParent, ns)...)
+		}
+	}
+	return conflicts
+}
+
+// getConflictingObjectsOfType returns a list of namespaced objects if there's
+// any conflict between the new ancestors and the descendants.
+func (v *Hierarchy) getConflictingObjectsOfType(gvk schema.GroupVersionKind, newParent, ns *forest.Namespace) []string {
+	// Get all the source objects in the new ancestors that would be propagated
+	// into the descendants.
+	newAnsSrcObjs := make(map[string]bool)
+	for _, o := range newParent.GetPropagatingObjects(gvk) {
+		newAnsSrcObjs[o.GetName()] = true
+	}
+
+	// Look in the descendants to find if there's any conflict.
+	cos := []string{}
+	dnses := append(ns.DescendantNames(), ns.Name())
+	for _, dns := range dnses {
+		for _, o := range v.Forest.Get(dns).GetOriginalObjects(gvk) {
+			if newAnsSrcObjs[o.GetName()] {
+				co := fmt.Sprintf("Namespace %q: %s (%v)", dns, o.GetName(), gvk)
+				cos = append(cos, co)
+			}
+		}
+	}
+
+	return cos
 }
 
 type serverCheckType int
--- a/incubator/hnc/internal/validators/hierarchy_test.go
+++ b/incubator/hnc/internal/validators/hierarchy_test.go
@@ -7,7 +7,9 @@ import (
 	. "github.com/onsi/gomega"
 	authn "k8s.io/api/authentication/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/reconcilers"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/foresttest"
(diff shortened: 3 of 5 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

/hold
Until #1085 is merged, because this PR uses the `createSecret()` function there.

adrianludwin:
> ```
> Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because
> it would overwrite the following object(s): [/v1, Kind=Secret/my-creds /v1, Kind=Secret/my-creds2]. Please rename or
> remove the conflicting objects first.
> ```

This doesn't say what namespace the affected objects are in, but it's not obvious, right? When you move a subtree from one place to another, the object that gets overwritten could be _anywhere_ in that subtree.

Also, the format of the object names looks a bit weird - the commas and spaces don't clearly show the difference between the different objects. E.g. it makes it look like the Kind is `Secret/my-creds /v1`, which isn't correct.

yiqigao217:
> > ```
> > Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update
> > hierarchy because it would overwrite the following object(s):
> > [/v1, Kind=Secret/my-creds /v1, Kind=Secret/my-creds2]. Please rename or remove the conflicting
> > objects first.
> > ```
> 
> This doesn't say what namespace the affected objects are in, but it's not obvious, right? When you move a subtree from one place to another, the object that gets overwritten could be _anywhere_ in that subtree.
> 
> Also, the format of the object names looks a bit weird - the commas and spaces don't clearly show the difference between the different objects. E.g. it makes it look like the Kind is `Secret/my-creds /v1`, which isn't correct.

What about this?
```
Could not update the hierarchical configuration of team-b.
Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s): [
Secret team-b/my-creds 
Secret team-b/my-creds2]. 
Please rename or remove the conflicting objects first.
```

adrianludwin:
> What about this?
> 
> ```
> Could not update the hierarchical configuration of team-b.
> Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: Cannot update hierarchy because it would overwrite the following object(s): [
> Secret team-b/my-creds 
> Secret team-b/my-creds2]. 
> Please rename or remove the conflicting objects first.
> ```

It might be nice to indent them and specify what's the namespace, and also get rid of that square bracket, e.g.:

```
Cannot update hierarchy because it would overwrite the following object(s):
  * Namespace "team-b": my-creds (Kind: Secret; apiVersion: v1)
  * Namespace "team-c": my-custom-resource (Kind: Foobar; apiVersion: acme.com/v1alpha1)
```

That puts the most relevant information (namespace and name) first. A good trick to do this is a fmt statement like:

```
msg := fmt.Sprintf("Cannot update:\n  * %s\nTo fix this, ...", conflictStr)
```
And then use `conflictStr := string.Join("\n  * ", conflicts)` to insert all the newlines and bullet points. But that's just an idea.

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,55 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
 		return deny(metav1.StatusReasonConflict, "Illegal parent: "+reason)
 	}
 
+	// Prevent overwriting source objects in the descendants after the hierarchy change.
+	if yes, co := v.hasConflictObject(newParent, ns); yes {
```
adrianludwin, line 216: You _could_ just return `co`:

`if co := v.getOverwrittenObjects(newParent, ns); co != nil {`

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,55 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
+		coStr := strings.Join(co, "\n  * ")
+		msg := fmt.Sprintf("\nCannot update hierarchy because it would overwrite the following object(s):\n  * %s\nTo fix this, please rename or remove the conflicting objects first.", coStr)
```
adrianludwin, line 218: I'm not sure about that _leading_ newline, I've never seen that before so it's probably a good idea to have at least the first part of the message in the main line.

Nit: also, it might be more clear to construct the message like this:

```
msg := "Cannot update hierarchy because it would overwrite the following object(s):\n"
msg += "  * " + coStr + "\n"
msg += "To fix this, please rename or remove the conflicting objects first."
```

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,55 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
+func (v *Hierarchy) hasConflictObject(pns, cns *forest.Namespace) (bool, []string) {
+	// Get all types with 'Propagate' mode.
+	ts := []forest.TypeSyncer{}
+	for _, t := range v.Forest.GetTypeSyncers() {
+		if t.GetMode() == api.Propagate {
+			ts = append(ts, t)
+		}
+	}
+
+	// Get all the original objects in the new ancestors (including the parent).
```
adrianludwin, line 235: s/original/source/

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,55 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
+	existsInAns := make(map[schema.GroupVersionKind]map[string]bool)
```
adrianludwin, line 236: Maybe `newSrcObjs`?

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,55 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
 
+// hasConflictObject returns true and a list of objects if there's a conflict.
+func (v *Hierarchy) hasConflictObject(pns, cns *forest.Namespace) (bool, []string) {
+	// Get all types with 'Propagate' mode.
+	ts := []forest.TypeSyncer{}
+	for _, t := range v.Forest.GetTypeSyncers() {
+		if t.GetMode() == api.Propagate {
+			ts = append(ts, t)
+		}
+	}
```
adrianludwin, line 233: This code might be more clear if you split it into two functions: one to iterate over the GVKs, and another to construct the conflicting strings just for that GVK:

```
func (v *Hierarchy) getConflictingObjects(pns, cns *forest.Namespace) []string {
  var conflicts []string
  for _, t := range v.Forest.GetTypeSyncers() {
    if t.GetMode() == api.Propagate {
      conflicts = append(conflicts, v.getConflictingObjectsOfType(t.GetGVK(), pns, cns))
    }
  }
  return conflicts
}
```

Then your lower-level functions only need to deal with one type, which should make them much simpler.

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,55 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
+
+	// Get all the original objects in the new ancestors (including the parent).
+	existsInAns := make(map[schema.GroupVersionKind]map[string]bool)
+	for _, t := range ts {
+		gvk := t.GetGVK()
+		existsInAns[gvk] = make(map[string]bool)
+		os := append(pns.GetPropagatedObjects(gvk), pns.GetOriginalObjects(gvk)...)
```
adrianludwin, line 240: This line probably deserves a comment. Also, once we have exceptions, we'll need something smarter than this, and you can comment that too (cc @GinnyJI ).

incubator/hnc/internal/validators/hierarchy_test.go
```diff
@@ -96,6 +98,54 @@ func TestChangeParentOnManagedBy(t *testing.T) {
...
 
+func TestChangeParentWithConflict(t *testing.T) {
+	f := foresttest.Create("-a-c") // a <- b; c <- d
+
+	// Set secret to "Propagate" mode.
+	or := &reconcilers.ObjectReconciler{
+		GVK:  schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"},
+		Mode: api.Propagate,
+	}
+	f.AddTypeSyncer(or)
```
adrianludwin, line 109: Why not just use rolebindings? Seems easier, but if there's a reason, comment it

incubator/hnc/internal/validators/hierarchy_test.go
```diff
@@ -96,6 +98,54 @@ func TestChangeParentOnManagedBy(t *testing.T) {
...
+	l := zap.Logger(false)
+
+	tests := []struct {
+		name string
+		nnm  string
+		pnm  string
+		fail bool
+	}{
+		{name: "conflict in subtree root and new parent", nnm: "a", pnm: "d", fail: true},
+		{name: "conflict in itself and new parent", nnm: "d", pnm: "a", fail: true},
```
adrianludwin, line 125: What's the difference between these? Just that `a` is a root and `d` is a leaf?

incubator/hnc/internal/validators/hierarchy.go
yiqigao217: Updated with `newAnsSrcObjs`?

adrianludwin: Are there ancestor objects that can't be sources? Just wondering why we need both descriptions. Also "Ans" is a bit of a hard short form to read, while both "src" and "obj" are fairly well-understood.

incubator/hnc/internal/validators/hierarchy.go
yiqigao217: Added it as a new function in forest with `TODO`:
```
// GetPropagatingObjects returns all the source objects to be propagated into the
// descendants of this namespace.
// TODO update this function to reflect the changes introduced by the future HNC
//  exceptions implementation.
func (ns *Namespace) GetPropagatingObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
	return append(ns.GetPropagatedObjects(gvk), ns.GetOriginalObjects(gvk)...)
}
```

incubator/hnc/internal/validators/hierarchy_test.go
yiqigao217: The test forest doesn't have `Role` and `RoleBinding` as `TypeSyncer` by default either. So it's the same for RB and secret, and we have an existing `createSecret()` func. Will comment that.

adrianludwin: sg

incubator/hnc/internal/validators/hierarchy_test.go
yiqigao217: Ah right, changed the new parent in the 2nd case to `b` and updated the first test name:
```
{name: "conflict in itself and the new parent", nnm: "a", pnm: "d", fail: true},
{name: "conflict in itself and a new ancestor (not the parent)", nnm: "d", pnm: "b", fail: true},
```
So that the 4 failed tests covers all combinations of conflicts between`itself` and `parent/ancestor`, `subtree leaf` and `parent/ancestor`.

incubator/hnc/internal/forest/namespaceobjects.go
```diff
@@ -63,6 +63,14 @@ func (ns *Namespace) GetPropagatedObjects(gvk schema.GroupVersionKind) []*unstru
...
 }
 
+// GetPropagatingObjects returns all the source objects to be propagated into the
+// descendants of this namespace.
+// TODO update this function to reflect the changes introduced by the future HNC
+//  exceptions implementation.
+func (ns *Namespace) GetPropagatingObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
+	return append(ns.GetPropagatedObjects(gvk), ns.GetOriginalObjects(gvk)...)
+}
+
```
yiqigao217, line 73: @adrianludwin lgty?

incubator/hnc/internal/validators/hierarchy.go
yiqigao217: Because we want to compare the source objects in all NEW ancestors for a parent change, so it's not just source objects in one namespace but new ancestors.

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,54 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
...
+		msg += "  * " + strings.Join(co, "\n  * ") + "\n"
+		msg += "To fix this, please rename or remove the conflicting objects first."
+		return deny(metav1.StatusReasonConflict, msg)
+	}
+
 	return allow("")
 }
 
+// hasConflictObject returns a list of namespaced objects if there's any conflict.
+func (v *Hierarchy) hasConflictObject(pns, cns *forest.Namespace) []string {
```
adrianludwin, line 227: Two nits:
* "has" suggests a boolean, but this returns a slice. `getConflictingObjects` would also match up well with `getConflictingObjectsOfType`, below.
* Instead of `pns` and `cns`, I'd keep the `newParent` and `ns` variable names from the caller, both for consistency but also because IMO it more clearly shows which is the "main" namespace (`ns`/`cns`) and what's the thing that's being tested against it.

adrianludwin (commented):
Two last nits then lgtm, but I'll approve now so Ryan can take a look

/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/internal/validators/hierarchy.go
adrianludwin: Note that `pns/cns` is used in this function and also `getConflictingObjectsOfType`

yiqigao217: sgtm, updated!

rjbez17:
sorry, those buttons are too close to each other.

/lgtm
/approve
/hold cancel

PR #1093, opened 2020-09-10
Upgrade controller-tools to enable unservedversion

yiqigao217:
Upgrade controller-tools to v0.2.8 where kubebuilder:unservedversion
marker is landed. See https://github.com/kubernetes-sigs/controller-tools/issues/462

Tested by make test.

Part of #1035


yiqigao217:
> That was a surprisingly clean change!

Great!

PR #1095, opened 2020-09-10
Fix make target kind-deploy for KIND environment

### LINKED ISSUE #1084: HNC: make target kind-deploy is failing due to empty PROJECT_ID
While deploying HNC in a kind cluster using `make kind-deploy` or `CONFIG=kind make deploy`, the deployment fails with below error.
```
$ make kind-deploy        
CONFIG=kind make deploy
make[1]: Entering directory '/home/mugdha/codespace/mugdha-adhav/multi-tenancy/incubator/hnc'
/home/mugdha/codespace/mugdha-adhav/multi-tenancy/incubator/hnc/bin/controller-gen object:headerFile=./hack/boilerplate.go.txt paths=./api/...
go fmt ./api/... ./cmd/... ./internal/... ./third_party/...
go vet ./api/... ./cmd/... ./internal/... ./third_party/...
Warning: this does not run tests. Run 'make test' to ensure tests are passing.
docker build . -t gcr.io//hnc-manager:kind-local
invalid argument "gcr.io//hnc-manager:kind-local" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
make[1]: *** [Makefile:212: docker-build] Error 125
make[1]: Leaving directory '/home/mugdha/codespace/mugdha-adhav/multi-tenancy/incubator/hnc'
make: *** [Makefile:230: kind-deploy] Error 2
```

The error occurs as `PROJECT_ID` variable is unset in case of kind deployment. The variable value is retrieved by running gcloud command which is not required for kind deployment.

mugdha-adhav:
Make target `kind-deploy` failing due to empty PROJECT_ID set for kind deployment.

This change skips the PROJECT_ID value if it is not set

Fixes #1084

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -22,7 +22,9 @@ PROJECT_ID := ${shell gcloud config get-value project 2>/dev/null}
 # By default, we use the Container Registry in the current GCP project, but you
 # can set it to anything, UNLESS you are calling 'make release' since the image
 # needs to be built in the same project as the GCB instance that builds it.
-HNC_REGISTRY ?= gcr.io/${PROJECT_ID}
+ifdef PROJECT_ID
+  HNC_REGISTRY ?= gcr.io/${PROJECT_ID}
+endif
 
 # The image name is the *base* name - excluding the registry and the tag.
 HNC_IMG_NAME ?= hnc-manager
@@ -44,7 +46,11 @@ endif
 #
 # Note that if you're using Kind, this image will never actually be pushed to
 # the registry.
-HNC_IMG = ${HNC_REGISTRY}/${HNC_IMG_NAME}:${HNC_IMG_TAG}
+ifdef HNC_REGISTRY
+	HNC_IMG = ${HNC_REGISTRY}/${HNC_IMG_NAME}:${HNC_IMG_TAG}
+else
+	HNC_IMG = ${HNC_IMG_NAME}:${HNC_IMG_TAG}
+endif
 
 # `make` must be called from the HNC root, or all kinds of things will break
 # (starting with this).


incubator/hnc/Makefile
```diff
@@ -22,7 +22,11 @@ PROJECT_ID := ${shell gcloud config get-value project 2>/dev/null}
 # By default, we use the Container Registry in the current GCP project, but you
 # can set it to anything, UNLESS you are calling 'make release' since the image
 # needs to be built in the same project as the GCB instance that builds it.
-HNC_REGISTRY ?= gcr.io/${PROJECT_ID}
+ifdef PROJECT_ID
+  HNC_REGISTRY ?= gcr.io/${PROJECT_ID}
+else
+  HNC_REGISTRY ?= gcr.io
+endif
```
adrianludwin, line 29: `gcr.io` on its own isn't really a registry. How about this:

* If `PROJECT_ID` doesn't exist, `HNC_REGISTRY` is an empty string
* On line 51, if `HNC_REGISTRY` is undefined, we just omit it when creating `HNC_IMG`.

E.g., if `PROJECT_ID` is my-proj, then `HNC_IMG` will be `gcr.io/my-proj/hnc-manager:kind-local` by default (if we're configured for KIND). But if `PROJECT_ID` is missing, then `HNC_IMG` will be `hnc-manager:kind-local` which is a perfectly valid image name for an image that will never leave the current computer.

I know it's a bit more work but it prevents us from having a registry that _looks_ valid but doesn't actually mean anything. Does this make sense?

adrianludwin (commented):
Thanks for this! It looks good but I think we can make it even cleaner by not setting HNC_REGISTRY at all - see comments for more.

incubator/hnc/Makefile
mugdha-adhav: Yes, definitely it does.

mugdha-adhav: Make the requested changes.

adrianludwin:
Thanks! The code looks good, do you mind making two last changes?

1. Can you please squash your two commits into one? (Let me know if you don't know how to do this).
2. Can you please amend the commit message to say how you tested this? E.g., it would be nice for you to run this both with and without `PROJECT_ID` defined so you have some confidence that it's working properly, _and_ for you to have a record of that in the commit message. E.g., see [this commit message](https://github.com/kubernetes-sigs/multi-tenancy/commit/ce983662e87264c76f92dbfbab7cef7bd6128837) for an example of how we document what we've tested.

Thanks again for this contribution!

/ok-to-test

mugdha-adhav:
I have already tested all the edge cases. Mentioning it in the commit message.

Also squashing the commits.

adrianludwin:
Thanks! Just force-push your changes to update the branch when you're ready.

mugdha-adhav:
Done!

Thanks for your support and quick response.

adrianludwin (commented):
/lgtm
/approve

Thanks for this change! We really appreciate it!

PR #1099, opened 2020-09-11
VC: Use optional env var for Docker registry in Makefile

salaxander:
Very small change! I was trying to do some local development, and noticed in the build script we can set the `VC_DOCKER_REGISTRY` variable when building the images.

I added this functionality to the Makefile as well to make pushing to a custom repo and testing very easy. It still defaults to `virtual cluster`


Fei-Guo:
Nice improvement!

/lgtm
/approve

Issue #1102, opened 2020-09-14
HNC: Update hncconfig validator to avoid overwriting source


yiqigao217:
Part of #1076 , see http://bit.ly/hnc-propagation-conflict.

Overwriting source could happen when a type is changed from another mode to `Propagate` mode and there happens to be an ancestor and a descendant objects with the same name.

The current hncconfig validator only checks if the _new_ configuration meets the restrictions, e.g. no duplicate types, `Propagate` mode for `Role` and `Rolebinding`, etc. Now since we need to detect a mode change, we will need to get the old configuration, which can be done by adding a `checkForest()` func.

We also need to get all the objects, especially the object names, of the types that are changed from others to `Propagate`. The current forest (in-memory data-structure) only stores user-created objects if the type mode is `Propagate`. Therefore, we cannot get the objects from the forest for those types switched to `Propagate` from others.

A **proposal** is that we can

-  Add a new field called `objectNames` (`map[schema.GroupVersionKind]map[string]bool`) to the `forest.Namespace` struct, to store all the names of the user-created objects by GVK _as long as there's an object reconciler for the GVK_. 
- Ask users, who wants to change a type to `Propagate`, to first change it to `Remove` mode so that an object reconciler can be created for this type. Or we can just update the user guide to say that it requires changing to `Remove` mode first for any mode changes between `Ignore` and `Propagate`, since we already require that for `Propagate => Ignore` switch.
- In object reconciliations, always `forest.ns.AddObjectName()` in `syncSource()` func and `forest.ns.RemoveObjectName()` in `syncMissingObject()` func.

With `objectNames` field in the forest, the hncconfig validator can detect the object name conflicts for non-`Propagate` types. As for the conflicts, the data-structure will look like this:
```
type conflicts map[schema.GroupVersionKind]map[string]affectedNamespaces
type affectedNamespaces map[string][]string
```
with specified GVK and object name, and groups of `affectedNamespaces` grouped by the first ancestor namespace name, considering the conflicts can happen on different branches in the forest and users have to resolve them separatedly.

yiqigao217:
/cc @adrianludwin 
/cc @rjbez17 
/cc @GinnyJI 

I tried implementing it and it seemed working fine and I will do more tests on it. Please let me know if there's any concerns of the changes in the forest (new `ns.objectNames` field), object reconciler (updating `ns.objectNames`) and the hncconfig validator (conflict datastructure), etc.

[...]

PR #1103, opened 2020-09-14
Update hncconfig webhook to avoid overwriting source

yiqigao217:
Add new field `objectNames` to `forest.Namespace` struct to store all
the names of user-created objects by GVK as long as there's an object
reconciler for that GVK.

Update object reonciler to add/remove reconciled source object for the
`objectNames` field in `syncSource()` and `syncMissingObject()` funcs.

Add new hncconfig webhook rules and new test cases. Check if changing
type mode(s) to `Propagate` would cause overwriting source by looking
up the `objectNames` data. Deny the request and output all the conflicts
if it would cause overwriting.

Tested by "make test" and manually on a GKE cluster with multiple
objects with the same name in the same tree or different trees.

The webhook message would look like this:
```
Cannot update configuration because setting types to 'Propagate' mode would overwrite user-created object(s):
 * Type: "/v1, Kind=Secret", Conflict(s):
   - "my-creds" in namespace "acme-org" would overwrite the one(s) in [team-a, team-b].
   - "my-creds" in namespace "bcme-org" would overwrite the one(s) in [team-c].
   - "my-creds2" in namespace "acme-org" would overwrite the one(s) in [team-b].
```
Part of #1102, #1076

### DIFF
--- a/incubator/hnc/internal/validators/hncconfig.go
+++ b/incubator/hnc/internal/validators/hncconfig.go
@@ -3,6 +3,7 @@ package validators
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"k8s.io/api/admission/v1beta1"
@@ -12,6 +13,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
@@ -29,6 +31,7 @@ const (
 
 type HNCConfig struct {
 	Log       logr.Logger
+	Forest    *forest.Forest
 	validator gvkValidator
 	decoder   *admission.Decoder
 }
@@ -103,7 +106,10 @@ func (c *HNCConfig) handle(ctx context.Context, inst *api.HNCConfiguration) admi
 	if !roleBindingExist {
 		return deny(metav1.StatusReasonInvalid, "Configuration for RoleBinding is missing")
 	}
-	return allow("")
+
+	// Lastly, check if changing a type to "Propagate" mode would cause
+	// overwriting user-created objects.
+	return c.checkForest(inst)
 }
 
 // Validate if the configuration of a type already exists. Each type should only have one configuration.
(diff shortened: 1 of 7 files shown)


incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,160 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+	for gvk, _ := range gvks {
+		// We will keep one object set for each type, because objects with the same
+		// name but different types won't affect each other.
+		obs := objSet{}
+
+		// Traverse all the namespaces.
+		for _, nnm := range c.Forest.GetNamespaceNames() {
+			ns := c.Forest.Get(nnm)
+
+			// Traverse all the original objects in each namespace. Please note that
```
adrianludwin, line 181: I found this section pretty hard to read. Also, I'm not sure that it will work with exceptions. How about doing something like this?

```
c := []conflicts{}
foreach ns {
  c = append(c, getConflictsFromTree(ns)
}

func getConflictsFromTree(ns) []conflicts {
  c := []conflicts{}
  foreach object in ns
    // when we get exceptions, hasConflictingAncestor will take the current
    // ns as a param too
    if src := hasConflictingAncestor(object); src != nil {
      c = append(c, src)
    }
  return c
}
```

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -197,6 +200,55 @@ func TestNonRBACTypes(t *testing.T) {
...
+		{
+			name:         "Objects with the same name existing in namespaces that one is not an ancestor of the other would not cause overwriting conflict",
+			inNamespaces: []string{"b", "c"},
+			allow:        true,
+		},
+		{
+			name:         "Objects with the same name existing in namespaces that one is an ancestor of the other would have overwriting conflict",
+			inNamespaces: []string{"a", "b"},
+			allow:        false,
+		}}
```
adrianludwin, line 221: Inconsistent indenting. Either:
```
{
  {
     ...
   },
   {
      ...
   },
}
```

or (my preference):

```
{{
  ...
}, {
  ...
}}
```

adrianludwin (commented):
As discussed offline, I think it would be better _not_ to introduce a new field into the forest, and instead to always store the source object and change the rest of the object to suit (e.g. don't propagate objects just because they're in the forest - we'll need this logic for exceptions anyway).

yiqigao217:
Closing this since as discussed offline with @adrianludwin , we will update the hncconfig webhook design (https://github.com/kubernetes-sigs/multi-tenancy/issues/1102#issue-701126978) to use the updated forest field `sourceObjects` with all objects stored even for non-`Propagate` mode types as long as there once is an object reconciler created for it.

#1106 is the replacement of this PR with two commits: 
1) refactor `sourceObjects` field and update object reconciler dependencies; 
2) update hncconfig webhook to handle object overwriting with the updated `sourceObjects` field.

PR #1106, opened 2020-09-16
Update hncconfig webhook to handle object overwriting and refactor forest `sourceObjects` field

yiqigao217:
Update forest field `sourceObjects` to store all the names of user-created
objects by GVK as long as there's an object reconciler for that GVK.

Change object reonciler behaviors dependent on the previous forest field
`originalObjects`, which only stores source objects for 'Propagate' types.
Add test cases for corner cases.

Add new hncconfig webhook rules and new test cases. Check if changing
type mode(s) to `Propagate` would cause overwriting source by looking
up the `objectNames` data. Deny the request and output all the conflicts
if it would cause overwriting.

Tested by "make test" and manually on a GKE cluster with multiple
objects with the same name in the same tree or different trees. Passed
all e2e tests too.

The webhook message would look like this:
```
Cannot update configuration because setting type "/v1, Kind=Secret" to 'Propagate' mode would overwrite user-created object(s):
  Object "my-creds2" in namespace "acme-org" would overwrite the one in "team-a"
  Object "my-creds2" in namespace "acme-org" would overwrite the one in "team-b"
  Object "my-creds" in namespace "bcme-org" would overwrite the one in "team-c"
  Object "my-creds2" in namespace "bcme-org" would overwrite the one in "team-c"
```
Part of #1102, #1076

### DIFF
--- a/incubator/hnc/internal/forest/forest.go
+++ b/incubator/hnc/internal/forest/forest.go
@@ -96,6 +96,19 @@ func (f *Forest) GetNamespaceNames() []string {
 	return names
 }
 
+// GetRoots returns all the root namespaces in the cluster. Any possible cycles
+// are omitted since we look for namespaces with no parent and cycles must
+// always be at roots.
+func (f *Forest) GetRoots() []*Namespace {
+	nses := []*Namespace{}
+	for _, ns := range f.namespaces {
+		if ns.parent == nil {
+			nses = append(nses, ns)
+		}
+	}
+	return nses
+}
+
 // AddTypeSyncer adds a reconciler to the types list.
 func (f *Forest) AddTypeSyncer(nss TypeSyncer) {
 	f.types = append(f.types, nss)
--- a/incubator/hnc/internal/forest/forest.go
+++ b/incubator/hnc/internal/forest/forest.go
@@ -77,11 +77,11 @@ func (f *Forest) Get(nm string) *Namespace {
 		return ns
 	}
 	ns = &Namespace{
-		forest:          f,
-		name:            nm,
-		children:        namedNamespaces{},
-		conditions:      conditions{},
-		originalObjects: objects{},
+		forest:        f,
+		name:          nm,
+		children:      namedNamespaces{},
+		conditions:    conditions{},
+		sourceObjects: objects{},
 	}
 	f.namespaces[nm] = ns
 	return ns
--- a/incubator/hnc/internal/forest/namespaceobjects.go
+++ b/incubator/hnc/internal/forest/namespaceobjects.go
@@ -5,80 +5,79 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
-// SetOriginalObject updates or creates the original object in the namespace in the forest.
-func (ns *Namespace) SetOriginalObject(obj *unstructured.Unstructured) {
+// SetSourceObject updates or creates the source object in forest.namespace.
+func (ns *Namespace) SetSourceObject(obj *unstructured.Unstructured) {
 	gvk := obj.GroupVersionKind()
 	name := obj.GetName()
-	_, ok := ns.originalObjects[gvk]
+	_, ok := ns.sourceObjects[gvk]
 	if !ok {
-		ns.originalObjects[gvk] = map[string]*unstructured.Unstructured{}
+		ns.sourceObjects[gvk] = map[string]*unstructured.Unstructured{}
 	}
-	ns.originalObjects[gvk][name] = obj
+	ns.sourceObjects[gvk][name] = obj
 }
 
-// GetOriginalObject gets an original object by name. It returns nil, if the object doesn't exist.
-func (ns *Namespace) GetOriginalObject(gvk schema.GroupVersionKind, nm string) *unstructured.Unstructured {
-	return ns.originalObjects[gvk][nm]
+// GetSourceObject gets a source object by name. If it doesn't exist, return nil.
+func (ns *Namespace) GetSourceObject(gvk schema.GroupVersionKind, nm string) *unstructured.Unstructured {
+	return ns.sourceObjects[gvk][nm]
 }
 
-// HasOriginalObject returns if the namespace has an original object.
-func (ns *Namespace) HasOriginalObject(gvk schema.GroupVersionKind, oo string) bool {
-	return ns.GetOriginalObject(gvk, oo) != nil
+// HasSourceObject returns if the namespace has a source object.
+func (ns *Namespace) HasSourceObject(gvk schema.GroupVersionKind, oo string) bool {
+	return ns.GetSourceObject(gvk, oo) != nil
 }
 
-// DeleteOriginalObject deletes an original object by name.
-func (ns *Namespace) DeleteOriginalObject(gvk schema.GroupVersionKind, nm string) {
-	delete(ns.originalObjects[gvk], nm)
+// DeleteSourceObject deletes a source object by name.
+func (ns *Namespace) DeleteSourceObject(gvk schema.GroupVersionKind, nm string) {
+	delete(ns.sourceObjects[gvk], nm)
 	// Garbage collection
-	if len(ns.originalObjects[gvk]) == 0 {
-		delete(ns.originalObjects, gvk)
+	if len(ns.sourceObjects[gvk]) == 0 {
+		delete(ns.sourceObjects, gvk)
 	}
 }
 
-// GetOriginalObjects returns all original objects in the namespace.
-func (ns *Namespace) GetOriginalObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
+// GetSourceObjects returns all source objects in the namespace.
+func (ns *Namespace) GetSourceObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
 	o := []*unstructured.Unstructured{}
-	for _, obj := range ns.originalObjects[gvk] {
+	for _, obj := range ns.sourceObjects[gvk] {
 		o = append(o, obj)
 	}
 	return o
 }
 
-// GetNumOriginalObjects returns the total number of original objects of a specific GVK in the namespace.
-func (ns *Namespace) GetNumOriginalObjects(gvk schema.GroupVersionKind) int {
-	return len(ns.originalObjects[gvk])
+// GetNumSourceObjects returns the total number of source objects of a specific
+// GVK in the namespace.
+func (ns *Namespace) GetNumSourceObjects(gvk schema.GroupVersionKind) int {
+	return len(ns.sourceObjects[gvk])
 }
 
-// GetPropagatedObjects returns all original copies in the ancestors.
-func (ns *Namespace) GetPropagatedObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
-	o := []*unstructured.Unstructured{}
+// GetAncestorSourceObjects returns all source objects with the specified name
+// in the ancestors (including itself) from top down. If the name is not
+// specified, all the source objects in the ancestors will be returned.
+func (ns *Namespace) GetAncestorSourceObjects(gvk schema.GroupVersionKind, name string) []*unstructured.Unstructured {
+	// The namespace could be nil when we use this function on "ns.Parent()" to
+	// get the source objects of the ancestors excluding itself without caring if
+	// the "ns.Parent()" is nil.
+	if ns == nil {
+		return nil
+	}
+
+	// Get the source objects in the ancestors from top down.
+	objs := []*unstructured.Unstructured{}
 	ans := ns.AncestryNames()
 	for _, n := range ans {
-		// Exclude the original objects in this namespace
-		if n == ns.name {
-			continue
-		}
-		o = append(o, ns.forest.Get(n).GetOriginalObjects(gvk)...)
-	}
-	return o
-}
-
-// GetPropagatingObjects returns all the source objects to be propagated into the
-// descendants of this namespace.
-// TODO update this function to reflect the changes introduced by the future HNC
-//  exceptions implementation.
-func (ns *Namespace) GetPropagatingObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
-	return append(ns.GetPropagatedObjects(gvk), ns.GetOriginalObjects(gvk)...)
-}
-
-// GetSource returns the original copy in the ancestors if it exists.
-// Otherwise, return nil.
-func (ns *Namespace) GetSource(gvk schema.GroupVersionKind, name string) *unstructured.Unstructured {
-	pos := ns.GetPropagatedObjects(gvk)
-	for _, po := range pos {
-		if po.GetName() == name {
-			return po
+		nsObjs := ns.forest.Get(n).GetSourceObjects(gvk)
+		if name == "" {
+			// Get all the source objects if the name is not specified.
+			objs = append(objs, ns.forest.Get(n).GetSourceObjects(gvk)...)
+		} else {
+			// If a name is specified, return the matching objects.
+			for _, o := range nsObjs {
+				if o.GetName() == name {
+					objs = append(objs, o)
+				}
+			}
 		}
 	}
-	return nil
+
+	return objs
 }
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -274,14 +274,16 @@ func (r *ObjectReconciler) skipNamespace(ctx context.Context, log logr.Logger, i
 // which indicates that we need to call the regular syncObject method. Note that this method may
 // modify `inst`.
 func (r *ObjectReconciler) syncMissingObject(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) syncAction {
-	// If the object exists, skip
+	// If the object exists, skip.
 	if inst.GetCreationTimestamp() != (v1.Time{}) {
 		return actionUnknown
 	}
 
+	ns := r.Forest.Get(inst.GetNamespace())
 	// If it's a source, it must have been deleted. Update the forest and enqueue all its
 	// descendants, but there's nothing else to do.
-	if r.forestHasSource(inst) {
+	if ns.HasSourceObject(r.GVK, inst.GetName()) {
+		ns.DeleteSourceObject(r.GVK, inst.GetName())
 		r.syncUnpropagatedSource(ctx, log, inst)
 		return actionNop
 	}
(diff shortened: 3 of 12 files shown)


incubator/hnc/internal/forest/namespaceobjects.go
```diff
@@ -5,76 +5,77 @@ import (
...
-func (ns *Namespace) GetPropagatedObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
+// GetAncestorSourceObjects returns all source objects in the ancestors.
+func (ns *Namespace) GetAncestorSourceObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
 	o := []*unstructured.Unstructured{}
 	ans := ns.AncestryNames()
 	for _, n := range ans {
 		// Exclude the original objects in this namespace
 		if n == ns.name {
 			continue
 		}
```
adrianludwin, line 61: How about we remove lines 58-61, get rid of `GetAncestorAndLocalSourceObjects`, and then the (one current) caller to this function just says `ns.GetParent().GetAncestorSourceObjects()` instead of `ns.GetAncestorSourceObjects()`? You could change this function to return nil if `ns` is nil.

This would simplify the implementation and reduce the confusion between having _three_ similar functions instead of just two. Maybe at that point, you could have `GetLocalSourceObjects` and `GetAncestorSourceObjects`.

Also, [according to the Intro to Algorithms book](https://stackoverflow.com/a/3078231/158143), a node considered to be one of its own ancestors. So this is a defensible choice and also matches the behaviour of `ns.AncestryNames()` (maybe we should change that to `AncestorNames`).

incubator/hnc/internal/forest/namespaceobjects.go
```diff
@@ -5,76 +5,77 @@ import (
...
 }
 
-// GetSource returns the original copy in the ancestors if it exists.
-// Otherwise, return nil.
-func (ns *Namespace) GetSource(gvk schema.GroupVersionKind, name string) *unstructured.Unstructured {
-	pos := ns.GetPropagatedObjects(gvk)
+// GetAncestorSourceObject returns the source object in the ancestors if it
+// exists. Otherwise, return nil.
+func (ns *Namespace) GetAncestorSourceObject(gvk schema.GroupVersionKind, name string) *unstructured.Unstructured {
+	pos := ns.GetAncestorSourceObjects(gvk)
```
adrianludwin, line 78: I don't think this is correct anymore because it doesn't account for the fact that the most recent source object with the same name might not have been propagated (and this will be even more important with exceptions). This should probably return a list, starting with any object in the most recent ancestor.

You could even combine this with `GetAncestorSourceObjects` (since this is no longer returning a single object anymore). If `name` is empty, it returns everything; otherwise it _only_ matches objects with the same name.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -332,11 +333,13 @@ func (r *ObjectReconciler) syncObject(ctx context.Context, log logr.Logger, inst
...
 	// see if there's a conflicting source.
-	srcInst := r.Forest.Get(inst.GetNamespace()).GetSource(r.GVK, inst.GetName())
+	srcInst := r.Forest.Get(inst.GetNamespace()).GetAncestorSourceObject(r.GVK, inst.GetName())
 
 	// The object is a source without conflict if a copy of the source is not
-	// found in the forest or itself is found.
-	if srcInst == nil || srcInst.GetNamespace() == inst.GetNamespace() {
+	// found in the forest or itself is found. There won't be any conflict for
+	// types with sync modes other than 'Propagate' either.
+	mode := r.Forest.GetTypeSyncer(inst.GroupVersionKind()).GetMode()
```
adrianludwin, line 341: Can use `r.GVK` instead of `inst.GroupVersionKind()` here.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -332,11 +333,13 @@ func (r *ObjectReconciler) syncObject(ctx context.Context, log logr.Logger, inst
...
+	if srcInst == nil || srcInst.GetNamespace() == inst.GetNamespace() || mode != api.Propagate {
```
adrianludwin, line 342: How could `srcInst.GetNamespace() == inst.GetNamespace()` today? Can't we get rid of this clause?

I also wonder if this code would now be written more naturally in the reverse (I had a bit of a hard time following the logic):

```
if srcInst != nil && mode == api.Propagate {
   // Since this is a conflict, overwrite
   return r.syncPropagated(ctx, log, inst)
}

// This object is not being overwritten; register the new source object and don't take any other action.
r.syncSource(ctx, log, inst)
return actionNop, nil
```

Or maybe even combine both calls in syncPropagated to one in this function by changing `hasPropagatedLabel` to `isPropagated`? That way you don't then have to "rediscover" the source object inside `syncPropagated` as we do today. E.g.:

```
if srcInst, isPropagated := r.isPropagated(inst); isPropagated {
  // srcInst can be nil if the source was removed
  return r.syncPropagated(ctx, log, srcInst, inst)
}
r.syncSource(ctx, log, inst)
return actionNop, nil
```
```

**Review by adrianludwin — COMMENTED** [review]
This is my review for the first commit only. Overall I really like it; I've left some comments that could simplify it even further (and possibly identifies one bug).

File: incubator/hnc/internal/validators/hncconfig.go [hunk 5]
```diff
@@ -141,6 +147,171 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+		return deny(metav1.StatusReasonConflict, printMsgWithConflicts(conflicts))
+	}
+	return allow("")
+}
+
+// checkConflictsForGVK checks object naming conflicts for a specific type.
+func (c *HNCConfig) checkConflictsForGVK(gvk schema.GroupVersionKind, conflicts conflicts) {
+	// We will keep one object set for each type, because objects with the same
+	// name but different types won't affect each other.
+	obs := objNameSet{}
```
adrianludwin, line 186: I find this logic quite hard to reason about, because we process the namespaces in a random order but maintain state between visiting each one.

As an alternative, how about this: do a _top-down_ traversal with the only information being passed from parent to child being the list of objects and the namespaces that they're from (since not all source objects are propagated, especially with exceptions, there can be multiple ancestors with objects with the same name). E.g. something like this:

```
// ancestorObjects maps an object name to the ancestor namespace(s) in which it's defined
type ancestorObjects map[string][]string
func (a ancestorObjects) copy() ancestorObject {... makes a copy ...}

func (...) checkConflictsForGVK(gvk) []string {
  conflicts := []string{}
  for _, ns := range c.Forest.GetRoots() {
    conflicts = append(conflicts, c.checkConflictsForNS(gvk, ancestorObjects{}, ns)
  }
}

func (...) checkConflictsForNS(gvk, ao, ns) []string {
  conflicts := []string{}
  objs := ao.copy()
  for _, o := ns.GetSourceObjects(gvk) {
    onm := o.GetName()
    if len(objs[onm]) > 0 {
      // TODO: check if this is a real conflict
      conflicts = append(conflicts, fmt.Sprintf("Object %q in namespace %q overwrites object in %q", onm, <todo>, ns.GetName())
      continue
    }
    objs[onm] = append(objs[onm], ns.GetName())
  }
  for _, cns := ns.GetChildren() {
    conflicts = append(conflicts, r.checkConflictsForNS(gvk, objs, cns)
  }
  return cns
}
```

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -197,6 +201,56 @@ func TestNonRBACTypes(t *testing.T) {
 	}
 }
 
+func TestPropagateConflict(t *testing.T) {
+	// Create a forest with "a" as the parent and "b" and "c" as the children.
+	f := foresttest.Create("-aa")
+
+	tests := []struct {
+		name         string
+		inNamespaces []string
```
adrianludwin, line 210: Since each namespace name is a single character, this can just be a string (that was on purpose 🙂 )

Although with that said - when we start testing exceptions, this code could become complicated enough that we'll need maps/lists anyway.

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -197,6 +201,56 @@ func TestNonRBACTypes(t *testing.T) {
...
+			for _, nsn := range tc.inNamespaces {
+				f.Get(nsn).SetSourceObject(inst)
+			}
+			got := config.handle(context.Background(), c)
+
+			// Clean up source objects before next test case.
+			for _, nsn := range tc.inNamespaces {
+				f.Get(nsn).DeleteSourceObject(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"}, "my-creds")
+			}
+
```
adrianludwin, line 247: If you create the forest inside the test, there's nothing to clean up.

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -197,6 +201,56 @@ func TestNonRBACTypes(t *testing.T) {
...
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewGomegaWithT(t)
+			configs := []api.TypeSynchronizationSpec{
+				{APIVersion: "rbac.authorization.k8s.io/v1", Kind: "Role", Mode: "Propagate"},
+				{APIVersion: "rbac.authorization.k8s.io/v1", Kind: "RoleBinding", Mode: "Propagate"},
+				{APIVersion: "v1", Kind: "Secret", Mode: "Propagate"}}
+			c := &api.HNCConfiguration{Spec: api.HNCConfigurationSpec{Types: configs}}
+			c.Name = api.HNCConfigSingleton
+			v := fakeGVKValidator{"CronTab"}
```
adrianludwin, line 231: CronTab isn't used in this test, you can just leave it out.

```
config := &HNCConfig{validator: fakeGVKValidator{}, Forest: f}
```

incubator/hnc/internal/forest/namespaceobjects.go
yiqigao217: sgtm!

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,171 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
 
+// objNameSet is a set of namespaced objects with object names as the key. It
+// also stores if the object has been checked for naming conflict.
+type objNameSet map[string]inNamespace
+type inNamespace struct {
+	namespace string
+	checked   bool
+}
+
+// conflicts stores object naming conflicts by GVK, object name and groups of
```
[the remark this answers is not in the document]
yiqigao217: Agreed. Removed this and used the recursion below. It's much easier now.

incubator/hnc/internal/forest/namespaceobjects.go
yiqigao217: True, I updated it with a list but still starting with the top ancestor since the reconciler behavior is to propagate the top "propagatable" ancestor (with no finalizer).

incubator/hnc/internal/reconcilers/object.go
yiqigao217: Updated all occurrences.

incubator/hnc/internal/reconcilers/object.go
yiqigao217: sgtm, reversed the logic!

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -191,6 +191,29 @@ var _ = Describe("Secret", func() {
 		Eventually(objectInheritedFrom(ctx, "Role", barName, "bar-role")).Should(Equal(fooName))
 	})
 
+	It("should overwrite conflicting source with the top source that can propagate", func() {
+		// Create a 'baz-role' in 'foo' that cannot propagate because of the finalizer.
+		makeObject(ctx, "Role", fooName, "baz-role")
+		Eventually(hasObject(ctx, "Role", fooName, "baz-role")).Should(BeTrue())
+		setFinalizer(ctx, fooName, "baz-role", true)
```
yiqigao217, line 198: Added line 197 before `setFinalizer`. The previous flaky "baz-role" not found error should be fixed.

adrianludwin (commented):
A few comments for the first commit - will come back and finish in a bit (just need to take a break)

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -325,34 +327,65 @@ func (r *ObjectReconciler) syncObject(ctx context.Context, log logr.Logger, inst
...
 	}
 
-	// Since there's a conflict that another source with the same name is found in
-	// the ancestors, this instance will be treated as propagated objects and will
-	// be overwritten by the source in the ancestor.
-	return r.syncPropagated(ctx, log, inst)
+	// If there's a conflicting source in the ancestors (excluding itself) and the
+	// the type has 'Propagate' mode, the object will be overwritten.
+	mode := r.Forest.GetTypeSyncer(r.GVK).GetMode()
+	if mode == api.Propagate && r.hasConflictingSource(log, inst) {
```
adrianludwin, line 351: Does it make sense to a) return the conflicting source here and b) return it from `shouldSyncAsPropagated` so we don't have to find this source a second time? I'm not worried about the runtime, it just seems a touch unfortunate to go through this logic twice.

But if it makes the code more convoluted, don't worry about it.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -325,34 +327,65 @@ func (r *ObjectReconciler) syncObject(ctx context.Context, log logr.Logger, inst
...
+		return true
+	}
+
+	return false
+}
+
+// hasConflictingSource returns true if there's any conflicting source in the
+// ancestors that would cause overwriting this source object.
+func (r *ObjectReconciler) hasConflictingSource(log logr.Logger, inst *unstructured.Unstructured) bool {
```
adrianludwin, line 360: Should this just be:

```
return r.getTopSourceToPropagate(log, inst) != nil
```
?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -382,9 +415,29 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+		}
+		return obj
+	}
+	return nil
+}
+
 // syncSource syncs the copy in the forest with the current source object. If there's a change,
 // enqueue all the descendants to propagate the new source.
 func (r *ObjectReconciler) syncSource(ctx context.Context, log logr.Logger, src *unstructured.Unstructured) {
+	// Update or create a copy of the source object in the forest.
```
adrianludwin, line 437: Maybe comment why we're doing this even if we're not propagating it.

incubator/hnc/internal/forest/namespaceobjects.go
```diff
@@ -5,80 +5,76 @@ import (
...
-// GetPropagatedObjects returns all original copies in the ancestors.
-func (ns *Namespace) GetPropagatedObjects(gvk schema.GroupVersionKind) []*unstructured.Unstructured {
-	o := []*unstructured.Unstructured{}
+// GetAncestorSourceObjects returns all source objects with the specified name
+// in the ancestors (including itself) from top down. If the name is not
+// specified, all the source objects in the ancestors will be returned.
+func (ns *Namespace) GetAncestorSourceObjects(gvk schema.GroupVersionKind, name string) []*unstructured.Unstructured {
+	if ns == nil {
+		return nil
+	}
```
adrianludwin, line 59: Maybe comment why we have this here - eg because it allows you to say `ns.GetParent().GetAncestorSourceObjects()` without caring if the namespace is a root.

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -246,15 +246,15 @@ func (v *Hierarchy) getConflictingObjectsOfType(gvk schema.GroupVersionKind, new
 	// Get all the source objects in the new ancestors that would be propagated
 	// into the descendants.
 	newAnsSrcObjs := make(map[string]bool)
-	for _, o := range newParent.GetPropagatingObjects(gvk) {
+	for _, o := range newParent.GetAncestorSourceObjects(gvk, "") {
 		newAnsSrcObjs[o.GetName()] = true
 	}
```
adrianludwin, line 251: TODO (maybe for Ginny): this doesn't obey the shouldPropagate rules, but it probably should (and will definitely have to once we have exceptions).

adrianludwin (commented):
Ok I'm now fully finished with the first commit :) On to the second commit

incubator/hnc/internal/forest/forest.go
```diff
@@ -96,6 +96,17 @@ func (f *Forest) GetNamespaceNames() []string {
 	return names
 }
 
+// GetRoots returns all the root namespaces in the cluster.
```
adrianludwin, line 99: Comment that any cycles will be omitted.

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+
+	// Get types that are changed from other modes to "Propagate" mode.
+	gvks := c.getNewPropagateTypes(inst)
+
+	// Check if user-created objects would be overwritten by these mode changes.
+	for gvk, _ := range gvks {
+		conflicts := c.checkConflictsForGVK(gvk)
+		if len(conflicts) != 0 {
+			msg := fmt.Sprintf("Cannot update configuration because setting type %q to 'Propagate' mode would overwrite user-created object(s):\n", gvk)
+			msg += strings.Join(c.checkConflictsForGVK(gvk), "\n")
```
adrianludwin, line 162: Should this just be `conflicts` instead of calling the function again?

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+
+// ancestorObjects maps an object name to the ancestor namespace(s) in which
+// it's defined.
+type ancestorObjects map[string][]string
+
+// checkConflictsForGVK looks for conflicts from top down for each tree.
+func (c *HNCConfig) checkConflictsForGVK(gvk schema.GroupVersionKind) []string {
+	conflicts := []string{}
+	for _, ns := range c.Forest.GetRoots() {
+		conflicts = append(conflicts, c.checkConflictsForNS(gvk, ancestorObjects{}, ns)...)
```
adrianludwin, line 179: Nit: consider checkConflictsForTree?

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+}
+
+func (a ancestorObjects) copy() ancestorObjects {
+	copy := ancestorObjects{}
+	for k, v := range a {
+		copy[k] = v
+	}
+	return copy
+}
+
```
adrianludwin, line 234: Nit: move this to just after where `ancestorObjects` is defined (or move `ancestorObjects` down here)

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+	}
+	return conflicts
+}
+
+func (c *HNCConfig) checkConflictsForNS(gvk schema.GroupVersionKind, ao ancestorObjects, ns *forest.Namespace) []string {
+	conflicts := []string{}
+	objs := ao.copy()
+	for _, o := range ns.GetSourceObjects(gvk) {
+		onm := o.GetName()
+		objs[onm] = append(objs[onm], ns.Name())
```
adrianludwin, line 189: Consider `objs.add(ns, onm)`?

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+		// If there are more than just the one (current) namespace we just added,
+		// there may be conflicts.
+		if len(objs[onm]) > 1 {
```
adrianludwin, line 192: Consider `objs.hasConflict(onm)`?

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+			// Currently the top conflicting ancestor would overwrite the object.
+			// TODO: check if this is a real conflict
+			conflicts = append(conflicts, fmt.Sprintf("  Object %q in namespace %q would overwrite the one in %q", onm, objs[onm][0], ns.Name()))
```
adrianludwin, line 195: Consider `objs.source(onm)` or `objs.top(onm)` or something like that instead of `objs[onm][0]`?

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -141,6 +147,91 @@ func (c *HNCConfig) validateType(ctx context.Context, t api.TypeSynchronizationS
...
+		}
+	}
+	for _, cnm := range ns.ChildNames() {
```
adrianludwin, line 198: Comment on why this can't cause a cycle :D

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -48,7 +51,8 @@ func TestDeletingOtherObject(t *testing.T) {
 }
 
 func TestRBACTypes(t *testing.T) {
-	config := &HNCConfig{}
+	f := forest.NewForest()
+	config := &HNCConfig{Forest: f}
```
adrianludwin, line 55: Delete?

adrianludwin (commented):
Second commit lgtm apart from some relatively minor suggestions

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -274,14 +274,16 @@ func (r *ObjectReconciler) skipNamespace(ctx context.Context, log logr.Logger, i
...
 		return actionUnknown
 	}
 
+	ns := r.Forest.Get(inst.GetNamespace())
 	// If it's a source, it must have been deleted. Update the forest and enqueue all its
 	// descendants, but there's nothing else to do.
-	if r.forestHasSource(inst) {
+	if ns.HasSourceObject(r.GVK, inst.GetName()) {
+		ns.DeleteSourceObject(r.GVK, inst.GetName())
 		r.syncUnpropagatedSource(ctx, log, inst)
```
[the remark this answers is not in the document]
yiqigao217: It has reconcile HNCConfig and `enqueueDescendants()` to delete the obsolete copies.

incubator/hnc/internal/reconcilers/object.go
yiqigao217: Updated to return `srcInst`.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -325,34 +327,65 @@ func (r *ObjectReconciler) syncObject(ctx context.Context, log logr.Logger, inst
...
 
-	// Since there's a conflict that another source with the same name is found in
-	// the ancestors, this instance will be treated as propagated objects and will
-	// be overwritten by the source in the ancestor.
-	return r.syncPropagated(ctx, log, inst)
+	// If there's a conflicting source in the ancestors (excluding itself) and the
+	// the type has 'Propagate' mode, the object will be overwritten.
+	mode := r.Forest.GetTypeSyncer(r.GVK).GetMode()
+	if mode == api.Propagate && r.hasConflictingSource(log, inst) {
+		return true
```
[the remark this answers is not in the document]
yiqigao217: sgtm updated.

incubator/hnc/internal/reconcilers/object.go
yiqigao217: Removed this dup function.

incubator/hnc/internal/reconcilers/object.go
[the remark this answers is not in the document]
adrianludwin: No, I just mean inline its features into this function since it does so little.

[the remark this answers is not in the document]
yiqigao217: `syncUnpropagatedSource` func is used twice and it contains two parts. I feel better to keep this logic stand alone, WDYT?
```
// syncUnpropagatedSource enqueues any propagated copies of the source.
//
// The method can be called when the source was deleted by users, or if it will no longer be
// propagated, e.g. because the user's changed the sync mode in the HNC Config.
func (r *ObjectReconciler) syncUnpropagatedSource(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) {
	// Signal the config reconciler for reconciliation because it is possible that the source object is
	// deleted on the apiserver.
	hnccrSingleton.requestReconcile("possible deleted source object")

	r.enqueueDescendants(ctx, log, inst, "possibly deleted source object")
}
```

incubator/hnc/internal/validators/hncconfig.go
yiqigao217: Ah right.

incubator/hnc/internal/validators/hncconfig_test.go
yiqigao217: It will have nil pointer error if we have forest as empty.

incubator/hnc/internal/reconcilers/object.go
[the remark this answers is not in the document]
adrianludwin: Ah sgtm

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 

Hey Ryan, this is a big PR with two commits. I'd recommend reviewing them individually:
1. Commit #1 is about _always_ saving _all_ source objects in the forest, even if we're not propagating objects. This enables...
2. Commit #2 which checks the forest for objects that are about to be overwritten if we change the mode to Propagate.

Note that #2 will only have an effect if the reconciler is currently not in the Ignore mode for that type of object - which it typically will be. A future change will guide users to _first_ change to the Remove mode, and only _later_ change to the Propagate mode.

yiqigao217:
Update: Passed all e2e tests.

Initially I got:
```
Summarizing 3 Failures:

1.[Fail] Demo [It] Should test basic functionalities in demo 
/usr/local/google/home/yiqigao/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils/testutils.go:110

2. [Fail] HNC should delete and create a new Rolebinding instead of updating it [It] Should delete and create a Rolebinding when HNC is undeployed - issue #798 
/usr/local/google/home/yiqigao/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils/testutils.go:110

3. [Fail] When deleting CRDs [It] should not delete subnamespaces 
/usr/local/google/home/yiqigao/go/src/sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils/testutils.go:69

Ran 29 of 29 Specs in 1030.354 seconds
FAIL! -- 26 Passed | 3 Failed | 0 Pending | 0 Skipped
```
1 & 2 passed when I ran it manually. 3 passed by itself if it's a stand-alone test (with `Fit`).

The 1st demo failure also failed on master. I suspect there's a bug in the test that it should use `RoleBinding` instead of `Role` here: https://github.com/kubernetes-sigs/multi-tenancy/blob/f990be3264cac6edb7618880b859522d25d40989/incubator/hnc/test/e2e/demo_test.go#L55

/assign @adrianludwin 
/cc @GinnyJI

adrianludwin:
Yes, @GinnyJI is looking into the e2e test failures on `master`. I'm comfortable that this change is good and doesn't break anything.

PR #1115, opened 2020-09-17
Cherrypick: Handle source object overwriting conflict

yiqigao217:
See issue at #1076 

The fix should include 4 changes, but we only include the first 2 fixes in this cherry-pick into HNC v0.5.
1. Object webhook (see #1079)
2. HierarchyConfiguration wehbook
3. Hncconfig webhook (too complicated to do a cherry-pick, see #1102)
4. HNC exceptions (not yet implemented, so irrelevant to v0.5)

Tested by `make test`

### DIFF
--- a/incubator/hnc/internal/validators/hierarchy.go
+++ b/incubator/hnc/internal/validators/hierarchy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/go-logr/logr"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
@@ -12,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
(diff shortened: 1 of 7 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 
/cc @GinnyJI 

This is the cherry-pick of the Object and HierarchyConfig webhooks to HNC v0.5, to prevent creating conflicting objects and changing parents that would cause this conflict.

adrianludwin:
This has four commits, but I thought it was only supposed to have two. Is that intended?

yiqigao217:
> This has four commits, but I thought it was only supposed to have two. Is that intended?

Yes, besides the object & hc webhook changes, there are also one for object reconciler behavior change to overwriting conflicting objects, and another one to make the test less flaky with a longer sleep time for the new tests included.

adrianludwin:
Given that this is a change onto a stable branch, can you please do a bit more testing on it? It would be nice to see the full suite of e2e tests at least. Thanks!

Other than that this lgtm. @rjbez17 , what do you think of checking this into v0.5.3? It solves _most_ of the problem that we identified that you can have silently overridden policies while not being so dangerous that they all get overwritten. The one dangerous part of this change is that if you turn on propagation for a _new_ object type, any conflicting objects will get swept away; we'll fix that in v0.6 but it's too hard to fix in v0.5.x. But given the overall stability of HNC, I think that's an acceptable compromise.

yiqigao217:
> Given that this is a change onto a stable branch, can you please do a bit more testing on it? It would be nice to see the full suite of e2e tests at least. Thanks!

Sure, will do. BTW just confirming that we don't have e2e test checked in in v0.5, I will have to manually add the e2e test files on my local branch to test, right? @adrianludwin

adrianludwin:
As discussed offline, at least some of them are there:
https://github.com/kubernetes-sigs/multi-tenancy/tree/hnc-v0.5/incubator/hnc/test/e2e.
You may need to say `make e2e-test` instead of `make test-e2e` IIRC.

yiqigao217:
Update: among 24 e2e tests, 22 passed and 2 failed:
```
[Fail] Issues [It] Should remove obsolete conditions CannotPropagateObject and CannotUpdateObject - issue #328 
[Fail] Issues [It] Should clear CannotUpdate conditions in descendants when a hierarchy changes - issue #605 
```
They both failed on webhook internal error on `kubectl hns set sub-child --root`:
```
Running:  [kubectl hns set sub-child --root]
Output:  Unsetting the parent of sub-child (was previously child)

Could not update the hierarchical configuration of sub-child.
Reason: Internal error occurred: failed calling webhook "hierarchyconfigurations.hnc.x-k8s.io": Post https://hnc-webhook-service.hnc-system.svc:443/validate-hnc-x-k8s-io-v1alpha1-hierarchyconfigurations?timeout=30s: stream error: stream ID 13139; INTERNAL_ERROR
```

Please note other HC changes such as `kubectl hns set sub-child --parent child` all passed. I'm looking into the issue now.

adrianludwin:
INTERNAL_ERROR sounds like it could be a real error - check the logs? All
tests were passing robustly on 0.5.2.

yiqigao217:
Update: the e2e test error is fixed and cherry-picked into this PR too. I reran the e2e tests focusing (`FIt()`) on the two previously failed tests and it passed.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 

This is a larger change (both LoC and behaviour) than I'd usually want to put on a branch, but I think it makes sense. Our current behaviour re conflicting objects is clearly broken and this seems like a reasonable approach. Ideally a couple of us will play around with this before we release 0.5.3 to see if any other issues crop up with it.

@rjbez17 , wdyt?

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -210,9 +212,59 @@ func (v *Hierarchy) checkParent(ns, curParent, newParent *forest.Namespace) admi
 		return deny(metav1.StatusReasonConflict, "Illegal parent: "+reason)
 	}
 
+	// Prevent overwriting source objects in the descendants after the hierarchy change.
+	if co := v.getConflictingObjects(newParent, ns); len(co) != 0 {
```
rjbez17, line 216: This seems like a pretty fundamental change in a patch release no? Granted we aren't k8s here, their definition of what is allowed in a patch release: https://github.com/kubernetes/community/blob/master/contributors/design-proposals/release/versioning.md#patch-releases

I totally see why we want this in here, but I'm a bit concerned with setting the precedence.

incubator/hnc/internal/validators/object.go
```diff
@@ -106,8 +108,14 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	oldSource, oldInherited := metadata.GetLabel(oldInst, api.LabelInheritedFrom)
 	newSource, newInherited := metadata.GetLabel(inst, api.LabelInheritedFrom)
 
-	// If the object wasn't and isn't inherited, it's none of our business.
+	// If the object wasn't and isn't inherited, we will check to see if the
+	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
+		if yes, dnses := o.hasConflict(inst); yes {
```
rjbez17, line 114: same as other comment.

incubator/hnc/internal/validators/hierarchy.go
adrianludwin: Agreed, this is a large change. Ordinarily I wouldn't want to patch it, but a) it will be a while until v0.6 still and b) I don't think we have enough customers to be _too_ worried yet.

Maybe ask the mailing list and slack channel what they think?

rjbez17: Yeah, I think that is a fair compromise.

adrianludwin: I think we have enough feedback that we can go ahead with this now. But we'll avoid this as HNC matures, and we'll try hard not to get stuck with long-lived branches as well.

PR #1121, opened 2020-09-18
Fix a nil pointer issue in HC webhook when setting as root

yiqigao217:
The bug was found when running e2e tests that the webhook got
INTERNAL_ERROR when setting a namespace as root. Fix it by checking if
the new parent is nil. If yes, early exit since it's impossible to
introduce new naming conflicts.

Add a new unit test case for it. The test failed before and passed after
the change.

Tested by make test and manually.


yiqigao217:
/assign @adrianludwin 
/cc @rjbez17 

This is a bug fix. The bug was found when running e2e tests in #1115 . Will cherry-pick this PR into #1115 after it's merged.

adrianludwin (commented):
/lgtm
/approve

PR #1122, opened 2020-09-18
Add exception when using 'select'

GinnyJI:
Add `labels` field in namespace struct within Forest. When propagating
an object, we first check if a `select` annotation exists. If this
annotation exists and doesn't match the current namespace, this object
will not be propagated.

Modify the object_test.go to use `Role` instead of `Secret` in
propagating tests.

Tested: make test

### DIFF
--- a/incubator/hnc/api/v1alpha2/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha2/hierarchy_types.go
@@ -31,11 +31,15 @@ const (
 
 // Constants for labels and annotations
 const (
-	MetaGroup                = "hnc.x-k8s.io"
-	LabelInheritedFrom       = MetaGroup + "/inheritedFrom"
-	FinalizerHasSubnamespace = MetaGroup + "/hasSubnamespace"
-	LabelTreeDepthSuffix     = ".tree." + MetaGroup + "/depth"
-	AnnotationManagedBy      = MetaGroup + "/managedBy"
+	MetaGroup                 = "hnc.x-k8s.io"
+	LabelInheritedFrom        = MetaGroup + "/inheritedFrom"
+	FinalizerHasSubnamespace  = MetaGroup + "/hasSubnamespace"
+	LabelTreeDepthSuffix      = ".tree." + MetaGroup + "/depth"
+	AnnotationManagedBy       = MetaGroup + "/managedBy"
+	AnnotationPropagatePrefix = "propagate." + MetaGroup
+
+	AnnotationSelector     = AnnotationPropagatePrefix + "/select"
+	AnnotationTreeSelector = AnnotationPropagatePrefix + "/treeSelect"
 )
 
 // Condition codes. *All* codes must also be documented in the comment to Condition.Code, be
--- a/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
+++ b/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
@@ -20,14 +20,10 @@ spec:
       description: Hierarchy is the Schema for the hierarchies API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
           type: string
         kind:
-          description: 'Kind is a string value representing the REST resource this
-            object represents. Servers may infer this from the endpoint the client
-            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
           type: string
         metadata:
           type: object
@@ -35,8 +31,7 @@ spec:
           description: HierarchySpec defines the desired state of Hierarchy
           properties:
             allowCascadingDelete:
-              description: AllowCascadingDelete indicates if the subnamespaces of
-                this namespace are allowed to cascading delete.
+              description: AllowCascadingDelete indicates if the subnamespaces of this namespace are allowed to cascading delete.
               type: boolean
             parent:
               description: Parent indicates the parent of this namespace, if any.
@@ -46,20 +41,17 @@ spec:
           description: HierarchyStatus defines the observed state of Hierarchy
           properties:
             children:
-              description: Children indicates the direct children of this namespace,
-                if any.
+              description: Children indicates the direct children of this namespace, if any.
               items:
                 type: string
               type: array
             conditions:
-              description: Conditions describes the errors and the affected objects,
-                if any.
+              description: Conditions describes the errors and the affected objects, if any.
               items:
                 description: Condition specifies the condition and the affected objects.
                 properties:
                   affects:
-                    description: Affects is a list of group-version-kind-namespace-name
-                      that uniquely identifies the object(s) affected by the condition.
+                    description: Affects is a list of group-version-kind-namespace-name that uniquely identifies the object(s) affected by the condition.
                     items:
                       description: AffectedObject defines uniquely identifiable objects.
                       properties:
@@ -76,56 +68,10 @@ spec:
                       type: object
                     type: array
                   code:
-                    description: "Describes the condition in a machine-readable string
-                      value. The currently valid values are shown below, but new values
-                      may be added over time. This field is always present in a condition.
-                      \n All codes that begin with the prefix `Crit` indicate that
-                      all HNC activities (e.g. propagating objects, updating labels)
-                      have been paused in this namespaces. HNC will resume updating
-                      the namespace once the condition has been resolved. Non-critical
-                      conditions typically indicate some kind of error that HNC itself
-                      can ignore, but likely indicates that the hierarchical structure
-                      is out-of-sync with the users' expectations. \n If the validation
-                      webhooks are working properly, there should typically not be
-                      any conditions on any namespaces, although some may appear transiently
-                      when the HNC controller is restarted. These should quickly resolve
-                      themselves (<30s). However, validation webhooks are not perfect,
-                      especially if multiple users are modifying the same namespace
-                      trees quickly, so it's important to monitor for critical conditions
-                      and resolve them if they arise. See the user guide for more
-                      information. \n Currently, the supported values are: \n - \"CritParentMissing\":
-                      the specified parent is missing and the namespace is an orphan.
-                      \n - \"CritCycle\": the namespace is a member of a cycle. For
-                      example, if namespace B says that its parent is namespace A,
-                      but namespace A says that its parent is namespace B, then A
-                      and B are in a cycle with each other and both of them will have
-                      the CritCycle condition. \n - \"CritDeletingCRD\": The HierarchyConfiguration
-                      CRD is being deleted. No more objects will be propagated into
-                      or out of this namespace. It is expected that the HNC controller
-                      will be stopped soon after the CRDs are fully deleted. \n -
-                      \"CritAncestor\": a critical error exists in an ancestor namespace,
-                      so this namespace is no longer being updated either. \n - \"SubnamespaceAnchorMissing\":
-                      this namespace is a subnamespace, but the anchor referenced
-                      in its `subnamespaceOf` annotation does not exist in the parent.
-                      \n - \"CannotPropagateObject\": this namespace contains an object
-                      that couldn't be propagated *out* of this namespace, to one
-                      or more of this namespace's descendants. If the object couldn't
-                      be propagated to *any* descendants - for example, because it
-                      has a finalizer on it (HNC can't propagate objects with finalizers),
-                      the `Affects` field will point to the object in this namespace.
-                      Otherwise, if it couldn't be propagated to *some* descendants,
-                      `Affects` will contain a list of the objects in those descendants
-                      that couldn't be created or updated. \n - \"CannotUpdateObject\":
-                      this namespace has an object that couldn't be propagated *into*
-                      this namespace - that is, it couldn't be created in the first
-                      place, or it couldn't be updated. The `Affects` field will point
-                      to the source object, which will always be in a namespace that's
-                      an ancestor of this namespace."
+                    description: "Describes the condition in a machine-readable string value. The currently valid values are shown below, but new values may be added over time. This field is always present in a condition. \n All codes that begin with the prefix `Crit` indicate that all HNC activities (e.g. propagating objects, updating labels) have been paused in this namespaces. HNC will resume updating the namespace once the condition has been resolved. Non-critical conditions typically indicate some kind of error that HNC itself can ignore, but likely indicates that the hierarchical structure is out-of-sync with the users' expectations. \n If the validation webhooks are working properly, there should typically not be any conditions on any namespaces, although some may appear transiently when the HNC controller is restarted. These should quickly resolve themselves (<30s). However, validation webhooks are not perfect, especially if multiple users are modifying the same namespace trees quickly, so it's important to monitor for critical conditions and resolve them if they arise. See the user guide for more information. \n Currently, the supported values are: \n - \"CritParentMissing\": the specified parent is missing and the namespace is an orphan. \n - \"CritCycle\": the namespace is a member of a cycle. For example, if namespace B says that its parent is namespace A, but namespace A says that its parent is namespace B, then A and B are in a cycle with each other and both of them will have the CritCycle condition. \n - \"CritDeletingCRD\": The HierarchyConfiguration CRD is being deleted. No more objects will be propagated into or out of this namespace. It is expected that the HNC controller will be stopped soon after the CRDs are fully deleted. \n - \"CritAncestor\": a critical error exists in an ancestor namespace, so this namespace is no longer being updated either. \n - \"SubnamespaceAnchorMissing\": this namespace is a subnamespace, but the anchor referenced in its `subnamespaceOf` annotation does not exist in the parent. \n - \"CannotPropagateObject\": this namespace contains an object that couldn't be propagated *out* of this namespace, to one or more of this namespace's descendants. If the object couldn't be propagated to *any* descendants - for example, because it has a finalizer on it (HNC can't propagate objects with finalizers), the `Affects` field will point to the object in this namespace. Otherwise, if it couldn't be propagated to *some* descendants, `Affects` will contain a list of the objects in those descendants that couldn't be created or updated. \n - \"CannotUpdateObject\": this namespace has an object that couldn't be propagated *into* this namespace - that is, it couldn't be created in the first place, or it couldn't be updated. The `Affects` field will point to the source object, which will always be in a namespace that's an ancestor of this namespace."
                     type: string
                   msg:
-                    description: A human-readable description of the condition, if
-                      the `code` and `affects` fields are not sufficiently clear on
-                      their own.
+                    description: A human-readable description of the condition, if the `code` and `affects` fields are not sufficiently clear on their own.
                     type: string
                 required:
                 - code
--- a/incubator/hnc/internal/forest/namespace.go
+++ b/incubator/hnc/internal/forest/namespace.go
@@ -27,6 +27,9 @@ type Namespace struct {
 	exists               bool
 	allowCascadingDelete bool
 
+	// labels store the original namespaces' labels
+	labels map[string]string
+
 	// originalObjects store the objects created by users, identified by GVK and name.
 	// It serves as the source of truth for object controllers to propagate objects.
 	originalObjects objects
@@ -86,6 +89,17 @@ func (ns *Namespace) UnsetExists() bool {
 	ns.exists = false
 	ns.clean() // clean up if this is a useless data structure
 	return changed
+}
+
+func (ns *Namespace) GetLabels() map[string]string {
+	return ns.labels
+}
+
+func (ns *Namespace) SetLabels(labels map[string]string) {
+	ns.labels = make(map[string]string)
+	for key, val := range labels {
+		ns.labels[key] = val
+	}
 }
 
 // clean garbage collects this namespace if it has a zero value.
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -398,6 +398,7 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 		anc = anc.Parent()
 		depth++
 	}
+	ns.SetLabels(nsInst.Labels)
 }
 
 func (r *HierarchyConfigReconciler) syncConditions(log logr.Logger, inst *api.HierarchyConfiguration, ns *forest.Namespace, deletingCRD, hadCrit bool) {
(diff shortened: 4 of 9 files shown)


GinnyJI:
tracks #1123

incubator/hnc/internal/forest/namespace.go
```diff
@@ -27,6 +27,10 @@ type Namespace struct {
 	exists               bool
 	allowCascadingDelete bool
 
+	// labels store the original namespaces's labels, and we add the propogate label to identify itself
```
yiqigao217, line 30: r/namespaces's/namespaces'

incubator/hnc/internal/forest/namespace.go
```diff
@@ -27,6 +27,10 @@ type Namespace struct {
...
+	// in the exception process
+	labels map[string]string
```
yiqigao217, line 32: Do we store all the labels in the forest or just the propagate labels?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -398,6 +404,9 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 		anc = anc.Parent()
 		depth++
 	}
+
+	labels[propagatePrefix+nsInst.Name] = "true"
```
yiqigao217, line 408: I only saw `propagate.hnc.x-k8s.io/all=true` or `propagate.hnc.x-k8s.io/none=true` annotations in the [design doc](https://docs.google.com/document/d/17J8icBEDvLLoPT4kQ4ArZcCerRweDY-TpJ48DJKpHJ0/edit#). What does `propagate.hnc.x-k8s.io/current_ns_name=true` do?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -53,22 +53,22 @@ var _ = Describe("Secret", func() {
...
 
 		// Create a secret that does NOT propogate to bazName
 		a := map[string]string{"propagate.hnc.x-k8s.io/select": "!propagate.hnc.x-k8s.io/" + bazName}
-		makeObjectWithAnnotation(ctx, "Secret", fooName, "fooSecret", a)
+		makeObjectWithAnnotation(ctx, "Role", fooName, "testrole", a)
 
-		Eventually(hasObject(ctx, "Secret", barName, "fooSecret")).Should(BeTrue())
-		Consistently(hasObject(ctx, "Secret", bazName, "fooSecret")).Should(BeFalse())
+		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
+		Consistently(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
```
yiqigao217, line 71: Just a question here, what's the difference between `Eventually` and `Consistently`?

incubator/hnc/internal/forest/namespace.go
GinnyJI: done

incubator/hnc/internal/reconcilers/hierarchy_config.go
GinnyJI: I found out that I had to add the `propagate.hnc.x-k8s.io/current_ns_name` label manually for each ns, otherwise the selector can't identify whether to propagate or not. The `true` value is just to say that this ns has the name `current_ns_name`

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: `Eventually` is to say we are able to get value A at some point within a period of time. `Consistently` means we are able to get value A for a stable period of time (e.g. the `False` doesn't get to changed to `True` after like 1s or so)

yiqigao217: Thanks for the explanation!

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: IIUC the selector will look for `current_ns_name.tree.hnc.x-k8s.io/depth: 0` label instead of a new `propagate.hnc.x-k8s.io/current_ns_name: true` label. WDYT? @adrianludwin

adrianludwin: Label design needs to be done _very_ carefully. @GinnyJI , why won't the depth labels that Yiqi mentioned work?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -43,6 +43,8 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/stats"
 )
 
+const propagatePrefix = "propagate.hnc.x-k8s.io/"
```
adrianludwin, line 46: IMO this should go into the API package, since the prefix is part of the API.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -398,6 +404,9 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 		anc = anc.Parent()
 		depth++
 	}
+
+	labels[propagatePrefix+nsInst.Name] = "true"
+	ns.SetLabels(labels)
```
adrianludwin, line 409: This doesn't look correct to me. It looks like we're _only_ storing the labels that HNC itself has created, but we should be storing _all_ the labels on the namespace, including the ones set by the users.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -53,6 +54,9 @@ const (
 	actionNop     syncAction = "no-op"
 
 	unknownSourceNamespace = "<unknown-source-namespace>"
+
+	selectorAnnot     = "propagate.hnc.x-k8s.io/select"
+	treeSelectorAnnot = "propagate.hnc.x-k8s.io/treeSelect"
```
adrianludwin, line 59: These constants should be in the API package. Also, you shouldn't be repeating yourself so much - "hnc.x-k8s.io" is already called something like `api.MetaGroup`, and you should have a constant for the `propagate.` part as well.

adrianludwin: https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/api/v1alpha2/hierarchy_types.go#L32

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -359,18 +363,20 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
-	// source instance. Note that DeepEqual could return `true` even if the object doesn't exist if
-	// the source object is trivial (e.g. a completely empty ConfigMap).
-	if !exists ||
-		!reflect.DeepEqual(object.Canonical(inst), object.Canonical(srcInst)) ||
-		inst.GetLabels()[api.LabelInheritedFrom] != srcInst.GetNamespace() {
-		metadata.SetLabel(inst, api.LabelInheritedFrom, srcInst.GetNamespace())
-		return actionWrite, srcInst
+	selector := r.getSelector(ctx, log, srcInst)
+
+	if selector == nil || selector.Matches(labels.Set(r.Forest.Get(inst.GetNamespace()).GetLabels())) {
```
adrianludwin, line 370: I think an early exit might be much more appropriate here (e.g. above line 366). This is very hard to read, with two levels of nested `if` statements (and also this line is much too long, IMO - you can/should introduce some temp variables here).

Also - what happens if `srcInst` _used_ to be propagated here, but now has a new selector that prevents it from being propagated here? This will execute line 382 which says "the object already exists and doesn't need to be updated" - does that seem correct?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -382,6 +388,24 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 	return actionNop, nil
 }
 
+func (r *ObjectReconciler) getSelector(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
```
adrianludwin, line 391: `ctx` doesn't appear to be used in this function?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -382,6 +388,24 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+	selectorstr, present := annot[selectorAnnot]
+	if present == false {
+		return nil
+	}
+	labelSelector, err := v1.ParseToLabelSelector(selectorstr)
+	if err != nil {
+		log.Error(err, "Could not parse selector annotation to labelSelector")
+		return nil
+	}
+	selector, err2 := v1.LabelSelectorAsSelector(labelSelector)
```
adrianludwin, line 402: No need to introduce a new name - `err` should work fine.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -382,6 +388,24 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 	return actionNop, nil
 }
 
+func (r *ObjectReconciler) getSelector(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	annot := inst.GetAnnotations()
+	selectorstr, present := annot[selectorAnnot]
```
adrianludwin, line 393: Prefer camel case (`selectorStr`). Also, if you liked, you could say:

```
selectorStr, ok := inst.GetAnnotations()[selectorAnnot]
```

The `ok` name is often used to see if an entry exists in a map.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -382,6 +388,24 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+	if present == false {
```
adrianludwin, line 394: Style: `!present`, not `present == false`

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -382,6 +388,24 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+		return nil
+	}
+	labelSelector, err := v1.ParseToLabelSelector(selectorstr)
+	if err != nil {
+		log.Error(err, "Could not parse selector annotation to labelSelector")
+		return nil
```
adrianludwin, line 400: We need a better way to surface these errors. @yiqigao217 can you please work with Ginny to determine one?

incubator/hnc/internal/forest/namespace.go
GinnyJI: Changed to store all the labels

incubator/hnc/internal/reconcilers/hierarchy_config.go
GinnyJI: As discussed offline, the test is updated and this label is removed

incubator/hnc/internal/reconcilers/hierarchy_config.go
GinnyJI: Removed

incubator/hnc/internal/reconcilers/hierarchy_config.go
GinnyJI: Done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: It's also not used in `syncPropagated`, so I thought I need to keep it even though it's not used directly

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Discussed with Yiqi offline and left a `TODO` here. Yiqi will draft a design doc and I'll implement at another PR

incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -36,6 +36,10 @@ const (
 	FinalizerHasSubnamespace = MetaGroup + "/hasSubnamespace"
 	LabelTreeDepthSuffix     = ".tree." + MetaGroup + "/depth"
 	AnnotationManagedBy      = MetaGroup + "/managedBy"
+	AnnotationPropagate      = "propagate." + MetaGroup
```
yiqigao217, line 39: Use `AnnotationPropagatePrefix`?

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
```diff
@@ -20,23 +20,18 @@ spec:
       description: Hierarchy is the Schema for the hierarchies API
       properties:
         apiVersion:
-          description: 'APIVersion defines the versioned schema of this representation
-            of an object. Servers should convert recognized schemas to the latest
-            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
yiqigao217, line 23: You will need to rebase and run `make manifest` to remove the diffs on the `.yaml` files.

incubator/hnc/internal/forest/namespace.go
```diff
@@ -27,6 +27,10 @@ type Namespace struct {
 	exists               bool
 	allowCascadingDelete bool
 
+	// labels store the original namespaces' labels, and we add the propogate label to identify itself
```
yiqigao217, line 30: Update the comment since we don't have propagate labels now after the first push.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -377,6 +377,7 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 	// and cycles.
 	anc := ns
 	depth := 0
+
```
yiqigao217, line 380: Maybe remove this line?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -359,9 +360,20 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+
+	instLabels := labels.Set(r.Forest.Get(inst.GetNamespace()).GetLabels())
+	if selector != nil && !selector.Matches(instLabels) {
+		// The object already exists and doesn't need to be updated. This will typically happen when HNC
+		// is restarted - all the propagated objects already exist on the apiserver. Record that it exists
+		// for our statistics.
+		r.recordPropagatedObject(log, inst.GetNamespace(), inst.GetName())
+
+		// Nothing more needs to be done.
+		return actionNop, nil
```
yiqigao217, line 375: The logic seems not right to me. Why do we move this part up? I think it should be that if it doesn't match we will just early exit with `actionNop`. The rest of the logic to `actionWrite` if there's a diff and `actionNop` if there's not should remain the same.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -370,16 +382,27 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 		!reflect.DeepEqual(object.Canonical(inst), object.Canonical(srcInst)) ||
 		inst.GetLabels()[api.LabelInheritedFrom] != srcInst.GetNamespace() {
 		metadata.SetLabel(inst, api.LabelInheritedFrom, srcInst.GetNamespace())
-		return actionWrite, srcInst
 	}
+	return actionWrite, srcInst
+}
```
yiqigao217, line 387: Same as the comment above, this should still be in the if clause.

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -53,22 +53,22 @@ var _ = Describe("Secret", func() {
...
 	})
 
-	PIt("should propagate object only to selected namespace using select", func() {
+	FIt("should propagate object only to selected namespace using select", func() {
```
yiqigao217, line 62: Remove the `F` in `FIt` before checking in.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -370,16 +382,27 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
-	// The object already exists and doesn't need to be updated. This will typically happen when HNC
-	// is restarted - all the propagated objects already exist on the apiserver. Record that it exists
-	// for our statistics.
-	r.recordPropagatedObject(log, inst.GetNamespace(), inst.GetName())
-
-	// Nothing more needs to be done.
-	return actionNop, nil
+func (r *ObjectReconciler) getSelector(ctx context.Context, log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	annot := inst.GetAnnotations()
+	selectorStr, ok := annot[api.AnnotationSelector]
```
yiqigao217, line 391: Just confirming: Is it that `AnnotationTreeSelector` is not yet implemented in the PR?

GinnyJI: Yes, it's not part of this pr

incubator/hnc/api/v1alpha2/hierarchy_types.go
GinnyJI: Done

incubator/hnc/internal/reconcilers/object.go
adrianludwin: Maybe it used to be used there but was removed. I'd take it out of any place it's not (currently) needed, it's easy to add it back.

incubator/hnc/internal/forest/namespace.go
```diff
@@ -88,6 +92,14 @@ func (ns *Namespace) UnsetExists() bool {
...
 }
 
+func (ns *Namespace) GetLabels() map[string]string {
+	return ns.labels
+}
+
+func (ns *Namespace) SetLabels(labels map[string]string) {
+	ns.labels = labels
+}
+
```
adrianludwin, line 102: If we never do anything smart with these fields, maybe we should just make `ns.Labels` public?

adrianludwin: Or should we make a deep copy of the labels that get passed in so they can't be modified later? (`ns.labels = labels` is a shallow copy)

incubator/hnc/internal/reconcilers/object.go
adrianludwin: @yiqigao217 how much will this logic change with the changes you're making around object overwriting? It may not be worth fine-tuning this function as it exists today if you're about to change the way `syncObject` works.

With that said, I mostly agree with Yiqi. This should be an early-exit, but it shouldn't be a no-op - the object might exist and need to be removed. Your existing code doesn't account for this either - so you should probably add a new test which is to ensure that if you _add_ a label selector to an _existing_ object, that it gets _deleted_ from any namespace in which it no longer belongs.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -359,9 +360,20 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 	if srcInst == nil {
 		return actionRemove, nil
 	}
-
 	// If an object doesn't exist, assume it's been deleted or not yet created.
 	exists := inst.GetCreationTimestamp() != v1.Time{}
+	selector := r.getSelector(ctx, log, srcInst)
+
+	instLabels := labels.Set(r.Forest.Get(inst.GetNamespace()).GetLabels())
```
adrianludwin, line 367: This is not the `instLabels`, it's the namespace's labels. So let's call this `nsLabels`.

incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
GinnyJI: Done

incubator/hnc/internal/forest/namespace.go
GinnyJI: Done

incubator/hnc/internal/reconcilers/hierarchy_config.go
GinnyJI: Done

adrianludwin:
Is `pull-hnc-test` just failing because there's a focused test?

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/forest/namespace.go
GinnyJI: done. Made deep copies for `ns.labels`

incubator/hnc/internal/forest/namespace.go
```diff
@@ -88,6 +91,17 @@ func (ns *Namespace) UnsetExists() bool {
 	return changed
 }
 
+func (ns *Namespace) GetLabels() map[string]string {
+	return ns.labels
+}
+
+func (ns *Namespace) SetLabels(labels map[string]string) {
+	ns.labels = make(map[string]string)
```
adrianludwin, line 99: Maybe just comment why we're doing a deep copy

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -359,9 +360,26 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
 	// If an object doesn't exist, assume it's been deleted or not yet created.
 	exists := inst.GetCreationTimestamp() != v1.Time{}
+	selector := r.getSelector(ctx, log, srcInst)
+
+	nsLabels := labels.Set(r.Forest.Get(inst.GetNamespace()).GetLabels())
+	if selector != nil && !selector.Matches(nsLabels) {
+		if exists {
+			// The object already exists but it shouldn't be here according to the updated selector, so we remove it
+			r.deleteObject(ctx, log, inst)
+			return actionRemove, nil
```
adrianludwin, line 372: Returning `actionRemove` tells the _caller_ that it needs to delete the object, so don't call `r.deleteObject` here. We never make any commands to the apiserver while we hold the global mutex since network calls are slow (seconds) whereas we want to be able to lock/unlock the mutex hundreds of times a second.

Also, you don't have any error handling here, but `r.deleteObject()` return an error and we should never just ignore errors (without a very good reason why).

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -359,9 +360,26 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
...
+		} else {
+			// The object already exists and doesn't need to be updated. This will typically happen when HNC
+			// is restarted - all the propagated objects already exist on the apiserver. Record that it exists
+			// for our statistics.
+			r.recordPropagatedObject(log, inst.GetNamespace(), inst.GetName())
+
```
adrianludwin, line 378: Why is this needed?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -373,15 +391,29 @@ func (r *ObjectReconciler) syncPropagated(ctx context.Context, log logr.Logger,
 		return actionWrite, srcInst
 	}
 
-	// The object already exists and doesn't need to be updated. This will typically happen when HNC
-	// is restarted - all the propagated objects already exist on the apiserver. Record that it exists
-	// for our statistics.
-	r.recordPropagatedObject(log, inst.GetNamespace(), inst.GetName())
-
```
adrianludwin, line 380: Why is this removed?

GinnyJI:
Submitting new PRs instead of this one. Closing this PR.

PR #1124, opened 2020-09-20
Regenerate CRD yamls

adrianludwin:
Just ran `make manifests` but nothing else.

Tested: installed HNC


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 

I think only one person needs to approve.

yiqigao217 (commented):
/lgtm
/hold

Just confirming that this is because of the controller-gen upgrade? Please feel free to release the hold.

adrianludwin:
No, this isn't related to controller-gen. I'm honestly not sure why I'm
getting different results from what's checked in.

yiqigao217:
Discussed offline: it should be from #1093 (upgrade controller-tools), and there's no API conversions PRs merged after that so we didn't run `make manifests`.

/hold cancel

PR #1125, opened 2020-09-21
Fix certs in K8s 1.19

adrianludwin:
This change incorporates
https://github.com/open-policy-agent/gatekeeper/pull/813 and
https://github.com/open-policy-agent/gatekeeper/pull/811 to allow HNC's
internal cert generator to work on K8s 1.19, which uses Go 1.15.

Tested: HNC cannot be installed on KIND 1.19 without this change, and
can successfully be installed and have all e2e test run with it.


yiqigao217 (commented):
/lgtm
/hold

/assign @rjbez17

adrianludwin:
This doesn't require two approvals IMO. You only need to hold if you're not
confident

yiqigao217:
Thanks for the explanation!

/hold cancel

PR #1126, opened 2020-09-21
Update to controller-runtime v0.6.3

adrianludwin:
Also fixed minor e2e test issues

Tested: make test and make test-e2e without repairs. Four e2e tests
failed both with an without these changes; at least two of them are
exactly the same (the two demo tests) and I _believe_ the other two
failing tests are the same as well. But I think it's clear that the
update to controller-runtime isn't breaking anything fundamental.


yiqigao217 (commented):
Question: does this upgrade resolves the webhook startup issue https://github.com/kubernetes-sigs/controller-runtime/issues/1148?

adrianludwin:
I believe Ginny checked this in by mistake and I didn't catch it during the
review. "Focused" means "none of the other tests get run" which should
~never be checked in.

yiqigao217:
> I believe Ginny checked this in by mistake and I didn't catch it during the review. "Focused" means "none of the other tests get run" which should ~never be checked in.

SG. Ah I see, it's only possible to have `PIt` checked in but never `Fit`. What about the other question: 



> Question: does this upgrade resolves the webhook startup issue [kubernetes-sigs/controller-runtime#1148](https://github.com/kubernetes-sigs/controller-runtime/issues/1148)?

adrianludwin:
> Question: does this upgrade resolves the webhook startup issue [kubernetes-sigs/controller-runtime#1148](https://github.com/kubernetes-sigs/controller-runtime/issues/1148)?

Sorry I missed that. No, the fix hasn't been checked into controller-runtime yet; my next PR pulls in a patch to fix it. This is just getting us onto the latest _released_ controller-runtime.

yiqigao217:
> > Question: does this upgrade resolves the webhook startup issue [kubernetes-sigs/controller-runtime#1148](https://github.com/kubernetes-sigs/controller-runtime/issues/1148)?
> 
> Sorry I missed that. No, the fix hasn't been checked into controller-runtime yet; my next PR pulls in a patch to fix it. This is just getting us onto the latest _released_ controller-runtime.

/lgtm

PR #1128, opened 2020-09-22
Use official cert-controller and support K8s 1.19

adrianludwin:
This change removes our copy of cert-controller and replaces it with the
"official" copy from the Gatekeeper repo. Since the official copy has
been updated to support Go 1.15 (and hence K8s 1.19), this means that
HNC now support K8s 1.19 as well.

Since Gatekeeper hasn't defined any formal releases yet, we'll
periodically just update to use their head but at some point we'll work
with GK to setup up a formal release process.

Tested: installs cleanly on a fresh Kind 1.19 cluster, which was
previously broken on HEAD (e2e tests showed 8 failures but there are 3-4
failures on HEAD on GKE already and additional known issues with Kind).
Also installs cleanly on GKE 1.17 and e2e tests show the expected 3
failures.


rjbez17:
/lgtm
/approve
/hold 

@yiqigao217 feel free to remove hold after you take a look (since you did the original work here).

incubator/hnc/go.mod
```diff
@@ -14,11 +14,11 @@ require (
 	github.com/gregjones/httpcache v0.0.0-20190611155906-901d90724c79 // indirect
 	github.com/onsi/ginkgo v1.12.1
 	github.com/onsi/gomega v1.10.1
-	github.com/pkg/errors v0.8.1
```
yiqigao217, line 17: Question: this probably was used in our /third-party code, so it's gone after using cert-controller directly?

yiqigao217 (commented):
/lgtm
/hold cancel

Just a quick question.

incubator/hnc/go.mod
adrianludwin: That's my assumption too. I generally just like `go mod tidy` or `go mod vendor` add and remove things without caring too much, unless I know I care about it deeply (e.g. replacing the version of `controller-runtime`).

PR #1131, opened 2020-09-23
Fix some e2e tests

GinnyJI:
Update the output format; change the wait time; add verification to make
sure the ns anchor is created; add verification in RecoverHNC().

Tested: make test-e2e

### DIFF
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -257,6 +257,13 @@ func RecoverHNC() {
 	}
 	// give HNC enough time to repair
 	time.Sleep(5 * time.Second)
+	// Verify and wait till HNC is fully repaired, sometimes it takes up to 30s.
+	// The `kubectl hns create` command will fail is HNC is broken, so we confirm that HNC is back by successfully 
+	// running this command. 
+	CleanupNamespaces("a", "b")
+	MustRun("kubectl create ns a")
+	RunShouldContain("Successfully created", 30, "kubectl hns create b -n a")
+	CleanupNamespaces("a", "b")
 }
 
 func WriteTempFile(cxt string) string {
--- a/incubator/hnc/test/e2e/demo_test.go
+++ b/incubator/hnc/test/e2e/demo_test.go
@@ -43,6 +43,9 @@ var _ = Describe("Demo", func() {
 		// make acme-org the parent of team-a, and team-a the parent of service-1.
 		MustRun("kubectl hns set", nsTeamA, "--parent", nsOrg)
 		MustRun("kubectl hns set", nsService1, "--parent", nsTeamA)
+		// Verify and wait for HNC to pick up the change, otherwise the following commands will fail
+		RunShouldContain("Parent: "+nsOrg, defTimeout, "kubectl hns describe", nsTeamA)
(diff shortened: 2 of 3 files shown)


incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -257,6 +257,11 @@ func RecoverHNC() {
 	}
 	// give HNC enough time to repair
 	time.Sleep(5 * time.Second)
+	// verify
```
adrianludwin, line 260: Can you please add a much longer comment here about what it is you're trying to do? (But good solution!)

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -43,6 +43,9 @@ var _ = Describe("Demo", func() {
 		// make acme-org the parent of team-a, and team-a the parent of service-1.
 		MustRun("kubectl hns set", nsTeamA, "--parent", nsOrg)
 		MustRun("kubectl hns set", nsService1, "--parent", nsTeamA)
+		// verify
```
adrianludwin, line 46: Can you comment why this is needed for _functional_ correctness and not just because you want to add an additional check? E.g. "wait for HNC to pick up the changes"

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -52,7 +55,7 @@ var _ = Describe("Demo", func() {
 		// Now, if we check service-1 again, we’ll see all the rolebindings we expect:
 		RunShouldContainMultiple([]string{"hnc.x-k8s.io/inheritedFrom=acme-org", "hnc.x-k8s.io/inheritedFrom=team-a"},
 			propogationTimeout, "kubectl -n", nsService1, "describe roles")
-		RunShouldContainMultiple([]string{"Role/acme-org-sre", "Role/team-a-sre"}, propogationTimeout, "kubectl -n", nsService1, "get rolebindings")
+		RunShouldContainMultiple([]string{"acme-org-sre", "team-a-sre"}, propogationTimeout, "kubectl -n", nsService1, "get rolebindings")
```
adrianludwin, line 58: Do you have any idea what changed here? `kubectl -n <foo> get rolebindings` isn't an HNC thing that's changed. Also, why do we not see failures in v0.5? Is this test not checked into v0.5?

adrianludwin (commented):
Fixes look good, thanks! I just want to understand a few more things (and/or get them commented in the code)

incubator/hnc/pkg/testutils/testutils.go
GinnyJI: done

incubator/hnc/test/e2e/demo_test.go
GinnyJI: done

incubator/hnc/test/e2e/demo_test.go
GinnyJI: I'm still not sure where this format change come from, but no this test is not checked into v0.5

adrianludwin: Weird. Maybe it never passed, or it passed with an old version of kubectl? But knowing it's not in v0.5 is probably good enough.

adrianludwin (commented):
/lgtm
/approve

PR #1132, opened 2020-09-23
Add gofmt enforcement

### LINKED ISSUE #1129: HNC: ensure `go fmt` was called before presubmit
We should modify the CI presubmit test to run `go fmt` and fail if it changes any files. That way, we'll be sure that developers have always run `make build` before checking in.

/good-first-issue

angelbarrera92:
Closes #1129 

Open to discuss the solution :)

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -72,6 +72,7 @@ HNC_KREW_TAR_SHA256=$(shell sha256sum bin/kubectl-hns.tar.gz | cut -d " " -f 1)
 
 # The directories to run Go command on (excludes /vendor)
 DIRS=./api/... ./cmd/... ./internal/...
+GOFMT_DIRS=$(DIRS:/...=)
 
 all: test docker-build
 
@@ -135,7 +136,11 @@ manifests: controller-gen
 
 # Run go fmt against code
 fmt:
-	go fmt ${DIRS}
+	gofmt -l -w ${GOFMT_DIRS}
+
+# check-fmt: Checks gofmt/go fmt has been ran. gofmt -l lists files whose formatting differs from gofmt's, so it fails if there are unformatted go code.
+check-fmt:
+	@test -z $(shell gofmt -l ${GOFMT_DIRS})
 
 # Run go vet against code
 vet:
--- a/incubator/hnc/hack/ci-test.sh
+++ b/incubator/hnc/hack/ci-test.sh
@@ -35,6 +35,6 @@ mv kubebuilder_${kb}_linux_amd64 /usr/local/kubebuilder
 
 hack_dir=$(dirname ${BASH_SOURCE})
 
-echo "Running 'make tests'"
+echo "Running 'make check-fmt test'"
 cd "$hack_dir/.."
-make test
+make check-fmt test


incubator/hnc/Makefile
```diff
@@ -69,6 +69,7 @@ KREW_CKSM=$(shell sha256sum bin/kubectl-hns.tar.gz | cut -d " " -f 1)
 
 # The directories to run Go command on (excludes /vendor)
 DIRS=./api/... ./cmd/... ./internal/... ./third_party/...
+GOFMT_DIRS=./api ./cmd ./internal ./third_party
```
adrianludwin, line 72: Why are these two different?

incubator/hnc/cloudbuild.yaml
```diff
@@ -12,7 +12,16 @@ steps:
     cd multi-tenancy
     echo Checking out hnc-$_HNC_IMG_TAG
     git checkout hnc-$_HNC_IMG_TAG
-# Build the manifests and the kubectl plugin
+# go fmt enforcement
+- name: golang:1.14
```
adrianludwin, line 16: We only use this file (and Cloud Build) when we're generating an official release. The file that gets run on every commit by Github is https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/hack/ci-test.sh - can you please move this logic there?

incubator/hnc/Makefile
```diff
@@ -155,7 +156,11 @@ manifests: controller-gen
 
 # Run go fmt against code
 fmt:
-	go fmt ${DIRS}
+	gofmt -w ${GOFMT_DIRS}
```
adrianludwin, line 159: Why the switch from `go fmt` to `gofmt`?

incubator/hnc/Makefile
```diff
@@ -155,7 +156,11 @@ manifests: controller-gen
...
+
+# Check if fmt was ran
+check-fmt:
+	@test -z $(shell gofmt -l ${GOFMT_DIRS})
```
adrianludwin, line 163: Can you comment a bit more how this command works? Thanks!

adrianludwin:
Left a few comments. Also, can you please squash your commits? We generally only have one commit per PR. See: https://github.com/kubernetes-sigs/multi-tenancy/tree/master/incubator/hnc#development-workflow

Thanks for this change!

incubator/hnc/Makefile
angelbarrera92: In short, `go fmt` runs `gofmt -l -w` on the packages specified by the arguments.

As we include `check-fmt` make target using `gofmt`, I thought it was a good idea to use the same command.

incubator/hnc/Makefile
angelbarrera92: `gofmt -l` list files whose formatting differs from gofmt's.
`test -z`: True if the length of string is zero.

So if there are no files differing from `gofmt` command, `test` returns `True` otherwise fails.

incubator/hnc/cloudbuild.yaml
angelbarrera92: My fault! Doing

incubator/hnc/Makefile
angelbarrera92: `gofmt` does not understand the `./api/...` syntax. Instead, it works with the simple: `./api`.
I saw other make targets using the `DIRS` variable `(make vet)`, so I opted to create a new variable for the `go fmt` related make targets.

incubator/hnc/Makefile
adrianludwin: Sorry, I mean "comment in the makefile," not in the PR review :) This is to make life easier for future maintainers.

incubator/hnc/Makefile
adrianludwin: Sorry for not understanding. Why do we _not_ want to use the `-l` flag where we were using it before? And why can't we use `-l` in `check-fmt:`? Per the docs:

```
    -l
    	Do not print reformatted sources to standard output.
    	If a file's formatting is different from gofmt's, print its name
    	to standard output.
```

How about this: in the CI script, we just run `make fmt` and see if it prints anything to stdout? If nothing happens, then we know the files were all in a good state. We don't need any modifications to the makefile at all?

(Context: I'm trying to avoid having two vars with duplicated data that need to be kept in sync, as these directories do change fairly frequently)

incubator/hnc/hack/ci-test.sh
```diff
@@ -27,14 +27,19 @@ set -e
 export PATH=$(go env GOPATH)/bin:$PATH
 mkdir -p $(go env GOPATH)/bin
 
+hack_dir=$(dirname ${BASH_SOURCE})
```
adrianludwin, line 30: Any reason to do this before installing kubebuilder? I don't think it actually _depends_ on it in any way, but it's a bit simpler IMO to say that the script a) installs dependencies and b) runs all tests.

That would also eliminate the need to say `cd -` after the first check.

incubator/hnc/Makefile
rjbez17: wouldn't `GOFMT_DIRS=$(DIRS:/...=)` work here? Would protect against changes to DIRS in the future. Although, it isn't foolproof.

adrianludwin:
My bash knowledge is not that strong 🙂

incubator/hnc/hack/ci-test.sh
angelbarrera92: Ye, I like to fail fast if the developer didn't apply `gofmt` in the commit.
But I understand it could be better to just: `make check-fmt test` after installing the dependencies.

Changing!

incubator/hnc/Makefile
angelbarrera92: I've added a more descriptive comment :)

incubator/hnc/Makefile
angelbarrera92: Thanks @rjbez17 that code worked!

incubator/hnc/Makefile
angelbarrera92: Ye, it was my fault. I missed the `-l` in the `gofmt -w` command. Solved.
I don't like to implement this kind of logic in the CI script, `make` targets are cleaner IMHO.

Also, we solved the issue of having two variables. Thanks to @rjbez17 who gave us the key!

adrianludwin (commented):
/lgtm
/approve

Thanks for this change!

PR #1136, opened 2020-09-23
Build Krew archive and manifest

adrianludwin:
Tested: `make krew-install` and verified that the plugin worked
correctly and the documentation link was correct. Built a test release
to my personal repo and verified that I could download the manifest and
install from it correctly.

Fixes #1030

PR #1137, opened 2020-09-23
Build Krew archive

adrianludwin:
Unlike in 0.6+, this change doesn't attempt to build the manifest as a
part of Cloud Build. That can be done by hand after the release; too
much has changed in the makefiles between the two versions. So this just
focuses on creating the archive.

Tested: this is virtually impossible to test given the state of the
build process in v0.5 (it's much better in v0.6). I'll revert the
changes to cloudbuild.yaml if I can't build a release candidate, and the
changes to the hack/krew*.yaml files don't matter to anyone.


incubator/hnc/cloudbuild.yaml
```diff
@@ -32,6 +32,11 @@ steps:
     # Build plugin
     GOOS=linux GOARCH=amd64 go build -o kubectl-hns_linux_amd64 ../cmd/kubectl/main.go
     GOOS=darwin GOARCH=amd64 go build -o kubectl-hns_darwin_amd64 ../cmd/kubectl/main.go
+    mkdir kubectl
+    cp kubectl-hns_linux_amd64 kubectl/
+    cp kubectl-hns_darwin_amd64 kubectl/
+    cp ../LICENSE kubectl/
+    tar czvf kubectl-hns.tar.gz kubectl/
```
yiqigao217, line 39: Why didn't see this part in #1136 but here?

adrianludwin: In v0.6, cloudbuild.yaml defers to the makefile for almost everything. In v0.5 it replicates what's in the makefile, so I had to do the same.

incubator/hnc/hack/krew-kubectl-hns.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: krew.googlecontainertools.github.com/v1alpha2
```
rjbez17, line 1: In the 0.6 PR you use sed to replace the tag, checksum, etc in this file. I don't see and of that here, am I missing something?

rjbez17: oh I see, you meant to do it by hand. Maybe add a comment indicating as such? Or purposely making it invalid yaml so it isn't pushed accidentally?

adrianludwin: Big scary comment added :)

PR #1139, opened 2020-09-23
Let subns with propagated objects be deleted

### LINKED ISSUE #1130: HNC: Deleting a subnamespace after creating resources in the parent namespace results in error
The following sequence doesn't work:
1. Creating a subnamespace by creating a SubnamespaceAnchor resource (the name of the parent ns is hnc-try, and the name of the subnamespace is hnc-try-sub-1).
2. Creating some RBAC resources in the parent namespace (role + rolebinding).
3. Deleting the SubnamespaceAnchor resource.

The subnamespace gets stuck in the middle of deletion, and I see the following log:
```
{"level":"info","ts":1600856582.4537065,"logger":"validators.Object","msg":"Handled","nm":"","resource":{"group":"rbac.authorization.k8s.io","version":"v1","resource":"roles"},"nnm":"hnc-try-sub-1","op":"DELETE","allowed":false,"code":403,"reason":"Forbidden","message":"Cannot delete object propagated from namespace \"hnc-try\""}
```

By adding some logging to the validator, I got that the username which issued the `AdmissionRequest` is `system:serviceaccount:kube-system:namespace-controller`. However, this user isn't exempted from the webhook, and only `system:serviceaccounts:hnc-system` is exempted. So the `AdmissionRequest` isn't automatically allowed, and then it is rejected as the resource has the label `hnc.x-k8s.io/inheritedFrom: hnc-try` (hnc-try is the name of the parent namespace).

I would suggest to exclude the same service accounts as atlasian excludes:
https://github.com/atlassian/voyager/blob/11ade5a733008770a7df4b12aee2399dc2af5d12/pkg/admission/server.go#L61

### Environment
Kind version:
```
kind v0.7.0 go1.13.6 linux/amd64
```
Kubernetes version:
```
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2020-01-14T00:09:19Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```
Operator version - v0.5.2

adrianludwin:
See issue #1130. The HNC object validator was preventing the K8s
namespace controller from cleaning out a namespace to allow it to be
deleted. This change lets HNC understand that when a namespace is being
deleted, we should ignore the subnamespaceOf annotation. This allows the
HierarchyConfiguration object to be reset to its initial state (like a
full namespace would when it's being deleted), orphaning the namespace
and allowing all its propagated objects to be removed.

Tested: added new tests and ran them multiple times while watching the
logs. The subnamespace test failed without this change and passed every
time with it, with the logs showing the expected behaviour.

Fixes #1130

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -273,6 +273,21 @@ func (r *HierarchyConfigReconciler) syncSubnamespaceParent(log logr.Logger, inst
 	}
 
 	pnm := nsInst.Annotations[api.SubnamespaceOf]
+
+	// Issue #1130: as a subnamespace is being deleted (e.g. because its anchor was deleted), ignore
+	// the annotation. K8s will remove the HC, which will effectively orphan this namespace, prompting
+	// HNC to remove all propagated objects, allowing it to be deleted cleanly. Without this, HNC
+	// would continue to think that all propagated objects needed to be protected from deletion and
+	// would prevent K8s from emptying and removing the namespace.
+	//
+	// We could also add an exception to allow K8s SAs to override the object validator (and we
+	// probably should), but this prevents us from getting into a war with K8s and is sufficient for
+	// v0.5.
+	if pnm != "" && !nsInst.DeletionTimestamp.IsZero() {
+		log.Info("Subnamespace is being deleted; ignoring SubnamespaceOf annotation", "parent", inst.Spec.Parent, "annotation", pnm)
+		pnm = ""
+	}
+
 	if pnm == "" {
 		ns.IsSub = false
 		return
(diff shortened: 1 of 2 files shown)


adrianludwin:
NB: I'm currently running all e2e tests on this patch; I'll remove my hold when they all pass.

/hold
/assign @rjbez17 
/assign @yiqigao217 
/cc @GinnyJI

rjbez17:
/approve
/lgtm

assuming the e2e tests pass

yiqigao217 (commented):
/lgtm

adrianludwin:
All e2e tests passed, including the new ones.

/hold cancel

PR #1142, opened 2020-09-23
Regenerate CRD yamls with controller-tools v0.2.8

yiqigao217:
The diff is a reverse of #1124 plus the controller-gen version changed
from v0.2.5 to v0.2.8.

Update the CRD yamls with v0.2.8 controller-tools (controller-gen). This
should be the stable generated CRD yamls. If you get a different
version, you can delete your local 'bin/controller-gen' and rerun 'make
manifests'. It will install controller-gen v0.2.8 for you.

Tested by make test.

We got #1124 before probably because we have the go.mod updated in
#1093 but we didn't upgrade our local controller-gen.

### DIFF
--- a/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
+++ b/incubator/hnc/config/crd/bases/hnc.x-k8s.io_hierarchyconfigurations.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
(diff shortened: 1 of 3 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

The diff is a reverse of #1124 plus the controller-gen version changed from v0.2.5 to v0.2.8.

We got #1124 before probably because we have the go.mod updated in #1093 but we didn't upgrade our local controller-gen.

adrianludwin:
The root cause for all this confusion is fixed in #1143 .

/close

PR #1143, opened 2020-09-23
Ensure controller-gen is rebuilt properly

adrianludwin:
The Makefile wasn't correctly regenerating controller-gen when the
version in go.mod (and hence vendor/) was updated. This led to me
mistakenly submitting #1124. This change _always_ rebuilds
controller-gen from scratch to ensure that leftover tool binaries won't
cause more problems.

Tested: before this change, ran `make manifests` and observed that the
version of controller-gen (as seen by `--version`) didn't match the
version in go.mod. After this change, observed that the reported version
now matched that in go.mod and the CRD YAMLs were rebuilt. Installed
HNC and verified that it started up ok.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -145,11 +145,9 @@ vet:
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
 
-# find or download controller-gen
-# download controller-gen if necessary
-controller-gen: $(CONTROLLER_GEN)
-
-$(CONTROLLER_GEN):
+# Use the version of controller-gen that's checked into vendor/ (see
+# hack/tools.go to see how it got there).
+controller-gen:
 	go build -o $(CONTROLLER_GEN) sigs.k8s.io/controller-tools/cmd/controller-gen
 
 ###################### DEPLOYABLE ARTIFACTS AND ACTIONS #########################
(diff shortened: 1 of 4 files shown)


incubator/hnc/Makefile
```diff
@@ -145,11 +145,9 @@ vet:
...
 
-# find or download controller-gen
-# download controller-gen if necessary
-controller-gen: $(CONTROLLER_GEN)
-
-$(CONTROLLER_GEN):
+# Use the version of controller-gen that's checked into vendor/ (see
+# hack/tools.go to see how it got there).
+controller-gen:
 	go build -o $(CONTROLLER_GEN) sigs.k8s.io/controller-tools/cmd/controller-gen
```
yiqigao217, line 151: Is it that if `CONTROLLER_GEN` (/bin/controller-gen) is not empty on line 150, line 153 won't be run?

yiqigao217 (commented):
lgtm just a confusion here.

incubator/hnc/Makefile
yiqigao217: Oh you will have to view it in the commit diff (the line numbers are for the old makefile).

yiqigao217:
@adrianludwin BTW are you still able to run `make test` after this change? I got the same output as https://github.com/slipway-gitops/slipway/issues/21 when I work on #1074 , not sure if it's because of the controller-gen version change in the crd/bases yamls. Could you please help me confirm that?

incubator/hnc/Makefile
adrianludwin: Yes that's exactly it. We're not using makefiles "properly" so that they understand timestamps, and even if we were, timestamps are really brittle and Go compiles really quickly so the best thing for us to do is always rebuild.

adrianludwin (commented):
Yes `make test` passes for me after this change.

incubator/hnc/Makefile
yiqigao217: Thanks for the explanation!

yiqigao217 (commented):
/lgtm

PR #1145, opened 2020-09-24
Set crd:trivialVersions=false for easier API upgrade

yiqigao217:
We were going to set `crd:trivialVersions=false` for per-version
validation in 0.6. However, we cannot patch CRD if the one in v0.5 is
still using top-level schema. We would get "`Forbidden: top-level and`
`per-version schemas are mutually exclusive`" error when upgrading HNC,
specifically CRDs.

Tested by API upgrading from this to the v0.6 change in a GKE cluster.
It passed after this change and failed with the above error before. Also
tested by `make test` to ensure it doesn't break v0.5.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -30,8 +30,11 @@ HNC_RELEASED_IMG ?= "gcr.io/k8s-staging-multitenancy/hnc/controller:${HNC_IMG_TA
 
 CONTROLLER_GEN ?= "./bin/controller-gen"
 
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
+# No version conversion yet, but have per-version schema ready for version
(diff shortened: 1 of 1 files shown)


yiqigao217:
One thing I don't understand is that the CRD yamls are not changed. I would expect there's any schema change there.

However, it did make a difference that the `Forbidden: top-level and per-version schemas are mutually exclusive` error is gone.

yiqigao217:
/hold
Will do more tests on this.

yiqigao217:
False alarm: the exclusive schema error only happens when I mistakenly use **create** instead of **apply**: `kubectl create -f {hnc 0.5 yaml}` and then `kubectl apply -f manifest/hnc-manager.yaml`. I should use `kubectl apply -f {hnc 0.5 yaml}` instead (the error is gone).

I searched online and found there's a difference but not sure how it affects this: `kubectl create` uses Imperative Management; `kubectl apply` uses Declarative approach.

adrianludwin:
Thanks for clearing that up!

PR #1147, opened 2020-09-25
VC: adding healthz for vc-syncer,vc-manager

miamipanther:


incubator/virtualcluster/cmd/manager/main.go
```diff
@@ -100,6 +102,15 @@ func main() {
...
 	}
 
+	// start a health http server.
+	log.Info("Starting a health http server")
+	mux := http.NewServeMux()
+	healthz.InstallHandler(mux)
+	if err := http.ListenAndServe(":8080", mux); err != nil {
+		log.Error(err, "unable to start health http server")
+		os.Exit(1)
+	}
```
christopherhein, line 112: I believe we need to wrap this entire thing in a goroutine so that the rest of the controller can start.

incubator/virtualcluster/cmd/vn-agent/app/server.go
```diff
@@ -136,5 +137,10 @@ func Run(c *config.Config, serverOption *options.ServerOption, stopCh <-chan str
 		klog.Errorf("server listen error %v", err)
 	}
 
+	// start a health http server.
+	mux := http.NewServeMux()
+	healthz.InstallHandler(mux)
+	klog.Fatal(http.ListenAndServe(":8080", mux))
```
christopherhein, line 143: Likewise, this needs to be run in a goroutine.

christopherhein (changes requested):
This holds the server open for healthchecks but isn't actually running the controllers anymore. all the http.ListenAndServe() need to be run in goroutines.

incubator/virtualcluster/cmd/manager/main.go
miamipanther: Good catch, `ListenAndServe` is block call and should be put in go routine

incubator/virtualcluster/cmd/vn-agent/app/server.go
miamipanther: Put it in goroutine

christopherhein:
Make sure you complete the CLA as well.

incubator/virtualcluster/cmd/vn-agent/app/server.go
```diff
@@ -136,5 +137,12 @@ func Run(c *config.Config, serverOption *options.ServerOption, stopCh <-chan str
 		klog.Errorf("server listen error %v", err)
 	}
 
+	go func() {
+		// start a health http server.
+		mux := http.NewServeMux()
+		healthz.InstallHandler(mux)
+		klog.Fatal(http.ListenAndServe(":8080", mux))
```
Fei-Guo, line 144: I think vn-agent will use host network. The 8080 port may not be available.

miamipanther:
> Make sure you complete the CLA as well.

CLA completed

incubator/virtualcluster/cmd/vn-agent/app/server.go
miamipanther: changed port to 8088

zhuangqh: it is unnecessary to open a health port for vn-agent. firstly it occupy a node port. node port should be strictly managed for security consideration.  And it may conflict with other component.
Using tcpSocket type liveness from kubelet is enough for vn-agent.

zhuangqh:
it is unnecessary to open a health port for vn-agent. firstly it occupy a node port. node port should be strictly managed for security consideration. And it may conflict with other component.
Using tcpSocket type liveness from kubelet is enough for vn-agent.

miamipanther:
> it is unnecessary to open a health port for vn-agent. firstly it occupy a node port. node port should be strictly managed for security consideration. And it may conflict with other component.
> Using tcpSocket type liveness from kubelet is enough for vn-agent.

Agreed tcpSocket liveness is good enough for vn-agent; The port added here is to be used by external probe monitoring and alert service, I will remove vn-agent and we might address it in the future.

Issue #1149, opened 2020-09-25
HNC: anchor in "conflict" mode can still delete subns


adrianludwin:
Steps to reproduce:
1. Create subns
2. Delete subnamespaceOf annotation
3. Delete anchor

Expected: namespace is not deleted
Actual: namespace is deleted

PR #1150, opened 2020-09-25
Don't delete subns if anchor has conflict

adrianludwin:
See issue #1149. This affects the ability to turn a subnamespace into a
full namespace. If you remove the subnamespaceOf annotation, move the
namespace to a new parent, and *then* clean up the anchor, everything's
ok. But if you delete the anchor _before_ moving the namespace, the
namespace gets deleted, which is clearly wrong.

This change is the minimal safe change needed to fix this in v0.5. In
v0.6, we should restructure these functions to be more heavily based on
the explicit state of the anchor, not the implied state.

Tested: new e2e test fails (hangs, actually) without this change and
passes with it. All e2e tests pass on GKE 1.17.

### DIFF
--- a/incubator/hnc/internal/reconcilers/anchor.go
+++ b/incubator/hnc/internal/reconcilers/anchor.go
@@ -17,10 +17,11 @@ package reconcilers
 
 import (
 	"context"
+	"errors"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -65,7 +66,7 @@ func (r *AnchorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// purged.
 	inst, err := r.getInstance(ctx, pnm, nm)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, err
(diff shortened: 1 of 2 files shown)


incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -118,13 +119,19 @@ func (r *AnchorReconciler) onDeleting(ctx context.Context, log logr.Logger, inst
 		return false, err
 	}
 
+	// Update the state of the anchor
+	//
+	// TODO: call this in the parent function after v0.5 so there's no special code path for the
```
rjbez17, line 124: Is there already an issue created for this?

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -152,6 +161,27 @@ func (r *AnchorReconciler) shouldDeleteSubns(inst *api.SubnamespaceAnchor, nsIns
 		return false
 	}
 
+	// If the anchor isn't in the "ok" state, don't delete the namespace. If it's "conflict," we
+	// *definitely* don't want to delete it; if it's missing (and possibly being created), then there
+	// isn't really a good option but we'll eventually put a condition on the namespace so it's not a
+	// big deal.
+	//
+	// TODO: much of the remaining portion of this function can probably be replaced by this switch
```
rjbez17, line 169: Same question about an issue.

rjbez17:
Wouldn't hurt to create issues and link them in the TODOs but otherwise lgtm.

/lgtm
/approve
/hold 

feel free to remove hold if you don't plan to link issues.

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: I'm going to do it as I cherrypick this change into master, starting ~now :)

adrianludwin:
The change was just to fix the format of a log message

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -118,13 +119,19 @@ func (r *AnchorReconciler) onDeleting(ctx context.Context, log logr.Logger, inst
 		return false, err
 	}
 
+	// Update the state of the anchor
+	//
+	// TODO: call this in the parent function after v0.5 so there's no special code path for the
+	// onDeleting case.
+	r.updateState(log, inst, snsInst)
```
yiqigao217, line 126: I'm not sure if you added the `updateState()` here just to make the change small in v0.5. It seemed the same for me if just moving the existing `updateState()` up before `onDeleting()` in the `Reconcile()`.

yiqigao217 (commented):
lgtm just 2 small questions.

The other question is that: Will this change in any case (for `conflict` or `missing`) prevent a should-delete subns from being deleted?

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: Yes, I wanted to make this change as localized as possible in v0.5. In the master branch that I'm working on, I move this to Reconcile.

adrianludwin:
> The other question is that: Will this change in any case (for `conflict` or `missing`) prevent a should-delete subns from being deleted?

The "conflict" case is a change - that's the bug I'm trying to fix. The "missing" case shouldn't be a change - we've _just_ verified that the namespace is truly missing, so there's nothing to delete.

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: Update: fixed in #1151

yiqigao217:
> > The other question is that: Will this change in any case (for `conflict` or `missing`) prevent a should-delete subns from being deleted?
> 
> The "conflict" case is a change - that's the bug I'm trying to fix. The "missing" case shouldn't be a change - we've _just_ verified that the namespace is truly missing, so there's nothing to delete.

Discussed offline, I was confused with allowing user to delete anchors when it has `conflict` to resolve issues. This PR is only for deleting sub-"namespace".

/lgtm

PR #1152, opened 2020-09-25
Fix a very rare flaky test on Prow

yiqigao217:
This flaky test cannot be reproduced when `make test` for 10+ times.
Update the `Expect` to `Eventually` to make it less likely to fail.

Tested by 'make test'.

See failed Prow [log](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_multi-tenancy/1144/pull-hnc-test/1309517615571406848).


adrianludwin (commented):
/lgtm
/approve

PR #1157, opened 2020-09-29
Update "subnamespace-of" sub-namespace annotation

yiqigao217:
Update the sub-namespace annotation into `subnamespace-of` from
`subnamespaceOf` in v1alpha2. Update the namespace annotation in the
hierarchyConfiguration reconciler.

Add documentation that `Ok` is the only good state of anchor.

Update conversion tests.

Update util funcs CleanupNamespace() and RecoverHNC() to report failure
in cleanup.

Tested by `make test-conversion`, `make test` and manually.

Part of #868

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -212,6 +212,13 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	// Set external tree labels in the forest if this is an external namespace.
 	r.syncExternalNamespace(log, nsInst, ns)
 
+	// In v0.5.x, HNC uses v1alpha1 and the annotation on any subnamespace is
+	// called "subnamespaceOf". In v1alpha2, the namespace annotation is upgraded
+	// into "subnamespace-of". If the deprecated "subnamespaceOf" annotation still
+	// exists, upgrade API and early exit since updating the subnamespace will
+	// enqueue the singleton again.
+	r.upgradeSubnamespaceAnnotation(nsInst)
+
 	// If this is a subnamespace, make sure .spec.parent is set correctly. Then sync the parent to the
 	// forest, and finally notify any relatives (including the parent) that might have been waiting
 	// for this namespace to be synced.
@@ -265,6 +272,28 @@ func (r *HierarchyConfigReconciler) syncExternalNamespace(log logr.Logger, nsIns
 	}
 	etls[ns.Name()] = 0
 	ns.ExternalTreeLabels = etls
+}
+
+// upgradeSubnamespaceAnnotation replaces the deprecated "subnamespaceOf"
+// annotation with the new "subnamespace-of" annotation. If both exist, we will
+// only delete the deprecated annotation.
+func (r *HierarchyConfigReconciler) upgradeSubnamespaceAnnotation(inst *corev1.Namespace) {
+	a := inst.GetAnnotations()
+	sOf, oldAnnotationExists := a[api.DeprecatedSubnamespaceOf]
+	if !oldAnnotationExists {
+		return
+	}
+
+	// Remove old annotation
+	delete(a, api.DeprecatedSubnamespaceOf)
+
+	// Add new annotation key with the old value if it doesn't exist. There's a
+	// corner case if the namespace has the new (unknown to HNC v0.5) annotation
+	// before upgrading API, the new annotation will take over after upgrading.
+	if _, newAnnotationExists := a[api.SubnamespaceOf]; !newAnnotationExists {
+		a[api.SubnamespaceOf] = sOf
+	}
+	inst.SetAnnotations(a)
 }
 
 // syncSubnamespaceParent sets the parent to the owner and updates the SubnamespaceAnchorMissing
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -215,9 +215,17 @@ func CleanupNamespaces(nses ...string) {
 	// has cascading deletion so we can delete any of its subnamespace descendants, and
 	// make sure that it's not a subnamespace itself so we can delete it directly.
 	for _, ns := range nses {
-		TryRunQuietly("kubectl hns set", ns, "-a")
-		TryRunQuietly("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
-		TryRunQuietly("kubectl delete ns", ns)
+		if err := TryRunQuietly("kubectl get ns", ns); err == nil {
+			MustRunWithTimeout(30, "kubectl hns set", ns, "-a")
+			// 'subnamespaceOf' is the old subnamespace annotation used in v0.5. We
+			// still need to clean up this old annotation because this util func is
+			// also used in the API conversion test to clean up namespaces in v0.5.
+			// TODO: remove this line after v0.6 branches and we are no longer
+			//  supporting v1alpha1 conversion.
+			MustRunWithTimeout(30, "kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
+			MustRunWithTimeout(30, "kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespace-of-")
+			MustRunWithTimeout(30, "kubectl delete ns", ns)
+		}
 	}
 }
 
@@ -270,19 +278,7 @@ func RecoverHNC() {
 		a = "recover-test-a"
 		b = "recover-test-b"
 	)
-	// Do NOT use CleanupNamespaces because that just assumes that if it can't delete a namespace that
-	// everthing's fine, but this is a poor assumption if HNC has just been repaired.
-	//
-	// TODO: if CleanupNamespaces ever starts using labels to select namespaces to delete, then get
-	// rid of this hack.
-	if err := TryRunQuietly("kubectl get ns", a); err == nil {
-		MustRunWithTimeout(30, "kubectl hns set", a, "-a")
-		MustRunWithTimeout(30, "kubectl delete ns", a)
-	}
-	if err := TryRunQuietly("kubectl get ns", b); err == nil {
-		MustRunWithTimeout(30, "kubectl annotate ns", b, "hnc.x-k8s.io/subnamespaceOf-")
-		MustRunWithTimeout(30, "kubectl delete ns", b)
-	}
+	CleanupNamespaces(a, b)
 	// Ensure validators work
 	MustRunWithTimeout(30, "kubectl create ns", a)
 	// Ensure reconcilers work
--- a/incubator/hnc/test/conversion/conversion_test.go
+++ b/incubator/hnc/test/conversion/conversion_test.go
@@ -27,6 +27,10 @@ const (
 	configCRD       = "hncconfigurations.hnc.x-k8s.io"
 	hierSingleton   = "hierarchy"
 	configSingleton = "config"
+
+	namspacePrefix = "e2e-conversion-test-"
+	nsA            = namspacePrefix + "a"
+	nsB            = namspacePrefix + "b"
 )
 
 func TestConversion(t *testing.T) {
@@ -41,9 +45,6 @@ var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
 	const (
 		// hncFromVersion has to be a version using v1alpha1.
 		hncFromVersion = "v0.5.0"
-		namspacePrefix = "e2e-conversion-test-"
-		nsA            = namspacePrefix + "a"
-		nsB            = namspacePrefix + "b"
 	)
 
 	BeforeEach(func() {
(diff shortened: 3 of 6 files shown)


incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -87,6 +87,14 @@ func (r *AnchorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	}
 
+	// In v0.5.x, HNC uses v1alpha1 and the annotation is called "subnamespaceOf".
+	// In v1alpha2, the namespace annotation is upgraded into "subnamespace-of".
+	// If the deprecated "subnamespaceOf" annotation still exists, upgrade API and
+	// early exit since updating the subnamespace will enqueue the anchor again.
+	if oldSOf := snsInst.GetAnnotations()[api.DeprecatedSubnamespaceOf]; oldSOf != "" {
+		return ctrl.Result{}, r.upgradeSubnamespaceOfAPI(ctx, log, snsInst, oldSOf)
+	}
+
```
adrianludwin, line 97: This feels unnatural to me, tbh. This reconciler has only ever created and deleted namespaces, not updated them - the HC reconciler has a much better flow for making incremental changes to namespaces. It would also allow you to leave most of this file completely untouched - you wouldn't need to change the mapping function or add a getSubnamespaceOf() function (when HNC v0.6 is installed, all anchors would briefly go into a Conflict state until their annotations were fixed, but that's fine IMO).

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -106,6 +114,25 @@ func (r *AnchorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+// upgradeSubnamespaceOfAPI replaces the deprecated "subnamespaceOf" annotation
+// with the new "subnamespace-of" annotation.
+func (r *AnchorReconciler) upgradeSubnamespaceOfAPI(ctx context.Context, log logr.Logger, inst *corev1.Namespace, sOf string) error {
+	// Remove old annotation.
+	a := inst.GetAnnotations()
+	delete(a, api.DeprecatedSubnamespaceOf)
+	inst.SetAnnotations(a)
+
+	// Add new annotation.
+	metadata.SetAnnotation(inst, api.SubnamespaceOf, sOf)
```
adrianludwin, line 126: Here, it looks like if both are present, the old one will overwrite the new one. Is that intended?

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: sg will look into it.

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: Ah, it should be that if both exists, we will just remove the deprecated annotation directly, right?

adrianludwin: That would make sense, I think

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: In your other PR changing the anchor reconciler flow, if the state is `Conflict`, it will be untouched and nothing will be deleted, right? If so, it should be fine to have `Conflict` state for a short period at HNC startup.

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: Done.

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: Moved the change to HC reconciler and leave anchor reconciler untouched. All tests passed. Just need a confirmation for my above comment from @adrianludwin .

adrianludwin: Yes, that's right. If an anchor is in the "conflict" state, it just means that if you delete the anchor, the namespace won't be deleted. As soon as the annotation is fixed, that problem goes away. Note that this doesn't affect the _parent_ of the subnamespace, which remains unchanged throughout.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -106,6 +106,15 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
...
 
+	// In v0.5.x, HNC uses v1alpha1 and the annotation on any subnamespace is
+	// called "subnamespaceOf". In v1alpha2, the namespace annotation is upgraded
+	// into "subnamespace-of". If the deprecated "subnamespaceOf" annotation still
+	// exists, upgrade API and early exit since updating the subnamespace will
+	// enqueue the singleton again.
+	if oldSOf := nsInst.GetAnnotations()[api.DeprecatedSubnamespaceOf]; oldSOf != "" {
+		return r.upgradeSubnamespaceOfAPI(ctx, log, nsInst, oldSOf)
+	}
+
```
adrianludwin, line 117: Why do an early return here? We already know how to write the instances (line 144), so I'd just put this into r.syncWithForest() or something like that. You just need to modify nsInst and then it will be written automatically.

I'd do something like this:

```
func syncWithForest(... nsInst ...) {
   ...
   upgradeSubnamespaceAnnotation(... nsInst ...)
   ...
}

func upgradeSubnamespaceAnnotation(... nsInst ...) {
  if old annotation doesn't exist {
     return
  }
  update nsInst
}
```

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -59,7 +59,7 @@ var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
 		TearDownHNC("")
 	})
 
-	It("should convert subnamespace anchors", func() {
+	It("should convert subnamespace anchors and subnamespace-of annotation", func() {
```
adrianludwin, line 62: Maybe include a test case where _both_ are defined?

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: done.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Added and found that we will need to decide which one takes precedence if both exists in the conversion. PTAL at the new test and the reconciler behavior.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -267,6 +274,33 @@ func (r *HierarchyConfigReconciler) syncExternalNamespace(log logr.Logger, nsIns
 	ns.ExternalTreeLabels = etls
 }
 
+// upgradeSubnamespaceAnnotation replaces the deprecated "subnamespaceOf"
+// annotation with the new "subnamespace-of" annotation. If both exist, we will
+// only delete the deprecated annotation.
+func (r *HierarchyConfigReconciler) upgradeSubnamespaceAnnotation(inst *corev1.Namespace) {
+	sOf, exist := inst.GetAnnotations()[api.DeprecatedSubnamespaceOf]
```
adrianludwin, line 281: This method interacts with `inst`'s annotations in three separate ways - once in an expression (this line), once as a temporary variable (line 287) and once through a helper function (line 300). This makes it quite difficult to follow the flow of information. Do you think you could find a way to make these all look more similar?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -267,6 +274,33 @@ func (r *HierarchyConfigReconciler) syncExternalNamespace(log logr.Logger, nsIns
...
+		return
+	}
+
+	// Remove old annotation
+	a := inst.GetAnnotations()
+	delete(a, api.DeprecatedSubnamespaceOf)
+	inst.SetAnnotations(a)
+
+	// Add new annotation key with the old value if it doesn't exist or this is
+	// the first time this namespace gets reconciled, which is a corner case if
```
adrianludwin, line 292: I don't understand this "first time the namespace gets reconciled." Why should it matter if it's the first time or not? Generally speaking, the reconciler should be as stateless as possible and we shouldn't care what's come before (the only exception I know of is forcing _other_ objects to be re-reconciled once we have more information).

Also, what happens if this is the second time we've reconciled this namespace, but there was an error the first time we reconciled it and the namespace was never written back to the apiserver?

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -216,6 +216,7 @@ func CleanupNamespaces(nses ...string) {
 	for _, ns := range nses {
 		TryRunQuietly("kubectl hns set", ns, "-a")
 		TryRunQuietly("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespaceOf-")
+		TryRunQuietly("kubectl annotate ns", ns, "hnc.x-k8s.io/subnamespace-of-")
```
adrianludwin, line 219: Why are we keeping the old annotation if (as of this change) it should never exist or do anything meaningful?

adrianludwin: Also, can you please update the similar code I added in the rolebinding test?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -79,6 +72,53 @@ metadata:
...
+	})
+
+	It("should respect the old 'subnamespaceOf' annotation AT conversion if both exists", func() {
+		// Before conversion, create namespace A and a subnamespace B.
+		createSampleV1alpha1Subnamespace()
+		// Add `subnamespace-of` annotation, which should be unknown to HNC v0.5.
+		MustRun("kubectl annotate ns", nsB, "hnc.x-k8s.io/subnamespace-of=wrongvalue")
+		// Both subnamespace annotations should be there.
+		FieldShouldContain("ns", "", nsB, ".metadata.annotations", "hnc.x-k8s.io/subnamespaceOf:"+nsA)
+		FieldShouldContain("ns", "", nsB, ".metadata.annotations", "hnc.x-k8s.io/subnamespace-of:wrongvalue")
```
adrianludwin, line 85: Hmm, I'm not sure if I agree with this. It seems to make the logic more complicated but for not much gain. We could just say "if the new one exists, it _always_ overrules the old one" because I can't really imagine any reasonable cases where it would accidentally have the wrong value.

incubator/hnc/pkg/testutils/testutils.go
yiqigao217: Oh this function cleans up namespaces in both v0.5 and v0.6. Updating using your RB test code sgtm.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: sgtm.

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Done.

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: Hmm, maybe comment why we still need to clean up the v0.5 versions and when this can be removed?

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: RE second time reconciling: the complicated logic is mostly for lowering the risk at API conversion (when HNC restarts).

SGTM, updated the logic to always respect the new annotation.

incubator/hnc/pkg/testutils/testutils.go
yiqigao217: Updated the code and comment.

adrianludwin: I'm not seeing the rolebinding code updated in the latest commit?

adrianludwin:
lgtm after fixing the rolebinding test and clarifying the comment on when we can stop cleaning up the old annotation

@rjbez17 can you PTAL?

incubator/hnc/pkg/testutils/testutils.go
yiqigao217: What rolebinding code you are talking about? I thought it's https://github.com/kubernetes-sigs/multi-tenancy/pull/1164/commits/6665340e7956fccc73e98212447c6f91301174ec that we will have `TryRunQuietly()` to make sure the namespace exists and use `MustRunWithTimeout()`. is it not what you want?

adrianludwin: Resolved offline, my change is in testutils not rolebinding, and Yiqi had to rebase to see it (but there was no line conflict so Github didn't alert her that she needed to rebase).

adrianludwin: (the change is in `RecoverHNC()`)

yiqigao217: Updated and tested rolebinding with `FIt` in `make test-e2e`.

adrianludwin (commented):
/lgtm
/hold
/approve
/assign @rjbez17

Issue #1123, opened 2020-09-18
HNC: exceptions feature


GinnyJI:
This issue is to track the implementation of exceptions, the design doc can be found at https://docs.google.com/document/d/17J8icBEDvLLoPT4kQ4ArZcCerRweDY-TpJ48DJKpHJ0

adrianludwin:
This isn't _required_ for HNC v0.6 to moving to v0.7. However much we can get done in v0.6 is what we'll release.

adrianludwin:
I think we're done here - nice work!

/close

PR #1161, opened 2020-09-30
Implement selector feature using 'select' annotation

GinnyJI:
Tested: make test

### DIFF
--- a/incubator/hnc/api/v1alpha2/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha2/hierarchy_types.go
@@ -31,11 +31,14 @@ const (
 
 // Constants for labels and annotations
 const (
-	MetaGroup                = "hnc.x-k8s.io"
-	LabelInheritedFrom       = MetaGroup + "/inheritedFrom"
-	FinalizerHasSubnamespace = MetaGroup + "/hasSubnamespace"
-	LabelTreeDepthSuffix     = ".tree." + MetaGroup + "/depth"
-	AnnotationManagedBy      = MetaGroup + "/managedBy"
+	MetaGroup                 = "hnc.x-k8s.io"
+	LabelInheritedFrom        = MetaGroup + "/inheritedFrom"
+	FinalizerHasSubnamespace  = MetaGroup + "/hasSubnamespace"
+	LabelTreeDepthSuffix      = ".tree." + MetaGroup + "/depth"
+	AnnotationManagedBy       = MetaGroup + "/managedBy"
+	AnnotationPropagatePrefix = "propagate." + MetaGroup
+
+	AnnotationSelector = AnnotationPropagatePrefix + "/select"
 )
 
 // Condition codes. *All* codes must also be documented in the comment to Condition.Code, be
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -413,6 +413,9 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 		anc = anc.Parent()
 		depth++
 	}
+	// Update the labels in the forest so that we can quickly access the labels and
+	// compare if they match the given selector
+	ns.SetLabels(nsInst.Labels)
 }
 
 func (r *HierarchyConfigReconciler) syncConditions(log logr.Logger, inst *api.HierarchyConfiguration, ns *forest.Namespace, deletingCRD, hadCrit bool) {
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -284,7 +285,7 @@ func (r *ObjectReconciler) syncMissingObject(ctx context.Context, log logr.Logge
 	// descendants, but there's nothing else to do.
 	if ns.HasSourceObject(r.GVK, inst.GetName()) {
 		ns.DeleteSourceObject(r.GVK, inst.GetName())
-		r.syncUnpropagatedSource(ctx, log, inst)
+		r.syncPropagation(ctx, log, inst)
 		return actionNop
 	}
 
@@ -375,7 +376,7 @@ func (r *ObjectReconciler) getTopSourceToPropagate(log logr.Logger, inst *unstru
 		// TODO: add a webhook rule to prevent e.g. removing a source finalizer that
 		//  would cause overwrting the source objects in the descendents.
 		//  See https://github.com/kubernetes-sigs/multi-tenancy/issues/1120
-		if !r.shouldPropagateSource(log, obj) {
+		if !r.shouldPropagateSource(log, obj, inst.GetNamespace()) {
 			continue
 		}
 		return obj
@@ -431,23 +432,8 @@ func (r *ObjectReconciler) syncSource(ctx context.Context, log logr.Logger, src 
 	ns := r.Forest.Get(src.GetNamespace())
 	ns.SetSourceObject(src.DeepCopy())
 
-	if !r.shouldPropagateSource(log, src) {
-		// If an object was *previously* propagated by HNC, it will already be in the forest, and all
-		// the propagated copies will think they can stick around too. If, for some reason, we _now_
-		// want to stop propagating it (e.g. because we've changed the sync mode in the HNCConfig
-		// object), we need to delete it from the forest and enqueue all its propagated copies. The
-		// propagated copies will then see that that the source is missing from the forest, and delete
-		// themselves.
-		r.syncUnpropagatedSource(ctx, log, src)
-		return
-	}
-
-	// Signal the config reconciler for reconciliation because it is possible that a source object is
-	// added to the apiserver.
-	hnccrSingleton.requestReconcile("possible new source object")
-
-	// Enqueue all the descendant copies
-	r.enqueueDescendants(ctx, log, src, "possible new or modified source object")
+	// Enqueue propagated copies for this possibly deleted source
+	r.syncPropagation(ctx, log, src)
 }
 
 func (r *ObjectReconciler) enqueueDescendants(ctx context.Context, log logr.Logger, src *unstructured.Unstructured, reason string) {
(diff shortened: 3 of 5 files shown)


adrianludwin:
```
Ran 2 of 58 Specs in 7.171 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 56 Skipped
PASS | FOCUSED
FAIL	sigs.k8s.io/multi-tenancy/incubator/hnc/internal/reconcilers	7.222s
```

@GinnyJI I think the problem is that you checked in some focused tests? So the tests you focused on passed, but the test _suite_ failed because (I'm guessing) you're not supposed to check in focused tests.

GinnyJI:
> ```
> Ran 2 of 58 Specs in 7.171 seconds
> SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 56 Skipped
> PASS | FOCUSED
> FAIL	sigs.k8s.io/multi-tenancy/incubator/hnc/internal/reconcilers	7.222s
> ```
> 
> @GinnyJI I think the problem is that you checked in some focused tests? So the tests you focused on passed, but the test _suite_ failed because (I'm guessing) you're not supposed to check in focused tests.

Oh good point. Just removed the focused test.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -413,6 +413,7 @@ func (r *HierarchyConfigReconciler) syncLabel(log logr.Logger, nsInst *corev1.Na
 		anc = anc.Parent()
 		depth++
 	}
+	ns.SetLabels(nsInst.Labels)
```
adrianludwin, line 416: This is probably worth a comment ("why are we doing this?")

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -763,7 +764,13 @@ func (r *ObjectReconciler) syncUnpropagatedSource(ctx context.Context, log logr.
 // - Objects of a type whose mode is set to "remove" in the HNCConfiguration singleton
 // - Objects with nonempty finalizer list
 // - Service Account token secrets
-func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured) bool {
+func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst, destInst *unstructured.Unstructured) bool {
```
adrianludwin, line 767: The only information from `destInst` we're using here is its namespace name. It's a good idea to minimize the amount of information that you pass around the system, so I'd suggest just passing in the destination namespace name as a string.

Also very minor nit - we usually use `dst` instead of `dest` because it lines up nicely with `src`. I'd either call the string argument `dst` or `dstNS` but I think `dst` is clear enough from the context.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -763,7 +764,13 @@ func (r *ObjectReconciler) syncUnpropagatedSource(ctx context.Context, log logr.
...
+	var selector labels.Selector
+	var nsLabels labels.Set
+	if destInst != nil {
+		selector = r.getSelector(log, inst)
+		nsLabels = labels.Set(r.Forest.Get(destInst.GetNamespace()).GetLabels())
```
adrianludwin, line 772: Should `getLabels()` return a `labels.Set`, if that's the only way this is ever used?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -793,6 +804,26 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
 
+// getSelector returns the selector on a given object if it exists
+func (r *ObjectReconciler) getSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	annot := inst.GetAnnotations()
+	selectorStr, ok := annot[api.AnnotationSelector]
+	if !ok {
+		return nil
+	}
+	labelSelector, err := v1.ParseToLabelSelector(selectorStr)
+	// TODO: surface the error messages here
```
adrianludwin, line 815: File issue?

adrianludwin (commented):
Looking nice!

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Agreed. Is it to update the labels in the forest?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -431,7 +432,7 @@ func (r *ObjectReconciler) syncSource(ctx context.Context, log logr.Logger, src
 	ns := r.Forest.Get(src.GetNamespace())
 	ns.SetSourceObject(src.DeepCopy())
 
-	if !r.shouldPropagateSource(log, src) {
+	if !r.shouldPropagateSource(log, src, nil) {
```
yiqigao217, line 435: Add a comment here that it only checks if the source itself is propagatable in general but not to a specific destination. We will enqueue all the descendants if it's propagatable and decide whether to propagate when reconciling the descendants.

incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -31,11 +31,15 @@ const (
...
-	AnnotationManagedBy      = MetaGroup + "/managedBy"
+	MetaGroup                 = "hnc.x-k8s.io"
+	LabelInheritedFrom        = MetaGroup + "/inheritedFrom"
+	FinalizerHasSubnamespace  = MetaGroup + "/hasSubnamespace"
+	LabelTreeDepthSuffix      = ".tree." + MetaGroup + "/depth"
+	AnnotationManagedBy       = MetaGroup + "/managedBy"
+	AnnotationPropagatePrefix = "propagate." + MetaGroup
+
+	AnnotationSelector     = AnnotationPropagatePrefix + "/select"
+	AnnotationTreeSelector = AnnotationPropagatePrefix + "/treeSelect"
```
yiqigao217, line 42: Is `AnnotationTreeSelector` not used anywhere in this PR yet?

incubator/hnc/internal/reconcilers/object.go
yiqigao217: Also FYI I was writing a "HNC Exceptions - handling errors" doc on this with possible events or webhooks.

incubator/hnc/api/v1alpha2/hierarchy_types.go
GinnyJI: yes

incubator/hnc/internal/reconcilers/hierarchy_config.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go

incubator/hnc/internal/reconcilers/object.go

incubator/hnc/internal/reconcilers/object.go

incubator/hnc/internal/reconcilers/object.go

adrianludwin:
> Tested: make test

Can you please update this commit message with more information (e.g. link to the design doc, summary of the changes) as well as more information about the testing? E.g. why is `make test` sufficient?

incubator/hnc/api/v1alpha2/hierarchy_types.go
adrianludwin: If it's not used, it should probably be removed. It's best not to add stuff before you need it.

incubator/hnc/internal/reconcilers/hierarchy_config.go
adrianludwin: This comment doesn't answer the question of _why_ we're doing this - it just restates _what_ the code is doing but in a different way. Usually, you can figure out the "what" just by reading the code more carefully, but "why" is much harder to figure out.

Imagine you're looking at this code in a year and you've completely forgotten why you wrote it. Can you leave yourself a hint so that you know what to look at to understand why you can't delete it?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -431,7 +432,10 @@ func (r *ObjectReconciler) syncSource(ctx context.Context, log logr.Logger, src
 	ns := r.Forest.Get(src.GetNamespace())
 	ns.SetSourceObject(src.DeepCopy())
 
-	if !r.shouldPropagateSource(log, src) {
+	// We are only checking if the source itself is propagatable, but not to a specific destination.
+	// We will enqueue all the descendants if it's propagatable and decide whether to propagate when
+	// reconciling the descendants.
+	if !r.shouldPropagateSource(log, src, "") {
 		// If an object was *previously* propagated by HNC, it will already be in the forest, and all
```
adrianludwin, line 439: This comment is out of date now - can you chat with Yiqi to update it?

GinnyJI: Updated

GinnyJI:
tracked at #1123

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -431,13 +432,11 @@ func (r *ObjectReconciler) syncSource(ctx context.Context, log logr.Logger, src
...
 	ns.SetSourceObject(src.DeepCopy())
 
-	if !r.shouldPropagateSource(log, src) {
-		// If an object was *previously* propagated by HNC, it will already be in the forest, and all
-		// the propagated copies will think they can stick around too. If, for some reason, we _now_
-		// want to stop propagating it (e.g. because we've changed the sync mode in the HNCConfig
-		// object), we need to delete it from the forest and enqueue all its propagated copies. The
-		// propagated copies will then see that that the source is missing from the forest, and delete
-		// themselves.
+	// We are only checking if the source itself is propagatable, but not to a specific destination.
```
adrianludwin, line 435: Hmm, I'm wondering if we can simplify the rest of this function. Right now it's:

```
if !should propagate source {
  sync unpropagated source()
  // this does an hncconfig reconcile and enqueues all descendants
  return
}
hncconfig reconcile
enqueue descendants
```

But both the `if` and (implied) `else` clause here are the same! So we can just get rid of the call to `shouldPropagateSource`, and since (I think) this is the only place where we call that function with the empty string as the namespace, we can remove the logic to handle "all namespaces" too. Does that look correct to you?

adrianludwin (commented):
Looking good but can we make it even simpler?

incubator/hnc/internal/reconcilers/object.go
GinnyJI: yea that looks right, done

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -431,23 +432,9 @@ func (r *ObjectReconciler) syncSource(ctx context.Context, log logr.Logger, src
...
-
-	// Signal the config reconciler for reconciliation because it is possible that a source object is
-	// added to the apiserver.
-	hnccrSingleton.requestReconcile("possible new source object")
-
-	// Enqueue all the descendant copies
-	r.enqueueDescendants(ctx, log, src, "possible new or modified source object")
+	// Enqueue propagated copies for this possibly deleted source
+	r.syncUnpropagatedSource(ctx, log, src)
+	return
```
adrianludwin, line 437: No need for a "bare" return statement at the end of a function

yiqigao217, line 437: If you combined the "syncUnpropagatedSource" and the "syncPropagatedSource" (old line 445-450) maybe update the `syncUnpropagatedSource(arg..)` name to `syncPropagation(arg.., bool)` with a new boolean to indicate whether to propagate or delete since the enqueue message is different. Or change the enqueue message to more general without saying `deleted source` or `new or modified source`. 

Overall, simplifying the logic lgtm. Just need a little cleanup on the `syncUnpropagatedSource()` func.

adrianludwin: +1 to all of this, especially making the enqueue message more general.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -31,11 +31,14 @@ const (
...
-	MetaGroup                = "hnc.x-k8s.io"
-	LabelInheritedFrom       = MetaGroup + "/inheritedFrom"
-	FinalizerHasSubnamespace = MetaGroup + "/hasSubnamespace"
-	LabelTreeDepthSuffix     = ".tree." + MetaGroup + "/depth"
-	AnnotationManagedBy      = MetaGroup + "/managedBy"
+	MetaGroup                 = "hnc.x-k8s.io"
+	LabelInheritedFrom        = MetaGroup + "/inheritedFrom"
+	FinalizerHasSubnamespace  = MetaGroup + "/hasSubnamespace"
+	LabelTreeDepthSuffix      = ".tree." + MetaGroup + "/depth"
+	AnnotationManagedBy       = MetaGroup + "/managedBy"
```
rjbez17, line 38: nit: maybe link issue to rename to `managed-by`?

rjbez17:
/lgtm

feel free to remove hold if we don't need the issue for managed-by (I think we plan to have a PR for it in the next day or two?)

incubator/hnc/api/v1alpha2/hierarchy_types.go
adrianludwin: That's not really related to this change, but I'm working on it now anyway.

adrianludwin:
> feel free to remove hold if we don't need the issue for managed-by (I think we plan to have a PR for it in the next day or two?)

Yup I'm working on it now.

/hold cancel

PR #1162, opened 2020-09-30
Create minimial v1a2 to v1a1 converters

adrianludwin:
While investigating another issue, I looked at the logs from the K8s
apiserver and observed that we were seeing tons of errors as the
apiserver asked HNC to convert the v1a2 objects back to v1a1 -
specifically, that their names has changed. I have no idea why K8s is
asking for these objects to be backconverted, but after glancing at the
K8s source code, I guessed that copying the object metadata would work
to fix the issue, and it did.

Tested: observed no more logspam from the apiserver


yiqigao217 (commented):
/lgtm

PR #1163, opened 2020-09-30
Ensure all objects are reconciled at least once

adrianludwin:
While looking into a sporadic but reproducible (maybe 1/5 times) failure
of the rolebinding recreation test, I observed that the objects in the
`parent` namespace were occasionally not being reconciled after HNC
restarted. This appeared to be because the HC reconciler was not
modifying either the HC or the namespace, and so was skipping the update
of the object reconcilers. This resulted in source objects not being
inserted into the forest, and as a result, propagated objects were being
removed.

By adding a check that the object reconcilers are called when a
namespace is first added to the forest, we can be certain that all
objects are present in the forest.

Tested: ran the rolebinding e2e test 10/10 times successfully.

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -129,10 +129,13 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 	r.updateFinalizers(ctx, log, inst, nsInst, anms)
 
 	// Sync the Hierarchy singleton with the in-memory forest.
-	r.syncWithForest(log, nsInst, inst, deletingCRD, anms)
+	initial := r.syncWithForest(log, nsInst, inst, deletingCRD, anms)
 
-	// Write back if anything's changed. Early-exit if we just write back exactly what we had.
-	if updated, err := r.writeInstances(ctx, log, origHC, inst, origNS, nsInst); !updated || err != nil {
+	// Write back if anything's changed. Early-exit if we just write back exactly what we had and this
+	// isn't the first time we're syncing.
+	updated, err := r.writeInstances(ctx, log, origHC, inst, origNS, nsInst)
+	updated = updated || initial
+	if !updated || err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


yiqigao217 (commented):
> While looking into a sporadic but reproducible (maybe 1/5 times) failure
> of the rolebinding recreation test, I observed that the objects in the
> `parent` namespace were occasionally not being reconciled after HNC
> restarted. This appeared to be because the HC reconciler was not
> modifying either the HC or the namespace, and so was skipping the update
> of the object reconcilers.

Why it's flaky but not consistent failure if HC reconciler was not modifying objects? When would it modify objects and cause them to be enqueued to update forest and passed the test?

adrianludwin:
> > While looking into a sporadic but reproducible (maybe 1/5 times) failure
> > of the rolebinding recreation test, I observed that the objects in the
> > `parent` namespace were occasionally not being reconciled after HNC
> > restarted. This appeared to be because the HC reconciler was not
> > modifying either the HC or the namespace, and so was skipping the update
> > of the object reconcilers.
> 
> Why it's flaky but not consistent failure if HC reconciler was not modifying objects? When would it modify objects and cause them to be enqueued to update forest and passed the test?

I _believe_ it's because there's a race condition between reconciling the objects and the namespace. If the namespace is reconciled first, then when we reconcile the source object, it gets inserted into the forest. But if we reconcile the source object first, then the namespace doesn't exist in the forest and we just ignore the object, under the assumption that when we _do_ sync the namespace, we'll enqueue all the objects in that namespace.

yiqigao217:
> I _believe_ it's because there's a race condition between reconciling the objects and the namespace. If the namespace is reconciled first, then when we reconcile the source object, it gets inserted into the forest. But if we reconcile the source object first, then the namespace doesn't exist in the forest and we just ignore the object, under the assumption that when we _do_ sync the namespace, we'll enqueue all the objects in that namespace.

Sorry that I got a little confused. It seems to me that even before your change, if namespace is reconciled first, if there's no change, `r.updateObjects(ctx, log, nm)` won't get chance to be called?

adrianludwin:
> Sorry that I got a little confused. It seems to me that even before your change, if namespace is reconciled first, if there's no change, `r.updateObjects(ctx, log, nm)` won't get chance to be called?

That's correct - if the namespace is reconciled first, then `r.updateObjects` won't be called. But the problem is here:

https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/object.go#L263

If the namespace is reconciled first, then even though `r.updateObjects` isn't called, when the object is reconciled _second_ it will be put into the forest, and everything is fine. But if the _object_ is reconciled first, then `ns.Exists()` returns false and the object is _not_ placed into the forest. And then, because `r.updateObjects` isn't called, it's never reconciled again. This fix addresses that second case.

Does this make sense?

yiqigao217:
> > Sorry that I got a little confused. It seems to me that even before your change, if namespace is reconciled first, if there's no change, `r.updateObjects(ctx, log, nm)` won't get chance to be called?
> 
> That's correct - if the namespace is reconciled first, then `r.updateObjects` won't be called. But the problem is here:
> 
> https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/object.go#L263
> 
> If the namespace is reconciled first, then even though `r.updateObjects` isn't called, when the object is reconciled _second_ it will be put into the forest, and everything is fine. But if the _object_ is reconciled first, then `ns.Exists()` returns false and the object is _not_ placed into the forest. And then, because `r.updateObjects` isn't called, it's never reconciled again. This fix addresses that second case.
> 
> Does this make sense?

Yes, makes sense now. So if namespace reconciles first, it would be fine since the object reconciler would not early exit with `actionNop`. And this PR is to make sure we don't early exit in the HC reconciler if the namespace is first-time synced and make sure objects are enqueued.

/lgtm

adrianludwin:
Exactly!

PR #1164, opened 2020-09-30
Make e2e test repairs more robust

adrianludwin:
Made several debuggability and functional improvements:

* Added timestamps to all output to correlate with logs. Timestamps are
  of the form "seconds since Unix epoch" which isn't human friendly but
  is identical to the timestamps produced by HNC logs.
* Fully deleted the HNC deployment in RecoverHNC since HNC doesn't
  really seem to recover well if various things are changed without
  restarting the pod.
* Make the post-recovery test far more robust by not ignoring failures
  to delete the test namespaces.
* Stop force-deleting pods in the rolebinding test and instead just wait
  a few moments. I found that we don't actually need to wait for the pod
  object to be fully deleted on the server for it to stop getting in the
  way; the container in the pod appears to stop running ~instantly while
  the pod can occasionally hang around for over 60s.

All of these were added as I saw failures in the affected code.

Tested: Ran a set of five flaky tests with and without these changes
(while also including PR #1163). Without, at least one of them failed
virtually every time; with this change, they passed 5/5 times on GKE
plus 2/2 times on Kind.

### DIFF
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -203,7 +203,8 @@ func RunCommand(cmdln ...string) (string, error) {
 			args = append(args, strings.Split(subcmdln, " ")...)
 		}
 	}
-	GinkgoT().Log("Running: ", args)
+	prefix := fmt.Sprintf("[%d] Running: ", time.Now().Unix())
+	GinkgoT().Log(prefix, args)
 	cmd := exec.Command(args[0], args[1:]...)
 	stdout, err := cmd.CombinedOutput()
 	return string(stdout), err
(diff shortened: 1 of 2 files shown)


incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -257,13 +263,34 @@ func RecoverHNC() {
...
+	// TODO: if CleanupNamespaces ever starts using labels to select namespaces to delete, then get
+	// rid of this hack.
+	if err := TryRunQuietly("kubectl get ns", a); err == nil {
+		MustRunWithTimeout(30, "kubectl hns set", a, "-a")
+		MustRunWithTimeout(30, "kubectl delete ns", a)
+	}
+	if err := TryRunQuietly("kubectl get ns", b); err == nil {
+		MustRunWithTimeout(30, "kubectl annotate ns", b, "hnc.x-k8s.io/subnamespaceOf-")
+		MustRunWithTimeout(30, "kubectl delete ns", b)
+	}
```
yiqigao217, line 285: Hmm, this looked the same to me as `CleanupNamespaces()` except skipping de-annotating a and no setting AC for b. What makes a difference?

adrianludwin: CleanupNamespaces is TryRunQuietly, this is MustRunWithTimeout.

yiqigao217 (commented):
Ah I see.
/lgtm
/hold
/assign @rjbez17

adrianludwin:
Test-only changes don't need double-approval, though @rjbez17 please lmk if you have any suggestions and I'll follow them up.

/hold cancel

PR #1169, opened 2020-10-05
Add '--force' flag when switching from 'Ignore' to 'Propagate' mode

GinnyJI:
If the user switch the propagation mode directly from 'Ignore' to
'Propagate', it will fail with an error message telling the user that a
'--force' flag is needed. '-f' is also supported as a shorthand.

Tested: make test-e2e

### DIFF
--- a/incubator/hnc/internal/kubectl/configset.go
+++ b/incubator/hnc/internal/kubectl/configset.go
@@ -17,6 +17,7 @@ package kubectl
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 
@@ -37,12 +38,22 @@ var setTypeCmd = &cobra.Command{
 		flags := cmd.Flags()
 		apiVersion, _ := flags.GetString("apiVersion")
 		kind, _ := flags.GetString("kind")
+		force, _ := flags.GetBool("force")
 		config := client.getHNCConfig()
 
 		exist := false
 		for i := 0; i < len(config.Spec.Types); i++ {
 			t := &config.Spec.Types[i]
 			if t.APIVersion == apiVersion && t.Kind == kind {
+				if t.Mode == api.Ignore && mode == api.Propagate && !force {
+					fmt.Println("Switching directly from 'Ignore' to 'Propagate' mode could cause existing %s objects in "+
+						"child namespaces to be overwritten by objects from ancestor namespaces.", kind)
+					fmt.Println("If you are sure you want to proceed with this operation, use the '--force' flag.")
+					fmt.Println("If you are not sure and would like to see what source objects would be overwritten," +
+						"please switch to 'Remove' first. To see how to enable propagation safely, refer to " +
+						"https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/how-to.md#admin-types")
+					os.Exit(1)
+				}
 				t.Mode = mode
 				exist = true
 				break
@@ -65,5 +76,6 @@ var setTypeCmd = &cobra.Command{
 func newSetTypeCmd() *cobra.Command {
 	setTypeCmd.Flags().String("apiVersion", "", "API version of the kind")
 	setTypeCmd.Flags().String("kind", "", "Kind to be configured")
+	setTypeCmd.Flags().BoolP("force", "f", false, "Force to set the propagation mode")
 	return setTypeCmd
 }
--- a/incubator/hnc/test/e2e/kubectl_test.go
+++ b/incubator/hnc/test/e2e/kubectl_test.go
@@ -1,0 +1,16 @@
+package e2e
+
+import (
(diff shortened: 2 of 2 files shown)


adrianludwin:
/assign @yiqigao217 

Yiqi, can you please work with Ginny to fine-tune the error message? We should explain exactly why this feature exists, and maybe link to some documentation as well (which we might need to write).

incubator/hnc/test/e2e/hns_test.go
```diff
@@ -0,0 +1,14 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils"
+)
+
+var _ = Describe("HNS set-config", func() {
+	It("Should use '--force' flag to change from 'Ignore' to 'Propagate'", func() {
```
adrianludwin, line 9: You should test that none of the other transitions require `--force`, eg Ignore->Remove, Remove->Propagate, Propagate->anything. Or maybe that's overkill and at least verify that _one_ of them doesn't require `--force`.

adrianludwin (commented):
Basic code lgtm! We need to tweak the message, please work on this with Yiqi.

incubator/hnc/test/e2e/hns_test.go
GinnyJI: done

incubator/hnc/test/e2e/hns_test.go
```diff
@@ -0,0 +1,14 @@
+package e2e
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/internal/kubectl/configset.go
```diff
@@ -37,12 +38,21 @@ var setTypeCmd = &cobra.Command{
...
 		exist := false
 		for i := 0; i < len(config.Spec.Types); i++ {
 			t := &config.Spec.Types[i]
 			if t.APIVersion == apiVersion && t.Kind == kind {
+				if t.Mode == api.Ignore && mode == api.Propagate && !force {
+					fmt.Println("Switching directly from 'Ignore' to 'Propagate' mode might cause source overwriting conflict.")
+					fmt.Println("If you are sure you want to proceed with this operation, use the '--force' flag.")
+					fmt.Println("If you are not sure and would like to see what source objects would be overwritten," +
+						"please switch to 'Remove' first. Wait 10s before switching it to 'Propagate', which will" +
+						" list all the conflicts that you can resolve them by hand.")
```
adrianludwin, line 53: I'd add a link to https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/how-to.md#admin-types, and file a bug to update that topic to discuss this issue.

Also, I'm not sure that 10s is enough? In the doc you can go into more detail; maybe here I'd either say "a minute or two" or just say "to see how to enable propagation safely, see \<link\>."

GinnyJI: Done. Issue filed at https://github.com/kubernetes-sigs/multi-tenancy/issues/1172

incubator/hnc/internal/kubectl/configset.go
```diff
@@ -37,12 +38,21 @@ var setTypeCmd = &cobra.Command{
...
 		kind, _ := flags.GetString("kind")
+		force, _ := flags.GetBool("force")
 		config := client.getHNCConfig()
 
 		exist := false
 		for i := 0; i < len(config.Spec.Types); i++ {
 			t := &config.Spec.Types[i]
 			if t.APIVersion == apiVersion && t.Kind == kind {
+				if t.Mode == api.Ignore && mode == api.Propagate && !force {
+					fmt.Println("Switching directly from 'Ignore' to 'Propagate' mode might cause source overwriting conflict.")
```
adrianludwin, line 49: Last change - users won't know what "source overwriting conflict" means. Let's use more natural language: "could cause existing %s objects in child namespaces to be overwritten by objects from ancestor namespaces" (where `%s` is `t.Kind`). The rest of the message lgtm.

adrianludwin (commented):
lgtm after one last nit on the message, but I'll send this to Ryan now.

/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/internal/kubectl/configset.go
GinnyJI: done

yiqigao217 (commented):
Sorry missed this PR earlier. BTW seems we still need e2e tests on switching modes between others, e.g. allow Remove => Propagate, etc.

adrianludwin:
> Sorry missed this PR earlier. BTW seems we still need e2e tests on switching modes between others, e.g. allow Remove => Propagate, etc.

I don't think we need exhaustive testing, it doesn't have to be 100% black-box given how simple the feature is.

Issue #1167, opened 2020-10-02
HNC: Update object conditions into events in v1alpha2


yiqigao217:
Part of #868 

Feedback from API reviewers: Current object conditions in HierarchyConfiguration CRs are not intuitive. In addition, having all object conditions in HC would overflow the size and we would have to truncate it.

We also had a lot of trouble fixing bugs related to setting and unsetting conditions. Therefore for object conditions, we will replace it with events in v1alpha2.

We will eventually let users get object errors with `kubectl hns describe`.

yiqigao217:
I followed https://book-v1.book.kubebuilder.io/beyond_basics/creating_events.html to experiment with events. I found that for `pods`, we can get events just by `kubectl describe`. However, for other types, e.g. `role` or `secret`, I cannot get events directly by `kubectl describe` but only from `kubectl get events`:
```
$ k describe role -n a aa
Name:         aa
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources               Non-Resource URLs  Resource Names  Verbs
  ---------               -----------------  --------------  -----
  deployments.extensions  []                 []              [update]

$ k describe secrets -n a my-creds2 
Name:         my-creds2
Namespace:    a
Labels:       <none>
Annotations:  <none>

Type:  Opaque

Data
====
password:  8 bytes

$ k get events -n a
LAST SEEN   TYPE     REASON       OBJECT                       MESSAGE
5m16s       Normal   YG_Created   role/aa                      YG_Created deployment YG/YG
4m4s        Normal   YG_Created   secret/default-token-n7jtp   YG_Created deployment YG/YG
4m4s        Normal   YG_Created   secret/my-creds2             YG_Created deployment YG/YG
6m4s        Normal   YG_Created   role/team-a-sre              YG_Created deployment YG/YG
```

In that case, we can probably attach `kubectl get events -n` results to the current `kubectl hns describe` in kubectl implementation. WDYT? @adrianludwin 

/cc @GinnyJI

adrianludwin:
That sgtm. It would have been great if they appeared when describing the objects themselves but the current "conditions" method doesn't have that either, and the simplification to the implementation still makes it well worthwhile IMO.

yiqigao217:
The second half of this issue is to mask `k get events` (with different flags, e.g. `-n`, `--field-selector`) with `kubectl hns describe`, so that users can get object errors. We can probably append the events to the existing `k hns describe` output.

[...]

PR #1171, opened 2020-10-05
Replace object conditions with events

yiqigao217:
Instead of setting/unsetting object conditions in HC objects, create
events if there's any errors propagating objects.

Update integration tests and e2e tests.

Tesed by `make test`, `make test-e2e` and manually on a GKE cluster. The
output looks like:
```
$ k get events -n child --field-selector involvedObject.name=cred
LAST SEEN   TYPE      REASON                  OBJECT        MESSAGE
112s        Warning   CannotUpdateObject      secret/cred   Could not write from source namespace "parent": <error msg>.

$ k get events -n parent --field-selector involvedObject.name=cred
LAST SEEN   TYPE      REASON                  OBJECT        MESSAGE
2m43s       Warning   CannotPropagateObject   secret/cred   Could not write to destination namespace "child": <error msg>.
```

Part of #1167


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 
cc @GinnyJI

incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -47,25 +47,37 @@ const (
...
 
+const (
+	// CannotPropagate is one of the object propagation errors when a namespace
+	// contains an object that couldn't be propagated *out* of the namespace, to
+	// one or more of its descendants. If the object couldn't be propagated to
+	// *any* descendants - for example, because it has a finalizer on it (HNC
+	// can't propagate objects with finalizers), the error message will point to
+	// the object in this namespace. Otherwise, if it couldn't be propagated to
+	// *some* descendant, the error message will point to the descendant.
+	CannotPropagate           string = "CannotPropagateObject"
```
adrianludwin, line 73: Maybe give this an "Event" prefix so that it's clear how this should be used?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -601,69 +603,26 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
-// This function also enqueues any modified namespaces for reconciliation.
-func (r *ObjectReconciler) setConditions(log logr.Logger, srcInst, inst *unstructured.Unstructured, act syncAction, err error) {
-	r.Forest.Lock()
-	defer r.Forest.Unlock()
-	ao := api.NewAffectedObject(inst.GetObjectKind().GroupVersionKind(), inst.GetNamespace(), inst.GetName())
-	// This affected object is initialized with the source object if it exits.
-	sao := api.NewAffectedObject(inst.GetObjectKind().GroupVersionKind(), inst.GetLabels()[api.LabelInheritedFrom], inst.GetName())
-	ns := r.Forest.Get(inst.GetNamespace())
-
+// generateWarnings is called when the reconciler has performed all necessary
```
adrianludwin, line 606: generateEvents?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -82,7 +82,9 @@ var _ = Describe("Issues", func() {
 		MustRun("kubectl delete subns", nsChild, "-n", nsParent)
 	})
 
-	It("Should remove obsolete conditions CannotPropagateObject and CannotUpdateObject - issue #328", func() {
+	// This issue is obsolete after introducing events to surface errors in the
+	// object reconciliation, thus the e2e test case is obsolete too.
+	PIt("Should remove obsolete conditions CannotPropagateObject and CannotUpdateObject - issue #328", func() {
```
adrianludwin, line 87: Any reason not to just delete these?

adrianludwin (commented):
Looks great so far, just a few minor comments.

incubator/hnc/test/e2e/issues_test.go
yiqigao217: Deleted and have the issue IDs in the new `Should` clause saying `replacing obsolete issues x, y, z`.

adrianludwin: If the issues are truly obsolete IMO we don't need a record of them :)

adrianludwin:
Just one last question - do you know if these events get de-duplicated? E.g. if we get a dozen cannot-propagate events for one object, does that result in one Event or twelve?

yiqigao217:
> Just one last question - do you know if these events get de-duplicated? E.g. if we get a dozen cannot-propagate events for one object, does that result in one Event or twelve?

There's a `count` field to dedup the same event. Here's an example when I manually set `err = fmt.Errorf("YG")` after https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/object.go#L512 so it can throw events. I am not sure why the `count` is 15 though:
```
$ k get events -n child
LAST SEEN   TYPE      REASON               OBJECT                       MESSAGE
2s          Warning   CannotUpdateObject   secret/default-token-nq27d   Could not remove from source namespace "<unknown-source-namespace>": YG.

$ k get secrets -n parent 
NAME                  TYPE                                  DATA   AGE
cred                  Opaque                                1      3m6s
default-token-nq27d   kubernetes.io/service-account-token   3      3m39s

$ k get events -n child -oyaml
apiVersion: v1
items:
- apiVersion: v1
  count: 15
  eventTime: null
  firstTimestamp: "2020-10-06T14:39:44Z"
  involvedObject:
    apiVersion: v1
    kind: Secret
    name: default-token-nq27d
    namespace: child
  kind: Event
  lastTimestamp: "2020-10-06T14:41:06Z"
  message: 'Could not remove from source namespace "<unknown-source-namespace>": YG.'
  metadata:
    creationTimestamp: "2020-10-06T14:39:44Z"
    name: default-token-nq27d.163b6e21e8651e43
    namespace: child
    resourceVersion: "7575"
    selfLink: /api/v1/namespaces/child/events/default-token-nq27d.163b6e21e8651e43
    uid: 56899dbc-e3c7-4d10-af33-f15ca5189b90
  reason: CannotUpdateObject
  reportingComponent: ""
  reportingInstance: ""
  source:
    component: hnc.x-k8s.io
  type: Warning
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

adrianludwin:
> There's a `count` field to dedup the same event. Here's an example when I manually set `err = fmt.Errorf("YG")` after https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/object.go#L512 so it can throw events. I am not sure why the `count` is 15 though:

Cool, that seems perfect. I'm guessing it's 15 times because controller-runtime keeps requeuing when there's an error.

adrianludwin (commented):
/lgtm
/approve
/assign @rjbez17 
/hold

PR #1173, opened 2020-10-06
Switch to managed-by annotation

adrianludwin:
This change updates HNC to recognize the managed-by annotation, and
convert any existing managedBy annotation from v1alpha1 to the new name.
It also adds some minor cleanups to the conversion tests and makes the
CleanupNamespaces function a bit more robust to account for some errors
I saw before the new tests were working.

Tested: new conversion test; all existing conversion tests pass.

### DIFF
--- a/incubator/hnc/api/v1alpha2/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha2/hierarchy_types.go
@@ -35,7 +35,8 @@ const (
 	LabelInheritedFrom        = MetaGroup + "/inheritedFrom"
 	FinalizerHasSubnamespace  = MetaGroup + "/hasSubnamespace"
 	LabelTreeDepthSuffix      = ".tree." + MetaGroup + "/depth"
-	AnnotationManagedBy       = MetaGroup + "/managedBy"
+	AnnotationManagedBy       = MetaGroup + "/managed-by"
+	AnnotationManagedByV1A1   = MetaGroup + "/managedBy" // TODO: remove after v0.6 branches (#1177)
 	AnnotationPropagatePrefix = "propagate." + MetaGroup
 
 	AnnotationSelector = AnnotationPropagatePrefix + "/select"
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -230,22 +230,41 @@ func RunCommand(cmdln ...string) (string, error) {
 	return string(stdout), err
 }
 
+// CleanupNamespaces does everything it can to delete the passed-in namespaces. It also uses very
+// high timeouts (30s) since this function is often called after HNC has just been reinstalled, and
+// it can take a while of HNC to start allowing changes to namespaces again.
+//
+// TODO: also find a way to remove all finalizers on all HC objects (subns and hierarchy config).
+// HNC doesn't put finalizers on namespaces themselves; it blocks namespace deletion by blocking
+// deletion of the objects in it, but if HNC is damaged or missing, this can result in namespaces
+// never being deleted without admin action.
 func CleanupNamespaces(nses ...string) {
-	// Remove all possible objections HNC might have to deleting a namesplace. Make sure it
-	// has cascading deletion so we can delete any of its subnamespace descendants, and
-	// make sure that it's not a subnamespace itself so we can delete it directly.
+	const cleanupTimeout = 30
+
+	// Remove all objections HNC might have to deleting a namespace.
+	toDelete := []string{} // exclude missing namespaces
 	for _, ns := range nses {
-		if err := TryRunQuietly("kubectl get ns", ns); err == nil {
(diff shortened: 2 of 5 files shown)


incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -210,22 +210,35 @@ func RunCommand(cmdln ...string) (string, error) {
...
+		if err := TryRunQuietly("kubectl get ns", ns); err != nil {
+			continue
 		}
+
+		// Remove all possible objections HNC might have to deleting this namespace. If the HCs are
+		// present, turn on cascading deletion so that we can delete this namespace even if it has
+		// subnamespaces. Note that if HNC isn't properly installed or is installed at an older
+		// version in the conversion tests, this function will fail as it tried to update the HC at a
+		// specific version, so we ignore errors. TODO: find a more reliable way of doing this.
+		TryRun("kubectl hns set", ns, "-a")
```
yiqigao217, line 229: If we are removing namespaces one by one and we will remove annotation for each one, can we just remove `k hns set ns -a`? So `CleanupNamespaces` func can be version neutral.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -272,10 +306,8 @@ func setupV1alpha2(){
 	MustRun("kubectl apply -f ../../manifests/hnc-manager.yaml")
 	// Wait for the cert rotator to write caBundles in CRD conversion webhooks.
 	ensureCRDConvWHReady()
-}
 
-// Verify CRDs still have 'v1alpha1' in spec.versions but not in status.storedVersions.
-func verifyCRDConversion(){
+	// Verify CRDs still have 'v1alpha1' in spec.versions but not in status.storedVersions.
```
yiqigao217, line 310: Maybe still keep the `verifyCRDConversion()` func and put the func call at the end of `setupv1a2`?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -308,16 +340,17 @@ func ensureCRDConvWHReady(){
 
 // createSampleV1alpha1Tree creates a tree with 'a' as the root, 'b' as the child.
 func createSampleV1alpha1Tree(){
-	MustRun("kubectl create ns e2e-conversion-test-a")
-	MustRun("kubectl create ns e2e-conversion-test-b")
+	MustRun("kubectl create ns", nsA)
+	MustRun("kubectl create ns", nsB)
```
yiqigao217, line 344: Ah nice, good catch. The variables are now global.

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: Why? We only ever do the verification after the upgrade, and we always want to verify after an upgrade.

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: Yeah, good idea. Will do.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Ah right, sgtm.

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: Done

incubator/hnc/api/v1alpha2/hierarchy_types.go
```diff
@@ -35,7 +35,8 @@ const (
 	LabelInheritedFrom        = MetaGroup + "/inheritedFrom"
 	FinalizerHasSubnamespace  = MetaGroup + "/hasSubnamespace"
 	LabelTreeDepthSuffix      = ".tree." + MetaGroup + "/depth"
-	AnnotationManagedBy       = MetaGroup + "/managedBy"
+	AnnotationManagedBy       = MetaGroup + "/managed-by"
+	AnnotationManagedByOld    = MetaGroup + "/managedBy"
```
rjbez17, line 39: nit: When we changed to subnamespace-of we used deprecated instead of old. 

https://github.com/kubernetes-sigs/multi-tenancy/blob/b6089fff3909031ae593a739135474c4f9c173aa/incubator/hnc/api/v1alpha2/subnamespace_anchor.go#L28

rjbez17: Also, should we add a comment/issue link indicating when/if we can remove this old/deprecated annotation?

adrianludwin: Changed both to the more descriptive "V1A1" suffix and added link to issue.

rjbez17:
/lgtm
/approve
/hold

@yiqigao217 feel free to remove hold if your questions are answered

yiqigao217 (commented):
/lgtm
/hold cancel

PR #1174, opened 2020-10-06
Improve test failure messages; fix some tests

adrianludwin:
Prior to this change, the failure messages from our current e2e and
conversion webhooks were very poor - they don't identify the line number
from the _test_ that causes a failure, since most of the failures are
hidden several levels deep in the testutils package. In addition, almost
no information is printed out about the command that caused the failure
\- you need to dig through the logs. Finally, many failures are printed
in unnecessarily verbose ways.

This change fixes all three problems, by using Gomega's WithOffset
functions to get the test line from the stack trace, using a custom
Gomega matcher to print out nicer information, and printint out the
failing command on failure. Take together, these three changes change
an error like this:

```
  Expected success, but got an error:
      <*errors.errorString | 0xc000067430>: {
          s: "Missing: Ok\n\nGot: Conflict",
      }
      Missing: Ok

      Got: Conflict

  <redacted>/multi-tenancy/incubator/hnc/pkg/testutils/testutils.go:110
```

to this:

```
  Command: [kubectl get subnamespaceanchors.hnc.x-k8s.io e2e-conversion-test-b -n e2e-conversion-test-a -o template --template={{.status.status}}]
  did not output the expected substring(s): Ok
  and instead output: Conflict

  <redacted>/multi-tenancy/incubator/hnc/test/conversion/conversion_test.go:112
```

This should make debugging far easier by identifying the exact command
and line number that caused the failure. Even for more sophisticated
functions like CleanupNamespaces where the line number might not be
useful, printing out the exact failing command should go a long way to
make these failures easier to diagnose.

Finally, while testing this change, I discovered that #1157 had broken
some e2e tests, so I fixed them.

Tested: all basic (see below) e2e and conversion tests pass. Manually
caused several failures like the one shown above and verified that the
error messages were better. I found that the e2e tests that require
HNC_REPAIR were failing but this was due to an unrelated error that's
already addressed by #1173.

### DIFF
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -1,7 +1,6 @@
 package testutils
 
 import (
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -256,6 +276,9 @@ func CheckHNCPath() {
 }
 
 func RecoverHNC() {
+	if hncRecoverPath == "" {
+		return
+	}
```
yiqigao217, line 281: I thought it already existed somewhere?

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -20,59 +19,75 @@ const eventuallyTimeout = 5
...
 func FieldShouldContainWithTimeout(resource, ns, nm, field, want string, timeout float64){
-	FieldShouldContainMultipleWithTimeout(resource, ns, nm, field, []string{want}, timeout)
+	fieldShouldContainMultipleWithTimeout(1, resource, ns, nm, field, []string{want}, timeout)
 }
 
 func FieldShouldContainMultipleWithTimeout(resource, ns, nm, field string, want []string, timeout float64){
+	fieldShouldContainMultipleWithTimeout(1, resource, ns, nm, field, want, timeout)
+}
+
+func fieldShouldContainMultipleWithTimeout(offset int, resource, ns, nm, field string, want []string, timeout float64){
```
GinnyJI, line 37: Why do we want the 'offset', what's it used for?

GinnyJI (commented):
lgtm just one question

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: In the commit message, look at the filename and line numbers. In "before," they point to testutils.go, which isn't useful to figure out which part of the *test* failed. In "after," they identify the line of the test that failed.

The way we do that is to pass an offset to Gomega. This offset specifies the number of levels in the stack that it should pop to print the right filename and line number. See https://onsi.github.io/gomega/#making-assertions-in-helper-functions for more.

yiqigao217 (commented):
lgtm just a minor question.

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: In `BeforeEach` we call `Skip()` if this var isn't set, but I _believe_ based on what I saw that `AfterEach` is still called anyway, which makes sense because there might be things we want to clean up. I _think_ we haven't noticed this before because we ~always have `HNC_REPAIR` set when running tests (at least I do).

What my tests showed is that all tests _before_ the skipped tests worked fine, and all tests that ran _after_ the skipped tests failed because the service was broken. When I looked, the HNC deployment had been deleted, which is all this function will do if `HNC_REPAIR` is missing. Adding this line fixed the problem.

yiqigao217: Thanks for the explanation! It would be nice to have these as a comment too.

adrianludwin: Good point! Done

rjbez17:
lgtm, @GinnyJI can merge when satisfied with the answer to her question

PR #1175, opened 2020-10-06
Fix flaky presubmit tests

GinnyJI:
Add propagationTimeout to give enough time for an object to propagate to
child namespaces.

Tested: make test

### DIFF
--- a/incubator/hnc/internal/reconcilers/hnc_config_test.go
+++ b/incubator/hnc/internal/reconcilers/hnc_config_test.go
@@ -34,6 +34,10 @@ const (
 	// This issue is logged at https://github.com/kubernetes-sigs/multi-tenancy/issues/871
 	statusUpdateTime = 4 * time.Second
 
+	// propagationTimeout is the timeout for objects to propagate in HNC. It always passes when we
+	// set it to 10 secs
+	propagationTimeout = 10 * time.Second
+
 	// rbacAV is a nice short form of the RBAC APIVersion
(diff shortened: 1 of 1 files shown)


incubator/hnc/internal/reconcilers/hnc_config_test.go
```diff
@@ -34,6 +34,10 @@ const (
 	// This issue is logged at https://github.com/kubernetes-sigs/multi-tenancy/issues/871
 	statusUpdateTime = 4 * time.Second
 
+	// propagationTimeout is the timeout for objects to propagate in HNC. It always passes when we
+	// set it to 10 secs
+	propagationTimeout = 10 * time.Second
+
```
adrianludwin, line 40: This is a bit misleading - it wasn't (reliably) failing before you set this to 10s. It's probably better to tell the story of why you put this line here, what timeout it replaces, and how we might know whether we want to get rid of it in the future.

Ditto for the commit message title - instead of "fix flaky presubmit test" (which we don't actually know if we've done), just say something like "increase timeouts on presubmits" so that it's clear from the summary.

yiqigao217: +1. Also maybe create a gist of the failure and link it in this PR?

adrianludwin:
Let's hold on this PR for a bit while I look into it.

adrianludwin:
Replaced by #1179 

/close

PR #1176, opened 2020-10-06
Implement selector when using treeSelect annotation

GinnyJI:
When user set `{"propagate.hnc.x-k8s.io/treeSelect": "!baz"}`, HNC
should not propagate this object to `baz` namespace. This annotation
also supports multiple namespace select, which means if user set `{"propagate.hnc.x-k8s.io/treeSelect": "!baz, !bar"}`, HNC should not propagate this object to neither `baz` nor `bar`. These two cases are included in `object_test.go`.

When both `treeSelect` and `select` present, HNC will only propagate to
the namespace that matches both of the two selectors.

Tested: make test
The test cases covers the following situations:
- select/treeSelect to one single namespace AND multiple namespaces
- when select AND treeSelect both exists
- when select/treeSelect is updated

I believe these cover all of the selector test cases

### DIFF
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -658,6 +659,7 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 // - Service Account token secrets
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
 	selector := r.getSelector(log, inst)
+	treeSelector := r.getTreeSelector(log, inst)
 	nsLabels := r.Forest.Get(dst).GetLabels()
 
 	switch {
@@ -672,6 +674,10 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 
 	// Selector does not match
 	case selector != nil && !selector.Matches(nsLabels):
+		return false
+
+	// treeSelector does not match
+	case treeSelector != nil && !treeSelector.Matches(nsLabels):
 		return false
 
 	case r.GVK.Group == "" && r.GVK.Kind == "Secret":
@@ -693,6 +699,29 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 	}
 }
 
+// getTreeSelector is similar to a regular selector, except that it adds the LabelTreeDepthSuffix to every string
+// To transform a tree selector into a regular label selector, we follow these steps:
+// 1. get the treeSelector annotation if it exists
+// 2. convert the annotation string to a slice of strings seperated by comma, because user is allowed to put multiple selectors
+// 3. append the LabelTreeDepthSuffix to each of the treeSelector string
+// 4. combine them into a single string connected by comma
+func (r *ObjectReconciler) getTreeSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	annot := inst.GetAnnotations()
+	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
+	if !ok {
+		return nil
+	}
+	strs := strings.Split(treeSelectorStr, ",")
+	selectorStr := ""
+	for i, str := range strs {
+		selectorStr = selectorStr + str + api.LabelTreeDepthSuffix
+		if i < len(strs)-1 {
+			selectorStr = selectorStr + ", "
+		}
+	}
+	return r.getSelectorFromString(log, selectorStr)
+}
+
 // getSelector returns the selector on a given object if it exists
 func (r *ObjectReconciler) getSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
 	annot := inst.GetAnnotations()
@@ -700,7 +729,12 @@ func (r *ObjectReconciler) getSelector(log logr.Logger, inst *unstructured.Unstr
 	if !ok {
 		return nil
 	}
-	labelSelector, err := v1.ParseToLabelSelector(selectorStr)
+	return r.getSelectorFromString(log, selectorStr)
+}
+
+// getSelectorFromString converts the given string to a selector
+func (r *ObjectReconciler) getSelectorFromString(log logr.Logger, str string) labels.Selector {
+	labelSelector, err := v1.ParseToLabelSelector(str)
 	// TODO: surface the error messages here (https://github.com/kubernetes-sigs/multi-tenancy/issues/1165)
 	if err != nil {
 		log.Error(err, "Could not parse selector annotation to labelSelector")
--- a/incubator/hnc/internal/reconcilers/object_test.go
+++ b/incubator/hnc/internal/reconcilers/object_test.go
@@ -15,6 +15,178 @@ import (
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
 
+var _ = Describe("Exceptions", func() {
+	ctx := context.Background()
+
+	BeforeEach(func() {
+		// We want to ensure we're working with a clean slate, in case a previous tests objects still exist
+		cleanupObjects(ctx)
+	})
+
+	AfterEach(func() {
+		// Change current config back to the default value.
+		resetHNCConfigToDefault(ctx)
+		cleanupObjects(ctx)
+	})
+
+	Context("Add exception annotations", func() {
+		const (
+			p  = "parent"
+			c1 = "child1"
+			c2 = "child2"
+			c3 = "child3"
+		)
+		tests := []struct {
+			name         string
+			selector     string
+			treeSelector string
+			want         []string
+			notWant      []string
+		}{{
+			name:     "not propagate object to a negatively selected namespace",
+			selector: "!" + c1 + api.LabelTreeDepthSuffix,
+			want:     []string{c2, c3},
+			notWant:  []string{c1},
+		}, {
+			name:     "not propagate object to multiple negatively selected namespaces",
+			selector: "!" + c1 + api.LabelTreeDepthSuffix + ", !" + c2 + api.LabelTreeDepthSuffix,
+			want:     []string{c3},
+			notWant:  []string{c1, c2},
+		}, {
+			name:         "not propagate object to a negatively treeSelected namespace",
+			treeSelector: "!" + c1,
+			want:         []string{c2, c3},
+			notWant:      []string{c1},
+		}, {
+			name:         "not propagate object to multiple negatively treeSelected namespaces",
+			treeSelector: "!" + c1 + ", !" + c2,
+			want:         []string{c3},
+			notWant:      []string{c1, c2},
+		}, {
+			name:         "not propagate object to neither negatively selected or treeSelected namespaces",
+			selector:     "!" + c1 + api.LabelTreeDepthSuffix,
+			treeSelector: "!" + c2,
+			want:         []string{c3},
+			notWant:      []string{c1, c2},
+		}, {
+			name:         "only propagate object to the intersection of selected and treeSelected namespaces",
+			selector:     c1 + api.LabelTreeDepthSuffix,
+			treeSelector: c2,
+			want:         []string{},
+			notWant:      []string{c1, c2, c3},
+		}}
+
+		for _, tc := range tests {
+			// Making a local copy of tc is necessary to ensure the correct value is passed to the closure,
+			// for more details look at https://onsi.github.io/ginkgo/ and search for 'closure'
+			tc := tc
+			It("Should "+tc.name, func() {
+				// Set up namespaces
+				names := map[string]string{
+					p:  createNS(ctx, p),
+					c1: createNS(ctx, c1),
+					c2: createNS(ctx, c2),
+					c3: createNS(ctx, c3),
+				}
+				setParent(ctx, names[c1], names[p])
+				setParent(ctx, names[c2], names[p])
+				setParent(ctx, names[c3], names[p])
+
+				tc.selector = replaceStrings(tc.selector, names)
+				tc.treeSelector = replaceStrings(tc.treeSelector, names)
+
+				// Create a Role with the selector and treeSelector annotation
+				makeObjectWithAnnotation(ctx, "Role", names[p], "testrole", map[string]string{
+					"propagate.hnc.x-k8s.io/select":     tc.selector,
+					"propagate.hnc.x-k8s.io/treeSelect": tc.treeSelector,
+				})
+				for _, ns := range tc.want {
+					ns = replaceStrings(ns, names)
+					Eventually(hasObject(ctx, "Role", ns, "testrole")).Should(BeTrue(), "When propagating testrole to %s", ns)
+				}
+				for _, ns := range tc.notWant {
+					ns = replaceStrings(ns, names)
+					Consistently(hasObject(ctx, "Role", ns, "testrole")).Should(BeFalse(), "When propagating testrole to %s", ns)
+				}
+			})
+		}
+	})
+
+	Context("Update exception annotations", func() {
+		const (
+			p  = "parent"
+			c1 = "child1"
+			c2 = "child2"
+			c3 = "child3"
+		)
+		tests := []struct {
+			name         string
+			selector     string
+			treeSelector string
+			want         []string
+			notWant      []string
+		}{{
+			name:     "update select",
+			selector: "!" + c1 + api.LabelTreeDepthSuffix,
+			want:     []string{c2, c3},
+			notWant:  []string{c1},
+		}, {
+			name:         "update treeSelect",
+			treeSelector: "!" + c1,
+			want:         []string{c2, c3},
+			notWant:      []string{c1},
+		}}
+
+		for _, tc := range tests {
+			// Making a local copy of tc is necessary to ensure the correct value is passed to the closure,
+			// for more details look at https://onsi.github.io/ginkgo/ and search for 'closure'
+			tc := tc
+			It("Should "+tc.name, func() {
+				// Set up namespaces
+				names := map[string]string{
+					p:  createNS(ctx, p),
+					c1: createNS(ctx, c1),
+					c2: createNS(ctx, c2),
+					c3: createNS(ctx, c3),
+				}
+				setParent(ctx, names[c1], names[p])
+				setParent(ctx, names[c2], names[p])
+				setParent(ctx, names[c3], names[p])
+				tc.selector = replaceStrings(tc.selector, names)
+				tc.treeSelector = replaceStrings(tc.treeSelector, names)
+
+				// Create a Role and verify it's propagated
+				makeObject(ctx, "Role", names[p], "testrole")
+				for _, ns := range names {
+					Eventually(hasObject(ctx, "Role", ns, "testrole")).Should(BeTrue(), "When propagating testrole to %s", ns)
+				}
+
+				// update the role with the selector and treeSelector annotation
+				updateObjectWithAnnotation(ctx, "Role", names[p], "testrole", map[string]string{
+					"propagate.hnc.x-k8s.io/select":     tc.selector,
+					"propagate.hnc.x-k8s.io/treeSelect": tc.treeSelector,
+				})
+				// make sure the changes are propagated
+				for _, ns := range tc.notWant {
+					ns = replaceStrings(ns, names)
+					Eventually(hasObject(ctx, "Role", ns, "testrole")).Should(BeFalse(), "When propagating testrole to %s", ns)
+				}
+				// then check that the objects are kept in these namespaces
+				for _, ns := range tc.want {
+					ns = replaceStrings(ns, names)
+					Consistently(hasObject(ctx, "Role", ns, "testrole")).Should(BeTrue(), "When propagating testrole to %s", ns)
+				}
+
+				// remove the annotation and verify that the object is back for every namespace
+				updateObjectWithAnnotation(ctx, "Role", names[p], "testrole", map[string]string{})
+				for _, ns := range names {
+					Eventually(hasObject(ctx, "Role", ns, "testrole")).Should(BeTrue(), "When propagating testrole to %s", ns)
+				}
+			})
+		}
+	})
+})
+
 var _ = Describe("Secret", func() {
 	ctx := context.Background()
 
@@ -43,48 +215,7 @@ var _ = Describe("Secret", func() {
 		cleanupObjects(ctx)
 	})
 
-	// Exceptions have not been implemented, so we are ignoring the following four tests for now.
-	// They should be enabled once exception is done.
-	PIt("should propagate object only to selected namespace using treeSelect", func() {
-		setParent(ctx, barName, fooName)
-		setParent(ctx, bazName, fooName)
-
-		// Create a Role that does NOT propogate to bazName
-		a := map[string]string{"propagate.hnc.x-k8s.io/treeSelect": "!" + bazName}
-		makeObjectWithAnnotation(ctx, "Role", fooName, "testrole", a)
-
-		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
-		Consistently(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
-	})
-
-	It("should propagate object only to selected namespace using select", func() {
-		setParent(ctx, barName, fooName)
-		setParent(ctx, bazName, fooName)
-
-		// Create a Role that does NOT propogate to bazName
-		a := map[string]string{"propagate.hnc.x-k8s.io/select": "!" + bazName + ".tree.hnc.x-k8s.io/depth"}
-		makeObjectWithAnnotation(ctx, "Role", fooName, "testrole", a)
-
-		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
-		Consistently(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
-	})
-
-	It("should remove object in selected namespace when selector is updated", func() {
-		setParent(ctx, barName, fooName)
-		setParent(ctx, bazName, fooName)
-
-		makeObject(ctx, "Role", fooName, "testrole")
-		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
-		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeTrue())
-
-		// Update the Role so that it does NOT propogate to bazName
-		a := map[string]string{"propagate.hnc.x-k8s.io/select": "!" + bazName + ".tree.hnc.x-k8s.io/depth"}
-		Expect(updateObjectWithAnnotation(ctx, "Role", fooName, "testrole", a)).Should(Succeed())
-
-		Consistently(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
-		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
-	})
-
+	// This test should be converted to a table test after 'none' selector is implemented
 	PIt("should not propagate object to any namespace using none", func() {
 		setParent(ctx, barName, fooName)
 		setParent(ctx, bazName, fooName)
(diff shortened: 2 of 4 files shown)


adrianludwin:
> Tested: make test

Maybe say why you think this is sufficient testing

adrianludwin:
... `make test` was technically all I did for #1174 but the entire commit message is about the test!

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -752,6 +752,9 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 // - Service Account token secrets
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
 	selector := r.getSelector(log, inst)
+	if selector == nil {
+		selector = r.getTreeSelector(log, inst)
+	}
```
adrianludwin, line 757: Was this the behaviour we defined in the design doc? Ignore the tree selector if there's a non-tree selector? If not, do you think we should change what's in the doc, and if so, why?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -787,14 +790,29 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
 }
 
+// getSelector returns the selector on a given object if it exists
+func (r *ObjectReconciler) getTreeSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	annot := inst.GetAnnotations()
+	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
+	if !ok {
+		return nil
+	}
+	return r.getSelectorFromString(log, treeSelectorStr+api.LabelTreeDepthSuffix)
```
adrianludwin, line 800: This doesn't look right to me - we specified that there could be multiple comma-separated trees in the tree-selector, but this doesn't account for that, does it?

This probably shows that we need tests (both tree and regular) showing more complex selectors. Can you please add a more complex test for the tree selectors in this PR?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -59,6 +59,22 @@ var _ = Describe("Secret", func() {
 		Consistently(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
 	})
 
+	It("should remove object in selected namespace when treeSelector is updated", func() {
+		setParent(ctx, barName, fooName)
+		setParent(ctx, bazName, fooName)
+
+		makeObject(ctx, "Role", fooName, "testrole")
+		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
+		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeTrue())
```
adrianludwin, line 68: I've just discovered (in #1174) that you can comment the "Should" clause, e.g.:

`Eventually(hasObject(ctx, "Role", barName, "testrole)).Should(BeTrue(), "When propagating testrole to %s", barName)`

Maybe start annotating all your `Should` clauses like this so the errors are more descriptive?

adrianludwin (commented):
New test looks good! But I don't think this implements the design doc for treeselect. Perhaps some of that could be changed but we should be explicit about it.

incubator/hnc/internal/reconcilers/object.go
GinnyJI: After reading the design doc again, I think we should 'add' the two selectors together if they both exists. I'll update the logic. However, when the `none` selector exists at the same time with either `select` or `treeSelect` or both of them, do we prioritize the `none` selector?

adrianludwin: What do you mean by "add" - an AND operation or an OR operation?

If "adding" is hard we could also consider changing the doc. It probably doesn't make a huge difference as most people should only really be using one or the other.

incubator/hnc/internal/reconcilers/object.go
yiqigao217, line 800: Isn't `LabelTreeDepthSuffix` already included in the `treeSelectorStr`?

incubator/hnc/internal/reconcilers/object.go
GinnyJI: OR is easier in this case, do you have a preference if `treeSelector` or `selector` should be prioritized? For me they look equally important

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: No, the `treeSelectorStr` is a short-handed representation of `selectorStr`, which doesn't include the `LabelTreeDepthSuffix`, and we need to add that in

yiqigao217: Ah right, sorry for the confusion. Selector is the one with DepthSurfix and treeSelector is the one only with namespace names.

incubator/hnc/internal/reconcilers/object.go
adrianludwin: When I meant "OR," I mean "a namespace matches if it's in either selector." And "AND" means "matches if it's matched by _both_ selectors." But I think you're describing something different, which is to only look at one or the other. Is that what you meant?

If they're equally important, then maybe it's worth a quick investigation as to how you'd combine them using either an AND or OR operation. I prefer AND, I think, because that has very consistent behaviour with the `none` selector - if `none` is present, nothing ever gets propagated.

yiqigao217: The current test cases lgtm. BTW after this is addressed, a new test case with both `select` and `treeselect` will be added too, right?

GinnyJI: Re @adrianludwin: yes I was thinking using one of the selector annotations. Looking into a way to combine both of the selectors.

Re @yiqigao217: Good point, I'll add a test case for this

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -787,14 +791,37 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 	}
 }
 
+// getSelector returns the selector on a given object if it exists
```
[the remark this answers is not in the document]

incubator/hnc/internal/reconcilers/object.go

adrianludwin:
> > Tested: make test
> 
> Maybe say why you think this is sufficient testing

Ping!

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -787,14 +793,37 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
 }
 
+// getTreeSelector returns the selector on a given object if it exists
+func (r *ObjectReconciler) getTreeSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	annot := inst.GetAnnotations()
+	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
+	if !ok {
+		return nil
+	}
+	strs := strings.Split(treeSelectorStr, ",")
```
adrianludwin, line 803: Comment why we're doing this. It's fine to refer to the design doc, but it might also make sense to document this here as well.

The comment might make more sense at the function level (e.g. at line 796). Eg "A tree selectors is similar to a regular selector, except for the following things... To transform a tree selector into a regular label selector, we follow these steps..."

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -45,18 +45,46 @@ var _ = Describe("Secret", func() {
...
-		Consistently(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
+		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue(), "When propagating testrole to %s", barName)
+		Consistently(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse(), "When propagating testrole to %s", bazName)
+	})
+
+	It("should propagate object only to selected namespace using multiple treeSelect", func() {
+		setParent(ctx, barName, fooName)
+		setParent(ctx, bazName, fooName)
+
+		// Create a Role that does NOT propogate to bazName
```
adrianludwin, line 66: This comment is out of date. Also, the test might be stronger if `foo` had _three_ children, two of which are excluded but one of which is not.

Finally, all your tests use negation. It would be nice to have a test that _only_ propagated to 2/3 named namespaces. E.g., let's say you had three children (bar, baz, bam). The selectors might be:
* bar (should only propagate to bar)
* !bar (should only propagate to baz/bam)
* bar,baz (should only propagate to bar/baz)
* !bar,!baz (should only propagate to bam)

This looks a lot like a table-driven test - e.g. https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/validators/hierarchy_test.go#L18. Each test case is a single line and not an entire function. E.g. what I just wrote above could look like:

```
tests := []struct{
  selector string
  want []string
}{
  {selector: barName, want: []string{barName}},
  {selector: "!"+barName, want: []string{bazName, bamName}},
  ...
}
```

And then you write the test _once_ that sets up the object and then verifies its result. Does this make sense?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -45,18 +45,46 @@ var _ = Describe("Secret", func() {
...
+		makeObject(ctx, "Role", fooName, "testrole")
+		Eventually(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue(), "When propagating testrole to %s", barName)
+		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeTrue(), "When propagating testrole to %s", bazName)
+
+		// Update the Role so that it does NOT propogate to bazName
+		a := map[string]string{"propagate.hnc.x-k8s.io/treeSelect": "!" + bazName}
+		Expect(updateObjectWithAnnotation(ctx, "Role", fooName, "testrole", a)).Should(Succeed())
+
+		Consistently(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue(), "When propagating testrole to %s", barName)
+		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse(), "When propagating testrole to %s", bazName)
```
adrianludwin, line 87: I'd swap the order of these two (here and in similar tests). Line 87 requires the reconciler to actually do something; line 86 is hoping that the reconciler _doesn't_ do something. So waiting for Eventually to pass gives us a bit more confidence that if the reconciler was _going_ to delete the propagated object, it would have already.

Not a huge deal but probably something worth doing (and commenting).

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -67,24 +95,50 @@ var _ = Describe("Secret", func() {
...
 
-		Consistently(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue())
-		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse())
+		Consistently(hasObject(ctx, "Role", barName, "testrole")).Should(BeTrue(), "When propagating testrole to %s", barName)
+		Eventually(hasObject(ctx, "Role", bazName, "testrole")).Should(BeFalse(), "When propagating testrole to %s", bazName)
+	})
+
+	It("should propagate object only to selected namespace using both select and treeSelect when both are present", func() {
+		setParent(ctx, barName, fooName)
+		setParent(ctx, bazName, fooName)
```
adrianludwin, line 133: This would be a stronger test IMO again if there were three children. Use select to propagate to bar and baz; use treeSelect to propagate to baz and bam. Verify that it's only propagated to baz.

You might be able to set up a table test (as described above) that covers _all_ these cases. E.g.:

```
exceptionTests := []struct{
  name string
  select string
  treeSelect string
  want []string
}{
  {name: "simple treeSelect", treeSelect: barName, want: []string{barName}},
  {name: "negated treeSelect", treeSelect: "!"+barName, want: []string{bazName, bamName}},
  ... other simple one-line cases ...
  {
     name: "only propagate to namespaces selected by both select and treeSelect",
     select: barName+treeSuffix+","+bazName+treeSuffix,
     treeSelect: bazName+","+bamName,
     want: []string{bazName}
  },
  ... other complex cases ...
}
for _, tc := range exceptionTests {
  It("should handle exceptions: "+tc.name, func() {
    ... logic goes here ...
  }
}
```

adrianludwin (commented):
I made a suggestion to try to combine all these tests into one big test - maybe look through all the comments before getting started. This is a big change but should make the tests much easier to read and maintain IMO.

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: I'm not sure if I understood the structure of this test - does this function still goes under `object_test.go`? Do I also need to set up the test and hierarchy like in the example? I.e.```f := foresttest.Create("-a-") // a <- b; c
h := &Hierarchy{Forest: f}
l := zap.Logger(false)
```

Comment (adrianludwin) [reply incubator/hnc/internal/reconcilers/object_test.go:66]: No, those three setup lines aren't needed if you're in the reconciler integ tests. And yes, you should keep the tests in the file where they are now, just structure them differently.

You can think of it like this: instead of writing each test individually, you're writing a function to create the tests for you. See the other comment I made for how I _think_ it should work, but TBH I've never written a table test in Ginkgo before - what I wrote these is my best guess.

I'd recommend starting small. See if you can write a dumb table test with two subcases that don't do anything but make sure that they run, e.g.:

```
for i:=0; i <=1; i++ {
  FIt("should verify that i == 0", func() {
    Expect(i == 0)
  })
}
```

That should give you one passing and one failing test. See if that works! And if so, expand from there.

**Comment by GinnyJI** [comment]
Work in progress
/hold

File: incubator/hnc/internal/reconcilers/object.go (hunk 5 above)
Comment (GinnyJI) [reply incubator/hnc/internal/reconcilers/object.go:808]: done

File: incubator/hnc/internal/reconcilers/object_test.go [hunk 9]
```diff
@@ -14,19 +15,120 @@ import (
...
+			{name: "propagate object only to selected namespace using treeSelect",
+				selector:     "!" + c1 + api.LabelTreeDepthSuffix,
+				treeSelector: "",
+				want:         []string{c2, c3},
+				notWant:      []string{c1},
+			},
+		}
+
+		for _, tc := range tests {
+			FIt("Should handle exceptions: "+tc.name, func() {
```
yiqigao217, line 67: Remove the FIt?

yiqigao217 (commented):
A general question: Since `getTreeSelector()` and `getSelector()` get the same thing with `depthSuffix`, can we have them combined and do a `.Matches()`?

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: It was a draft, removed now

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

GinnyJI:
> > Tested: make test
> 
> Maybe say why you think this is sufficient testing

Done, please let me know if I can make it more clear

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

GinnyJI:
This PR is ready for review
/hold cancel

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+				}
+				for _, ns := range tc.notWant {
+					ns = replaceStrings(ns, indToNm)
+					Consistently(hasObject(ctx, "Role", ns, "testrole")).Should(BeFalse(), "When propagating testrole to %s", ns)
+				}
+			})
+		}
+	})
+
+	Context("Update exception annotations ", func() {
```
yiqigao217, line 130: What's the difference between the update cases with the add test cases?

adrianludwin (commented):
Can you please rewrap the commit message? Nice work on the testing documentation. It doesn't have to be quite that detailed, you _could_ just say something like "added tests for `treeSelect` and to test interaction with `select`." Up to you though.

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+		tests := []struct {
+			name         string
+			selector     string
+			treeSelector string
+			want         []string
+			notWant      []string
+		}{
+			{name: "simply using select",
+				selector:     "!" + c1 + api.LabelTreeDepthSuffix,
+				treeSelector: "",
```
adrianludwin, line 48: You can omit anything with the zero value (for a string, this is `""`)

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+				want:         []string{c2, c3},
+				notWant:      []string{c1},
```
adrianludwin, line 50: Maybe calculate `notWant` from `want`?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+		for _, tc := range tests {
+			// Making a local copy of tc is necessary to ensure the correct value is passed to the closure,
+			// for more details look at https://onsi.github.io/ginkgo/ and search for 'closure'
+			tc := tc
+			It("Should handle exceptions: "+tc.name, func() {
+				// Set up namespaces
+				fooName := createNS(ctx, "foo")
+				barName := createNS(ctx, "bar")
+				bazName := createNS(ctx, "baz")
+				quxName := createNS(ctx, "qux")
```
adrianludwin, line 93: Suggestion: line these up with the constants used to define the test, i.e. `child1`, etc.

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+			c3 = "child3"
+		)
+		tests := []struct {
+			name         string
+			selector     string
+			treeSelector string
+			want         []string
+			notWant      []string
+		}{
+			{name: "simply using select",
```
adrianludwin, line 46: Maybe pick the names so that they would flow naturally after the word "Should", so the It block could be `It("Should "+tc.name)`, e.g. "It should not appear in a negated namespace".

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+				if tc.treeSelector != "" {
+					a["propagate.hnc.x-k8s.io/treeSelect"] = tc.treeSelector
+				}
+				updateObjectWithAnnotation(ctx, "Role", fooName, "testrole", a)
+				// make sure the changes are propagated
+				for _, ns := range tc.notWant {
+					ns = replaceStrings(ns, indToNm)
+					Eventually(hasObject(ctx, "Role", ns, "testrole")).Should(BeFalse(), "When propagating testrole to %s", ns)
+				}
+				// then check that the objects are still reserved in these namespaces
```
adrianludwin, line 203: "reserved"?

Also it might be nice to try _removing_ the annotation and verify that the objects are re-propagated everywhere

incubator/hnc/internal/reconcilers/object_test.go
adrianludwin: Another suggestion:
```
names := map[string]string{
  p:  createNS(ctx, p),
  c1: createNS(ctx, c1),
  c2: createNS(ctx, c2),
  c3: createNS(ctx, c3),
}

setParent(ctx, names[c1], names[p])
setParent(ctx, names[c2], names[p])
setParent(ctx, names[c2], names[p])
tc.selector = replaceStrings(tc.selector, names)
tc.treeSelector = replaceStrings(tc.treeSelector, names)
```

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+				tc.treeSelector = replaceStrings(tc.treeSelector, indToNm)
+
+				// Create a Role with the selector and treeSelector annotation
+				a := make(map[string]string)
+				if tc.selector != "" {
+					a["propagate.hnc.x-k8s.io/select"] = tc.selector
+				}
+				if tc.treeSelector != "" {
+					a["propagate.hnc.x-k8s.io/treeSelect"] = tc.treeSelector
+				}
```
adrianludwin, line 116: What would happen if we made a selector with an empty string? Would that be "match everything" or "match nothing?" Do we define that properly in the design doc? If we treat an empty string as "missing" (i.e. match everything) then we could say:
```
makeObjectWithAnnotation(ctx, "Role", fooName, "testrole", map[string]string{
  "propagate.hnc.x-k8s.io/select": tc.selector,
  "propagate.hnc.x-k8s.io/treeSelect": tc.treeSelector,
})
```

which seems kind of nice.

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: The update case checks that a ns had `testrole` but got removed after we update `testrole` with selectors, while the add case never updates the object, it was created with the selector so we only check each ns once

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
 
+var _ = Describe("Exception", func() {
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+		cleanupObjects(ctx)
+	})
+
+	AfterEach(func() {
+		// Change current config back to the default value.
+		resetHNCConfigToDefault(ctx)
+		cleanupObjects(ctx)
+	})
+
+	Context("Add exception annotations ", func() {
```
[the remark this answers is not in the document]
GinnyJI: Ah the space should't be there

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: I did a brief search and seems like the only way is to iterate through the whole set (all three nses) for each element in the `want`. That looks a bit overkill for me (?) So I thought since there are only three elements we might just keep both sets

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,201 @@ import (
...
+			want         []string
+			notWant      []string
+		}{
+			{name: "simply using select",
+				selector:     "!" + c1 + api.LabelTreeDepthSuffix,
+				treeSelector: "",
+				want:         []string{c2, c3},
+				notWant:      []string{c1},
+			},
+			{name: "using multiple select",
```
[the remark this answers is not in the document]
GinnyJI: Ah, `gofmt` did this for me and I can't stop it doing this. But seems like it likes the second format :D

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: I tested it and seems like when converting string to selector, we do treat empty string as missing. Updated

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: Good to know! Would you prefer `0` over `""`?

incubator/hnc/internal/reconcilers/object_test.go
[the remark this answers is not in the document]
adrianludwin: Weird. Ok, I guess I was wrong about "standard format!"

GinnyJI:
> Can you please rewrap the commit message? Nice work on the testing documentation. It doesn't have to be quite that detailed, you _could_ just say something like "added tests for `treeSelect` and to test interaction with `select`." Up to you though.

Updated

incubator/hnc/internal/reconcilers/object_test.go
adrianludwin: No, I just mean: take out this line completely, and Go will auto-set `treeSelector` to `""`.

The term "zero value" just means "default value" in Go. For an `int`, the zero value is `0`; for a string, it's `""`, for a pointer it's `nil`, etc.

incubator/hnc/internal/reconcilers/object_test.go
adrianludwin:
```
for _, ns := range []string{c1, c2, c3} {
  if isIn(ns, tc.want) {
    ... want case ...
  } else {
    .. not-want case ...
  }
}
```

?

adrianludwin: But you can do that in a future PR, it doesn't have to be in this one

incubator/hnc/internal/reconcilers/object_test.go
adrianludwin: sgtm and can you please update the design doc to state that this is the intended behaviour?

adrianludwin:
> > Can you please rewrap the commit message? Nice work on the testing documentation. It doesn't have to be quite that detailed, you _could_ just say something like "added tests for `treeSelect` and to test interaction with `select`." Up to you though.
> 
> Updated

The commit message still doesn't look properly word-wrapped? I.e. some lines are much longer than others.

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,184 @@ import (
...
+		}}
+
+		for _, tc := range tests {
+			// Making a local copy of tc is necessary to ensure the correct value is passed to the closure,
+			// for more details look at https://onsi.github.io/ginkgo/ and search for 'closure'
+			tc := tc
+			It("Should handle exceptions: "+tc.name, func() {
+				// Set up namespaces
+				names := map[string]string{
+					p:  createNS(ctx, "parent"),
```
adrianludwin, line 153: Should be: `p: createNS(ctx, p)` since `p` already has the right value (also on line 90)

adrianludwin (commented):
lgtm after final changes are made. Nice!

/approve
/assign @rjbez17

GinnyJI:
> > > Can you please rewrap the commit message? Nice work on the testing documentation. It doesn't have to be quite that detailed, you _could_ just say something like "added tests for `treeSelect` and to test interaction with `select`." Up to you though.
> > 
> > 
> > Updated
> 
> The commit message still doesn't look properly word-wrapped? I.e. some lines are much longer than others.

Oh the line length was decided by vim, I actually don't know why some lines are longer. If I do a manually `enter` to separate the longer lines, vim shows me that other lines don't have a `return` symbol but just those longer lines. Do you know how to format line length in vim?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -15,6 +15,184 @@ import (
...
+			treeSelector: c2,
+			want:         []string{},
+			notWant:      []string{c1, c2, c3},
+		}}
+
+		for _, tc := range tests {
+			// Making a local copy of tc is necessary to ensure the correct value is passed to the closure,
+			// for more details look at https://onsi.github.io/ginkgo/ and search for 'closure'
+			tc := tc
+			It("Should handle exceptions: "+tc.name, func() {
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: I see! Done

adrianludwin:
> Oh the line length was decided by vim, I actually don't know why some lines are longer. If I do a manually `enter` to separate the longer lines, vim shows me that other lines don't have a `return` symbol but just those longer lines. Do you know how to format line length in vim?

Highlight all the relevant lines (eg using the `V` command) and then hit `gq`.

adrianludwin (commented):
/lgtm
/hold
/assign @rjbez17 

Nice work!

PR #1179, opened 2020-10-08
Improve HNC Config test diagnosis/stability

### LINKED ISSUE #1170: HNC: Presubmit tests are flaky
Logs at https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_multi-tenancy/1161/pull-hnc-test/1313137143345844224

adrianludwin:
The HNC Config tests were flaky when run in a loaded environment, such
as on Prow, and produced hard-to-reproduce errors. I was able to
reproduce them about 20% of the time using the `stress` utility to load
up my workstation.

The concrete proble I found was that the tests were not always waiting
for changes to the HNC config to be picked up - see my comments in #1170
for details. I've addressed this by adding a setTypeConfig function that
always waits for config changes to appear in the status before the test
can proceed.

I also added or fixed ...WithOffset() Gomega assertions now that I
actually understand how they work, and focused on never having any
_nested_ assertions since they make it impossible to get the correct
originating line number. I also changed some functions' names to better
match the way they're used (e.g. don't use "has" in the name if it
doesn't return a bool).

Tested: on a loaded workstation, the HNC config tests now pass 10/10
times. While I was working on the tests and had a bunch of problems, all
the line numbers were correct and useful.

FTR, on my 12-core workstation, the command I used to load it up was
`stress -m 22 -i 24 -c 24`. Anything higher and the apiserver wouldn't
start; anything lower and it didn't seem to have enough effect. This was
enough to increase the runtime of the tests from ~30s to ~50s so it
seemed to be doing a reasonable job.

Fixes #1170

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config_test.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config_test.go
@@ -2,12 +2,10 @@ package reconcilers_test
 
 import (
 	"context"
-	"fmt"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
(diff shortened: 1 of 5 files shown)


adrianludwin:
/assign @yiqigao217 
/assign @rjbez17 
/cc @GinnyJI 

This only needs one approver, I think.

yiqigao217 (commented):
/lgtm
/hold

Just holding because this is a big PR and I feel safer to have another eye. Feel free to remove the hold if it's not necessary.

adrianludwin:
/cc @rjbez17 

I'm guessing you're fine with this (it's purely a test change) but as always lmk if you want any changes and we can make them in a followup PR.

PR #1183, opened 2020-10-09
Add a cascadingDelete case for subnamespace e2e test

yiqigao217:
Seems the cascadingDeletion test doesn't have a case for itself.

Tested manully and by focused test.


yiqigao217:
This is covered in `Should cascading delete all subnamespaces if the parent is deleted and allows cascadingDeletion - issue #501` already

Issue #1118, opened 2020-09-18
[VC] proposal: kubectl vc command line tool


zhuangqh:
## Motivation

super cluster admin managed all of the virtual clusters. They usually switch between different vc. First find the kubeconfig, then switch to this cluster. We need a command line tool to manage multi clusters.

### kubectl vc create xxx

create a virtual cluster and waiting it to be running. export kubeconfig to specific path.

### kubectl vc delete xxx

delete a virtual cluster 

### kubectl vc config -n $vc_cr_ns $vc_cr_name

switch to a vc kubeconfig 

### kubectl vc -n $vc_cr_ns $vc_cr_name xxx

switch to a vc context and do normal kubectl command

zhuangqh:
@Fei-Guo @charleszheng44

Fei-Guo:
This is going to be very useful!

A few comments:

`kubectl vc config -n $vc_cr_ns $vc_cr_name` will be optional if we have `kubectl vc -n $vc_cr_ns $vc_cr_name xxx` been available.

We can combine "-n $vc_cr_ns $vc_cr_name" to a single "vcfullname" as input so the cmd can be as simple as

`Kubectl vc admin1/vc-sample1 get pod -A`

charleszheng44:
+1 for the vc subcommand :+1:

salaxander:
This was something I have been thinking about too! I had originally considered extending `vcctl`, but I like the idea of a `kubectl` plugin!

PR #1188, opened 2020-10-12
VC: migrate from vcctl to kubectl-vc

zhuangqh:
- kubectl vc: implement create command
- remove vcctl and update document
- remove command `delete`, because it can be done by straight kubectl

link #1118 

@charleszheng44 @salaxander


incubator/virtualcluster/cmd/kubectl-vc/root.go
```diff
@@ -0,0 +1,66 @@
...
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/urfave/cli"
```
christopherhein, line 23: Any reason we went with `urfave`? Isn’t the rest of vc using cobra?

zhuangqh: i think urfave is much easier to construct a CLI with multi subcommands 😂
kubectl vc is a very simple tool with clean implementation

christopherhein: Gotcha, as we dev this we might want to strive for some consistency across the components so that devs don't need to relearn and we can reuse the same packages.

christopherhein: I'm referring to things like having a shared flags implementation cause all of the components use different flags eg `master-kubeconfig` `kubeconfig`, while not in this cli yet I'm sure we're going to hit this eventually.

PR #1189, opened 2020-10-12
VC:  kubectl vc exec: switch workspace among virtualclusters

zhuangqh:
```bash
[root@vm virtualcluster]# kubectl vc
VirtualCluster Command tool

Usage:
  kubectl-vc [flags]
  kubectl-vc [command]

Available Commands:
  create      Create a new VirtualCluster
  exec        Switch to virtualcluster workspace
  help        Help about any command

Flags:
  -h, --help      help for kubectl-vc
      --version   version for kubectl-vc

Use "kubectl-vc [command] --help" for more information about a command.
[root@vm virtualcluster]# kubectl vc exec -h
Switch to virtualcluster workspace

Usage:
  kubectl-vc exec VC_NAME [flags]

Examples:

	# Switch to a virtualcluster
	kubectl vc exec -n foo bar

	# Specific vc by namespaced name
	kubectl vc exec foo/bar

	# Customize kubeconfig file path
	kubectl vc exec --kube-file-dir /path/to/file foo/bar

Flags:
  -h, --help                   help for exec
      --kube-file-dir string   The directory to place the kubeconfig of specific vc (default "/root/.kube/vc")
  -n, --namespace string       If present, the namespace scope for this CLI request (default "default")
[root@vm virtualcluster]# kubectl vc exec default/vc-sample-1
kubeconfig for virtualcluster default/vc-sample-1 is placed at:

	/root/.kube/vc/default-8c6873-vc-sample-1.kubeconfig

❗ You are now at VirtualCluster default/vc-sample-1
❗ use regular kubectl commands to operate vc in this temporary workspace
❗ type 'exit' to exit
[root@vc:default/vc-sample-1 virtualcluster]# k cluster-info
Kubernetes master is running at https://10.0.2.15:31510

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
[root@vc:default/vc-sample-1 virtualcluster]# exit
exit
❗ exit VirtualCluster default/vc-sample-1
[root@vm virtualcluster]#
```

In legacy terminal
```bash
[root@vm virtualcluster]# TERM=dumb kubectl vc exec vc-sample-1
kubeconfig for virtualcluster default/vc-sample-1 is placed at:

	/root/.kube/vc/default-b59386-vc-sample-1.kubeconfig

!! You are now at VirtualCluster default/vc-sample-1
!! use regular kubectl commands to operate vc in this temporary workspace
!! type 'exit' to exit
[root@vc:default/vc-sample-1 virtualcluster]# exit
exit
!! exit VirtualCluster default/vc-sample-1
```

-  refactor to cobra
-  kubectl vc exec: switch workspace among virtualclusters

@salaxander @charleszheng44

### DIFF
--- a/incubator/virtualcluster/cmd/kubectl-vc/create.go
+++ b/incubator/virtualcluster/cmd/kubectl-vc/create.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"time"
 
 	"github.com/pkg/errors"
@@ -30,7 +31,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
@@ -131,11 +131,6 @@ func createVirtualCluster(cli client.Client, vccli vcclient.Interface, vc *tenan
 		return nil, errors.Wrapf(err, "required cluster version not found")
 	}
 
-	apiSvcPort, err := getAPISvcPort(cv.Spec.APIServer.Service)
-	if err != nil {
-		return nil, err
-	}
-
 	// fail early, if service type is not supported
 	svcType := cv.Spec.APIServer.Service.Spec.Type
 	if svcType != v1.ServiceTypeNodePort &&
@@ -155,23 +150,23 @@ func createVirtualCluster(cli client.Client, vccli vcclient.Interface, vc *tenan
 	}); err != nil {
 		return nil, fmt.Errorf("cannot find sts/etcd in ns %s: %s", ns, err)
 	}
-	klog.Info("etcd is ready")
+	log.Println("etcd is ready")
 
 	if err := retryIfNotFound(5, 2, func() error {
 		return kubeutil.WaitStatefulSetReady(cli, ns, "apiserver", pollStsTimeoutSec, pollStsPeriodSec)
 	}); err != nil {
 		return nil, fmt.Errorf("cannot find sts/apiserver in ns %s: %s", ns, err)
 	}
-	klog.Info("apiserver is ready")
+	log.Println("apiserver is ready")
 
 	if err := retryIfNotFound(5, 2, func() error {
 		return kubeutil.WaitStatefulSetReady(cli, ns, "controller-manager", pollStsTimeoutSec, pollStsPeriodSec)
 	}); err != nil {
 		return nil, fmt.Errorf("cannot find sts/controller-manager in ns %s: %s", ns, err)
 	}
-	klog.Info("controller-manager is ready")
+	log.Println("controller-manager is ready")
 
-	return genKubeConfig(ns, cli, svcType, apiSvcPort)
+	return genKubeConfig(cli, vc, cv)
 }
 
 // getAPISvcPort gets the apiserver service port if not specifed
@@ -225,7 +220,8 @@ func getVcKubeConfig(cli client.Client, clusterNamespace, srtName string) ([]byt
 }
 
 // genKubeConfig generates the kubeconfig file for accessing the virtual cluster
-func genKubeConfig(clusterNamespace string, cli client.Client, svcType v1.ServiceType, apiSvcPort int) ([]byte, error) {
+func genKubeConfig(cli client.Client, vc *tenancyv1alpha1.VirtualCluster, cv *tenancyv1alpha1.ClusterVersion) ([]byte, error) {
+	clusterNamespace := conversion.ToClusterKey(vc)
 	kbCfgBytes, err := getVcKubeConfig(cli, clusterNamespace, "admin-kubeconfig")
 	if err != nil {
 		return nil, err
@@ -236,8 +232,13 @@ func genKubeConfig(clusterNamespace string, cli client.Client, svcType v1.Servic
 		return nil, err
 	}
 
+	apiSvcPort, err := getAPISvcPort(cv.Spec.APIServer.Service)
+	if err != nil {
+		return nil, err
+	}
+
 	// replace the server address in kubeconfig based on service type
-	kubecfg, err = replaceServerAddr(kubecfg, cli, clusterNamespace, svcType, apiSvcPort)
+	kubecfg, err = replaceServerAddr(kubecfg, cli, clusterNamespace, cv.Spec.APIServer.Service.Spec.Type, apiSvcPort)
 	if err != nil {
 		return nil, err
 	}
--- a/incubator/virtualcluster/cmd/kubectl-vc/create.go
+++ b/incubator/virtualcluster/cmd/kubectl-vc/create.go
@@ -29,13 +29,15 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/yaml"
 
 	tenancyv1alpha1 "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/apis/tenancy/v1alpha1"
 	vcclient "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/client/clientset/versioned"
+	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/client/clientset/versioned/scheme"
 	kubeutil "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/controller/util/kube"
 	netutil "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/controller/util/net"
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/conversion"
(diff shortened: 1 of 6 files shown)


incubator/virtualcluster/cmd/kubectl-vc/use.go
```diff
@@ -0,0 +1,132 @@
...
+	err = ioutil.WriteFile(kbFilePath, kbBytes, 0644)
+
+	return kbFilePath, err
+}
+
+func enterVCShell(kbFilePath, ns, name string) error {
+	fmt.Printf("❗ You are now at VirtualCluster %s/%s\n", ns, name)
+	fmt.Println("❗ type 'exit' to exit this temporary workspace")
+
+	bashBinPath, err := exec.LookPath("bash")
```
charleszheng44, line 115: Can we use $SHELL instead of bash? somthing like, exec.LookPath(os.Getenv("SHELL"))

charleszheng44 (changes requested):

incubator/virtualcluster/cmd/kubectl-vc/use.go
zhuangqh: good idea

incubator/virtualcluster/cmd/kubectl-vc/use.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
```
christopherhein, line 27: I missed reviewing the previous pr before it was merged, why did we choose to add a new cli tool when syncer and vn-agent uses  `cobra` https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/virtualcluster/cmd/syncer/app/server.go#L27 & https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/virtualcluster/cmd/vn-agent/app/server.go#L27

incubator/virtualcluster/cmd/kubectl-vc/use.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		kbFilePath, err := placeVCKubeconfig(ns, name)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("kubeconfig for virtualcluster %s/%s is placed at:\n\n\t%s\n\n", ns, name, kbFilePath)
+
+		if cctx.Bool("interactive") {
+			return enterVCShell(kbFilePath, ns, name)
```
christopherhein, line 67: Should this be a separate command like `kubectl vc exec ....`?

Fei-Guo: I see one problem of having another command is that it is not very clear which kubeconfig is using at the moment of typing "kubectl vc exec" especially when you have a lot of tenants.

incubator/virtualcluster/cmd/kubectl-vc/use.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+const vcKubeconfigHome = ".kube/vc/"
+
+var useCommand = cli.Command{
+	Name:      "use",
+	Usage:     "Switch kubeconfig among virtualclusters",
+	ArgsUsage: "VC_NAME",
+	Flags: []cli.Flag{
+		cli.StringFlag{
```
Fei-Guo, line 40: If I understand correctly, this name is the vc cr ns name.   How about let user specify the fullname of the vc cr so that we can remove one input parameter and avoid mistakes. 

e.g., Kubectl vc use tenant1-admin/vc-sample1

incubator/virtualcluster/cmd/kubectl-vc/use.go
```diff
@@ -0,0 +1,128 @@
...
+			return errors.New("virtualcluster name is required")
+		}
+
+		kbFilePath, err := placeVCKubeconfig(ns, name)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("kubeconfig for virtualcluster %s/%s is placed at:\n\n\t%s\n\n", ns, name, kbFilePath)
+
+		if cctx.Bool("interactive") {
```
Fei-Guo, line 66: In my opinion, the -i command is mandatory. There is not much of use case people just want to generate the kubeconfig and don't want to operate on it. 

I would suggest to always call enterVCShell and again remove one input parameter and make the tool easier to use. 

My point is that if 99% of time people need to input a bool parameter which is true, that bool behavior should be added to code by default.

incubator/virtualcluster/cmd/kubectl-vc/use.go
```diff
@@ -0,0 +1,128 @@
...
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/conversion"
+)
+
+const vcKubeconfigHome = ".kube/vc/"
```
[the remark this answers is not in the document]
Fei-Guo: Agree. This one can be added to input parameter.

incubator/virtualcluster/cmd/kubectl-vc/use.go
Fei-Guo: I don't have strong opinion about cli tool but it will be nice to be consistent. Maybe you can explain the benefits of one over another and covert the less beneficial one.

incubator/virtualcluster/cmd/kubectl-vc/use.go
christopherhein: could we have a required flag or params being `ns/clustername` either as `--namespace` and `--cluster-name` or using them as parts of the actual command like `kubectl vc exec -n [ns] [cluster]`

incubator/virtualcluster/cmd/kubectl-vc/use.go
christopherhein: I like that but it does sort of change peoples dev exp around kubeconfigs.

incubator/virtualcluster/cmd/kubectl-vc/use.go
christopherhein: I added a couple notes here - https://github.com/kubernetes-sigs/multi-tenancy/pull/1188#discussion_r503426564 but to summarize.

As we dev this we might want to strive for some consistency across the components so that devs don't need to relearn and we can reuse the same packages. I'm referring to things like having a shared flags implementation cause all of the components use different flags eg `master-kubeconfig` & `kubeconfig`, while not in this cli yet I'm sure we're going to hit this eventually.

incubator/virtualcluster/cmd/kubectl-vc/use.go
zhuangqh: good idea

incubator/virtualcluster/cmd/kubectl-vc/use.go
zhuangqh: i am ok to change to cobra. The main reason for using urfave is that the container engine world is so used. And I am much more familiar with it...😂

Fei-Guo:
Mostly LG. 

Let Chris give another pass. @christopherhein

incubator/virtualcluster/cmd/kubectl-vc/create.go
```diff
@@ -48,64 +48,87 @@ const (
...
+func (o *CreateOptions) Run() error {
+	fileBytes, err := getYamlContent(o.fileName)
 	if err != nil {
-		return nil, errors.Wrapf(err, "required cluster version not found")
+		return errors.Wrapf(err, "read \"%s\"", o.fileName)
 	}
 
-	apiSvcPort, err := getAPISvcPort(cv.Spec.APIServer.Service)
+	vc := &tenancyv1alpha1.VirtualCluster{}
+	if err = yaml.Unmarshal(fileBytes, vc); err != nil {
```
christopherhein, line 109: Do you want to use the apimachinery scheme to decode this for you instead of having to use the yaml parsers. All you have to do is use the scheme to add the tenancy scheme then call `DecodeInto` and pass a runtime.Object and it will parse for you.

```
	content, err := ioutil.ReadFile(o.fileName)
	// handle err
	codecs := serializer.NewCodecFactory(scheme)
	err := runtime.DecodeInto(codecs.UniversalDecoder(), content, vc)
        // handle err
```

for a full example I'm doing this here - https://github.com/kubernetes-sigs/controller-runtime/blob/e6c724240cc44ff8cb96ffeae019341f6d44f24d/pkg/config/config.go#L99-L120

incubator/virtualcluster/cmd/kubectl-vc/exec.go
```diff
@@ -0,0 +1,155 @@
...
+		Short:   "Switch to virtualcluster workspace",
+		Example: execExample,
+		Run: func(cmd *cobra.Command, args []string) {
+			CheckErr(o.Complete(f, cmd, args))
+			CheckErr(o.Run())
+		},
+	}
+
+	cmd.Flags().StringVarP(&o.namespace, "namespace", "n", metav1.NamespaceDefault, "If present, the namespace scope for this CLI request")
+	cmd.Flags().StringVar(&o.kubeFileDir, "kube-file-dir", filepath.Join(os.Getenv("HOME"), ".kube/vc/"), "The directory to place the kubeconfig of specific vc")
```
christopherhein, line 70: should this be `kubeconfig-file-dir`?

incubator/virtualcluster/cmd/kubectl-vc/exec.go
```diff
@@ -0,0 +1,155 @@
...
+	execExample = `
+	# Switch to a virtualcluster
+	kubectl vc exec -n foo bar
+
+	# Specific vc by namespaced name
+	kubectl vc exec foo/bar
+
+	# Customize kubeconfig file path
+	kubectl vc exec --kube-file-dir /path/to/file foo/bar`
+)
```
christopherhein, line 46: Just have to say, this is such a cool idea. Nice work!

christopherhein:
Reviewed, only a couple more comments. Nice work on this!

incubator/virtualcluster/cmd/kubectl-vc/create.go
zhuangqh: can this codec recognize the json or yaml format?

zhuangqh: oh i see, json works fine as well

zhuangqh:
@christopherhein all of the problem addressed is resolved. thx for your suggestion ~

zhuangqh:
link #1118

Fei-Guo:
Indeed nice work. 

/lgtm
/approve

PR #1190, opened 2020-10-13
Implement noneSelector and add test cases

GinnyJI:
If the noneSelector is set to true, the object should not br propagated
to any child namspaces regardless of other selectors

Tested: make test
The test cases test the noneSelector by itself, noneSelector combined
with other selectors, and setting then resetting the noneSelector

### DIFF
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -660,6 +661,7 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
 	selector := r.getSelector(log, inst)
 	treeSelector := r.getTreeSelector(log, inst)
+	noneSelector := r.getNoneSelector(log, inst)
 	nsLabels := r.Forest.Get(dst).GetLabels()
 
 	switch {
@@ -670,6 +672,10 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 
 	// Object with nonempty finalizer list is not propagated
 	case hasFinalizers(inst):
+		return false
+
+	// None selector is set to true
+	case noneSelector:
 		return false
 
 	// Selector does not match
@@ -697,6 +703,29 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 		// Everything else is propagated
 		return true
 	}
+}
+
+// getNoneSelector returns true indicates that user do not want this object to be propagated
+func (r *ObjectReconciler) getNoneSelector(log logr.Logger, inst *unstructured.Unstructured) bool {
+	annot := inst.GetAnnotations()
+	noneSelectorStr, ok := annot[api.AnnotationNoneSelector]
+	if !ok {
+		return false
+	}
+	// Empty string is treated as 'false'. In other selector cases, the empty string is auto converted to
+	// a selector that matches everything.
+	if noneSelectorStr == "" {
+		return false
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// TODO: surface the error
+		log.Error(err, "Invalid noneSelector value", "It should be either true or false, but got", noneSelectorStr)
+		// When the user put an invalid noneSelector, we choose not to propagate this object to any child
+		// namespace to protect any object in the child namespaces to be overwritten
+		return true
+	}
+	return noneSelector
 }
 
 // getTreeSelector is similar to a regular selector, except that it adds the LabelTreeDepthSuffix to every string
@@ -733,6 +762,7 @@ func (r *ObjectReconciler) getSelector(log logr.Logger, inst *unstructured.Unstr
 }
 
 // getSelectorFromString converts the given string to a selector
+// Note: any invalid Selector value will cause this object not propagating to any child namespace
 func (r *ObjectReconciler) getSelectorFromString(log logr.Logger, str string) labels.Selector {
 	labelSelector, err := v1.ParseToLabelSelector(str)
 	// TODO: surface the error messages here (https://github.com/kubernetes-sigs/multi-tenancy/issues/1165)
--- a/incubator/hnc/internal/reconcilers/object_test.go
+++ b/incubator/hnc/internal/reconcilers/object_test.go
@@ -40,6 +40,7 @@ var _ = Describe("Exceptions", func() {
 			name         string
 			selector     string
 			treeSelector string
+			noneSelector string
 			want         []string
 			notWant      []string
 		}{{
@@ -53,6 +54,14 @@ var _ = Describe("Exceptions", func() {
 			want:     []string{c3},
 			notWant:  []string{c1, c2},
 		}, {
+			// When the user input an invalid selector and we don't understand what the users' intention is,
+			// we choose not to propagate this object to any child namespace to protect any object in the child
+			// namespaces to be overwritten. Same for treeSelectors and noneSelector.
+			name:     "not propagate to any namespace with a bad selector",
+			selector: "random",
+			want:     []string{},
+			notWant:  []string{c1, c2, c3},
+		}, {
 			name:         "not propagate object to a negatively treeSelected namespace",
 			treeSelector: "!" + c1,
 			want:         []string{c2, c3},
@@ -63,6 +72,11 @@ var _ = Describe("Exceptions", func() {
 			want:         []string{c3},
 			notWant:      []string{c1, c2},
 		}, {
+			name:         "not propagate to any namespace with a bad treeSelector",
+			treeSelector: "random",
+			want:         []string{},
+			notWant:      []string{c1, c2, c3},
+		}, {
 			name:         "not propagate object to neither negatively selected or treeSelected namespaces",
 			selector:     "!" + c1 + api.LabelTreeDepthSuffix,
 			treeSelector: "!" + c2,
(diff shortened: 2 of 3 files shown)


GinnyJI:
tracked at #1123

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -672,6 +674,10 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 	case hasFinalizers(inst):
 		return false
 
+	// None selector is set to true
+	case noneSelector == true:
```
adrianludwin, line 678: You can just say: `case nonSelector:`

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -699,6 +705,21 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
+	noneSelectorStr, ok := annot[api.AnnotationNoneSelector]
+	if !ok {
+		return false
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		log.Error(err, "Invalid noneSelector value, it should be either true or false.")
+		return false
+	}
+	return noneSelector
```
adrianludwin, line 720: Remind me if we agreed on this? Another option here is "any non-empty string means `true`, empty string or missing means `false`." I'm not sure what the K8s standard is though.

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -74,13 +76,25 @@ var _ = Describe("Exceptions", func() {
...
+			want:         []string{},
+			notWant:      []string{c1, c2, c3},
 		}}
 
 		for _, tc := range tests {
 			// Making a local copy of tc is necessary to ensure the correct value is passed to the closure,
 			// for more details look at https://onsi.github.io/ginkgo/ and search for 'closure'
 			tc := tc
-			It("Should "+tc.name, func() {
+			FIt("Should "+tc.name, func() {
```
adrianludwin, line 97: Don't check in

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -40,6 +41,7 @@ var _ = Describe("Exceptions", func() {
 			name         string
 			selector     string
 			treeSelector string
+			noneSelector bool
```
adrianludwin, line 44: Better to make this a string like the others, that way you can test values other than `true` or `false` which is probably a good test to add here anyway.

adrianludwin (commented):
Looks good! Just a few changes needed.

adrianludwin:
Oh also, can you please explain in the commit message why `make test` is sufficient.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -699,6 +705,21 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
+// getNoneSelector returns true indicates that user do not want this object to be propagated
+func (r *ObjectReconciler) getNoneSelector(log logr.Logger, inst *unstructured.Unstructured) bool {
+	annot := inst.GetAnnotations()
+	noneSelectorStr, ok := annot[api.AnnotationNoneSelector]
+	if !ok {
+		return false
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		log.Error(err, "Invalid noneSelector value, it should be either true or false.")
```
yiqigao217, line 717: BTW should this error be surfaced? Seems yes? If so, maybe add the same todo comment?

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Here also includes the case where user inputs an invalid value (i.e. neither "true" or "false"), and I don't think these should be allowed.

For the empty string, do we want to treat it as `false` or not-allowed?

incubator/hnc/internal/reconcilers/object.go

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

GinnyJI:
> Oh also, can you please explain in the commit message why `make test` is sufficient.

done

incubator/hnc/internal/reconcilers/object.go
adrianludwin: Ok maybe let's keep this as-is. At least it's clear :)

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -74,6 +75,18 @@ var _ = Describe("Exceptions", func() {
 			treeSelector: c2,
 			want:         []string{},
 			notWant:      []string{c1, c2, c3},
+		}, {
+			name:         "not propagate to any object when noneSelector is set to true",
+			noneSelector: "true",
```
adrianludwin, line 80: Try an explicit `false` and a bad value too? Actually all of them should have bad values to show that they're ignored.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -699,6 +705,26 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
+		return false
+	}
+	// Empty string is treated as 'false'
+	if noneSelectorStr == "" {
+		return false
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// TODO: surface the error
+		log.Error(err, "Invalid noneSelector value, it should be either true or false.")
```
adrianludwin, line 722: Does `err` include the actual value of `noneSelectorStr`? If not, we should add it to the log.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -699,6 +705,26 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
+func (r *ObjectReconciler) getNoneSelector(log logr.Logger, inst *unstructured.Unstructured) bool {
+	annot := inst.GetAnnotations()
+	noneSelectorStr, ok := annot[api.AnnotationNoneSelector]
+	if !ok {
+		return false
+	}
+	// Empty string is treated as 'false'
+	if noneSelectorStr == "" {
+		return false
+	}
```
adrianludwin, line 718: This explicit `return false` isn't part of the other two functions. If you pass an empty string as `selector` or `treeSelector`, do those count as "match anything" or "match nothing?" Given that the tests are passing, I'm guessing that we interpret an empty string as "match everything" but we should probably specify that somewhere.

incubator/hnc/internal/reconcilers/object.go
GinnyJI: added

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: done. Also added bad values for selector and treeselector

incubator/hnc/internal/reconcilers/object.go
GinnyJI: The empty string selector matches everything - the reason why I add this line here is because we are using `noneselector` as a boolean value (not a `label.selector`), so we need to specifically handle the empty string case

I've added some comments here

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: Ah I just realized that we are not currently checking the validity of selector and treeSelector. For example if we put a random string to either of the two selectors, it will turn out not to match anything. Maybe this is to be done with the webhook?

adrianludwin: You don't need to improve the testing of selector and treeSelector in this PR, but you should add tests for them. We need to know what will happen (and document it) if the user gets past the webhook. Webhooks are for usability but we never assume they're there.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -699,6 +705,27 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
+	// Empty string is treated as 'false'. In other selector cases, the empty string is auto converted to
+	// a selector that matches everything.
+	if noneSelectorStr == "" {
+		return false
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// TODO: surface the error
+		log.Error(err, "Invalid noneSelector value, it should be either true or false, but got: "+noneSelectorStr)
```
adrianludwin, line 723: Remember to use structured logging, not string concatenation.

adrianludwin (commented):
lgtm once the test is passing

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object_test.go
GinnyJI: The current behavior is that invalid selector value will result in this object not propagating to any ns.

Do you think we need to check and ignore any selector value that doesn't map to the child namespaces?

GinnyJI: I document the current behavior before `getSelectorFromString()`

incubator/hnc/internal/reconcilers/object.go
adrianludwin: sg

incubator/hnc/internal/reconcilers/object_test.go
adrianludwin: lgtm

adrianludwin (commented):
I think this change is ready to go in, but in your next PR can you please add more testing for bad values of all three selectors? Thanks!

/lgtm
/approve
/hold
/assign @rjbez17

GinnyJI:
> I think this change is ready to go in, but in your next PR can you please add more testing for bad values of all three selectors? Thanks!
> 
> /lgtm
> /approve
> /hold
> /assign @rjbez17

Added test cases for bad values

adrianludwin:
Hey @rjbez17 let us know if you had any concern with this, but it's just a continuation of all the other changes you've already approved :)

/hold cancel

PR #1191, opened 2020-10-13
Delete bad anchors and fix e2e tests

adrianludwin:
While running the e2e tests, the CleanupNamespace command was hanging
because we weren't allowing deletion of anchors in the Conflict state;
we were just checking whether the would-be subnamespace had the correct
parent, but not whether the subnamespace anchor was actually in the 'ok'
state. This was resulting in HNC blocking the deletion of a bad anchor,
which we should always allow (this should have been done as a part of
PR #1151).

I also made three minor fixes to the e2e tests themselves:
1. Turn propagation on for new types with the --force option
2. Skip the network policy demo if we detect that network policies
   aren't enabled on the cluster.
3. Delete the webhook configs when HNC might be in a bad state so that
   they don't try to block cleanup of namespaces.

Tested: the e2e tests hang repeatedly and all the demo tests fail
without this fix; with this fix, they pass.

### DIFF
--- a/incubator/hnc/internal/validators/anchor_test.go
+++ b/incubator/hnc/internal/validators/anchor_test.go
@@ -62,6 +62,7 @@ func TestAllowCascadingDeleteSubnamespaces(t *testing.T) {
 		acd  string
 		pnm  string
 		cnm  string
+		stt  api.SubnamespaceAnchorState // anchor state, "ok" by default
 		fail bool
 	}{
 		{name: "set in parent", acd: "c", pnm: "c", cnm: "d"},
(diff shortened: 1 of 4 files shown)


incubator/hnc/internal/validators/anchor_test.go
```diff
@@ -62,14 +62,15 @@ func TestAllowCascadingDeleteSubnamespaces(t *testing.T) {
...
 		cnm  string
+		stt  api.SubnamespaceAnchorState // anchor state, "ok" by default
 		fail bool
 	}{
 		{name: "set in parent", acd: "c", pnm: "c", cnm: "d"},
 		{name: "set in non-leaf", acd: "d", pnm: "c", cnm: "d"},
 		{name: "set in ancestor that is not the first full namespace", acd: "a", pnm: "c", cnm: "d"},
 		{name: "unset in leaf", pnm: "d", cnm: "e"},
 		{name: "unset in non-leaf", pnm: "c", cnm: "d", fail: true},
-		{name: "unset in non-leaf but bad anchor", pnm: "b", cnm: "d"},
```
yiqigao217, line 72: I think we still need to keep this one. This is the case for deleting a bad (nonsense) anchor. Maybe add `stt:api.Conflict` to it and keep it?

yiqigao217 (commented):
Nice work on setting type mode `--force` and checking networkpolicy!

adrianludwin:
/hold

Whoops, I reran the e2e tests and got a related failure from the namespace admission controller. This needs one more upgrade.

incubator/hnc/internal/validators/anchor_test.go
adrianludwin: Done

adrianludwin:
I don't think the e2e failure was caused by this change but I diagnosed and fixed it since I'm fixing other e2e problems anyway.

/hold cancel

yiqigao217 (commented):
/lgtm
/hold

Please feel free to remove the hold.

adrianludwin:
I think this is a fairly small change so @rjbez17 I'm letting this go in now. LMK if you have any concerns.

/hold cancel

PR #1193, opened 2020-10-15
Add e2e tests acting-as different accounts

### LINKED ISSUE #783: HNC: add e2e tests acting-as different accounts
Add a couple of end-to-end tests that make use of act-as features to prove that certain users can take actions that others can't. They don't need to be as comprehensive as the unit tests - just good enough to ensure that the auth system isn't completely broken on any one flavour of K8s (eg GKE, EKS, AKS, Kind, etc).

mbtamuli:
Closes #783. 

Add test to act as different users.

Just ran the single test using `go test -ginkgo.focus 'Acting-as'`, test passed.

Signed-off-by: Mriyam Tamuli <mbtamuli@gmail.com>

### DIFF
--- a/incubator/hnc/test/e2e/acting-as.go
+++ b/incubator/hnc/test/e2e/acting-as.go
@@ -1,0 +1,52 @@
+package e2e
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "sigs.k8s.io/multi-tenancy/incubator/hnc/pkg/testutils"
+)
+
+var _ = Describe("Acting-as", func() {
+	const (
+		nsTarget = "target"
+		saTeamA  = "team-a"
+		saTeamB  = "team-b"
+	)
+
+	BeforeEach(func() {
+		CleanupNamespaces(nsTarget)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsTarget)
+	})
+
+	It("should allow acting as different service accounts", func() {
+		MustRun("kubectl create ns", nsTarget)
+		MustRun("kubectl get ns", nsTarget)
+		MustRun("kubectl get sa -n", nsTarget)
+
+		// create service accounts
+		MustRun("kubectl -n", nsTarget, "create sa", saTeamA)
+		MustRun("kubectl -n", nsTarget, "create sa", saTeamB)
+
+		// fail to create secret
+		MustNotRun("kubectl -n", nsTarget, "create secret generic hnc-secret --from-literal=password=testingsa --as system:serviceaccount:"+nsTarget+":"+saTeamA)
+
+		// allow creation of secret using team-a service account
+		MustRun("kubectl -n", nsTarget, "create role "+saTeamA+"-sre --verb=create --resource=secrets")
+		MustRun("kubectl -n", nsTarget, "create rolebinding "+saTeamA+"-sres --role "+saTeamA+"-sre --serviceaccount="+nsTarget+":"+saTeamA)
+
+		// secret should get created
+		MustRun("kubectl -n", nsTarget, "create secret generic hnc-secret --from-literal=password=testingsa --as system:serviceaccount:"+nsTarget+":"+saTeamA)
+
+		// fail to delete secret
+		MustNotRun("kubectl -n", nsTarget, "delete secret hnc-secret --as system:serviceaccount:"+nsTarget+":"+saTeamB)
+
+		// allow deletion of secret using team-b service account
+		MustRun("kubectl -n", nsTarget, "create role "+saTeamB+"-sre --verb=delete --resource=secrets")
+		MustRun("kubectl -n", nsTarget, "create rolebinding "+saTeamB+"-sres --role "+saTeamB+"-sre --serviceaccount="+nsTarget+":"+saTeamB)
+
+		// secret should get deleted
+		MustRun("kubectl -n", nsTarget, "delete secret hnc-secret --as system:serviceaccount:"+nsTarget+":"+saTeamB)
+	})
+})
--- a/incubator/hnc/test/e2e/acting-as.go
+++ b/incubator/hnc/test/e2e/acting-as.go
@@ -7,46 +7,36 @@ import (
 
 var _ = Describe("Acting-as", func() {
 	const (
-		nsTarget = "target"
-		saTeamA  = "team-a"
-		saTeamB  = "team-b"
+		parent = "parent"
+		child  = "child"
 	)
 
 	BeforeEach(func() {
-		CleanupNamespaces(nsTarget)
+		CleanupNamespaces(parent, child)
 	})
 
 	AfterEach(func() {
-		CleanupNamespaces(nsTarget)
+		CleanupNamespaces(parent, child)
 	})
 
 	It("should allow acting as different service accounts", func() {
-		MustRun("kubectl create ns", nsTarget)
-		MustRun("kubectl get ns", nsTarget)
-		MustRun("kubectl get sa -n", nsTarget)
+		// create the namespaces
+		MustRun("kubectl create ns", parent)
+		MustRun("kubectl create ns", child)
 
-		// create service accounts
-		MustRun("kubectl -n", nsTarget, "create sa", saTeamA)
-		MustRun("kubectl -n", nsTarget, "create sa", saTeamB)
+		// fail to set parent
+		MustNotRun("kubectl hns set", child, "--parent", parent, "--as system:serviceaccount:"+child+":default")
 
-		// fail to create secret
-		MustNotRun("kubectl -n", nsTarget, "create secret generic hnc-secret --from-literal=password=testingsa --as system:serviceaccount:"+nsTarget+":"+saTeamA)
+		// set the proper roles giving the required permissions
+		MustRun("kubectl -n", child, "create role "+child+"-hnc-role --verb=get,create,update --resource=HierarchyConfiguration")
+		MustRun("kubectl -n", child, "create rolebinding "+child+"-hnc-rolebinding --role "+child+"-hnc-admin --serviceaccount="+child+":default")
+		MustRun("kubectl -n", parent, "create role "+parent+"hnc-role --verb=update --resource=HierarchyConfiguration")
+		MustRun("kubectl -n", parent, "create rolebinding "+parent+"hnc-rolebinding --role "+parent+"-hnc-admin --serviceaccount="+child+":default")
+		MustRun("kubectl -n", child, "create role "+child+"-ns-role --verb=get --resource=namespaces")
+		MustRun("kubectl -n", child, "create rolebinding "+child+"-ns-rolebinding --role "+child+"-ns-get --serviceaccount="+child+":default")
 
-		// allow creation of secret using team-a service account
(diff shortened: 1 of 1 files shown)


incubator/hnc/test/e2e/acting-as.go
```diff
@@ -0,0 +1,52 @@
...
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(nsTarget)
+	})
+
+	It("should allow acting as different service accounts", func() {
+		MustRun("kubectl create ns", nsTarget)
+		MustRun("kubectl get ns", nsTarget)
+		MustRun("kubectl get sa -n", nsTarget)
```
adrianludwin, line 26: What's the purpose of these two lines? If `nsTarget` doesn't get created instantly, line 29 will just repeat until it does.

adrianludwin (commented):
Thanks for taking a crack at this! What #783 is intended to address is the HNC authz features, not the K8s builtin features (which this test addresses). But I'm pretty sure you can modify your test fairly easily to test these HNC features.

The features that we're trying to test are described:
* [In the design doc](https://docs.google.com/document/d/10MZfFfbQMm33CBboMq2bfrEtXkJQQT4-UH4DDXZRrKY/edit#heading=h.mh5rr9a4jxat)
* [In the user documentation](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/how-to.md#use-full)
* [In the unit tests for the validating webhook](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/validators/hierarchy_test.go#L152)
  * This might take a bit of careful reading to fully understand - look into the `internal/foresttest` package for guidance.
  * This is *way* overkill for what we need to test in the e2e tests. Having one "success" and one "failure" is probably more than good enough for e2e.

Does this better help explain what we were aiming to test? Again, thanks for doing this!

mbtamuli:
Hey @adrianludwin, I initially misunderstood the task.

But now that I started testing the HNC authz features, I plan on doing these tests.
1. Creation of subnamespace
2. Setting X to be parent of Y

> Having one "success" and one "failure" is probably more than good enough for e2e.

Do you mean saying first try doing with a SA and fail, then allow that SA the required permissions and it should pass?

Let me know if anything else needs to be considered.

adrianludwin:
> But now that I started testing the HNC authz features, I plan on doing these tests.
> 
> 1. Creation of subnamespace

One minor correction - you're creating [full namespaces, not subnamespaces](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/concepts.md#basic-subns), because subnamespaces can have their parents changed, while full namespaces cannot.

> > Having one "success" and one "failure" is probably more than good enough for e2e.
> 
> Do you mean saying first try doing with a SA and fail, then allow that SA the required permissions and it should pass?

Yup, exactly. Thanks!

mbtamuli:
@adrianludwin  I am still confused on this 
> One minor correction - you're creating full namespaces, not subnamespaces, because subnamespaces can have their parents changed, while full namespaces cannot.


So what I have tried till now is,
This refers to the first point I mentioned **Creation of subnamespace** 
So this is testing for creation of subnamespace, right?
```
# create parent namespace
kubectl create ns acme-org

# create service accounts
kubectl -n acme-org create sa user-a

# fail to create subnamespace
kubectl hns create sub1 -nacme-org --as system:serviceaccount:acme-org:user-a

# allow creation of subnamespace using user-a service account
kubectl -n acme-org create role org-sre --verb=create --resource=subnamespaceanchor.hnc.x-k8s.io
kubectl -n acme-org create rolebinding org-sres --role org-sre --serviceaccount=acme-org:user-a

# subnamespace should get created
kubectl hns create sub1 -nacme-org --as system:serviceaccount:acme-org:user-a
```

Now for the second point, **Setting X to be parent of Y**
I saw this part of the doc

> Any regular namespace in a cluster may have its parent set. That is, you may create a namespace via kubectl create namespace foo, and then set its parent via kubectl hns set foo --parent bar. However, this first step always requires cluster-level privileges, which may not be widely granted in your cluster.

> In general, to change the parent of your namespace N from A to B, you must have the following privileges:

> You must be the admin of the highest namespace that will no longer be an ancestor of namespace N after this change, in order to confirm that you are happy to lose your privileges in namespace N.
You must be the admin of namespace B, in order to acknowledge that sensitive objects from B may be copied into namespace N.

I assumed I should be able to create two namespaces and try to make one the parent of the other one, I am not able to proceed just by giving the `update` privileges for `HierarchyConfiguration`

```
# Outside tree (as continuation of the above)
kubectl create ns child-a
kubectl create ns child-b

# fail to set parent
kubectl hns set child-b --parent child-a --as system:serviceaccount:child-b:default

# Make child-b:default the "admin" of child-b namespace
kubectl -n child-b create role team-b-sre --verb=update --resource=HierarchyConfiguration
kubectl -n child-b create rolebinding team-b-sres --role team-b-sre --serviceaccount=child-b:default
```

Even after this, I am not able to set parent
```
kubectl hns set child-b --parent child-a --as system:serviceaccount:child-b:default
```

Am I doing something wrong? Should I even try doing the second part or just the first part should be fine?

adrianludwin:
> So this is testing for creation of subnamespace, right?

This test (while it looks good and it's correct) isn't actually needed, because you're just basically testing that K8s RBAC works. So you can leave this first part out.

> Now for the second point, **Setting X to be parent of Y**

Just before we get into this - you used the namespaces `child-a` and `child-b`, but one of them's the parent and the other is the child, so I'd suggest calling them `parent` and `child` instead.

> > In general, to change the parent of your namespace N from A to B, you must have the following privileges:
> 
> > You must be the admin of the highest namespace that will no longer be an ancestor of namespace N after this change, in order to confirm that you are happy to lose your privileges in namespace N.

In this case, this doesn't mean anything to you because `child` doesn't have an existing parent.

> > You must be the admin of namespace B, in order to acknowledge that sensitive objects from B may be copied into namespace N.

I think this is why your test isn't passing. You made the service account an admin of `child` (by giving it `update` permission in the child) but you _didn't_ make it an admin of `parent`. This is why HNC is rejecting it.

> I assumed I should be able to create two namespaces and try to make one the parent of the other one, I am not able to proceed just by giving the `update` privileges for `HierarchyConfiguration`

That's expected :) If it were that simple, we wouldn't need a dedicated test for it, since that would just be regular K8s RBAC functionality with nothing special for HNC to do.

> Am I doing something wrong? Should I even try doing the second part or just the first part should be fine?

As I mentioned, it's the reverse - we only need the second part, not the first part.

Does this help?

mbtamuli:
Hey @adrianludwin this worked for me. I will be adding this in the tests. 

Should I add any more cases?

```
kubectl create ns parent
kubectl create ns child

# fail to set the parent
kubectl hns set child --parent parent --as system:serviceaccount:child:default

kubectl -n child create role child-hnc-admin --verb=get,create,update --resource=HierarchyConfiguration
kubectl -n child create rolebinding child-hnc-admins --role child-hnc-admin --serviceaccount=child:default
kubectl -n parent create role parent-hnc-admin --verb=update --resource=HierarchyConfiguration
kubectl -n parent create rolebinding parent-hnc-admins --role parent-hnc-admin --serviceaccount=child:default
kubectl -n child create role child-ns-get --verb=get --resource=namespaces
kubectl -n child create rolebinding child-ns --role child-ns-get --serviceaccount=child:default

# after setting the above permissions, now able to set the parent
kubectl hns set child --parent parent --as system:serviceaccount:child:default

```

incubator/hnc/test/e2e/acting-as.go
```diff
@@ -0,0 +1,42 @@
...
+
+	BeforeEach(func() {
+		CleanupNamespaces(parent, child)
+	})
+
+	AfterEach(func() {
+		CleanupNamespaces(parent, child)
+	})
+
+	It("should allow acting as different service accounts", func() {
```
adrianludwin, line 22: What we're really testing here is that HNC will _reject_ calls if your service account has the wrong permissions. The "act-as" is just a feature _of the test_ that we use to test this.

I'd suggest something like `It("should reject hierarchy changes for users without the correct permissions")`.

incubator/hnc/test/e2e/acting-as.go
```diff
@@ -0,0 +1,42 @@
...
+		// create the namespaces
+		MustRun("kubectl create ns", parent)
+		MustRun("kubectl create ns", child)
+
+		// fail to set parent
+		MustNotRun("kubectl hns set", child, "--parent", parent, "--as system:serviceaccount:"+child+":default")
+
+		// set the proper roles giving the required permissions
+		MustRun("kubectl -n", child, "create role "+child+"-hnc-role --verb=get,create,update --resource=HierarchyConfiguration")
```
adrianludwin, line 31: I think you can use the built-in `admin` clusterrole instead of creating a new role here; this will simplify the test. You can grant cluster roles in per-namespace role bindings, which only grants the permissions within a single namespace.

incubator/hnc/test/e2e/acting-as.go
```diff
@@ -0,0 +1,42 @@
...
+		MustRun("kubectl -n", child, "create rolebinding "+child+"-hnc-rolebinding --role "+child+"-hnc-admin --serviceaccount="+child+":default")
```
adrianludwin, line 32: Nit: instead of string manipulation, you can just use commas, which may be more natural, e.g.:

```
MustRun("kubectl -n", child, "create rolebinding", child+"-hnc-rolebinding", "--role", child+"-hnc-admin", "--serviceaccount", child+":default")
```

or if you're using the `admin` clusterrole:

```
MustRun("kubectl -n", child, "create rolebinding local-admin --clusterrole admin --serviceaccount", child+":default")
```

Note that the rolebinding name doesn't need to include `child` in the name; "local-admin" would be fine.

incubator/hnc/test/e2e/acting-as.go
```diff
@@ -0,0 +1,42 @@
...
+		MustRun("kubectl -n", parent, "create role "+parent+"hnc-role --verb=update --resource=HierarchyConfiguration")
+		MustRun("kubectl -n", parent, "create rolebinding "+parent+"hnc-rolebinding --role "+parent+"-hnc-admin --serviceaccount="+child+":default")
+		MustRun("kubectl -n", child, "create role "+child+"-ns-role --verb=get --resource=namespaces")
```
adrianludwin, line 35: Why was this needed?

adrianludwin (commented):
This test is almost correct, but it wouldn't actually catch a regression if HNC suddenly stopped testing authz at all. The reason is that this test verify that the user can't update the hierarchy before granting _any_ permissions at all, including the one used to modify the HierarchyConfiguration in the first place. The test then grants _all_ the required permissions before testing again.

As a result, it would be good to give the permission to update the HC in the child namespace _first_, then verify that this isn't enough to change the parent field in the HC in the child namespace. Then give the permission to update the HC in the _parent_ namespace, and verify that it works.

Does this make sense?

rjbez17:
Hi @mbtamuli  any updates on this?

adrianludwin:
Closing for now, please re-open if you're planning on doing more work on it. Thanks!

/close

PR #1194, opened 2020-10-15
Update HNCConfig API with GR for restmapper lookup

yiqigao217:
In v1alpha2, we will replace the `apiVersion` and `kind` fields in the
hncconfig spec with `group` and `resource` to use restmapper lookup to
see if the type exists and remove version in the spec.

Still keep the object reconcilers and in-memory forest to use GVK. In
HNCConfig reconciler and validator, always get the mapping from GR to
GVK first so it can check forest or create object reconcilers with GVK.

Add `version` field in the HNCConfig status to show users which version
is picked. Add `TypeNotFound` condition for HNCConfig when the type is
not found, which was previously reported with
`ObjectReconcilerCreationFailed`. `ObjectReconcilerCreationFailed` will
only be for errors happened when creating the reconciler.

Hack the CRD conversion with string manipulation from `apiVersion` to
`group` and `kind` to `resource`.

Tested by manual testing on GKE cluster, `make test` and `make
test-conversion`. `make test-e2e` passed eventually (had some failure
when running all tests and then `FIt` the failed tests and they all
passed).

Part of #868

### DIFF
--- a/incubator/hnc/api/v1alpha1/hnc_config_conversion.go
+++ b/incubator/hnc/api/v1alpha1/hnc_config_conversion.go
@@ -16,6 +16,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"strings"
+
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 
 	v1a2 "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
@@ -39,8 +41,21 @@ func (src *HNCConfiguration) ConvertTo(dstRaw conversion.Hub) error {
 	dstSpecTypes := []v1a2.TypeSynchronizationSpec{}
 	for _, st := range srcSpecTypes {
 		dt := v1a2.TypeSynchronizationSpec{}
-		dt.APIVersion = st.APIVersion
-		dt.Kind = st.Kind
+		// Hack the group from APIVersion by removing the version, e.g.
+		// 1) "rbac.authorization.k8s.io/v1" => "rbac.authorization.k8s.io";
+		// 2) "v1" => "" (for core type).
+		gv := strings.Split(st.APIVersion, "/")
+		if len(gv) == 2 {
+			dt.Group = gv[0]
+		}
+		// Hack the resource from Kind by using the lower case and plural form, e.g.
+		// 1) "Role" => "roles"
+		// 2) "NetworkPolicy" => "networkpolicies"
+		lk := strings.ToLower(st.Kind)
+		if strings.HasSuffix(lk, "y") {
+			lk = strings.TrimSuffix(lk, "y") + "ie"
+		}
+		dt.Resource = lk + "s"
 		dtm, ok := toV1A2[st.Mode]
 		if !ok {
 			// This should never happen with the enum schema validation.
--- a/incubator/hnc/internal/reconcilers/hnc_config.go
+++ b/incubator/hnc/internal/reconcilers/hnc_config.go
@@ -9,6 +9,9 @@ import (
 
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
@@ -51,8 +54,12 @@ type ConfigReconciler struct {
 	// not use it.
 	HierarchyConfigUpdates chan event.GenericEvent
 
-	// activeGVKs contains GVKs that are configured in the Spec.
-	activeGVKs gvkSet
+	// activeGVKMode contains GRs that are configured in the Spec and their mapping
+	// GVKs and configured modes.
+	activeGVKMode gr2gvkMode
+
+	// activeGR contains the mapped GVKs of the GRs configured in the Spec.
+	activeGR gvk2gr
 
 	// enqueueReasons maintains the list of reasons we've been asked to update ourselves. Functionally,
 	// it's just a boolean (nil or non-nil), everything else is just for logging.
@@ -60,8 +67,16 @@ type ConfigReconciler struct {
 	enqueueReasonsLock sync.Mutex
 }
 
-// gvkSet keeps track of a group of unique GVKs.
-type gvkSet map[schema.GroupVersionKind]bool
+type gvkMode struct {
+	gvk  schema.GroupVersionKind
+	mode api.SynchronizationMode
+}
+
+// gr2gvkMode keeps track of a group of unique GRs and the mapping GVKs and modes.
+type gr2gvkMode map[schema.GroupResource]gvkMode
+
+// gvk2gr keeps track of a group of unique GVKs with the mapping GRs.
+type gvk2gr map[schema.GroupVersionKind]schema.GroupResource
 
 // checkPeriod is the period that the config reconciler checks if it needs to reconcile the
 // `config` singleton.
@@ -78,6 +93,10 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, err
 	}
 	inst.Status.Conditions = nil
+
+	if err := r.reconcileTypes(inst); err != nil {
+		return ctrl.Result{}, err
+	}
 
 	// Create or sync corresponding ObjectReconcilers, if needed.
 	syncErr := r.syncObjectReconcilers(ctx, inst)
@@ -96,6 +115,47 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Retry reconciliation if there is a sync error.
 	return ctrl.Result{}, syncErr
+}
+
+// reconcileTypes makes sure there's no dup and the types exist. Update the type
+// set with GR to GVK mappings.
+func (r *ConfigReconciler) reconcileTypes(inst *api.HNCConfiguration) error {
+	// Get all resources for all groups.
+	allRes, err := GetAllResources(r.Manager.GetConfig())
+	if err != nil {
+		r.Log.Error(err, "while trying to get all resources")
+		return err
+	}
+
+	// Get valid settings in the `config` singleton.
+	r.activeGVKMode = gr2gvkMode{}
+	r.activeGR = gvk2gr{}
+	for _, t := range inst.Spec.Types {
+		gr := schema.GroupResource{Group: t.Group, Resource: t.Resource}
+		grm := fmt.Sprintf("Group: %s, Resource: %s, Mode: %s", t.Group, t.Resource, t.Mode)
+		// If there are multiple configurations of the same type, we will follow the
+		// first configuration and ignore the rest.
+		if _, exist := r.activeGVKMode[gr]; exist {
+			msg := fmt.Sprintf("Ignore the configuration: %q because the configuration of the type already exists; "+
+				"only the first configuration will be applied", grm)
+			r.Log.Info(msg)
+			r.writeCondition(inst, api.MultipleConfigurationsForOneType, msg)
+			continue
+		}
+
+		// Look if the resource exists in the API server.
+		gvk, err := GVKFor(gr, allRes)
+		if err != nil {
+			// If the type is not found, log error and write conditions but don't
+			// early exit since the other types can still be reconciled.
+			r.Log.Error(err, "while trying to reconcile the configuration", "configuration", grm)
+			r.writeCondition(inst, api.TypeNotFound, err.Error())
+			continue
+		}
+		r.activeGVKMode[gr] = gvkMode{gvk, t.Mode}
+		r.activeGR[gvk] = gr
+	}
+	return nil
 }
 
 // getSingleton returns the singleton if it exists, or creates a default one if it doesn't.
@@ -163,11 +223,11 @@ func (r *ConfigReconciler) validateRBACTypes(inst *api.HNCConfiguration) {
 }
 
 func (r *ConfigReconciler) isRole(t api.TypeSynchronizationSpec) bool {
-	return t.APIVersion == "rbac.authorization.k8s.io/v1" && t.Kind == "Role"
+	return t.Group == api.RBACGroup && t.Resource == api.RoleResource
 }
 
 func (r *ConfigReconciler) isRoleBinding(t api.TypeSynchronizationSpec) bool {
-	return t.APIVersion == "rbac.authorization.k8s.io/v1" && t.Kind == "RoleBinding"
+	return t.Group == api.RBACGroup && t.Resource == api.RoleBindingResource
 }
 
 // writeSingleton creates a singleton on the apiserver if it does not exist.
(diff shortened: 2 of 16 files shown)


yiqigao217:
Running e2e test on this now.

/assign @adrianludwin 
/assign @rjbez17

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -98,6 +110,92 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		"only the first configuration will be applied", specMsg)
+	r.writeCondition(inst, api.MultipleConfigurationsForOneType, msg)
+	return false
+}
+
+// addType sets the type with GR to GVK mapping. Return error if the resource is
+// not found.
+func (r *ConfigReconciler) addType(gr schema.GroupResource, allRes []*restmapper.APIGroupResources) error {
+	// Look for a matching resource from all resources.
+	for _, groupedResources := range allRes {
```
adrianludwin, line 177: This code looks duplicated in the validator. Factor this out so that it's exactly the same code running in both places?

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -163,11 +261,11 @@ func (r *ConfigReconciler) validateRBACTypes(inst *api.HNCConfiguration) {
 }
 
 func (r *ConfigReconciler) isRole(t api.TypeSynchronizationSpec) bool {
-	return t.APIVersion == "rbac.authorization.k8s.io/v1" && t.Kind == "Role"
+	return t.Group == "rbac.authorization.k8s.io" && t.Resource == "roles"
```
adrianludwin, line 264: Switch to const strings?

adrianludwin (commented):
Will finish the review tomorrow. Overall comments: looks good! I think the logic in the reconciler could be simplified, I'm not sure why we keep mapping between GRs and GVKs.

yiqigao217:
> Running e2e test on this now.

FYI the e2e tests all passed eventually. Still got failures when running **ALL** e2e tests. Added `FIt` on them and reran the tests, they all passed.

adrianludwin:
Sigh. Those e2e tests are hard to keep alive, did you make a note of which
ones failed?

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
```diff
@@ -39,8 +41,13 @@ func (src *HNCConfiguration) ConvertTo(dstRaw conversion.Hub) error {
 	dstSpecTypes := []v1a2.TypeSynchronizationSpec{}
 	for _, st := range srcSpecTypes {
 		dt := v1a2.TypeSynchronizationSpec{}
-		dt.APIVersion = st.APIVersion
-		dt.Kind = st.Kind
+		// Hack the group from APIVersion by removing the version.
+		gv := strings.Split(st.APIVersion, "/")
+		if len(gv) == 2 {
+			dt.Group = gv[0]
+		}
```
adrianludwin, line 48: What if it's not?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -232,7 +232,57 @@ spec:
...
+    mode: propagate
+  - apiVersion: v1
+    kind: Secret
+    mode: propagate
+  - apiVersion: v1
+    kind: ResourceQuota
+    mode: remove
+  - apiVersion: v1
+    kind: ConfigMap
+    mode: foobar`
```
adrianludwin, line 265: What should happen here? I think you could make the case that it should just be converted as-is, right?

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -98,6 +110,92 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	// Get all resources for all groups.
+	dc, err := discovery.NewDiscoveryClientForConfig(r.Manager.GetConfig())
+	if err != nil {
+		r.Log.Error(err, "while creating discovery client")
+		return nil, err
+	}
+	return restmapper.GetAPIGroupResources(dc)
+}
+
+// ensureNoDuplicateTypeConfigurations checks whether the configuration of a
```
adrianludwin, line 154: Any reason you moved this function around? It would be easier to compare if you left it where it was (just before syncRemovedReconcilers)

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -60,8 +65,11 @@ type ConfigReconciler struct {
 	enqueueReasonsLock sync.Mutex
 }
 
+// grSet keeps track of a group of unique GRs and their mappings to GVKs.
```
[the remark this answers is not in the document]
adrianludwin: I'm thinking something like this might be a bit more clear:

```
type gvkMode struct {
   gvk schema.GroupVersionKind
   mode api.Mode
}

type grSettings map[schema.GroupResource]gvkMode
```

Now, as you go through the `config.Spec.Types`, you're populating a `grSettings` object, and if it's already populated (with a different value), then you know you have a conflict. Make sense?

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -98,6 +110,92 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		group := groupedResources.Group
+		// Skip resources from a different group.
+		if group.Name != gr.Group {
+			continue
+		}
+		// Search in the resources of the group by version.
+		for _, version := range group.Versions {
+			for _, resource := range groupedResources.VersionedResources[version.Version] {
+				if resource.Name == gr.Resource {
+					// Set the type with the GR to GVK mapping.
```
adrianludwin, line 187: It's worth documenting in the code why we think it's safe to take the first version we find.

yiqigao217:
> Sigh. Those e2e tests are hard to keep alive, did you make a note of which ones failed?

See these 2 full run of the all the e2e tests (with 1 failure and 3 failures in each): https://gist.github.com/yiqigao217/4b00ba1108e8efe10b5fc731b738d30e

The failed tests passed with `FIt` or manually run on a GKE cluster with the same commands.

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
```diff
@@ -39,8 +41,13 @@ func (src *HNCConfiguration) ConvertTo(dstRaw conversion.Hub) error {
...
+		// Hack the resource from Kind by using the lower case.
+		dt.Resource = strings.ToLower(st.Kind) + "s"
```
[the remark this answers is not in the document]
yiqigao217: Yes, otherwise singles will be considered as not found. What about having a specific case for `networkpolicies` mapping?

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
yiqigao217: "v1" apiVersion will be converted to "" group (empty group).

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Ah I see, you mean still with `foobar` mode in the v1alpha2 and let the reconcilers decide on it (ignoring it).
Ah but we have the schema validation that it won't take anything other than `Ignore`, `Propagate` or `Remove`, so I guess we have to convert it and not leave it as-is. SGTY?

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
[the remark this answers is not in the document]
adrianludwin: I mean, anything that ends in "-y" could be replaced by "-ies." That's the standard English pattern and the two of those should catch ~99% of all cases, I think.

What happens if we convert something incorrectly?

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
adrianludwin: Maybe worth a comment

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: Ah right, if the mode is wrong, converting it to Ignore probably makes sense.

What about if the apiVersion or kind are wrong?

incubator/hnc/internal/reconcilers/hnc_config.go
yiqigao217: Ah it's in the order of being called after the caller function. Since it's now called when traversing all types at the beginning, instead of being called in `syncActiveReconcilers` after the active types are already got.

incubator/hnc/internal/reconcilers/hnc_config.go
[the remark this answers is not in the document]
yiqigao217: sgtm. Just confirming we will still keep the gvk2gr mapping for `setTypeStatus` from forest to hncconfig status, right?

incubator/hnc/internal/reconcilers/hnc_config.go
yiqigao217: sg will do.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: There's no schema validation on them yet (neither v1a1 or v1a2) but only webhooks to see if the type exists.

If the apiVersion or kind is wrong, the converted GR will not be found and will have `TypeNotFound` condition. Here's an current example of having `networking.k8s.io/v1/NetworkPolicy` converted to `networking.k8s.io/networkpolicys` (wrong resource, should be `networkpolicies`): 
```
spec:
  types:
  - group: networking.k8s.io
    mode: Propagate
    resource: networkpolicys
  - group: rbac.authorization.k8s.io
    mode: Propagate
    resource: roles
  - group: rbac.authorization.k8s.io
    mode: Propagate
    resource: rolebindings
status:
  conditions:
  - code: TypeNotFound
    msg: Resource "networkpolicys.networking.k8s.io" not found
  types:
  - group: rbac.authorization.k8s.io
    mode: Propagate
    numPropagatedObjects: 0
    numSourceObjects: 0
    resource: rolebindings
    version: v1
  - group: rbac.authorization.k8s.io
    mode: Propagate
    numPropagatedObjects: 0
    numSourceObjects: 0
    resource: roles
    version: v1
```

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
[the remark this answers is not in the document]
yiqigao217: it will have status.conditions `TypeNotFound` and not listed in status.types. See https://github.com/kubernetes-sigs/multi-tenancy/pull/1194#discussion_r506473410 for the full spec and status for wrong `networkpolicys` example for now.

SG will do `y` to `ies` mapping!

incubator/hnc/internal/reconcilers/hnc_config.go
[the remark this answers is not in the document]
adrianludwin: Ah yes, that makes sense.

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: sg, as long as we allow the "bad" type to be created with a condition rather than just omitting it. But we should definitely test that in a conversion test.

incubator/hnc/internal/reconcilers/hnc_config.go
adrianludwin: Yeah, makes sense. When I'm making a big change like this, I generally try to keep things in the same order to make it easy to see what's changed in the diff. Then you can do a separate "rearrange the functions" cleanup PR later. Up to you if you want to change it though.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: sg done.

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -98,6 +110,92 @@ func (r *ConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		// configuration and ignore the rest.
+		if !r.ensureNoDuplicateTypeConfigurations(inst, t, exist) {
+			continue
+		}
+		// Look if the resource exists in the API server.
+		gr := schema.GroupResource{Group: t.Group, Resource: t.Resource}
+		if err := r.addType(gr, allRes); err != nil {
+			// If the type is not found, log error and write conditions but don't
+			// early exit since the other types can still be reconciled.
+			r.Log.Error(err, "while trying to get the resource")
```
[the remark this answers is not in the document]
yiqigao217: Updated to trying to reconcile the type.

incubator/hnc/internal/reconcilers/hnc_config.go
yiqigao217: Removed this func and inlined it with the factored out `GVKFor()` func.

incubator/hnc/internal/validators/hncconfig.go
```diff
@@ -76,83 +76,71 @@ func (c *HNCConfig) Handle(ctx context.Context, req admission.Request) admission
...
+func (c *HNCConfig) validateTypes(inst *api.HNCConfiguration, ts gvkSet) admission.Response {
 	roleExist := false
 	roleBindingExist := false
-	ts := gvkSet{}
 	for _, t := range inst.Spec.Types {
-		if rp := c.isTypeConfigured(t, ts); !rp.Allowed {
-			return rp
+		// Validate the type exists in the apiserver. If yes, convert GR to GVK.
+		gr := schema.GroupResource{Group: t.Group, Resource: t.Resource}
+		gvk, err := c.validator.GVKFor(gr)
```
adrianludwin, line 97: Why do we need to index this on the gvk? Can't we index it on the gr?

adrianludwin (commented):
Code lgtm, still need to finish reviewing the tests.

@rjbez17 , this is a big change but I don't think there's any easy way to break it up. Do you want to start taking a look?

incubator/hnc/internal/validators/hncconfig.go
yiqigao217: We will need `ts` (types with mode) to be stored as `map[gvk]mode` because we will `checkForest()` next to see if any source overwriting would happen if the mode is changed.

adrianludwin: Maybe add that to the comment on line 95? Say _why_ we're converting from GR to GVK as opposed to just stating that we're doing it.

incubator/hnc/internal/reconcilers/test_helpers_test.go
```diff
@@ -18,17 +18,17 @@ import (
...
-	"RoleBinding":   {Group: "rbac.authorization.k8s.io", Version: "v1", Kind: "RoleBinding"},
-	"NetworkPolicy": {Group: "networking.k8s.io", Version: "v1", Kind: "NetworkPolicy"},
-	"ResourceQuota": {Group: "", Version: "v1", Kind: "ResourceQuota"},
-	"LimitRange":    {Group: "", Version: "v1", Kind: "LimitRange"},
-	"ConfigMap":     {Group: "", Version: "v1", Kind: "ConfigMap"},
-	// CronTab is a custom resource.
-	"CronTab": {Group: "stable.example.com", Version: "v1", Kind: "CronTab"},
+	"secrets":               {Group: "", Version: "v1", Kind: "Secret"},
+	api.RoleResource:        {Group: api.RBACGroup, Version: "v1", Kind: api.RoleKind},
+	api.RoleBindingResource: {Group: api.RBACGroup, Version: "v1", Kind: api.RoleBindingKind},
```
adrianludwin, line 25: For future reference - I don't think we needed to make the change to the callers of these functions in this PR - leaving the _existing_ tests with "RoleBinding" when calling `makeObject` and `hasObject` would have made this PR quite a bit smaller without breaking functionality, and then you could have added it in a followup PR. But I don't think it's worth _undoing_ that change either.

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -16,6 +16,15 @@ import (
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
 
+var (
+	gr2gvk = map[schema.GroupResource]schema.GroupVersionKind{
```
adrianludwin, line 20: Comment why we have this - eg to implement a fake discovery client

incubator/hnc/internal/validators/hncconfig_test.go
```diff
@@ -244,15 +253,15 @@ func TestPropagateConflict(t *testing.T) {
 	}
 }
 
-// fakeGVKValidator implements gvkValidator. Any kind that are in the slice are denied; anything else
+// fakeGVKValidator implements grValidator. Any kind that are in the slice are denied; anything else
```
adrianludwin, line 256: Nit: Change name to `fakeGRValidator`? Although it's not really a validator anymore - it's more like a translator. Maybe the interface name should be changed to `grTranslator` or `grLookup` or something like that (here and in the real code)? Before it really just checked if something existed to "validate" made sense, although even then (I now realize in retrospect) that name's a bit confusing since this entire package is called "validators."

incubator/hnc/test/e2e/demo_test.go
```diff
@@ -76,7 +76,7 @@ var _ = Describe("Demo", func() {
 
 	It("Should propagate different types", func() {
 		// ignore Secret in case this demo is run twice and the secret has been set to 'Propagate'
-		MustRun("kubectl hns config set-type --apiVersion v1 --kind Secret Ignore")
+		MustRun("kubectl hns config set-type --resource secrets Ignore")
```
adrianludwin, line 79: File bug to fix documentation?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -232,7 +232,68 @@ spec:
...
+		MustRun("kubectl -n", nsA, "create secret generic my-creds-1 --from-literal=password=iama")
+		RunShouldContain("my-creds-1", propagationTime, "kubectl get secrets -n", nsB)
+
+		// Convert
+		setupV1alpha2()
+
+		// Verify type conversion.
+		FieldShouldNotContain(configCRD, "", configSingleton, ".spec.types", "apiVersion")
+		FieldShouldNotContain(configCRD, "", configSingleton, ".spec.types", "kind")
+		FieldShouldNotContain(configCRD, "", configSingleton, ".spec.types", "v1")
```
adrianludwin, line 283: This might be worth a comment, e.g.: "Look for some text that was legal in v1a1 but isn't legal in v1a2 as a good first check to ensure that everything was converted."

BTW this is a great check.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -232,7 +232,68 @@ spec:
...
+		// Check NetworkPolicy to networkpolicies conversion
+		FieldShouldContain(configCRD, "", configSingleton, ".spec.types", "group:networking.k8s.io mode:Remove resource:networkpolicies")
+		// Check WrongType conversion with TypeNotFound condition.
+		FieldShouldNotContain(configCRD, "", configSingleton, ".status.types", "wrongtypes")
+		FieldShouldContain(configCRD, "", configSingleton, ".status.conditions", "code:TypeNotFound msg:Resource \"wrongtypes\" not found")
```
adrianludwin, line 288: Can you make this more targeted? E.g. can you say `.status.types[wrongtypes]` or something like that in the field name?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -232,7 +232,68 @@ spec:
...
+		// Just randomly pick some other types to check.
+		FieldShouldContain(configCRD, "", configSingleton, ".spec.types", "group:rbac.authorization.k8s.io mode:Propagate resource:rolebindings")
+		FieldShouldContain(configCRD, "", configSingleton, ".spec.types", "group: mode:Remove resource:resourcequotas")
+		FieldShouldContain(configCRD, "", configSingleton, ".spec.types", "group: mode:Propagate resource:secrets")
+
```
adrianludwin, line 293: Can we be a bit more systematic about this? E.g. it would be nice to see a line here for every type that's in the v1a1 spec.

adrianludwin (commented):
lgtm other than some nits and a little more testing needed in the conversion test.

incubator/hnc/internal/reconcilers/test_helpers_test.go
yiqigao217: Thanks for the suggestion, agreed. I will resolve this without any changes.

incubator/hnc/test/e2e/demo_test.go
yiqigao217: Filed #1207 under #868 . Won't add `todo` in any code for now since it's well tracked and not related to code any more.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: No for these two cases `status.types` and `status.conditions` because they are both lists not map that I cannot get a more specific handle.

adrianludwin: Great, thanks.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

rjbez17:
this lgtm other than the hack for the kind -> resource name. We may want (at a minimum) to create a ticket to circle back to it and link in a comment.

incubator/hnc/api/v1alpha1/hnc_config_conversion.go
```diff
@@ -39,8 +41,21 @@ func (src *HNCConfiguration) ConvertTo(dstRaw conversion.Hub) error {
...
-		dt.APIVersion = st.APIVersion
-		dt.Kind = st.Kind
+		// Hack the group from APIVersion by removing the version, e.g.
+		// 1) "rbac.authorization.k8s.io/v1" => "rbac.authorization.k8s.io";
+		// 2) "v1" => "" (for core type).
+		gv := strings.Split(st.APIVersion, "/")
+		if len(gv) == 2 {
+			dt.Group = gv[0]
+		}
+		// Hack the resource from Kind by using the lower case and plural form, e.g.
```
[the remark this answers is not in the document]
yiqigao217: Agreed it's a good concern. The only thing is that if we don't use hack, we would have to use a client in the conversion webhook here to look up the right resource name for the kind from apiserver. I didn't spend too much time on exploring that option though. Probably for now I will do what you suggested, filing an issue and circle back?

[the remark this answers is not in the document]
adrianludwin: So I'm in favour of leaving the hack as-is, because the cost of getting it wrong is very low - you just need to review the HNCConfiguration after installing v0.6 and fix any mistakes. If it works for all the builtin policy types, I feel like that's a win. None of this code will survive to v0.7 so I'm not too excited about spending a lot of additional time on it.

IIRC Jordan's first recommendation was not to do _any_ automatic conversion at all! So I feel this is already kind of above and beyond what's expected of a project at this level of maturity.

Thoughts?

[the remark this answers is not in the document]
rjbez17: sgtm

PR #1195, opened 2020-10-17
kubectl hns config set-type example: `Error: unknown flag:` failed due to case sensitive

frbimo:
This PR fix kubectl hns config set-type's example case-sensitive issue. 
```
✘ bee@thebees  ~   kubectl hns config set-type --apiversion stable.example.com/v1 --kind CronTab propagate
Error: unknown flag: --apiversion
Usage:
  kubectl-hns config set-type --apiVersion X --kind Y <Propagate|Remove|Ignore> [flags]

Examples:
  # Set configuration of a core type
  kubectl hns config set-type --apiVersion v1 --kind Secret ignore

  # Set configuration of a custom type
  kubectl hns config set-type --apiversion stable.example.com/v1 --kind CronTab propagate

Flags:
      --apiVersion string   API version of the kind
  -f, --force               Force to set the propagation mode
  -h, --help                help for set-type
      --kind string         Kind to be configured
...
```

### DIFF
--- a/incubator/hnc/internal/kubectl/configset.go
+++ b/incubator/hnc/internal/kubectl/configset.go
@@ -29,9 +29,9 @@ var setTypeCmd = &cobra.Command{
 		api.Propagate, api.Remove, api.Ignore),
 	Short: "Sets the HNC configuration of a specific resources type",
 	Example: fmt.Sprintf("  # Set configuration of a core type\n" +
-		"  kubectl hns config set-type --apiVersion v1 --kind Secret ignore\n\n" +
+		"  kubectl hns config set-type --apiVersion v1 --kind Secret Ignore\n\n" +
 		"  # Set configuration of a custom type\n" +
-		"  kubectl hns config set-type --apiversion stable.example.com/v1 --kind CronTab propagate"),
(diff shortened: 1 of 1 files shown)


adrianludwin (commented):
This code is about to change more or less completely due to #1194, so _most_ of it isn't needed.

I do think it makes sense for kubectl to accept the mode as either lower case or capitalized (e.g. either `propagate` or `Propagate`), even though the API only accepts the capitalized case. Do you want to try making that change?

frbimo:
> This code is about to change more or less completely due to #1194, so _most_ of it isn't needed.
> 
> I do think it makes sense for kubectl to accept the mode as either lower case or capitalized (e.g. either `propagate` or `Propagate`), even though the API only accepts the capitalized case. Do you want to try making that change?

i see. Should i wait for #1194 to finish first?

adrianludwin:
#1194 is now merged, so you can sync with that and go ahead. Thanks for waiting!

adrianludwin:
Hi @frbimo , lmk if you need a hand with this

frbimo:
i will open a new PR for this @adrianludwin

adrianludwin:
sounds good!

PR #1205, opened 2020-10-19
Add validator check for 'select' annotation

GinnyJI:
If the user input value for "propagate.hnc.x-k8s.io/select" can not be
parsed as a valid labelSelector, we will deny this operation.
Also moved all selector related functions into a package.

Tested:
```
$ k delete role test -n a;
k create role test -n a --verb=update --resource=deployments;
k annotate role test -n a "propagate.hnc.x-k8s.io/select"="cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv"

role.rbac.authorization.k8s.io "test" deleted
role.rbac.authorization.k8s.io/test created
The request is invalid: : Invalid "propagate.hnc.x-k8s.io/select" value: cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv, which can not be parsed as a labelSelector

$ k delete role test -n a;
k create role test -n a --verb=update --resource=deployments;
k annotate role test -n a "propagate.hnc.x-k8s.io/select"="/"

role.rbac.authorization.k8s.io "test" deleted
role.rbac.authorization.k8s.io/test created
The request is invalid: : Invalid "propagate.hnc.x-k8s.io/select" value: /, which can not be parsed as a labelSelector
```

### DIFF
--- a/incubator/hnc/internal/pkg/selectors/selectors.go
+++ b/incubator/hnc/internal/pkg/selectors/selectors.go
@@ -1,0 +1,69 @@
+// Package selectors contains exceptions related utilities, including check the validity of selector,
+// treeSelector, and noneSelector, and parsing these three selectors.
+package selectors
+
+import (
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+
+	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
+)
+
+func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
+	annot := inst.GetAnnotations()
+	return annot[api.AnnotationSelector]
+}
+
+// GetTreeSelector is similar to a regular selector, except that it adds the LabelTreeDepthSuffix to every string
+// To transform a tree selector into a regular label selector, we follow these steps:
+// 1. get the treeSelector annotation if it exists
+// 2. convert the annotation string to a slice of strings seperated by comma, because user is allowed to put multiple selectors
+// 3. append the LabelTreeDepthSuffix to each of the treeSelector string
+// 4. combine them into a single string connected by comma
+func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
+	annot := inst.GetAnnotations()
+	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
+	if !ok {
+		return nil, nil
+	}
+	strs := strings.Split(treeSelectorStr, ",")
+	selectorStr := ""
+	for i, str := range strs {
+		selectorStr = selectorStr + str + api.LabelTreeDepthSuffix
+		if i < len(strs)-1 {
+			selectorStr = selectorStr + ", "
+		}
+	}
+	treeSelector, err := getSelectorFromString(selectorStr)
+	if err != nil {
+		return nil, fmt.Errorf("while parsing %q: %w", api.AnnotationTreeSelector, err)
+	}
+	return treeSelector, nil
+}
+
+// GetSelector returns the selector on a given object if it exists
+func GetSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
+	selector, err := getSelectorFromString(GetSelectorAnnotation(inst))
+	if err != nil {
+		return nil, fmt.Errorf("while parsing %q: %w", api.AnnotationSelector, err)
+	}
+	return selector, nil
+}
+
+// getSelectorFromString converts the given string to a selector
+// Note: any invalid Selector value will cause this object not propagating to any child namespace
+func getSelectorFromString(str string) (labels.Selector, error) {
+	labelSelector, err := v1.ParseToLabelSelector(str)
+	if err != nil {
+		return nil, err
+	}
+	selector, err := v1.LabelSelectorAsSelector(labelSelector)
+	if err != nil {
+		return nil, err
+	}
+	return selector, nil
+}
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -20,14 +20,12 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
-	"strings"
 	"sync"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -43,6 +41,7 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/metadata"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/object"
+	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/pkg/selectors"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/stats"
 )
 
@@ -659,8 +658,14 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 // - Objects have a selector that doesn't match the destination namespace
 // - Service Account token secrets
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
-	selector := r.getSelector(log, inst)
-	treeSelector := r.getTreeSelector(log, inst)
+	selector, err := selectors.GetSelector(inst)
+	if err != nil {
+		log.Error(err, "Invalid value")
+	}
+	treeSelector, err := selectors.GetTreeSelector(inst)
+	if err != nil {
+		log.Error(err, "Invalid value")
+	}
 	noneSelector := r.getNoneSelector(log, inst)
 	nsLabels := r.Forest.Get(dst).GetLabels()
 
@@ -726,55 +731,6 @@ func (r *ObjectReconciler) getNoneSelector(log logr.Logger, inst *unstructured.U
 		return true
 	}
 	return noneSelector
-}
-
-// getTreeSelector is similar to a regular selector, except that it adds the LabelTreeDepthSuffix to every string
-// To transform a tree selector into a regular label selector, we follow these steps:
-// 1. get the treeSelector annotation if it exists
-// 2. convert the annotation string to a slice of strings seperated by comma, because user is allowed to put multiple selectors
-// 3. append the LabelTreeDepthSuffix to each of the treeSelector string
-// 4. combine them into a single string connected by comma
-func (r *ObjectReconciler) getTreeSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
-	annot := inst.GetAnnotations()
-	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
-	if !ok {
-		return nil
-	}
-	strs := strings.Split(treeSelectorStr, ",")
-	selectorStr := ""
-	for i, str := range strs {
-		selectorStr = selectorStr + str + api.LabelTreeDepthSuffix
-		if i < len(strs)-1 {
-			selectorStr = selectorStr + ", "
-		}
-	}
-	return r.getSelectorFromString(log, selectorStr)
-}
-
-// getSelector returns the selector on a given object if it exists
-func (r *ObjectReconciler) getSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
-	annot := inst.GetAnnotations()
-	selectorStr, ok := annot[api.AnnotationSelector]
-	if !ok {
-		return nil
-	}
-	return r.getSelectorFromString(log, selectorStr)
-}
-
-// getSelectorFromString converts the given string to a selector
-// Note: any invalid Selector value will cause this object not propagating to any child namespace
-func (r *ObjectReconciler) getSelectorFromString(log logr.Logger, str string) labels.Selector {
-	labelSelector, err := v1.ParseToLabelSelector(str)
-	// TODO: surface the error messages here (https://github.com/kubernetes-sigs/multi-tenancy/issues/1165)
-	if err != nil {
-		log.Error(err, "Could not parse selector annotation to labelSelector")
-		return nil
-	}
-	selector, err := v1.LabelSelectorAsSelector(labelSelector)
-	if err != nil {
-		log.Error(err, "Could not convert labelSelector to selector")
-	}
-	return selector
 }
 
 // recordPropagatedObject records the fact that this object has been propagated, so we can report
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/metadata"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/object"
+	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/pkg/selectors"
 )
 
 // ObjectsServingPath is where the validator will run. Must be kept in sync with the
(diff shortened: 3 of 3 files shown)


GinnyJI:
tracked at #1123

incubator/hnc/internal/pkg/selector/selector.go
```diff
@@ -0,0 +1,67 @@
+package selector
```
adrianludwin, line 1: A package-level comment is a good practice.

```
// Package selector is ...
```

You start with the words "Package \<packagename\>" for Godoc: https://blog.golang.org/godoc

incubator/hnc/internal/pkg/selector/selector.go
```diff
@@ -0,0 +1,67 @@
...
+// GetSelector returns the selector on a given object if it exists
+func GetSelector(log logr.Logger, inst *unstructured.Unstructured) labels.Selector {
+	selectorStr, ok := GetSelectorAnnotation(inst)
+	if !ok {
+		return nil
+	}
+	return GetSelectorFromString(log, selectorStr)
+}
+
+// getSelectorFromString converts the given string to a selector
```
adrianludwin, line 52: Why do we need to export this function? Why doesn't the validator just call `GetSelector` with `newInst`?

If we can't use `GetSelector`, I'd still like to hide this function because it seems like overkill to export it - I _think_ the only context it's used from outside of this package is to test whether a string is valid, not because we actually need the `labels.Selector` object. If that's correct, maybe we can export a `ValidateSelector` function instead, and leave this one as an unexported function.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -43,6 +43,7 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/metadata"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/object"
+	selectorUtil "sigs.k8s.io/multi-tenancy/incubator/hnc/internal/pkg/selector"
```
adrianludwin, line 46: Nit: consider renaming the package to `selectors` (like the Go standard library `strings` or `errors`) and then import it without renaming it here.

We should probably rename the `object` package to `objects` as well but not in this PR. ("metadata" is already plural, and there's usually only one "forest" so singular makes sense there IMO).

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -659,8 +660,8 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 // - Objects have a selector that doesn't match the destination namespace
 // - Service Account token secrets
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
-	selector := r.getSelector(log, inst)
-	treeSelector := r.getTreeSelector(log, inst)
```
adrianludwin, line 663: Did you actually delete these old functions?

incubator/hnc/internal/validators/object.go
```diff
@@ -111,18 +112,31 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	// If the object wasn't and isn't inherited, we will check to see if the
 	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
+		// check selector format
+		oldSelectorStr, _ := selector.GetSelectorAnnotation(oldInst)
+		newSelectorStr, _ := selector.GetSelectorAnnotation(inst)
+		sel := selector.GetSelectorFromString(log, newSelectorStr)
+		if oldSelectorStr != newSelectorStr && sel == nil {
```
adrianludwin, line 119: Maybe comment on _why_ we compare the old and new ones first.

incubator/hnc/internal/pkg/selector/selector.go
```diff
@@ -0,0 +1,67 @@
...
+// Note: any invalid Selector value will cause this object not propagating to any child namespace
+func GetSelectorFromString(log logr.Logger, str string) labels.Selector {
+	labelSelector, err := v1.ParseToLabelSelector(str)
+	// TODO: surface the error messages here (https://github.com/kubernetes-sigs/multi-tenancy/issues/1165)
+	if err != nil {
```
adrianludwin, line 57: We're throwing away valuable information from the label parser here by not returning the error - check to see if there's anything in there that would be useful to present to the user.

incubator/hnc/internal/validators/object.go
```diff
@@ -111,18 +112,31 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
+			return deny(metav1.StatusReasonInvalid, "Invalid \""+api.AnnotationSelector+"\" value: "+newSelectorStr+
+				", which can not be parsed as a labelSelector")
```
adrianludwin, line 121: Instead of `, which can not be parsed as a labelSelector`, I'd recommend ` is not a valid Kubernetes labelSelector`. If you Google "kubernetes labelselector" you immediately see the correct documentation as the first result, and this makes it clear that it's a K8s concept that the user needs to learn about, not an HNC concept.

Also, "can not be parsed as" suggests there might be an error with the parser, but "is not a valid labelselector" makes it clear that the user has put in a bad value.

The error returned by the parser would probably be very useful here.

Finally, it probably makes sense to change this to a format string which would be much easier to read:

```
msg := fmt.Sprintf("Invalid %s value: %q is not a valid Kubernetes labelSelector: %s", api.AnnotationSelector, newSelectorStr, err)
```
```

File: incubator/hnc/internal/pkg/selector/selector.go [hunk 8]
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+
+	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
+)
+
+func GetSelectorAnnotation(inst *unstructured.Unstructured) (string, bool) {
```
yiqigao217, line 14: Seems this func doesn't need export either, if we export `GetSelector` and `GetTreeSelector`?

incubator/hnc/internal/pkg/selector/selector.go
GinnyJI: Does `log.Error(err, "Could not parse selector annotation to labelSelector")` not log the `err`?

incubator/hnc/internal/pkg/selector/selector.go
GinnyJI: This is to check if a selector annotation exists. Maybe I should rename this function

incubator/hnc/internal/validators/object.go
GinnyJI: done

incubator/hnc/internal/pkg/selector/selector.go
GinnyJI: done

incubator/hnc/internal/pkg/selector/selector.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go

incubator/hnc/internal/reconcilers/object.go
GinnyJI: deleted now

incubator/hnc/internal/validators/object.go
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/go-logr/logr"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+
+	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
+)
+
+func GetSelectorAnnotation(inst *unstructured.Unstructured) (string, bool) {
```
adrianludwin, line 17: Didn't we agree that an empty selector is the same as a missing selector? If so, we don't need to return  `(string, bool)`, we can just return a string, right?

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -0,0 +1,70 @@
...
+// To transform a tree selector into a regular label selector, we follow these steps:
+// 1. get the treeSelector annotation if it exists
+// 2. convert the annotation string to a slice of strings seperated by comma, because user is allowed to put multiple selectors
+// 3. append the LabelTreeDepthSuffix to each of the treeSelector string
+// 4. combine them into a single string connected by comma
+func GetTreeSelector(log logr.Logger, inst *unstructured.Unstructured) (labels.Selector, error) {
+	annot := inst.GetAnnotations()
+	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
+	if !ok {
+		return nil, errors.New("TreeSelector not found.")
```
adrianludwin, line 33: It's not an error condition for an object to _not_ have a tree selector - in fact, that's the normal case. So we shouldn't return an error here IMO, just `nil, nil`.

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -0,0 +1,70 @@
...
+	return getSelectorFromString(log, selectorStr)
+}
+
+// getSelectorFromString converts the given string to a selector
+// Note: any invalid Selector value will cause this object not propagating to any child namespace
+func getSelectorFromString(log logr.Logger, str string) (labels.Selector, error) {
+	labelSelector, err := v1.ParseToLabelSelector(str)
+	// TODO: surface the error messages here (https://github.com/kubernetes-sigs/multi-tenancy/issues/1165)
+	if err != nil {
+		log.Error(err, "Could not parse selector annotation to labelSelector")
```
adrianludwin, line 61: As a general rule, errors should _either_ be returned _or_ logged. The reason is that you never want to see the same error cause lots of different log messages, which is what will happen if every function both logs and returns it.

What _is_ useful is [wrapping an error with more information](https://blog.golang.org/go1.13-errors). E.g., let's assume that both `v1.ParseToLabelSelector` and `v1.LabelSelectorAsSelector` both return errors that clearly identify which function returned the error (which is a good assumption). In that case, this function (`getSelectorFromString`) doesn't have any useful information to add. But the _callers_ do - they know if the error was caused by a `/selector` or `/treeSelector`. So, for example, in `GetSelector`, you might do something like this:

```
selector, err := getSelectorFromString(selectorStr)
if err != nil {
  return nil, fmt.Errorf("while parsing %q: %w", api.AnnotationSelector, err)
}
return selector, nil
```

In the validator, we can then pass this message back to the user (I _think_ the message will naturally show the object being rejected, so we don't need to add that information here). In the reconciler, we _don't_ want to return the error to controller-runtime, since it's a _user_ error and we don't want to re-reconcile the object. So that's a good place to log the error - and the logger already knows the name of the affected object, so we don't need to add it again there either.

Does this make sense?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -659,8 +658,8 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
...
+	selector, _ := selectors.GetSelector(log, inst)
+	treeSelector, _ := selectors.GetTreeSelector(log, inst)
```
adrianludwin, line 662: Throwing away return values is generally a "code smell" - not _necessarily_ an error, but a good indication that you should probably have a much closer look at what you're throwing away. Very likely, the thing you're throwing away isn't valuable (in which case, why are you returning it in the first place?) or it _is_ valuable (in which case, why are you throwing it away?) - in this case, I think the former applies.

incubator/hnc/internal/validators/object.go
```diff
@@ -111,18 +112,32 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	// If the object wasn't and isn't inherited, we will check to see if the
 	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
+		// check selector format
+		oldSelectorStr, _ := selectors.GetSelectorAnnotation(oldInst)
+		newSelectorStr, _ := selectors.GetSelectorAnnotation(inst)
+		sel, err := selectors.GetSelector(log, inst)
+		// If this is a selector change, and the new selector is not valid, we'll deny this operation
+		if oldSelectorStr != newSelectorStr && sel == nil {
+			msg := fmt.Sprintf("Invalid %s value: %q is not a valid Kubernetes labelSelector: \n  %s", api.AnnotationSelector, newSelectorStr, err)
```
adrianludwin, line 121: We usually don't put newlines in our validation error messages, so while it's a good idea, I think I'd take this out.

Also, if you follow my advice above about how to wrap the error message, this message can be much shorter since more of the information will already be in the error, e.g.:

```
msg := fmt.Sprintf("Invalid Kubernetes labelSelector: %s", err")
```

adrianludwin: ... also note the message that you're currently displaying:

```
The request is invalid: : Invalid propagate.hnc.x-k8s.io/select value: "cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv" is not a valid Kubernetes labelSelector:
  couldn't parse the selector string "cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv": unable to parse requirement: invalid label key "cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv": name part must be no more than 63 characters
```

There's a _lot_ of repeated information in that message. What do you think it _should_ look like?

incubator/hnc/internal/pkg/selector/selector.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: My understanding is that because in 'validators' we need the error to show it to the user, but here we have already logged the error so we don't need it in the 'reconcilers' anymore.

adrianludwin: Right, but see my other comments about where the error _should_ be logged.

incubator/hnc/internal/validators/object.go
GinnyJI: The part 
```
couldn't parse the selector string 
"cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv":
unable to parse requirement: invalid label key
"cqoornttdrkjacuiqagaqottaprzpjzfjudvlqziugeywbzgbdfvcxcimwiecodcpehldykmrikoxixv":
name part must be no more than 63 characters
```
comes from k8s. I'm thinking keeping it and just add `Invalid Kubernetes labelSelector: ` in the front. Basically what you suggested above. Wdyt?

incubator/hnc/internal/pkg/selectors/selectors.go

incubator/hnc/internal/pkg/selectors/selectors.go

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: I see! Done

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done

incubator/hnc/internal/validators/object.go
adrianludwin: That sgtm!

incubator/hnc/internal/validators/object.go
```diff
@@ -111,18 +112,37 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
 			msg := fmt.Sprintf("\nCannot create %q (%s) in namespace %q because it would overwrite objects in the following descendant namespace(s):\n  * %s\nTo fix this, choose a different name for the object, or remove the conflicting objects from the above namespaces.", inst.GetName(), inst.GroupVersionKind(), inst.GetNamespace(), dnsesStr)
 			return deny(metav1.StatusReasonConflict, msg)
 		}
 		return allow("source object")
 	}
+	// This is a propagated object.
+	return o.handleInherited(op, newSource, oldSource, inst, oldInst)
+}
+
+func isInvalidSelectorChange(inst, oldInst *unstructured.Unstructured) (bool, error) {
```
adrianludwin, line 135: Why not just return `nil` if everything's ok, and not a bool? That also makes the caller much more clear - right now, you could imagine someone asking "what if `invalid` is true but `err` is nil?"

```
func validateSelectorChange(inst, oldInst *u.U) error {
  oldStr := ...
  newStr := ...
  if oldStr == newStr {
    return nil
  }
  _, err := selectors.GetSelector(inst)
  return err
}
```

incubator/hnc/internal/pkg/selector/selector.go
GinnyJI: done

incubator/hnc/internal/validators/object.go
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	strs := strings.Split(treeSelectorStr, ",")
+	selectorStr := ""
+	for i, str := range strs {
+		selectorStr = selectorStr + str + api.LabelTreeDepthSuffix
+		if i < len(strs)-1 {
+			selectorStr = selectorStr + ", "
+		}
+	}
+	return getSelectorFromString(selectorStr)
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, fmt.Errorf("while parsing %q: %w", api.AnnotationSelector, err)
+	}
+	return selector, nil
+}
+
+// getSelectorFromString converts the given string to a selector
+// Note: any invalid Selector value will cause this object not propagating to any child namespace
+func getSelectorFromString(str string) (labels.Selector, error) {
+	labelSelector, err := v1.ParseToLabelSelector(str)
+	// TODO: surface the error messages here (https://github.com/kubernetes-sigs/multi-tenancy/issues/1165)
```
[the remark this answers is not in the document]
GinnyJI: removed

incubator/hnc/internal/validators/object.go
GinnyJI: done

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

PR #1212, opened 2020-10-21
Update HNCConfig conditions into standard schema

yiqigao217:
Update the {code, msg} schema into the standard {type, reason,
message...} schema. Update tests and add new conversion tests.

Tested manually on GKE cluster and by 'make test' and 'make
test-conversion'.

Example outputs:
```
$ k hns config describe
Synchronized types:
* Propagating: rolebindings (rbac.authorization.k8s.io/v1)
* Propagating: roles (rbac.authorization.k8s.io/v1)

Conditions:
BadConfiguration (MultipleConfigurationsForType): roles.rbac.authorization.k8s.io has multiple synchronization modes; all but one (Propagate) will be ignored.
```
HNCConfig.status.conditions:
```
status:
  conditions:
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: 'roles.rbac.authorization.k8s.io has multiple synchronization modes; all but
      one (Propagate) will be ignored.'
    reason: MultipleConfigurationsForType
    status: "True"
    type: BadConfiguration
```

Part of #868


adrianludwin:
Just a few quick high-level comments before I review the code: the way the conditions are printed out don't seem to match how we print conditions for the namespaces (e.g. the square brackets). Can we line those up?

Also, the most interesting information seems to be at the end of the message, rather than the beginning. How about something like this?

```
Conditions:
BadConfiguration (MultipleConfigurationsForOneType): rbac.authorization.k8s.io/roles has multiple settings; all but one will be ignored.
```

This puts all the relevant information at the front and will probably be easier to read IMO.

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -210,22 +210,23 @@ spec:
...
 
-	It("should still have HC condition if it exists in v1alpha1", func() {
+	It("should still have HC condition if it exists in v1alpha1, but with a new schema", func() {
 		// Before conversion, create namespace B with a missing parent A (have to
 		// create A first and then delete it because otherwise the webhook will deny
 		// setting a non-existing namespace as parent).
 		createSampleV1alpha1Tree()
 		MustRun("kubectl delete ns", nsA)
-		FieldShouldContain(hierCRD, nsB, hierSingleton, ".status.conditions", "CritParentMissing")
+		FieldShouldContain(hierCRD, nsB, hierSingleton, ".status.conditions", "code:CritParentMissing")
```
adrianludwin, line 219: Why is this part of this change?

incubator/hnc/api/v1alpha2/hnc_config.go
```diff
@@ -48,12 +48,17 @@ const (
...
-	TypeNotFound                     HNCConfigurationCode = "TypeNotFound"
-	ObjectReconcilerCreationFailed   HNCConfigurationCode = "ObjectReconcilerCreationFailed"
-	MultipleConfigurationsForOneType HNCConfigurationCode = "MultipleConfigurationsForOneType"
+	// Condition types.
+	ConditionBadTypeConfiguration string = "BadConfiguration"
+	ConditionOutOfSync            string = "OutOfSync"
+
+	// Condition reasons for BadConfiguration
+	ReasonMultipleConfigsForOne string = "MultipleConfigurationsForOneType"
+	ReasonTypeNotExist          string = "TypeNotExist"
```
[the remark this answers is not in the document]
yiqigao217: Oh `TypeNotExist` is what the doc says (but it's also what I wrote then), so I changed it from `TypeNotFound` to it. WDYT?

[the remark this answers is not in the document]
yiqigao217: Maybe call it `ResourceNotFound` since you are changing the  `spec.types` field into `spec.resources`?

incubator/hnc/api/v1alpha2/hnc_config.go
```diff
@@ -48,12 +48,17 @@ const (
...
+
+	// Condition reason for OutOfSync
+	ReasonCannotCreateReconciler string = "CannotCreateReconciler"
```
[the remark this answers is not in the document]
yiqigao217: sg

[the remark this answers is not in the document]
yiqigao217: Ah the reason `Unknown` may look weird with the type.
Does condition {type:OutOfSync reason:Unknown message: cannot create reconciler...} lgty?

incubator/hnc/api/v1alpha2/hnc_config.go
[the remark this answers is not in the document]
yiqigao217: Discussed offline - updated wiht `TypeNofFound` in this PR.

incubator/hnc/api/v1alpha2/hnc_config.go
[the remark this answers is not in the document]
yiqigao217: Discussed offline: use `Unknown`.

incubator/hnc/test/conversion/conversion_test.go
yiqigao217: Ah this is just making it more specific since the old schema uses `code`, although `CritParentMissing` naming with `Crit` prefix is already known as v1a1.

adrianludwin: sg!

incubator/hnc/internal/kubectl/configdescribe.go
```diff
@@ -44,7 +44,7 @@ var configDescribeCmd = &cobra.Command{
 		}
 		fmt.Print("\nConditions:\n")
 		for _, c := range config.Status.Conditions {
-			fmt.Printf("[%s] %s\n", string(c.Code), c.Msg)
+			fmt.Printf("%s (%s) %s\n", c.Type, c.Reason, c.Message)
```
adrianludwin, line 47: Shouldn't there be a colon in there?

`type (reason): msg`

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -132,14 +132,12 @@ func (r *ConfigReconciler) reconcileTypes(inst *api.HNCConfiguration) error {
...
 	for _, t := range inst.Spec.Types {
 		gr := schema.GroupResource{Group: t.Group, Resource: t.Resource}
-		grm := fmt.Sprintf("Group: %s, Resource: %s, Mode: %s", t.Group, t.Resource, t.Mode)
 		// If there are multiple configurations of the same type, we will follow the
 		// first configuration and ignore the rest.
-		if _, exist := r.activeGVKMode[gr]; exist {
-			msg := fmt.Sprintf("Ignore the configuration: %q because the configuration of the type already exists; "+
-				"only the first configuration will be applied", grm)
+		if gvkMode, exist := r.activeGVKMode[gr]; exist {
+			msg := fmt.Sprintf("%s has multiple settings; all but one (mode: %s) will be ignored.", gr, gvkMode.mode)
```
adrianludwin, line 138: s/settings/synchronization modes/ ? Then you can get rid of the "mode:" in the brackets.

adrianludwin (commented):
Two last small changes

incubator/hnc/internal/reconcilers/hnc_config.go
yiqigao217: Just confirming that we don't need to specify the mode is being used or ignored?

Some integration tests used to verify what modes are ignored and I changed it to verify what modes are used since we changed the msg. If removing the mode in the msg, we just verify a type has the condition. SGTY?

incubator/hnc/internal/kubectl/configdescribe.go
yiqigao217: ah right missed that. will add it now.

incubator/hnc/internal/reconcilers/hnc_config.go
adrianludwin: Sorry, that wasn't very clear!

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

PR #1184, opened 2020-10-09
Update conditions to use standard schema

yiqigao217:
Copy the standard condition schema to hierarhcy_configuration CRD.
Update the forest and the reconcilers to use the standard schema. Simply
the condition setting logic to always overwrite conditions in each
reconciliation.

Update the condition codes into types and reasons. Update the
kubectl-hns output and metrics accordingly.

Update namespace_conditions metrics keys into 'Condition' and 'Reason'.

Tested by 'make test', 'make test-conversion' and 'make test-e2e'.
Tested manually on GKE cluster and StackDriver metrics explorer.

Part of #868

PR #1215, opened 2020-10-21
Fix e2e tests after conditions updates

adrianludwin:
kubectl-hns was changed in #1184 but the e2e tests were still expecting
the old output; this change updates the e2e tests. In addition, #1171
moved object conditions to events, and those events included names from
an old ancestor, confusing the demo test that was expecting the old
ancestor to _not_ appear; this change fixes that too.

Tested: e2e tests failed but now pass


adrianludwin:
This only needs one person to lgtm

yiqigao217 (commented):
/lgtm

Ah right, they were updated in the initial commit and after addressing the comments, they were not updated again after changing the format.

PR #1216, opened 2020-10-21
Add validator check for `treeSelect`

GinnyJI:
We check if the raw string input from user is a valid DNS label. If this
is a valid DNS label but couldn't be parsed as a selector for some
reason, we will show user that it's an internal error, so that user can
reach us for help. In theory this should never happen.

Tested:
```
$k delete role test -n a
k create role test -n a --verb=update --resource=deployments
k annotate role test -n a "propagate.hnc.x-k8s.io/treeSelect"="1234567890123456789012345678901234567890123456789012345678901234567890"

role.rbac.authorization.k8s.io "test" deleted
role.rbac.authorization.k8s.io/test created
Error from server (BadRequest): admission webhook "objects.hnc.x-k8s.io" denied the request: Invalid HNC "propagate.hnc.x-k8s.io/treeSelect" value: 1234567890123456789012345678901234567890123456789012345678901234567890 is not a valid namespace name, must be no more than 63 characters
```

### DIFF
--- a/incubator/hnc/internal/pkg/selectors/selectors.go
+++ b/incubator/hnc/internal/pkg/selectors/selectors.go
@@ -9,6 +9,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/validation"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
@@ -18,6 +19,11 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
 	return annot[api.AnnotationSelector]
 }
 
+func GetTreeSelectorAnnotation(inst *unstructured.Unstructured) string {
+	annot := inst.GetAnnotations()
+	return annot[api.AnnotationTreeSelector]
+}
+
 // GetTreeSelector is similar to a regular selector, except that it adds the LabelTreeDepthSuffix to every string
 // To transform a tree selector into a regular label selector, we follow these steps:
 // 1. get the treeSelector annotation if it exists
@@ -25,24 +31,48 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
 // 3. append the LabelTreeDepthSuffix to each of the treeSelector string
 // 4. combine them into a single string connected by comma
 func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
-	annot := inst.GetAnnotations()
-	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
-	if !ok {
+	treeSelectorStr := GetTreeSelectorAnnotation(inst)
+	if treeSelectorStr == "" {
 		return nil, nil
 	}
-	strs := strings.Split(treeSelectorStr, ",")
+
+	segs := strings.Split(treeSelectorStr, ",")
 	selectorStr := ""
-	for i, str := range strs {
-		selectorStr = selectorStr + str + api.LabelTreeDepthSuffix
-		if i < len(strs)-1 {
+	for i, seg := range segs {
+		// check if it's a valid namespace name
+		if err := validateTreeSelectorSegment(seg); err != nil {
+			return nil, err
+		}
+
+		selectorStr = selectorStr + seg + api.LabelTreeDepthSuffix
+		if i < len(segs)-1 {
 			selectorStr = selectorStr + ", "
 		}
 	}
 	treeSelector, err := getSelectorFromString(selectorStr)
 	if err != nil {
-		return nil, fmt.Errorf("while parsing %q: %w", api.AnnotationTreeSelector, err)
+		// In theory this should never happen because we already checked DNS label before.
+		// If this happens, it's more likely that we have a bug in our code
+		return nil, fmt.Errorf("internal error while parsing %q: %w", api.AnnotationTreeSelector, err)
 	}
 	return treeSelector, nil
+}
+
+func validateTreeSelectorSegment(seg string) error {
+	seg = strings.TrimSpace(seg)
+	ns := ""
+	if seg[0] == '!' {
+		ns = seg[1:]
+	} else {
+		ns = seg
+	}
+	errStrs := validation.IsDNS1123Label(ns)
+	if len(errStrs) != 0 {
+		// If IsDNS1123Label() returns multiple errors, it will look like:
+		// "ns" is not a valid namespace name: err1; err2
+		return fmt.Errorf("%q is not a valid namespace name: %s", ns, strings.Join(errStrs, "; "))
+	}
+	return nil
 }
 
 // GetSelector returns the selector on a given object if it exists
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -117,9 +117,13 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 		// If this is a selector change, and the new selector is not valid, we'll deny this operation
 		if err != nil {
 			msg := fmt.Sprintf("Invalid Kubernetes labelSelector: %s", err)
-			return deny(metav1.StatusReasonInvalid, msg)
+			return deny(metav1.StatusReasonBadRequest, msg)
 		}
-
+		err = validateTreeSelectorChange(inst, oldInst)
+		if err != nil {
+			msg := fmt.Sprintf("Invalid HNC %q value: %s", api.AnnotationTreeSelector, err)
+			return deny(metav1.StatusReasonBadRequest, msg)
+		}
 		// TODO(@ginnyji): modify hasConflict so that it's aware of selectors
 		if yes, dnses := o.hasConflict(inst); yes {
 			dnsesStr := strings.Join(dnses, "\n  * ")
@@ -139,6 +143,16 @@ func validateSelectorChange(inst, oldInst *unstructured.Unstructured) error {
 		return nil
 	}
 	_, err := selectors.GetSelector(inst)
+	return err
+}
+
+func validateTreeSelectorChange(inst, oldInst *unstructured.Unstructured) error {
+	oldSelectorStr := selectors.GetTreeSelectorAnnotation(oldInst)
+	newSelectorStr := selectors.GetTreeSelectorAnnotation(inst)
+	if newSelectorStr == "" || oldSelectorStr == newSelectorStr {
+		return nil
+	}
+	_, err := selectors.GetTreeSelector(inst)
 	return err
 }
 
--- a/incubator/hnc/internal/validators/object_test.go
+++ b/incubator/hnc/internal/validators/object_test.go
@@ -336,6 +336,79 @@ func TestUserChanges(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "Deny creation of object with invalid treeSelect annotation",
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationTreeSelector: "foo, $bar",
+					},
+				},
+			},
+		},
+	}, {
+		name: "Allow creation of object with valid treeSelect annotation",
+		fail: false,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationTreeSelector: "foo, !bar",
+					},
+				},
+			},
+		},
+	}, {
+		name: "Deny creation of object with invalid selector and valid treeSelect annotation",
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector:     "$foo",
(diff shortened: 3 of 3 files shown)


adrianludwin:
That error message doesn't look right to me:

```
The request is invalid: : Invalid Kubernetes labelSelector: 
1234567890123456789012345678901234567890123456789012345678901234567890,
must be no more than 63 characters
```

There's an extra colon in there for some reason, but more importantly, this isn't a K8s labelSelector, it's an HNC tree selector with very different rules. We're not giving users the information that the problem is that the tree selector contains invalid namespace names.

GinnyJI:
> That error message doesn't look right to me:
> 
> ```
> The request is invalid: : Invalid Kubernetes labelSelector: 
> 1234567890123456789012345678901234567890123456789012345678901234567890,
> must be no more than 63 characters
> ```
> 
> There's an extra colon in there for some reason, but more importantly, this isn't a K8s labelSelector, it's an HNC tree selector with very different rules. We're not giving users the information that the problem is that the tree selector contains invalid namespace names.

Updated the error msg. I looked into the double colons, I don't _think_ it's from my change. I haven't found out the place that's causing this, still looking

adrianludwin:
> Updated the error msg. I looked into the double colons, I don't _think_ it's from my change. I haven't found out the place that's causing this, still looking

Out of interest, if you change the response code (e.g. from `metav1.StatusReasonInvalid` to something else, like `metav1.StatusReasonBadRequest`)? I agree that the code you picked is the correct one, I'm just wondering if there's something funny about the way it gets displayed to the user.

adrianludwin:
Also you need to identify the `treeSelect` annotation somewhere or users won't know which field is causing the problem.

incubator/hnc/internal/validators/object.go
```diff
@@ -142,6 +148,24 @@ func validateSelectorChange(inst, oldInst *unstructured.Unstructured) error {
...
+	oldSelectorStr := selectors.GetTreeSelectorAnnotation(oldInst)
+	newSelectorStr := selectors.GetTreeSelectorAnnotation(inst)
+	if newSelectorStr == "" || oldSelectorStr == newSelectorStr {
+		return nil
+	}
+	errStrs := validation.IsDNS1123Label(newSelectorStr)
+	if len(errStrs) != 0 {
+		return errors.New(newSelectorStr + ", " + strings.Join(errStrs, ". "))
+	}
+	_, err := selectors.GetTreeSelector(inst)
```
adrianludwin, line 161: The error message should be generated in the `selectors` package, not here, so that we can display the _same_ error from the reconciler if needed. `GetTreeSelector` should be the one that returns the error.

GinnyJI:
> > Updated the error msg. I looked into the double colons, I don't _think_ it's from my change. I haven't found out the place that's causing this, still looking
> 
> Out of interest, if you change the response code (e.g. from `metav1.StatusReasonInvalid` to something else, like `metav1.StatusReasonBadRequest`)? I agree that the code you picked is the correct one, I'm just wondering if there's something funny about the way it gets displayed to the user.

Yeah if we use `metav1.StatusReasonBadRequest` the error msg looks better formatted
```
Error from server (BadRequest): admission webhook "objects.hnc.x-k8s.io" denied the request: Invalid HNC treeSelect value: 1234567890123456789012345678901234567890123456789012345678901234567890, must be no more than 63 characters
```

adrianludwin:
> Yeah if we use `metav1.StatusReasonBadRequest` the error msg looks better formatted

Weird. I agree that `Invalid` is a better error code here, but maybe for now just use `BadRequest` and file a bug to look into how to change it back to `Invalid`? I suspect that there's something we can add to the webhook status message that we return that would format it better, but it's probably not worth looking into right now.

incubator/hnc/internal/validators/object.go
GinnyJI: done

GinnyJI:
> Also you need to identify the `treeSelect` annotation somewhere or users won't know which field is causing the problem.

done

GinnyJI:
> > Yeah if we use `metav1.StatusReasonBadRequest` the error msg looks better formatted
> 
> Weird. I agree that `Invalid` is a better error code here, but maybe for now just use `BadRequest` and file a bug to look into how to change it back to `Invalid`? I suspect that there's something we can add to the webhook status message that we return that would format it better, but it's probably not worth looking into right now.

Changed both of the errors for `select` and `treeSelect`, and filed the bug at https://github.com/kubernetes-sigs/multi-tenancy/issues/1218

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -18,18 +20,28 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
...
 func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
-	annot := inst.GetAnnotations()
-	treeSelectorStr, ok := annot[api.AnnotationTreeSelector]
-	if !ok {
+	treeSelectorStr := GetTreeSelectorAnnotation(inst)
+	if treeSelectorStr == "" {
 		return nil, nil
 	}
+	// check if it's a valid namespace name
+	errStrs := validation.IsDNS1123Label(treeSelectorStr)
```
adrianludwin, line 40: `a,b` is a valid tree selector string but is not a valid DNS label

GinnyJI: Done. The e2e test also caught this

GinnyJI:
tracked at #1123

adrianludwin:
Two comments about the latest error message (newlines added):

```
Error from server (BadRequest): admission webhook "objects.hnc.x-k8s.io" denied the request:
Invalid HNC treeSelect value:
1234567890123456789012345678901234567890123456789012345678901234567890
is not a valid namespace name, must be no more than 63 characters

```

1. Nothing here points the user to the annotation that caused the problem
2. What does this message look like if there are multiple validation errors? (Have you tested this?)

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -18,29 +20,47 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
...
 	selectorStr := ""
 	for i, str := range strs {
+		// check if it's a valid namespace name
+		if str[0] == '!' {
+			errStrs = validation.IsDNS1123Label(str[1:])
+		} else {
+			errStrs = validation.IsDNS1123Label(str)
+		}
+		if len(errStrs) != 0 {
+			return nil, errors.New(str + " is not a valid namespace name, " + strings.Join(errStrs, ". "))
```
adrianludwin, line 51: `str` here could include the exclamation mark, because you didn't actually strip it out on line 46. Did you want to include the exclamation mark here?

adrianludwin:
This change will also need tests in `internal/validators` before it goes in

incubator/hnc/internal/pkg/selectors/selectors.go

GinnyJI:
> Two comments about the latest error message (newlines added):
> 
> ```
> Error from server (BadRequest): admission webhook "objects.hnc.x-k8s.io" denied the request:
> Invalid HNC treeSelect value:
> 1234567890123456789012345678901234567890123456789012345678901234567890
> is not a valid namespace name, must be no more than 63 characters
> ```
> 
> 1. Nothing here points the user to the annotation that caused the problem
> 2. What does this message look like if there are multiple validation errors? (Have you tested this?)

Forgot to update the PR msg... updated now and it contains `treeSelect`

I wrote a brief comment in the code on what it will look like with multiple errors

adrianludwin:
> Forgot to update the PR msg... updated now and it contains `treeSelect`

Remember that one of your use cases is a user who's applying a YAML file they didn't write via `kubectl apply -f foo.yaml` and doesn't know anything about HNC. How would they know where to look if they just see the word "treeSelect"?

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -18,33 +19,62 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
...
+		if i < len(segs)-1 {
 			selectorStr = selectorStr + ", "
 		}
 	}
 	treeSelector, err := getSelectorFromString(selectorStr)
 	if err != nil {
-		return nil, fmt.Errorf("while parsing %q: %w", api.AnnotationTreeSelector, err)
+		// In theory this should never happen because we already checked DNS label before.
+		// If this happens, it's more likely that we have a bug in our code
+		return nil, fmt.Errorf("Internal error while parsing %q: %w", api.AnnotationTreeSelector, err)
```
adrianludwin, line 56: Error usually start with a lowercase letter since they'll often get wrapped, like you did here :)

Log message, by contrast, start with upper case letters because they can't be wrapped (same with user messages).

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -18,33 +19,62 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
...
 	}
 	return treeSelector, nil
 }
 
+func validateNamespaceName(seg string) error {
```
adrianludwin, line 61: You're not really passing a namespace to this function, you're passing a treeSelector segment. So maybe `validateTreeSelectorSegment`?

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -18,33 +19,62 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
...
+	if seg[0] == '!' {
+		ns = seg[1:]
+	} else {
+		ns = seg
+	}
+	errStrs := validation.IsDNS1123Label(ns)
+	if len(errStrs) != 0 {
+		// If IsDNS1123Label() returns multiple errors, it will look like:
+		// ""ns" is not a valid namespace name, err1; err2
+		return fmt.Errorf("%q is not a valid namespace name, %s", ns, strings.Join(errStrs, "; "))
```
adrianludwin, line 73: > ""ns" is not a valid namespace name, err1; err2

I assume that first double quote in the comment is just a typo?

This would look better IMO with a full colon (:), which is used to create the start of a list:

`"ns" is not a valid namespace name: error1; error 2`

incubator/hnc/internal/validators/object_test.go
```diff
@@ -146,6 +146,7 @@ func TestUserChanges(t *testing.T) {
 	f := forest.NewForest()
 	o := &Object{Forest: f}
 	l := zap.Logger(false)
+	invalidSelStr := "1234567890123456789012345678901234567890123456789012345678901234567890"
```
adrianludwin, line 149: This only gets used once so I'd recommend inlining it into the test.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +309,20 @@ func TestUserChanges(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "Deny creation of object with invalid treeSelect annotation",
```
adrianludwin, line 313: Probably also a good idea to have an "allow" test.

incubator/hnc/internal/validators/object_test.go
GinnyJI: I'm thinking of using it in the selector test as well

adrianludwin: Ah, I see. I think it might be better if each test had its own, different, bad selector :) E.g. "a b c" is a bad namespace name _and_ a bad selector.

Constants like this make the tests harder to read, because you can't just look at the test definition to know what's being tested. Constants can add clarity in same cases (e.g. you don't really have to look inside RBACGroup to know what's there) but "invalidSelector" could have anything in there, so it makes the test a bit harder to read IMO.

I know I sometimes say "please use constants!" and it's weird that I'm suggesting _not_ using them here, but I hope that explains why!

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Good to know! Corrected

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Done

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Done. Yea it was a typo

incubator/hnc/internal/validators/object_test.go
GinnyJI: Added

incubator/hnc/internal/validators/object_test.go
GinnyJI: I see what you mean. Yeah I agree, changed the test string here

GinnyJI:
> > Forgot to update the PR msg... updated now and it contains `treeSelect`
> 
> Remember that one of your use cases is a user who's applying a YAML file they didn't write via `kubectl apply -f foo.yaml` and doesn't know anything about HNC. How would they know where to look if they just see the word "treeSelect"?

ah I see, updated to use the full key

incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +308,79 @@ func TestUserChanges(t *testing.T) {
...
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector:     "foo",
+						api.AnnotationTreeSelector: "$bar",
+					},
+				},
+			},
+		},
+	}, {
+		name: "Allow creation of object with valid selector and treeSelect annotation",
```
adrianludwin, line 370: I thought we were going to disallow this? But if we haven't implemented it yet, that's fine for this PR

adrianludwin (commented):
Nice work!

/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/internal/validators/object_test.go
GinnyJI: Ah right. When it's implemented I'll update this test

yiqigao217: Just confirming that we decided to only allow one type of selector at a time and cannot mix any two, right?

adrianludwin: Yup!

PR #1219, opened 2020-10-22
Change HNC Config "types" to "resources"

adrianludwin:
We used the term "type" to represent GVKs, but now that we no longer
care about versions and have changed from kinds to resources, the term
"resources" now seems more appropriate. This commit makes that change.

This change also makes the `group` field optional for builtin types,
such as secrets. Finally, it changes the `hns config set-type` command
to `hns config set-resource` and changes the flags so that the ordered
arg is the resource name, and the propagation mode has an explicit
`--mode` flag (the opposite of what it used to be) since this reads more
naturally.

Also silenced two particularly noisy reconciler messages.

Tested: updated the HNCConfig conversion tests and they all passed.
Updated e2e tests and they passed too.  Played a bit with the HNC config
by hand and confirmed that everything looks good.


incubator/hnc/api/v1alpha2/hnc_config.go
```diff
@@ -55,18 +55,18 @@ const (
...
-// TypeSynchronizationSpec defines the desired synchronization state of a
+// ResourceSpec defines the desired synchronization state of a
 // specific resource.
-type TypeSynchronizationSpec struct {
+type ResourceSpec struct {
 	// Group of the resource defined below. This is used to unambiguously identify
-	// the resource.
-	Group string `json:"group"`
+	// the resource. It may be omitted for core resources (e.g. "secrets").
+	Group string `json:"group,omitempty"`
```
yiqigao217, line 69: Nice, it will turn, e.g.
```
spec:
  group: ""
  resource: secrets
  mode: Propagate
```
into
```
spec:
  resource: secrets
  mode: Propagate
```
right?

yiqigao217 (commented):
/lgtm
/hold

Please feel free to remove the hold if there's no need for a second eye since both of us already agreed on the field renaming.

incubator/hnc/api/v1alpha2/hnc_config.go
adrianludwin: Yup, exactly

adrianludwin:
I think this is fine to go in now, we've already discussed it and it's a fairly straightforward change.

/hold cancel

rjbez17:
Sorry for the delay, but lgtm as well.

PR #1221, opened 2020-10-22
Add test for invalid selector

GinnyJI:
Tested: go test ./internal/validators

### DIFF
--- a/incubator/hnc/internal/validators/object_test.go
+++ b/incubator/hnc/internal/validators/object_test.go
@@ -308,6 +308,34 @@ func TestUserChanges(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "Deny creation of object with invalid select annotation",
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector: "$foo",
+					},
+				},
+			},
+		},
+	}, {
+		name: "Allow creation of object with valid select annotation",
+		fail: false,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector: "foo",
+					},
+				},
+			},
+		},
 	}}
 
 	for _, tc := range tests {


incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +309,20 @@ func TestUserChanges(t *testing.T) {
...
+	}, {
+		name: "Deny creation of object with invalid select annotation",
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector: invalidSelStr,
```
adrianludwin, line 321: I feel like the string "a b c" would also be a good illegal labelSelector

incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +309,20 @@ func TestUserChanges(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "Deny creation of object with invalid select annotation",
```
adrianludwin, line 313: Should have an "allow" as well.

Also, how about the scenario where there are multiple selectors (eg `.../treeSelector` and `.../selector`) on the same object?

adrianludwin (commented):
Try testing your change with `go test -v ./internal/validators/...` and verifying that the error message that's returned by the validator is what you expect, and then document this fact in the commit message. We should probably start testing those messages as well but for now it's useful to have a human look at them at least once!

incubator/hnc/internal/validators/object_test.go
GinnyJI: Changed to `$foo`

incubator/hnc/internal/validators/object_test.go
GinnyJI: Added "allow" test. I'll add the multiple selector case in the other PR (as /treeSelect is not validated here yet)

incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +308,34 @@ func TestUserChanges(t *testing.T) {
...
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector: "$foo" + api.LabelTreeDepthSuffix,
```
adrianludwin, line 320: I don't think you _need_ to add the suffix here - `$foo` on its own is definitely not a legal label key :)

incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +308,34 @@ func TestUserChanges(t *testing.T) {
...
+	}, {
+		name: "Allow creation of object with valid select annotation",
+		fail: false,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector: "foo" + api.LabelTreeDepthSuffix,
```
adrianludwin, line 334: Ditto here, I'd leave out the tree suffix

adrianludwin (commented):
lgtm after one last nit

incubator/hnc/internal/validators/object_test.go
GinnyJI: Yes, you are right. I was trying to simulate the users' scenario more closely, because we are always expecting the user to input the full string. Do you think it's better without this suffix?

incubator/hnc/internal/validators/object_test.go
GinnyJI: for the same reason replied above

incubator/hnc/internal/validators/object_test.go
adrianludwin: Since `treeSelector` works so well for the HNC tree labels, I'd actually expect that people using `selector` would be searching for some _other_ kind of string, so I don't think it's necessary to check the HNC one.

But more to the point, our goal here isn't really to test the K8s label selector parsing code - it's that we do something sensible with the error that gets returned. As a result, making the test as simple as possible has a large benefit. Imagine being the next person to modify this test or debug a failure - if you see that input is `foo.tree.hnc.x-k8s.io/depth`, you might spend time looking into whether there's something about _that string_ that's causing a problem because "obviously" that complicated string must be what the test is testing, right? (Wrong!) But if the input string is just "foo" it's much easier to see that there's nothing special about the string.

This is what we mean when we say that most tests should be "minimal" - find the smallest, simplest input that could replicate a problem (or solution) so that if/when the test breaks, it's easier to debug.

GinnyJI: I see, agreed! Done

incubator/hnc/internal/validators/object_test.go
GinnyJI: Done

adrianludwin (commented):
/lgtm
/approve

This is a test-only change so I'm approving it directly.

GinnyJI:
tracked at #1123

PR #1225, opened 2020-10-23
Add validator for none selector

GinnyJI:
Tested: go test -v ./internal/validator

```
=== RUN   TestUserChanges/Deny_creation_of_object_with_invalid_noneSelect_annotation
    object_test.go:387: Got code 400, reason "BadRequest", message "invalid propagate.hnc.x-k8s.io/none value, strconv.ParseBool: parsing \"foo\": invalid syntax"
=== RUN   TestUserChanges/Allow_creation_of_object_with_valid_noneSelect_annotation
    object_test.go:387: Got code 0, reason "", message "source object"
```

### DIFF
--- a/incubator/hnc/internal/pkg/selectors/selectors.go
+++ b/incubator/hnc/internal/pkg/selectors/selectors.go
@@ -4,6 +4,7 @@ package selectors
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -22,6 +23,11 @@ func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
 func GetTreeSelectorAnnotation(inst *unstructured.Unstructured) string {
 	annot := inst.GetAnnotations()
 	return annot[api.AnnotationTreeSelector]
+}
+
+func GetNoneSelectorAnnotation(inst *unstructured.Unstructured) string {
+	annot := inst.GetAnnotations()
+	return annot[api.AnnotationNoneSelector]
 }
 
 // GetTreeSelector is similar to a regular selector, except that it adds the LabelTreeDepthSuffix to every string
@@ -96,4 +102,25 @@ func getSelectorFromString(str string) (labels.Selector, error) {
 		return nil, err
 	}
 	return selector, nil
+}
+
+// GetNoneSelector returns true indicates that user do not want this object to be propagated
+func GetNoneSelector(inst *unstructured.Unstructured) (bool, error) {
+	noneSelectorStr := GetNoneSelectorAnnotation(inst)
+	// Empty string is treated as 'false'. In other selector cases, the empty string is auto converted to
+	// a selector that matches everything.
+	if noneSelectorStr == "" {
+		return false, nil
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// Returning false here in accordance to the Go standard
+		return false,
+			fmt.Errorf("invalid %s value: %w",
+				api.AnnotationNoneSelector,
+				err,
+			)
+
+	}
+	return noneSelector, nil
 }
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"strconv"
 	"sync"
 
 	"github.com/go-logr/logr"
@@ -658,15 +657,19 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 // - Objects have a selector that doesn't match the destination namespace
 // - Service Account token secrets
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
-	selector, err := selectors.GetSelector(inst)
-	if err != nil {
-		log.Error(err, "Invalid value")
+	selector, errSelector := selectors.GetSelector(inst)
+	// TODO: generate events for the errors
+	if errSelector != nil {
+		log.Error(errSelector, "Cannot propagate")
 	}
-	treeSelector, err := selectors.GetTreeSelector(inst)
-	if err != nil {
-		log.Error(err, "Invalid value")
+	treeSelector, errTreeSelector := selectors.GetTreeSelector(inst)
+	if errTreeSelector != nil {
+		log.Error(errTreeSelector, "Cannot propagate")
 	}
-	noneSelector := r.getNoneSelector(log, inst)
+	noneSelector, errNoneSelector := selectors.GetNoneSelector(inst)
+	if errNoneSelector != nil {
+		log.Error(errNoneSelector, "Cannot propagate")
+	}
 	nsLabels := r.Forest.Get(dst).GetLabels()
 
 	switch {
@@ -677,6 +680,10 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 
 	// Object with nonempty finalizer list is not propagated
 	case hasFinalizers(inst):
+		return false
+
+	// Invalid selectors
+	case errSelector != nil || errTreeSelector != nil || errNoneSelector != nil:
 		return false
 
 	// None selector is set to true
@@ -708,29 +715,6 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 		// Everything else is propagated
 		return true
 	}
-}
-
-// getNoneSelector returns true indicates that user do not want this object to be propagated
-func (r *ObjectReconciler) getNoneSelector(log logr.Logger, inst *unstructured.Unstructured) bool {
-	annot := inst.GetAnnotations()
-	noneSelectorStr, ok := annot[api.AnnotationNoneSelector]
-	if !ok {
-		return false
-	}
-	// Empty string is treated as 'false'. In other selector cases, the empty string is auto converted to
-	// a selector that matches everything.
-	if noneSelectorStr == "" {
-		return false
-	}
-	noneSelector, err := strconv.ParseBool(noneSelectorStr)
-	if err != nil {
-		// TODO: surface the error
-		log.Error(err, "Invalid noneSelector value", "It should be either true or false, but got", noneSelectorStr)
-		// When the user put an invalid noneSelector, we choose not to propagate this object to any child
-		// namespace to protect any object in the child namespaces to be overwritten
-		return true
-	}
-	return noneSelector
 }
 
 // recordPropagatedObject records the fact that this object has been propagated, so we can report
--- a/incubator/hnc/internal/reconcilers/object_test.go
+++ b/incubator/hnc/internal/reconcilers/object_test.go
@@ -58,7 +58,7 @@ var _ = Describe("Exceptions", func() {
 			// we choose not to propagate this object to any child namespace to protect any object in the child
 			// namespaces to be overwritten. Same for treeSelectors and noneSelector.
 			name:     "not propagate to any namespace with a bad selector",
-			selector: "random",
+			selector: "$foo",
 			want:     []string{},
 			notWant:  []string{c1, c2, c3},
 		}, {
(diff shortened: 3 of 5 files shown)


GinnyJI:
tracked at #1123

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -67,3 +73,24 @@ func getSelectorFromString(str string) (labels.Selector, error) {
 	}
 	return selector, nil
 }
+
+// GetNoneSelector returns true indicates that user do not want this object to be propagated
```
yiqigao217, line 77: Maybe add 'or the selector cannot parse so HNC default it to not propagate`?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -666,7 +665,10 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 	if err != nil {
 		log.Error(err, "Invalid value")
 	}
-	noneSelector := r.getNoneSelector(log, inst)
+	noneSelector, err := selectors.GetNoneSelector(inst)
+	if err != nil {
+		log.Error(err, "Invalid value")
```
yiqigao217, line 670: Add surface error TODO with the issue link (if it's not linked nearby).

incubator/hnc/internal/pkg/selectors/selectors.go
adrianludwin: When I reviewed this in Ginny's repo, I suggested that if we return an error, the boolean value should be ignored, to be consistent with most other Go functions.

incubator/hnc/internal/reconcilers/object.go
adrianludwin: I think the log is sufficient surfacing here? The validator talks directly to the user, but the reconciler can't.

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Yeah I have updated according to Adrian's comment

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -33,6 +39,9 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
 	strs := strings.Split(treeSelectorStr, ",")
 	selectorStr := ""
 	for i, str := range strs {
+		if str == "" {
+			continue
+		}
```
adrianludwin, line 44: Are you going to remember why this is here in six months? :)

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -660,13 +659,16 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
 	selector, err := selectors.GetSelector(inst)
 	if err != nil {
-		log.Error(err, "Invalid value")
+		log.Error(err, "Cannot propagate")
 	}
```
adrianludwin, line 663: Why not `return false` immediately?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -660,13 +659,16 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
...
-	treeSelector, err := selectors.GetTreeSelector(inst)
-	if err != nil {
-		log.Error(err, "Invalid value")
+	treeSelector, err2 := selectors.GetTreeSelector(inst)
+	if err2 != nil {
+		log.Error(err2, "Cannot propagate")
+	}
```
adrianludwin, line 667: If you return early, you don't need to keep multiple `err` vars around.

incubator/hnc/internal/pkg/selectors/selectors.go
yiqigao217: sg

incubator/hnc/internal/reconcilers/object.go
yiqigao217: Do we still need to generate an event here?

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -67,3 +76,26 @@ func getSelectorFromString(str string) (labels.Selector, error) {
...
+	// a selector that matches everything.
+	if noneSelectorStr == "" {
+		return false, nil
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// When the user put an invalid noneSelector, we choose not to propagate this object to any child
+		// namespace to protect any object in the child namespaces to be overwritten
+		return true,
+			fmt.Errorf("invalid %q value: %q, %w",
```
adrianludwin, line 93: Are you sure that `ParseBool` doesn't include the bad string? If you are, then I'd suggest: "invalid %s value %q: %w"

What you have now will print something like:

`invalid "propagate.hnc.x-k8s.io/none" value: "ture", not a boolean string`

The initial double quotes aren't necessary because the annotation string won't be mistaken for a regular English word, and having a comma between the bad value and the error message isn't a strong separation between the different parts of the error. Instead, I'm suggesting: 

`invalid propagate.hnc.x-k8s.io/none value "ture": not a boolean string`

Now we're reserving the double quotes for the only part that could potentially be mistaken as an English word, and there's a strong separation between the error portions. Make sense?

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -67,3 +76,26 @@ func getSelectorFromString(str string) (labels.Selector, error) {
...
+	// Empty string is treated as 'false'. In other selector cases, the empty string is auto converted to
+	// a selector that matches everything.
+	if noneSelectorStr == "" {
+		return false, nil
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// When the user put an invalid noneSelector, we choose not to propagate this object to any child
+		// namespace to protect any object in the child namespaces to be overwritten
+		return true,
```
adrianludwin, line 92: I'd still go with `return false` here

incubator/hnc/internal/reconcilers/object.go
adrianludwin: Oh good point! Yes, we should.

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Added comment here

incubator/hnc/internal/reconcilers/object.go
GinnyJI: In the case there are multiple errors, I _think_ it will be helpful for the developers to see all the errors in the log, wdyt?

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Yeah I see, updated the comments. `ParseBool` does include the bad string so I removed it from our log

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -33,6 +39,12 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
 	strs := strings.Split(treeSelectorStr, ",")
 	selectorStr := ""
 	for i, str := range strs {
+		// There are cases that ok==true and treeSelectorStr is an empty string, and this
+		// will lead to selectorStr==api.LabelTreeDepthSuffix, which cannot be parsed as
+		// a valid label selector. So we need to skip the empty strings
+		if str == "" {
+			continue
+		}
```
adrianludwin, line 47: Putting the check here implies that any one part of the string could legitimately be empty, such as the string `,,,`, which probably shouldn't be legal.

How about putting this check on line 35/36 instead?

```
treeSelectorStr := annot[api.AnnotationTreeSelector]
if treeSelectorStr == "" {
  return nil, nil
}
```

You only need to the `x, ok := foo[bar]` form if you need to distinguish between `x` having the zero value (in the case of a string, `""`) and being missing completely. But in our case, we don't care which it is, so you can use the simpler `x := foo[bar]` form.

adrianludwin:
> Tested: go test -v ./internal/validator
> 
> ```
> === RUN   TestUserChanges/Deny_creation_of_object_with_invalid_noneSelect_annotation
>     object_test.go:387: Got code 400, reason "BadRequest", message "invalid propagate.hnc.x-k8s.io/none value, strconv.ParseBool: parsing \"foo\": invalid syntax"
> === RUN   TestUserChanges/Allow_creation_of_object_with_valid_noneSelect_annotation
>     object_test.go:387: Got code 0, reason "", message "source object"
> ```

It's worth doing at least check by hand

incubator/hnc/internal/pkg/selectors/selectors.go

incubator/hnc/internal/reconcilers/object.go
GinnyJI: updated to more descriptive err names

incubator/hnc/internal/reconcilers/object.go
GinnyJI: updated according to our discussion offline

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -67,3 +76,24 @@ func getSelectorFromString(str string) (labels.Selector, error) {
...
+	// Empty string is treated as 'false'. In other selector cases, the empty string is auto converted to
+	// a selector that matches everything.
+	if noneSelectorStr == "" {
+		return false, nil
+	}
+	noneSelector, err := strconv.ParseBool(noneSelectorStr)
+	if err != nil {
+		// Returning false here in accordance to the Go standard
+		return false,
+			fmt.Errorf("invalid %s value, %w",
```
adrianludwin, line 92: Please use a colon instead of a comma - wrapped errors should ~always be after a colon.

`invalid %s value: %w`

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -658,15 +657,21 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
...
-		log.Error(err, "Invalid value")
+	treeSelector, errTreeSelector := selectors.GetTreeSelector(inst)
+	if errTreeSelector != nil {
+		log.Error(errTreeSelector, "Cannot propagate")
+	}
+	noneSelector, errNoneSelector := selectors.GetNoneSelector(inst)
+	if errNoneSelector != nil {
+		log.Error(errNoneSelector, "Cannot propagate")
+		// We set noneSelector to true to protect any object in the child namespaces to be overwritten
+		noneSelector = true
```
adrianludwin, line 673: I don't think we need this anymore? Because if `errNoneSelector` is non-nil, we'll `return false` before we ever check the value of this.

adrianludwin (commented):
lgtm after two last nits

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: done

incubator/hnc/internal/reconcilers/object.go
yiqigao217: The only concern about adding it to metrics is that we need to unset it when it's gone.

What about object errors then? We have them as events now but not in the metrics yet, should we also add that? Maybe not for object events, since it's hard to unset them in metrics.

adrianludwin: Ok let's skip the metric - just an event will be fine.

I think events show up in Stackdriver as metrics and you can take the first derivative to be notified when something new happens.

yiqigao217: sg

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

Issue #1222, opened 2020-10-22
HNC: Add NamespaceConditions to `kubectl hns config describe` too


yiqigao217:
Currently `kubectl hns config describe`  only shows HNCConfig's own conditions `status.conditions`, but not the `status.namespaceConditions`, e.g.
`kubectl get hncconfigurations.hnc.x-k8s.io config -oyaml` returns
```
status:
  namespaceConditions:
  - namespaces:
    - b
    - c
    - d
    reason: ParentMissing
    type: ActivitiesHalted
```
while they are not included in:
```
$ kubectl hns config describe
Synchronized types:
* Propagating: secrets (/v1)
* Propagating: rolebindings (rbac.authorization.k8s.io/v1)
* Propagating: roles (rbac.authorization.k8s.io/v1)

Conditions:
```

yiqigao217:
Obsolete since we have removed `namespaceConditions` field and set them as conditions with `namespaceCondition` type. See #1227

PR #1227, opened 2020-10-23
Convert HNCConfig namespaceConditions into conditions

### LINKED ISSUE #1226: HNC: Limit conditions to list maximum 3 in v1alpha2
Part of #868 
See doc http://bit.ly/hnc_v1alpha2

There are 3 "conditions" fields in the HNC CRDs, `hc.status.conditions`, `hncconfig.status.conditions` and `hncconfig.status.namespaceConditions`.

I believe we only need to limit the length of the affected namespaces in `hncconfig.status.namespaceConditions`, since the other two only lists the local conditions that makes sense to list all. CMIIW

As for `hncconfig.status.namespaceConditions`, there's no good place to show in the CR that we only list first 3 affected namespaces if there are more. To make it clearer, I will
1. add a `numAffectedNamespaces` field to show the total number of affected namespaces;
2. add field description on `namespaces` in the CRD schema - "Namespaces is the list of affected namespaces in alphabetical order. Maximum of 3 namespaces will be listed if there are more than 3."

Thus, the `hncconfig.status.conditions` and `hncconfig.status.namespaceConditions` could look like this, with full list of `conditions` (possibly be more than 3) and maximum 3  in `namespaces` field in one namespace condition together with a total number of affected namespaces:
```
kind: HNCConfiguration
status:
  conditions:
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: Resource "wrongtype1s" not found
    reason: ResourceNotFound
    status: "True"
    type: BadConfiguration
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: Resource "wrongtype2s" not found
    reason: ResourceNotFound
    status: "True"
    type: BadConfiguration
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: Resource "wrongtype3s" not found
    reason: ResourceNotFound
    status: "True"
    type: BadConfiguration
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: Resource "wrongtype4s" not found
    reason: ResourceNotFound
    status: "True"
    type: BadConfiguration
  namespaceConditions:
  - numAffectedNamespaces: 4
    namespaces:
    - c
    - d
    - e
    reason: ParentMissing
    type: ActivitiesHalted
```

yiqigao217:
Convert HNCConfig namespaceConditions into conditions with
'NamespaceCondition' as the condition type and namespace condition types
as reasons, e.g. 'ActivitiesHalted'.

Only list the first three affected namespaces in the message with a
total number of affected namespaces. Also include the namespace
condition reason in the message.

Create a new conversion test case with four affected namespaces.

Tested manually on GKE and by 'make test' and 'make test-conversion'.

Part of #868 
Fixes #1226


adrianludwin:
High level comment: maybe we should get rid of `namespaceConditions` and just have this as a regular condition? Eg something like type=NamespaceCondition, reason=<the namespace condition type, like ActivitiesHalted> and then the message just indicates the first N namespaces affected by that condition and why?

yiqigao217:
> High level comment: maybe we should get rid of `namespaceConditions` and just have this as a regular condition? Eg something like type=NamespaceCondition, reason=<the namespace condition type, like ActivitiesHalted> and then the message just indicates the first N namespaces affected by that condition and why?

sgtm and they will show up in `k hns config describe` too (see #1222).

yiqigao217:
The message of 4, 3, 2, 1 affected namespace would look like this:
```
status:
  conditions:
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: 4 namespaces "b", "c", "d" ... are affected by "ParentMissing"
    reason: ActivitiesHalted
    status: "True"
    type: NamespaceCondition

status:
  conditions:
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: 3 namespaces "c", "d", "e" are affected by "ParentMissing"
    reason: ActivitiesHalted
    status: "True"
    type: NamespaceCondition

status:
  conditions:
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: 2 namespaces "d", "e" are affected by "ParentMissing"
    reason: ActivitiesHalted
    status: "True"
    type: NamespaceCondition

status:
  conditions:
  - lastTransitionTime: "1970-01-01T00:00:00Z"
    message: Namespaces "e" is affected by "ParentMissing"
    reason: ActivitiesHalted
    status: "True"
    type: NamespaceCondition
```

incubator/hnc/internal/reconcilers/hnc_config.go
```diff
@@ -469,20 +469,26 @@ func (r *ConfigReconciler) loadNamespaceConditions(inst *api.HNCConfiguration) {
...
 			sort.Strings(nsnms)
-			status = append(status, api.ConditionAndAffectedNamespaces{
-				Type:       tp,
-				Reason:     reason,
-				Namespaces: nsnms,
-			})
+			l := len(nsnms)
+			// Message for 2 or 3 affected namespaces, e.g.
+			// 2 namespaces "d", "e" are affected by "ParentMissing"
+			msg := fmt.Sprintf("%v namespaces \"%s\" are affected by %q", l, strings.Join(nsnms, "\", \""), reason)
```
adrianludwin, line 478: I avoid using `%v` in a user-facing message - use a specific one like `%d` instead

adrianludwin (commented):
One tiny nit but otherwise...

/lgtm
/approve
/hold
/assign @rjbez17

PR #1230, opened 2020-10-23
Reenable test for #1155

### LINKED ISSUE #1155: HNC: Incorrect default if delete and recreate a namespace with "allowCascadingDelete: true"
This issue happens on the current master.
How to reproduce:
```
$ kubectl create ns a
namespace/a created

$ kubectl hns create b -n a
Successfully created "b" subnamespace anchor in "a" namespace

$ kubectl hns set a -a
Allowing cascading deletion on 'a'
Succesfully updated 1 property of the hierarchical configuration of a

$ kubectl delete ns a
namespace "a" deleted
# namespace "b" is also cascading deleted in this step

############ The issue starts from here #############
$ kubectl create ns a
namespace/a created

$ k get hierarchyconfigurations.hnc.x-k8s.io -n c -oyaml
# You will see "allowCascadingDelete: true", which should default to "false" instead!
```

adrianludwin:
I wasn't able to get the original problem to reproduce itself, but I
added a log message and improve the test so that we can catch #1155 if
it ever reoccurs.

Tested: the newly enabled test passes with HNC as-is; if I break HNC to
manually re-enable allowCascadingDeletion, then the test fails as
expected. I also manually verified that the new log message is displayed
as expected.

Fixed #1155

### DIFF
--- a/incubator/hnc/internal/forest/namespace.go
+++ b/incubator/hnc/internal/forest/namespace.go
@@ -111,9 +111,14 @@ func (ns *Namespace) clean() {
 	delete(ns.forest.namespaces, ns.name)
 }
 
-// UpdateAllowCascadingDeletion updates if this namespace allows cascading deletion.
-func (ns *Namespace) UpdateAllowCascadingDeletion(acd bool) {
+// UpdateAllowCascadingDeletion updates if this namespace allows cascading deletion. It returns true
+// if the value has changed, false otherwise.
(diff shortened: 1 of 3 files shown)


adrianludwin:
/assign @yiqigao217 

I don't think @rjbez17 needs to approve this.

yiqigao217 (commented):
/lgtm

I was not reproduce it either on the new cluster. It's probably another case where the cluster is corrupted. Anyway the changes lgtm.

PR #1231, opened 2020-10-26
correct ipv6 domain name format

weiling61:
Ipv4 and ipv6 URL domain name is different in format. Make vc-manager compatible with ipv6 only cluster

### DIFF
--- a/incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
+++ b/incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
@@ -21,7 +21,6 @@ import (
 	"crypto/rsa"
 	"errors"
 	"fmt"
-	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


Fei-Guo (commented):
I feel it will be better to change generateKubeconfigUseCertAndKey since it is only the url that needs the '[]' format.

incubator/virtualcluster/pkg/controller/virtualcluster/master_provisioner_native.go
```diff
@@ -304,23 +305,28 @@ func (mpn *MasterProvisionerNative) createPKI(vc *tenancyv1alpha1.VirtualCluster
 	}
 	caGroup.APIServer = apiserverCAPair
 
-	finalAPIAddress := apiserverDomain
+	finalAPIDomain := apiserverDomain
 	if clusterIP != "" {
-		finalAPIAddress = clusterIP
+		if strings.Contains(clusterIP, ":") {
```
christopherhein, line 310: We probably want to make this slightly more robust as well and use something like:

```
import "net"

// ...

dr := net.ParseIP("2001:db8::68")
addr := net.ParseIP("192.0.2.1")

if addr.To4() != nil {
	// is ipv4
} else {
	// is ipv6
}
///
```

weiling61: corrected, thanks

PR #1234, opened 2020-10-26
Cap the backoff for the object reconciler to 10s

### LINKED ISSUE #1154: HNC: see why we can't propagate a RB without a role, and why this doesn't get resolved
https://www.youtube.com/watch?v=zAtaJ0x-ZwY&feature=youtu.be

adrianludwin:
See issue #1154. In some cases, objects cannot be propagated due to
conflicts with other objects - in the case of a rolebinding, because its
role doesn't already exist. By default, controller-runtime will retry
errors in such circumstances with an exponential backoff with a 1000s
cap, which could lead users to believe that HNC is broken and will never
propagate their objects even after the problem is fixed.

This change caps the backoff at 10s, which is probably high enough that
it won't flood HNC with retries and low enough that the problem will
likely be fixed within the next one or two checks.

Tested: create a rolebinding without a role and watched the logs.
Without this change, observed that the time between attempts doubled
continuously (I saw it get to about a minute) before I created the role,
and verified that the RB was _eventually_ propagated. With this change,
I observed that the time between events increased until it hit 5s and
then saw a new event every 5s, and after the role was create, the RB was
propagated very soon after that. (I later changed the timeout to 10s
during PR review but didn't retest).

Fixed #1154

### DIFF
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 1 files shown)


yiqigao217 (commented):
/lgtm
/hold

Feel free to remove the hold anyone!

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -764,6 +766,18 @@ func (r *ObjectReconciler) SetupWithManager(mgr ctrl.Manager, maxReconciles int)
...
+		// that do not directly cause reconciliations when they're modified - see, e.g.,
+		// https://github.com/kubernetes-sigs/multi-tenancy/issues/1154. To address this, replace the
+		// default exponential backoff with one with a 5s cap.
+		//
+		// I chose five seconds since I feel like that's about how much time it would take you to check
+		// to see if something's wrong, realize it hasn't been propagated, and then try again. This
+		// should be a rare enough case that I very much doubt that it would have much material impact
+		// on HNC, unless something's causing _every_ object propagation to fail (in which case the
+		// global rate limiter will kick in to protect the cluster).
+		RateLimiter: workqueue.NewItemExponentialFailureRateLimiter(250*time.Millisecond, 5*time.Second),
```
rjbez17, line 780: My only insight here is that during times of heavey load on a cluster, especially etcd, requests can take quite a while. I believe the etcd request timeout is 10 seconds (which we've been hit by a few times on our larger clusters). I also think the api-server timeout is something like 30 seconds?

So while 5 seconds is more than enough for a healthy cluster, it could cause further issues on an unhealthy cluster because we'd timeout and retry often-ish in the context of a slow cluster. 

I'm not sure it's worth changing now, but maybe something we'd come back to in the future.

rjbez17:
@adrianludwin and I discussed offline to consider changing the timeout to 10s. Regardless of 10s or 5s this lgtm. Remove hold whenever.

incubator/hnc/internal/reconcilers/object.go
adrianludwin: Changed to 10s

adrianludwin:
Changed to 10s, may I please get another lgtm?

/hold cancel

PR #1236, opened 2020-10-26
Don't block object deletion in dying namespaces

### LINKED ISSUE #1214: HNC: if a propagated object isn't deleted, there's no way to delete it later
The validating admission webhook won't let you delete anything with `inherited-from` and also won't let you delete that label. However, if the original object is now gone and the object reconciler didn't delete propagated copy, that makes it impossible to delete the object, and hence the namespace.

We should probably always allow object deletions if the namespace is being deleted.

adrianludwin:
See issue #1214. Our e2e tests were occasionally dying when a namespace
got messed up while propagated objects were in it; I *think* this is
because they got into an ActivityHalted state (so propagated objects
weren't cleaned up) but then the object validator would block deletion.
This caused the tests to hang indefinitely, and the objects couldn't
(easily) be removed without disbling the validator.

This change lets the validators know when the namespace is being
deleted, and unblocks deletion of all propagated objects.

Tested: a new e2e test hangs indefinitely without the fix and completes
successfully with it.

Fixes #1214

### DIFF
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -164,6 +164,16 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
 		return deny(metav1.StatusReasonForbidden, "Cannot create objects with the label \""+api.LabelInheritedFrom+"\"")
 
 	case admissionv1beta1.Delete:
+		if o.isDeletingNS(oldInst) {
+			// There are few things more irritating in (K8s) life than having some stupid controller stop
+			// your namespace from being deleted. If there's an object in here and we've decided that the
+			// namespace should be deleted, then don't block anything!
+			//
+			// It's probably slightly safer to create a K8s client, actually load the namespace and check
+			// its deletion timestamp here. But it's much easier to just record this information in the
+			// namespace/HC reconciler and read it here.
+			return allow("allowing deletion of propagated object since namespace is being deleted")
+		}
 		return deny(metav1.StatusReasonForbidden, "Cannot delete object propagated from namespace \""+oldSource+"\"")
 
 	case admissionv1beta1.Update:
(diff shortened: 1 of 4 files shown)


incubator/hnc/internal/validators/object.go
```diff
@@ -164,6 +164,16 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
...
+			// There are few things more irritating in (K8s) life than having some stupid controller stop
+			// your namespace from being deleted. If there's an object in here and we've decided that the
+			// namespace should be deleted, then don't block anything!
+			//
+			// It's probably slightly safer to create a K8s client, actually load the namespace and check
+			// its deletion timestamp here. But it's much easier to just record this information in the
+			// namespace/HC reconciler and read it here.
+			return allow("allowing deletion of propagated object since namespace is being deleted")
+		}
 		return deny(metav1.StatusReasonForbidden, "Cannot delete object propagated from namespace \""+oldSource+"\"")
```
yiqigao217, line 177: Question: Previously when a user deletes the namespace containing this copy, the object wehbook will deny it with this message? Or it's just the namespace is always terminating and we don't see this message?

yiqigao217 (commented):
The code lgtm but I have some questions. Also won't object reconciler remove obsolete copies even if the condition is `ActivitiesHalted`?

adrianludwin:
> The code lgtm but I have some questions. Also won't object reconciler remove obsolete copies even if the condition is `ActivitiesHalted`?

No, absolutely nothing happens in an `ActivitiesHalted` namespace: https://github.com/kubernetes-sigs/multi-tenancy/blob/8df8a461c02ac3b5aaf46fad9403c2700911f5da/incubator/hnc/internal/reconcilers/object.go#L248

incubator/hnc/internal/validators/object.go
adrianludwin: Yes, that's right - this was the error I was seeing when the e2e tests hung.

yiqigao217: sg

yiqigao217:
> > The code lgtm but I have some questions. Also won't object reconciler remove obsolete copies even if the condition is `ActivitiesHalted`?
> 
> No, absolutely nothing happens in an `ActivitiesHalted` namespace:
> 
> https://github.com/kubernetes-sigs/multi-tenancy/blob/8df8a461c02ac3b5aaf46fad9403c2700911f5da/incubator/hnc/internal/reconcilers/object.go#L248

Thanks for the explanation. Just another question out of curiosity, if the namespace only have `ParentMissing` condition and we set it as root, the obsolete copies should be removed by object reconciler, right? Just curious if that could be a way to get out of the stuck if not deleting VWH.

adrianludwin:
> Thanks for the explanation. Just another question out of curiosity, if the namespace only have `ParentMissing` condition and we set it as root, the obsolete copies should be removed by object reconciler, right? Just curious if that could be a way to get out of the stuck if not deleting VWH.

Yes, that would definitely fix it!

PR #1228, opened 2020-10-23
Add quickstart to docs and remove HNC v0.4

adrianludwin:
As we prepare to release HNC v0.6, get rid of all references to HNC v0.4
and convert the HNC demos to a "quickstart" format (thanks to Serge
Hartmann for the initial Markdown conversion).

Tested: staged at
https://github.com/adrianludwin/multi-tenancy/tree/quickstart/incubator/hnc/docs/user-guide.
Tested all links and did a quick visual inspection.

PR #1237, opened 2020-10-26
Add an FAQ with requirements and scaling info

### LINKED ISSUE #973: HNC: Question about some limitations of HNC
Currently, I'm working on the system which is based on the HNC. While working on it, I started to wonder about the limitation of HNC. I went through the guide and design doc but I could not find them. The followings are what I want to confirm:

- How many levels can we nest the namespaces?
- How many child namespaces one parent can hold?

If I miss the doc, please redirect me! Thanks.

### LINKED ISSUE #1006: HNC: document HNC's requirements (e.g. version, permissions)
HNC only works for K8s 1.15 and higher; we should make that explicit somewhere. We should also state that it requires very high privileges - this is alluded to in a bunch of places but not actually specified anywhere.

Thanks @rjbez17 for the suggestion.

adrianludwin:
These didn't seem to fit well in "concepts," "how-to" or "best
practices" so I put them into a catch-all topic.

Testing: staged at
https://github.com/adrianludwin/multi-tenancy/tree/docs/incubator/hnc/docs/user-guide

Fixes #973, fixes #1006

### DIFF
--- a/incubator/hnc/docs/user-guide/README.md
+++ b/incubator/hnc/docs/user-guide/README.md
@@ -20,6 +20,7 @@ see below.
 * [How-to](how-to.md): Perform common tasks when working with HNC
 * [Concepts](concepts.md): Learn more about the ideas behind HNC
 * [Best practices](best-practices.md): learn about the best ways to deploy HNC
+* [FAQs](faq.md): contact us, requirements, limitations and whatnot
 
 ## Older user guides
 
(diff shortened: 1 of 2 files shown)


adrianludwin:
Do not submit until #1228 is in
/hold
/assign @GinnyJI 
/assign @yiqigao217 
/assign @rjbez17

yiqigao217 (commented):
/lgtm

PR #1242, opened 2020-10-27
Check conflict with the selectors

GinnyJI:
If we have an object `A` already existing in child namespace `c1`, and
we are creating a new object also called `A` in the parent namespace,
and only _select_ it to propogate to another child namespace `c2`, we
should allow this operation, because `A` is not propagating to `c1` and
will not cause any conflict.

Test cases are also added in 'object_test.go'

Tested:
```
$ cat << EOF >> test.yaml
> apiVersion: v1
kind: Secret
metadata:
        annotations:
                propagate.hnc.x-k8s.io/treeSelect: c
        name: my-creds
        namespace: a
EOF
$ k delete secret my-creds -n a
$ k create -f test.yaml
```

### DIFF
--- a/incubator/hnc/internal/pkg/selectors/selectors.go
+++ b/incubator/hnc/internal/pkg/selectors/selectors.go
@@ -15,6 +15,23 @@ import (
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
 
+func ShouldPropagate(inst *unstructured.Unstructured, nsLabels labels.Set) (bool, error) {
+	if sel, err := GetSelector(inst); err != nil {
+		return false, err
+	} else if sel != nil && !sel.Matches(nsLabels) {
+		return false, nil
+	}
+	if sel, err := GetTreeSelector(inst); err != nil {
+		return false, err
+	} else if sel != nil && !sel.Matches(nsLabels) {
+		return false, nil
+	}
+	if none, err := GetNoneSelector(inst); err != nil || none {
+		return false, err
+	}
+	return true, nil
+}
+
 func GetSelectorAnnotation(inst *unstructured.Unstructured) string {
 	annot := inst.GetAnnotations()
 	return annot[api.AnnotationSelector]
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -650,20 +650,14 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
 // - Objects have a selector that doesn't match the destination namespace
 // - Service Account token secrets
 func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstructured.Unstructured, dst string) bool {
-	selector, errSelector := selectors.GetSelector(inst)
-	// TODO: generate events for the errors
-	if errSelector != nil {
-		log.Error(errSelector, "Cannot propagate")
-	}
-	treeSelector, errTreeSelector := selectors.GetTreeSelector(inst)
-	if errTreeSelector != nil {
-		log.Error(errTreeSelector, "Cannot propagate")
-	}
-	noneSelector, errNoneSelector := selectors.GetNoneSelector(inst)
-	if errNoneSelector != nil {
-		log.Error(errNoneSelector, "Cannot propagate")
-	}
 	nsLabels := r.Forest.Get(dst).GetLabels()
+	if ok, err := selectors.ShouldPropagate(inst, nsLabels); err != nil {
+		// TODO: generate events for the errors
+		log.Error(err, "Cannot propagate")
+		return false
+	} else if !ok {
+		return false
+	}
 
 	switch {
 	// Users can set the mode of a type to "remove" to exclude objects of the type
@@ -673,22 +667,6 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
 
 	// Object with nonempty finalizer list is not propagated
 	case hasFinalizers(inst):
-		return false
-
-	// Invalid selectors
-	case errSelector != nil || errTreeSelector != nil || errNoneSelector != nil:
-		return false
-
-	// None selector is set to true
-	case noneSelector:
-		return false
-
-	// Selector does not match
-	case selector != nil && !selector.Matches(nsLabels):
-		return false
-
-	// treeSelector does not match
-	case treeSelector != nil && !treeSelector.Matches(nsLabels):
 		return false
 
 	case r.GVK.Group == "" && r.GVK.Kind == "Secret":
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -246,7 +246,12 @@ func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
 	// Get a list of conflicting descendants if there's any.
 	for _, desc := range descs {
 		if o.Forest.Get(desc).HasSourceObject(gvk, nm) {
-			conflicts = append(conflicts, desc)
+			// If the user have chosen not to propagate the object to this descendant,
+			// there shouldn't be any conflict reported here
+			nsLabels := o.Forest.Get(inst.GetNamespace()).GetLabels()
+			if ok, _ := selectors.ShouldPropagate(inst, nsLabels); ok {
+				conflicts = append(conflicts, desc)
+			}
 		}
 	}
 
--- a/incubator/hnc/internal/validators/object_test.go
+++ b/incubator/hnc/internal/validators/object_test.go
@@ -471,6 +471,7 @@ func TestCreatingConflictSource(t *testing.T) {
 		conflictNamespace string
 		newInstName       string
 		newInstNamespace  string
+		newInstAnnotation map[string]string
 		fail              bool
 	}{{
 		name:              "Deny creation of source objects with conflict in child",
@@ -492,6 +493,33 @@ func TestCreatingConflictSource(t *testing.T) {
 		name:             "Allow creation of source objects with no conflict",
 		newInstName:      "secret-a",
 		newInstNamespace: "a",
+	}, {
+		name:              "Allow creation of source objects with treeSelector not matching the conflicting child",
+		forest:            "-aa",
+		conflictInstName:  "secret-b",
+		conflictNamespace: "b",
+		newInstName:       "secret-b",
+		newInstNamespace:  "a",
+		newInstAnnotation: map[string]string{api.AnnotationTreeSelector: "c"},
+		fail:              false,
+	}, {
+		name:              "Allow creation of source objects with selector not matching the conflicting child",
+		forest:            "-aa",
+		conflictInstName:  "secret-b",
+		conflictNamespace: "b",
+		newInstName:       "secret-b",
+		newInstNamespace:  "a",
+		newInstAnnotation: map[string]string{api.AnnotationSelector: "c" + api.LabelTreeDepthSuffix},
+		fail:              false,
+	}, {
+		name:              "Allow creation of source objects with noneSelector set",
+		forest:            "-aa",
+		conflictInstName:  "secret-b",
+		conflictNamespace: "b",
+		newInstName:       "secret-b",
+		newInstNamespace:  "a",
+		newInstAnnotation: map[string]string{api.AnnotationNoneSelector: "true"},
+		fail:              false,
 	}}
 
 	for _, tc := range tests {
@@ -507,6 +535,7 @@ func TestCreatingConflictSource(t *testing.T) {
 			inst.SetName(tc.newInstName)
 			inst.SetNamespace(tc.newInstNamespace)
 			inst.SetGroupVersionKind(schema.GroupVersionKind{Group: "", Version: "v1", Kind: "Secret"})
+			inst.SetAnnotations(tc.newInstAnnotation)
 			// Test
 			got := o.handle(context.Background(), l, op, inst, &unstructured.Unstructured{})
 			// Report


incubator/hnc/internal/validators/object.go
```diff
@@ -246,7 +246,19 @@ func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
 	// Get a list of conflicting descendants if there's any.
 	for _, desc := range descs {
 		if o.Forest.Get(desc).HasSourceObject(gvk, nm) {
-			conflicts = append(conflicts, desc)
+			// check if the selectors match
```
adrianludwin, line 249: This is a fairly low-level comment. Can you think of a higher-level description of what you're checking for here?

incubator/hnc/internal/validators/object.go
```diff
@@ -246,7 +246,19 @@ func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
...
+			selector, _ := selectors.GetSelector(inst)
+			treeSelector, _ := selectors.GetTreeSelector(inst)
+			noneSelector, _ := selectors.GetNoneSelector(inst)
```
adrianludwin, line 252: There seems to be a lot of overlap  between the code here and the code in the object reconciler (specifically `shouldPropagateSource`). It's actually the same logic but implemented twice, I think? Can we consolidate that all in one place?

incubator/hnc/internal/validators/object_test.go
```diff
@@ -492,6 +493,33 @@ func TestCreatingConflictSource(t *testing.T) {
...
+		name:              "Allow creation of source objects with exceptions on the conflicting child",
+		forest:            "-aa",
+		conflictInstName:  "secret-b",
+		conflictNamespace: "b",
+		newInstName:       "secret-b",
+		newInstNamespace:  "a",
+		newInstAnnotation: map[string]string{api.AnnotationTreeSelector: "c"},
+		fail:              false,
+	}, {
+		name:              "Allow creation of source objects with exceptions on the conflicting child",
```
adrianludwin, line 506: These names are identical - they should be different so you can tell which one failed

adrianludwin (commented):
We also need testcases for the other cases where we have overwriting prevention, e.g. when changing a namespace's parent or when switching the sync mode from `Remove` to `Propagate`.

Also, the testing description is good but I don't actually think you need that much detail for the manual testing, since it's all captured by the automated tests as well. I would just say something like "tested manually to verify my changes weren't rejected if I excluded the namespace with the conflict." It _would_ be nice to see a statement like "all the automated tests failed without the new code in the validators, and passed with it."

incubator/hnc/internal/validators/object.go
GinnyJI: updated the comment to "// If the user have chosen not to propagate the object to this descendant, there shouldn't be any conflict reported here"

incubator/hnc/internal/validators/object_test.go
GinnyJI: done

incubator/hnc/internal/validators/object.go
adrianludwin: Perfect

incubator/hnc/internal/validators/object.go
GinnyJI: Agreed, done

incubator/hnc/internal/validators/object.go
```diff
@@ -246,7 +246,19 @@ func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
...
+			nsLabels := o.Forest.Get(inst.GetNamespace()).GetLabels()
+			// We only check notMatch here because an empty selector annotation will be treated as an
+			// "all-match" selector
+			notMatch := (selector != nil && !selector.Matches(nsLabels)) ||
+				(treeSelector != nil && !treeSelector.Matches(nsLabels)) ||
+				noneSelector == true
+			if !notMatch {
```
yiqigao217: Seems it can be
```
match := (selector ==nil || selector.Matchers(l)) &&
 (treeSelector == nil || treeSelector.Matches(l)) &&
 noneSelector == false
```
so that you can use `if match {}` instead of `if !notMatch{}`?

yiqigao217 (commented):
+1 to Adrian's comments and just added one more.

GinnyJI:
> We also need testcases for the other cases where we have overwriting prevention, e.g. when changing a namespace's parent or when switching the sync mode from `Remove` to `Propagate`.
> 
> Also, the testing description is good but I don't actually think you need that much detail for the manual testing, since it's all captured by the automated tests as well. I would just say something like "tested manually to verify my changes weren't rejected if I excluded the namespace with the conflict." It _would_ be nice to see a statement like "all the automated tests failed without the new code in the validators, and passed with it."

I don't think the other two test cases is supported yet, my next thing would be working on the case: "switching the sync mode from `Remove` to `Propagate`"

incubator/hnc/internal/validators/object.go
GinnyJI: Ture, updated!

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -15,6 +15,27 @@ import (
...
+	noneSelector, err := GetNoneSelector(inst)
+	if err != nil {
+		return false, err
+	}
+	// We only check notMatch here because an empty selector annotation will be treated as an
+	// "all-match" selector
+	notMatch := (selector == nil || selector.Matches(nsLabels)) &&
+		(treeSelector == nil || treeSelector.Matches(nsLabels)) &&
+		noneSelector == false
+	return notMatch, nil
```
yiqigao217, line 36: Update comment and variable name to `match` lol?

adrianludwin:
> I don't think the other two test cases is supported yet, my next thing would be working on the case: "switching the sync mode from `Remove` to `Propagate`"

sgtm!

incubator/hnc/internal/validators/object_test.go
```diff
@@ -492,6 +493,33 @@ func TestCreatingConflictSource(t *testing.T) {
...
+		name:              "Allow creation of source objects with treeSelector on the conflicting child",
+		forest:            "-aa",
+		conflictInstName:  "secret-b",
+		conflictNamespace: "b",
+		newInstName:       "secret-b",
+		newInstNamespace:  "a",
+		newInstAnnotation: map[string]string{api.AnnotationTreeSelector: "c"},
+		fail:              false,
+	}, {
+		name:              "Allow creation of source objects with selector on the conflicting child",
```
yiqigao217, line 506: Update this and the previous case names with `Allow creation of source objects with selector/treeSelector not matching the conflicting child`?

incubator/hnc/internal/validators/object_test.go
```diff
@@ -492,6 +493,33 @@ func TestCreatingConflictSource(t *testing.T) {
...
+		forest:            "-aa",
+		conflictInstName:  "secret-b",
+		conflictNamespace: "b",
+		newInstName:       "secret-b",
+		newInstNamespace:  "a",
+		newInstAnnotation: map[string]string{api.AnnotationSelector: "c" + api.LabelTreeDepthSuffix},
+		fail:              false,
+	}, {
+		name:              "Allow creation of source objects with noneSelector on the conflicting child",
```
yiqigao217, line 515: Ditto, `Allow creation of source objects with noneSelector set`?

incubator/hnc/internal/pkg/selectors/selectors.go
adrianludwin, line 36: This logic isn't very clear to me. Since we're in a function and you can early-return, maybe take advantage of that?

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -650,20 +650,12 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
...
-	noneSelector, errNoneSelector := selectors.GetNoneSelector(inst)
-	if errNoneSelector != nil {
-		log.Error(errNoneSelector, "Cannot propagate")
-	}
 	nsLabels := r.Forest.Get(dst).GetLabels()
+	isMatch, err := selectors.IsSelectorsMatch(inst, nsLabels)
+	if err != nil {
+		// TODO: generate events for the errors
+		log.Error(err, "Cannot propagate")
+	}
```
adrianludwin, line 658: Sorry, remind me why we don't early-return here on error? A comment would be nice.

Honestly I'm not really sure about this whole switch statement anymore. When it started, it looked like there really would be a number of independent cases, but now it just looks like a bunch of `if` statements with early returns. I wonder if that would be more clear.

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -15,6 +15,27 @@ import (
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
 
+func IsSelectorsMatch(inst *unstructured.Unstructured, nsLabels labels.Set) (bool, error) {
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/internal/validators/object_test.go
GinnyJI: done

incubator/hnc/internal/validators/object_test.go
GinnyJI: done

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Updated to 
```

	if selector != nil && !selector.Matches(nsLabels) {
		return false, nil
	}
	if treeSelector != nil && !treeSelector.Matches(nsLabels) {
		return false, nil
	}
	if noneSelector {
		return false, nil
	}
	return true, nil
```
Does this look better for you?

incubator/hnc/internal/reconcilers/object.go
GinnyJI: You are right that it can be an early return. We didn't do it before because we want to log multiple errors, but now this can be done in the `DoSelectorsMatch()`. However, I didn't do it because last time we discussed that these two options both make sense, and logging multiple errors in the function seems like an overkill

GinnyJI:
tracked at #1123

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -15,6 +15,32 @@ import (
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 )
 
+func DoSelectorsMatch(inst *unstructured.Unstructured, nsLabels labels.Set) (bool, error) {
+	selector, err := GetSelector(inst)
+	if err != nil {
+		return false, err
+	}
```
adrianludwin, line 22: Nit: personally I'd group them by selector type, so you don't have to "remember" what all the variables are until halfway down the function, e.g.:

```
if sel, err := GetSelector(inst); err != nil {
  return false, err
} else if sel != nil && !sel.Matches(nsLabels) {
  return false, nil
}
// ... same for tree selector ...
if none, err := GetNoneSelector(inst); err != nil || none {
  return false, err
}

```

Note that this doesn't even introduce any temporary variables into the function scope.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -675,20 +668,8 @@ func (r *ObjectReconciler) shouldPropagateSource(log logr.Logger, inst *unstruct
...
-	case noneSelector:
-		return false
-
-	// Selector does not match
-	case selector != nil && !selector.Matches(nsLabels):
-		return false
-
-	// treeSelector does not match
-	case treeSelector != nil && !treeSelector.Matches(nsLabels):
+	// Selectord not match
```
adrianludwin, line 671: typo

incubator/hnc/internal/validators/object.go
```diff
@@ -246,7 +246,13 @@ func (o *Object) hasConflict(inst *unstructured.Unstructured) (bool, []string) {
 	// Get a list of conflicting descendants if there's any.
 	for _, desc := range descs {
 		if o.Forest.Get(desc).HasSourceObject(gvk, nm) {
-			conflicts = append(conflicts, desc)
+			// If the user have chosen not to propagate the object to this descendant,
+			// there shouldn't be any conflict reported here
+			nsLabels := o.Forest.Get(inst.GetNamespace()).GetLabels()
+			isMatch, _ := selectors.DoSelectorsMatch(inst, nsLabels)
```
adrianludwin, line 252: Having another look, I think `DoSelectorsMatch` is actually a bit confusing. Maybe just `selectors.ShouldPropagate` would be easier to read?

Also you can do this as a one line:

```
if conflict, _ := selectors.ShouldPropagate(inst, nsLabels); conflict {
  conflicts = append(conflicts, desc)
}
```

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -650,20 +650,13 @@ func (r *ObjectReconciler) syncPropagation(ctx context.Context, log logr.Logger,
...
-	if errNoneSelector != nil {
-		log.Error(errNoneSelector, "Cannot propagate")
-	}
 	nsLabels := r.Forest.Get(dst).GetLabels()
+	isMatch, err := selectors.DoSelectorsMatch(inst, nsLabels)
+	if err != nil {
+		// TODO: generate events for the errors
+		log.Error(err, "Cannot propagate")
+		return false
+	}
```
adrianludwin, line 659: You could also be even more aggresive about early returns here:

```
if ok, err := selectors.ShouldPropagate(inst, nsLabels); err != nil {
  // ... what you have now ...
} else if !ok {
  return false
}
```

And never even make it to the switch statement.

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Ah I see, updated

incubator/hnc/internal/reconcilers/object.go
GinnyJI: done

incubator/hnc/internal/validators/object.go
GinnyJI: We already have some other functions called 'shouldPropagate()', do you think this will make the function names confusing?

incubator/hnc/internal/reconcilers/object.go
GinnyJI: Done

incubator/hnc/internal/validators/object.go
adrianludwin: IMO it's fine to have similarly named functions in different packages. `shouldPropagate` in a reconciler might be making the overall decision, while `selectors.ShouldPropagate` is only about whether the _selectors_ think it's ok to propagate.

adrianludwin (commented):
lgtm if you decide you want to keep the DoSelectorsMatch name - lmk

incubator/hnc/internal/validators/object.go
GinnyJI: Makes sense, I'm changing it to ShouldPropagate

GinnyJI:
> lgtm if you decide you want to keep the DoSelectorsMatch name - lmk

Changed name to `ShouldPropagate`

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

adrianludwin:
@rjbez17 lmk if you had any concerns about this one but I know you're busy!

/hold cancel

PR #1245, opened 2020-10-28
Fix netpol e2e test (cherry-pick to master)

adrianludwin:
The test was failing incorrectly; it now passes if netpols are enabled
and is skipped if they're not.

Tested: on GKE with netpol enabled and on Kind with them disabled.


adrianludwin:
/assign @GinnyJI 
/assign @yiqigao217 

This is just the cherry pick, so one lgtm should be fine.

yiqigao217 (commented):
/lgtm

PR #1253, opened 2020-11-03
Check if more than one selector is created

GinnyJI:
Tested: make test

### DIFF
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -112,21 +112,22 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
 		// check selector format
-		err := validateSelectorChange(inst, oldInst)
 		// If this is a selector change, and the new selector is not valid, we'll deny this operation
-		if err != nil {
+		if err := validateSelectorChange(inst, oldInst); err != nil {
 			msg := fmt.Sprintf("Invalid Kubernetes labelSelector: %s", err)
 			return deny(metav1.StatusReasonBadRequest, msg)
 		}
-		err = validateTreeSelectorChange(inst, oldInst)
-		if err != nil {
+		if err := validateTreeSelectorChange(inst, oldInst); err != nil {
 			msg := fmt.Sprintf("Invalid HNC %q value: %s", api.AnnotationTreeSelector, err)
 			return deny(metav1.StatusReasonBadRequest, msg)
 		}
-		err = validateNoneSelectorChange(inst, oldInst)
-		if err != nil {
+		if err := validateNoneSelectorChange(inst, oldInst); err != nil {
 			return deny(metav1.StatusReasonBadRequest, err.Error())
 		}
+		if msg := validateSelectorUniqueness(inst, oldInst); msg != "" {
+			return deny(metav1.StatusReasonBadRequest, msg)
+		}
+
 		if yes, dnses := o.hasConflict(inst); yes {
 			dnsesStr := strings.Join(dnses, "\n  * ")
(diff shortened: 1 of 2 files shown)


incubator/hnc/internal/validators/object.go
```diff
@@ -112,22 +112,24 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
-		err = validateNoneSelectorChange(inst, oldInst)
-		if err != nil {
+		if err := validateNoneSelectorChange(inst, oldInst); err != nil {
 			return deny(metav1.StatusReasonBadRequest, err.Error())
 		}
-		// TODO(@ginnyji): modify hasConflict so that it's aware of selectors
+		if err := validateSelectorUniqueness(inst, oldInst); err != nil {
+			return deny(metav1.StatusReasonBadRequest, err.Error()+
+				", only one selector annotation should exist."+
+				" Please delete the existing one before adding a new one.")
```
adrianludwin, line 130: There might not be an "existing one" - this might just be a yaml file with two selectors. This will changes how you implement `validateSelectorUniqueness`.

Also, I'd just return the full string from `validateSelectorUniqueness` and display that, e.g.:

`if msg := validateSelectorUniqueness(...); msg != "" {`

There's no need to convert from a string to an error and back again.

yiqigao217: Just confirming the `validateSelectorUniqueness` implementation actually did implement the case for a new yaml file with two selectors. So do the tests, right?

GinnyJI: Yeah the test case currently covers the case where we creating multiple selectors at the same time. I modified the error msg to accommodate this case better

incubator/hnc/internal/validators/object_test.go
```diff
@@ -423,15 +423,45 @@ func TestUserChanges(t *testing.T) {
...
+					"annotations": map[string]interface{}{
+						api.AnnotationSelector:     "foo",
+						api.AnnotationTreeSelector: "!bar",
+					},
+				},
+			},
+		},
+	}, {
+		name: "Deny creation of object with both selector and noneSelect annotation",
+		fail: true,
```
adrianludwin, line 442: You don't need to test every possible combination of having multiple selectors IMO if the code itself doesn't care which ones you get.

incubator/hnc/internal/validators/object.go
```diff
@@ -139,6 +139,46 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
+	}
+	isValid := false
+	if sel != "" {
+		isValid = treeSel == "" && noneSel == ""
+	} else {
+		isValid = !(treeSel != "" && noneSel != "")
+	}
+	if isValid {
+		return ""
+	}
```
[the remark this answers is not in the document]
GinnyJI: done

incubator/hnc/internal/validators/object.go
```diff
@@ -139,6 +139,46 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
+
+	msg := "there is existing %s annotation" +
+		", only one selector annotation should exist." +
+		" Please delete the existing one before adding a new one."
```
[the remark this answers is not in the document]
GinnyJI: If there's not any existing selector, the error message will be `"cannot create more than one selector annotation, only one selector should exist"`

incubator/hnc/internal/validators/object_test.go
GinnyJI: Done

incubator/hnc/internal/validators/object.go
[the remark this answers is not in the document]
GinnyJI: Updated error msg

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/internal/validators/object.go
yiqigao217: sg

GinnyJI:
tracked at #1123

PR #1256, opened 2020-11-06
[vc] adding support for custom vn-agent ports

christopherhein:
This adds a flag to the syncer to allow you to set a "non-standard" vn-agent port with the syncer. You could previously supply a custom port to the vn-agent but it wouldn't work with how it was hardcoded when syncing.


christopherhein:
`make test` has been run, new images have been built and tested with

PR #1257, opened 2020-11-06
Update namespace deletion validator

### LINKED ISSUE #876: HNC: Revisit the subnamespace deletion rule in namespace validator
See [discussions](https://github.com/kubernetes-sigs/multi-tenancy/pull/870#discussion_r452959296)

Since we changed (**loosen**) the restriction to not requiring `allowCascadingDelete` set to be able to delete a leaf subnamespace, we probably need to revisit the subnamespace deletion validation to make sure it's **strict** enough that there are no other namespaces being deleted that user may not aware of.

GinnyJI:
The 'allowCascadingDelete' flag has to be set if the namespace to be 
deleted has any child namespace. If the namespace to be deleted has 
no child namespace, then no flag is required.

Tested: make test

Fix #876

### DIFF
--- a/incubator/hnc/internal/validators/namespace.go
+++ b/incubator/hnc/internal/validators/namespace.go
@@ -137,25 +137,12 @@ func (v *Namespace) cannotDeleteSubnamespace(req *nsRequest) admission.Response 
 }
 
 func (v *Namespace) illegalCascadingDeletion(ns *forest.Namespace) admission.Response {
-	// Early exit if the namespace allows cascading deletion.
-	if ns.AllowsCascadingDeletion() {
+	if ns.AllowsCascadingDeletion() || len(ns.ChildNames()) == 0 {
 		return allow("")
 	}
 
-	// Check if all children allow cascading deletion.
-	cantDelete := []string{}
-	for _, cnm := range ns.ChildNames() {
-		cns := v.Forest.Get(cnm)
-		if cns.IsSub && !cns.AllowsCascadingDeletion() {
-			cantDelete = append(cantDelete, cnm)
-		}
-	}
-	if len(cantDelete) != 0 {
-		msg := fmt.Sprintf("Please set allowCascadingDeletion first either in the parent namespace or in all the subnamespaces.\n  Subnamespace(s) without allowCascadingDelete set: %s.", cantDelete)
-		return deny(metav1.StatusReasonForbidden, msg)
-	}
-
-	return allow("")
+	msg := fmt.Sprintf("Please set allowCascadingDeletion in the parent namespace.")
+	return deny(metav1.StatusReasonForbidden, msg)
 }
 
 // decodeRequest gets the information we care about into a simple struct that's easy to both a) use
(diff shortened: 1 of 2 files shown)


incubator/hnc/internal/validators/namespace.go
```diff
@@ -151,7 +151,7 @@ func (v *Namespace) illegalCascadingDeletion(ns *forest.Namespace) admission.Res
 		}
 	}
 	if len(cantDelete) != 0 {
-		msg := fmt.Sprintf("Please set allowCascadingDeletion first either in the parent namespace or in all the subnamespaces.\n  Subnamespace(s) without allowCascadingDelete set: %s.", cantDelete)
+		msg := fmt.Sprintf("Please set allowCascadingDeletion in all the subnamespaces.\n  Subnamespace(s) without allowCascadingDelete set: %s.", cantDelete)
```
yiqigao217, line 154: Shouldn't the msg be `Please set ACD in this parent namespace`?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -137,9 +137,9 @@ func (v *Namespace) cannotDeleteSubnamespace(req *nsRequest) admission.Response
 }
 
 func (v *Namespace) illegalCascadingDeletion(ns *forest.Namespace) admission.Response {
-	// Early exit if the namespace allows cascading deletion.
-	if ns.AllowsCascadingDeletion() {
-		return allow("")
+	if !ns.AllowsCascadingDeletion() && len(ns.ChildNames()) > 0 {
+		msg := fmt.Sprintf("Please set allowCascadingDeletion first in the parent namespace.")
+		return deny(metav1.StatusReasonForbidden, msg)
```
yiqigao217, line 142: Maybe we still keep the early exit `if allowCD` and this one can just be `if len > 0`?

incubator/hnc/internal/validators/namespace.go
GinnyJI: This is where the parent has ACD set and the child doesn't

incubator/hnc/internal/validators/namespace.go
GinnyJI: I thought we want to have *both* ACD set ob parent and all child namespaces, and that's why we can't early return?

incubator/hnc/internal/validators/namespace.go
GinnyJI: Sry I mis-understood the logic, updated

incubator/hnc/internal/validators/namespace.go
GinnyJI: updated

yiqigao217: I think previously we allow `parent without ACD; all children with ACD` but now we just don't allow that since parent may not be aware of the cascading deletion when simply deleting the parent. But as long as `parent with ACD`, we would always allow it? (adding @adrianludwin to this discussion)

GinnyJI: Yeah I have updated the logic to allow deletion as long as parent has ACD

incubator/hnc/internal/validators/namespace.go
yiqigao217: (this change also depends on this discussion https://github.com/kubernetes-sigs/multi-tenancy/pull/1257#discussion_r518846183)

yiqigao217 (commented):
/lgtm
/hold
/assign @adrianludwin 
/assign @rjbez17

adrianludwin (commented):
/lgtm
/approve
/hold cancel

PR #1259, opened 2020-11-10
VC: introduce vnode provider

zhuangqh:
vnode provider enable developer to customize
the vnode object. eg. capability, address


Fei-Guo (approved):
/lgtm
/approve

Issue #1262, opened 2020-11-12
Fix issues with local development on MacOS


tsunny:
Noticed the following issues when building the project locally on MacOS

1. **kustomize** binary in hack folder is in linux format and will not work on mac. kustomize installed on MAC will be in **/usr/local/bin/kustomize**. Would be nice to pick this up automatically

2. **kubectl make target** - Works only for linux

3.  Few scripts in hack uses **sed** The sed installed on OSX is not the same as gnu version of sed. Running `make deploy` fails for this reason

PR #1263, opened 2020-11-12
Fixed issues with local development on OSX

tsunny:
- Kustomize binary path will be infered based on the OS
with default set to the binary in hack dir

- Kubectl target works

- Updated docs on how to install configure sed on OSX

Testing:
- [x] make deploy
- [x] make test

Issue: [#1262 ](https://github.com/kubernetes-sigs/multi-tenancy/issues/1262)

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -54,10 +54,18 @@ else
 	HNC_IMG = ${HNC_IMG_NAME}:${HNC_IMG_TAG}
 endif
 
+# Determine the OS Name, linux/darwin
+OS_NAME := $(shell go env GOOS)
+
 # `make` must be called from the HNC root, or all kinds of things will break
 # (starting with this).
 CURDIR = $(shell pwd)
 KUSTOMIZE ?= ${CURDIR}/hack/kustomize-3.8.1
+
+ifeq ($(OS_NAME),darwin)
+	KUSTOMIZE = $(shell which kustomize)
+endif
+
 CONTROLLER_GEN ?= ${CURDIR}/bin/controller-gen
 
 # Enable CRD per-version validation. As of Kubernetes 1.13, we can have multiple
@@ -107,7 +115,7 @@ clean: krew-uninstall
 
 # Installs the Linux kubectl plugin to $GOPATH/bin, assume that this is in your PATH already.
 kubectl: build
-	cp bin/kubectl/kubectl-hns_linux_amd64 ${GOPATH}/bin/kubectl-hns
+	cp bin/kubectl/kubectl-hns_$(OS_NAME)_amd64 $(GOBIN)/kubectl-hns
 
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: build
--- a/incubator/hnc/README.md
+++ b/incubator/hnc/README.md
@@ -121,6 +121,12 @@ is configured to use the project containing the registry you want to use, and
 that you've previously run `gcloud auth configure-docker` so that Docker can use
 your GCP credentials.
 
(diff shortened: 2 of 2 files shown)


incubator/hnc/Makefile
```diff
@@ -90,12 +98,8 @@ test: build
...
 build: generate fmt vet manifests
 	go build -o bin/manager ./cmd/manager/main.go
-	GOOS=linux GOARCH=amd64 go build \
-	     -o bin/kubectl/kubectl-hns_linux_amd64 \
-	     -ldflags="-X sigs.k8s.io/multi-tenancy/incubator/hnc/internal/version.Version=${HNC_IMG_TAG}" \
-	     ./cmd/kubectl/main.go
-	GOOS=darwin GOARCH=amd64 go build \
-	     -o bin/kubectl/kubectl-hns_darwin_amd64 \
+	GOOS=$(OS_NAME) GOARCH=amd64 go build \
+	     -o bin/kubectl/kubectl-hns_$(OS_NAME)_amd64 \
```
adrianludwin, line 102: I think we'd rather keep this as-is - every time we compile, we compile for multiple platforms so that we know if there's a problem early on. Also, we rely on this [in the release process](https://github.com/kubernetes-sigs/multi-tenancy/blob/775d0ca6bb7a9e1760be3eff46ad2eeffabba00f/incubator/hnc/cloudbuild.yaml#L28).

incubator/hnc/Makefile
```diff
@@ -107,7 +111,7 @@ clean: krew-uninstall
 
 # Installs the Linux kubectl plugin to $GOPATH/bin, assume that this is in your PATH already.
 kubectl: build
-	cp bin/kubectl/kubectl-hns_linux_amd64 ${GOPATH}/bin/kubectl-hns
+	cp bin/kubectl/kubectl-hns_$(OS_NAME)_amd64 $(GOBIN)/kubectl-hns
```
adrianludwin, line 114: Does GOBIN reliably default to GOPATH/bin if it's not explicitly set? If not, should we set it ourselves?

adrianludwin (commented):
Thanks for this! Just a few things I'd like to clear up before submitting it.

/ok-to-test

incubator/hnc/Makefile
tsunny: Yes. the `go env GOPATH` resolves to `<user-home-dir>/go` if GOPATH is unset. So, no need to set this explicitly.

incubator/hnc/Makefile
tsunny: Sure will revert this change.

incubator/hnc/Makefile
adrianludwin: Right, but this change uses `GOBIN`, not `GOPATH`. I double-checked and GOBIN is unset on my workstation, so this line will fail.

A more robust solution might be to say something like `GOBIN ?= $(shell go env GOPATH)/bin` which will set it to the default if it's unset.

tsunny: @adrianludwin - I believe that is how GOBIN is being set in the Makefile at the moment. https://github.com/kubernetes-sigs/multi-tenancy/blob/56839e8dbf3e8bdb676f7d1a86e414efab6850fc/incubator/hnc/Makefile#L79

And we are using the same variable here? Should work right?

adrianludwin: Whoops I forgot we already had that! Ok that lgtm

tsunny: Sure. Thanks for the review.

adrianludwin (commented):
/lgtm
/approve

PR #1266, opened 2020-11-12
Quick fix for missing space

GinnyJI:


adrianludwin (commented):
Technically that's actually correct! A one-character flag (which is usually introduced by a single dash, like `-n`) doesn't actually need a space. So all four of these are correct:

```
-nparent
-n parent
--namespace parent
--namespace=parent
```

But this is not (for obvious reasons): `--namespaceparent`

With that said, what you've written is definitely more clear.

/approve
/lgtm

GinnyJI:
> Technically that's actually correct! A one-character flag (which is usually introduced by a single dash, like `-n`) doesn't actually need a space. So all four of these are correct:
> 
> ```
> -nparent
> -n parent
> --namespace parent
> --namespace=parent
> ```
> 
> But this is not (for obvious reasons): `--namespaceparent`
> 
> With that said, what you've written is definitely more clear.
> 
> /approve
> /lgtm

Ah I wasn't aware of this, good to know!

Issue #1238, opened 2020-10-27
HNC: kubectl supports case-insensitive mode


frbimo:
Currently, kubectl-hnc doesn't support all-lowercase mode

```
Could not update the HNC Configuration: HNCConfiguration.hnc.x-k8s.io "config" is invalid: spec.resources.mode: Unsupported value: "propagate": supported values: "Propagate", "Ignore", "Remove"
```

PR #1267, opened 2020-11-14
HNC: kubectl supports case-insensitive mode

frbimo:
This normalize synchronize mode string format to match with current supported mode: Propagate, Ignore, Remove,

allows synchronize mode input to be case-insensitive.

Tested:
- exec `kubectl` command and set any mode

Closes:  #1238


adrianludwin:
/ok-to-test
/lgtm
/approve

Cool, thanks!

PR #1271, opened 2020-11-15
Better debug and error logging

adrianludwin:
This change reduces the differences between regular and debug logging
(it simply changes the logging level, as opposed to also the format) and
also gets rid of the (largely useless) stack traces that are printed out
during an error.

Tested: observed the better logging behaviour with the new switches.


rjbez17:
@adrianludwin is this ready for review?

adrianludwin:
Oh yeah - thanks!
/assign @rjbez17

PR #1272, opened 2020-11-16
Add post-submit test script

GinnyJI:
Test:

```
docker run -it -v /var/run/docker.sock:/var/run/docker.sock --entrypoint /bin/bash gcr.io/k8s-testimages/kubekins-e2e:latest-master
git clone https://github.com/GinnyJI/multi-tenancy.git
cd multi-tenancy/
git checkout postsubmit
incubator/hnc/hack/post-submit-test.sh
```
The e2e tests will run, but fail because this docker cluster does not
have dind (docker in docker) enabled.

### DIFF
--- a/incubator/hnc/hack/post-submit-test.sh
+++ b/incubator/hnc/hack/post-submit-test.sh
@@ -1,0 +1,23 @@
+# This file is run by Prow during all postsubmits
+export PATH=$(go env GOPATH)/bin:$PATH
+mkdir -p $(go env GOPATH)/bin
+
+echo "Installing 'kubebuilder' to include the Ginkgo test suite requirements"
+kb=2.3.1
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb}/kubebuilder_${kb}_linux_amd64.tar.gz
+tar -zxvf kubebuilder_${kb}_linux_amd64.tar.gz
+mv kubebuilder_${kb}_linux_amd64 /usr/local/kubebuilder
+
+# install and setup kind according to https://kind.sigs.k8s.io/docs/user/quick-start/
+echo "Setting up Kind"
+curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.8.1/kind-linux-amd64
+chmod +x ./kind
+mv ./kind /usr/local/bin/kind
+kind create cluster
+
+hack_dir=$(dirname ${BASH_SOURCE})
+
+echo "Running 'make test-e2e'"
+cd "$hack_dir/.."
+make kind-deploy
+make test-e2e


adrianludwin:
Can you find a way to enable Docker-in-Docker on your workstation? I thought it wouldn't be too hard.

incubator/hnc/hack/post-submit-test.sh
```diff
@@ -0,0 +1,23 @@
+# This file is run by Prow during all postsubmits
```
adrianludwin, line 1: Maybe include a link to the Prow github path

incubator/hnc/hack/post-submit-test.sh
```diff
@@ -0,0 +1,23 @@
...
+export PATH=$(go env GOPATH)/bin:$PATH
+mkdir -p $(go env GOPATH)/bin
+
+echo "Installing 'kubebuilder' to include the Ginkgo test suite requirements"
```
adrianludwin, line 5: Are you sure we need this? None of our test scripts use kubebuilder, and Ginkgo is included in our vendors/ directory.

incubator/hnc/hack/post-submit-test.sh
```diff
@@ -0,0 +1,23 @@
...
+chmod +x ./kind
+mv ./kind /usr/local/bin/kind
+kind create cluster
+
+hack_dir=$(dirname ${BASH_SOURCE})
+
+echo "Running 'make test-e2e'"
+cd "$hack_dir/.."
+make kind-deploy
+make test-e2e
```
adrianludwin, line 23: Do we need to put a sleep statement in between these two? I thought the e2e tests needed 30s or so for HNC to stabilize.

incubator/hnc/hack/post-submit-test.sh
```diff
@@ -0,0 +1,23 @@
...
+wget https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${kb}/kubebuilder_${kb}_linux_amd64.tar.gz
+tar -zxvf kubebuilder_${kb}_linux_amd64.tar.gz
+mv kubebuilder_${kb}_linux_amd64 /usr/local/kubebuilder
+
+# install and setup kind according to https://kind.sigs.k8s.io/docs/user/quick-start/
+echo "Setting up Kind"
+curl -Lo ./kind https://kind.sigs.k8s.io/dl/v0.8.1/kind-linux-amd64
+chmod +x ./kind
+mv ./kind /usr/local/bin/kind
+kind create cluster
```
rjbez17, line 16: nit: consider using `make kind-reboot` here? Incase our kind setup needs more steps later?

incubator/hnc/hack/post-submit-test.sh
rjbez17: Is it worthwhile to put a check/wait loop in the `make test-e2e` to ensure hnc is deployed and ready?

rjbez17:
lgtm after comments addressed

incubator/hnc/hack/post-submit-test.sh
adrianludwin: Maybe? It's a bit hard to tell when HNC is ready, and it's never really been an issue before now.

@GinnyJI - is there a pre-suite function we could call to wait for a) the webhooks' caBundles to be set and b) wait until the HNC pod is running? We'd probably want a brief sleep between those two since the pod exits after setting the caBundles and we would want to make sure we only started the wait _after_ that happened.

adrianludwin:
/remove-lifecycle stale
/assign @adrianludwin

I'm working on this now

adrianludwin:
Replaced by #1398

/close

PR #1281, opened 2020-11-20
Validate only one non-negated namespace for treeSelect

GinnyJI:
If multiple non-negated namespaces are set for treeSelect, this object
will not be propagated to anywhere, in which case the user should use
none selector instead.

Tested: make test

### DIFF
--- a/incubator/hnc/internal/pkg/selectors/selectors.go
+++ b/incubator/hnc/internal/pkg/selectors/selectors.go
@@ -61,10 +61,16 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
 
 	segs := strings.Split(treeSelectorStr, ",")
 	selectorStr := ""
+	nonNegatedNses := []string{}
 	for i, seg := range segs {
+		seg = strings.TrimSpace(seg)
 		// check if it's a valid namespace name
 		if err := validateTreeSelectorSegment(seg); err != nil {
 			return nil, err
+		}
+
+		if seg[0] != '!' {
+			nonNegatedNses = append(nonNegatedNses, seg)
 		}
 
 		selectorStr = selectorStr + seg + api.LabelTreeDepthSuffix
@@ -72,11 +78,18 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
 			selectorStr = selectorStr + ", "
 		}
 	}
+
 	treeSelector, err := getSelectorFromString(selectorStr)
 	if err != nil {
 		// In theory this should never happen because we already checked DNS label before.
 		// If this happens, it's more likely that we have a bug in our code
 		return nil, fmt.Errorf("internal error while parsing %q: %w", api.AnnotationTreeSelector, err)
+	}
+
+	// If there are more than one non-negated namespace, this object will not be propagated to
+	// any child namespace. We want to warn and stop user from doing this.
+	if len(nonNegatedNses) > 1 {
+		return nil, fmt.Errorf("should only have one non-negated namespace, but got multiple: %v", nonNegatedNses)
 	}
 	return treeSelector, nil
 }
--- a/incubator/hnc/internal/validators/object_test.go
+++ b/incubator/hnc/internal/validators/object_test.go
@@ -351,6 +351,34 @@ func TestUserChanges(t *testing.T) {
 			},
 		},
 	}, {
+		name: "Deny creation of object with multiple non-negated treeSelect annotation",
+		fail: true,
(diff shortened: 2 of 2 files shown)


incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -61,23 +61,35 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
...
 	segs := strings.Split(treeSelectorStr, ",")
 	selectorStr := ""
+	nonNegatedNses := []string{}
 	for i, seg := range segs {
 		// check if it's a valid namespace name
 		if err := validateTreeSelectorSegment(seg); err != nil {
 			return nil, err
 		}
 
+		if strings.TrimSpace(seg)[0] != '!' {
```
adrianludwin, line 71: Why are you trimming the spaces of `seg` here but not, say, on line 75?

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -61,23 +61,35 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
 
 	segs := strings.Split(treeSelectorStr, ",")
 	selectorStr := ""
+	nonNegatedNses := []string{}
```
adrianludwin, line 64: Rather than building a list, you could just early-return as soon as the _second_ non-negated name is found.

adrianludwin (commented):
lgtm with just a minor nit

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Do we want to include the non-negated nses in the error msg? Cause if we do then a list seems to be a better option

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: Because here we need to check if the leading char is `!` so we need to get rid of any leading whitespace. On line 75, `selectorStr` will be passed to `ParseToLabelSelector()`, and from experiments it will handle the whitespaces there.

adrianludwin: Ahh good to know. But. Hmm. If there were _trailing_ whitespace, though, line 75 wouldn't work.

I feel like it would both be more correct and also easier to read to insert this after line 65:

`seg = strings.TrimSpace(seg)`

And then this line just becomes `if seg[0] != '!'` which is way easier to read.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -350,6 +350,20 @@ func TestUserChanges(t *testing.T) {
 				},
 			},
 		},
+	}, {
+		name: "Deny creation of object with multiple non-negated treeSelect annotation",
```
adrianludwin, line 354: Do we have a test allowing multiple negated?

incubator/hnc/internal/pkg/selectors/selectors.go
GinnyJI: I agree, done!

incubator/hnc/internal/validators/object_test.go
GinnyJI: Added

incubator/hnc/internal/pkg/selectors/selectors.go
```diff
@@ -61,23 +61,36 @@ func GetTreeSelector(inst *unstructured.Unstructured) (labels.Selector, error) {
...
 	if err != nil {
 		// In theory this should never happen because we already checked DNS label before.
 		// If this happens, it's more likely that we have a bug in our code
 		return nil, fmt.Errorf("internal error while parsing %q: %w", api.AnnotationTreeSelector, err)
 	}
+
+	// If there are more than one non-negated namespace, this object will not be propagated to
+	// any child namespace. We want to warn and stop user from doing this.
+	if len(nonNegatedNses) > 1 {
+		return treeSelector, fmt.Errorf("should only have one non-negated namespace, but got multiple: %v", nonNegatedNses)
```
adrianludwin, line 92: Nit: would usually return `nil` as the first param here.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

incubator/hnc/internal/pkg/selectors/selectors.go

adrianludwin:
Oh right - Thanksgiving in the USA.

/hold cancel

Issue #1043, opened 2020-08-21
HNC: CleanupNamespaces shouldn't require an explicit list of namespaces


adrianludwin:
We should have dedicated `CreateNamespace` and `CreateSubnamespace` functions that add some kind of label (e.g. `hnc.x-k8s.io/testNamespace: true`) to all namespaces, so we can just search by label in `CleanupNamespaces` and not require an explicit list.

/good-first-issue
/cc @GinnyJI 
/cc @yiqigao217

mayankshah1607:
@adrianchung Can I take up this one if no one is working on it already? Thanks!
/assign

adrianludwin:
Go for it!

venkatnsrinivasan:
Hi
I attempted to try and work on this issue and opened a PR here 
https://github.com/kubernetes-sigs/multi-tenancy/pull/1283 . Any suggestions and guidance are welcome . hopefully the PR is on the right path

Thanks

adrianludwin:
This was fixed by #1283 - thanks @venkatnsrinivasan !

/close

PR #1283, opened 2020-11-21
CleanupNamespaces shouldn't require an explicit list of namespaces

venkatnsrinivasan:
See issue #1043. Added functions Createnamespace,Createsubnamespace,
CreateNamespaceWithLabels,CleanupTestNamespaces. The create functions
mark all the namespaces with a test label. Refactored all e2e test
cases to utilize the new functions. Preserved the older CleanupNamespace
function if it is useful for specific namespace deletion use case

Tested: Ran the e2e tests on k3d with k8s version v1.17.12. All tests passed
except the network policies test which was skipped.

### DIFF
--- a/incubator/hnc/pkg/testutils/testutils.go
+++ b/incubator/hnc/pkg/testutils/testutils.go
@@ -15,6 +15,8 @@ import (
 // The time that Eventually() will keep retrying until timeout
 // we use 5 seconds here because some tests require deleting a namespace, and shorter time period might not be enough
 const eventuallyTimeout = 5
+// The testing label marked on all namespaces created using the testing phase, offering ease when doing cleanups
+const testingNamespaceLabel = "hnc.x-k8s.io/testNamespace"
 
 var hncRecoverPath = os.Getenv("HNC_REPAIR")
 
@@ -230,7 +232,42 @@ func RunCommand(cmdln ...string) (string, error) {
 	return string(stdout), err
 }
 
-// CleanupNamespaces does everything it can to delete the passed-in namespaces. It also uses very
+// CreateNamespace creates the specified namespace with canned testing labels making it easier
+// to look up and delete later.
+func CreateNamespace( ns string)   {
+	MustRun("kubectl create ns", ns)
+	labelTestingNs(ns)
+}
+
+// CreateSubnamespace creates the specified namespace in the parent namespace with canned testing labels making it easier
+// to look up and delete later.
+func CreateSubnamespace( ns string,parent string)   {
+	MustRun( "kubectl hns create", ns, "-n", parent)
+	labelTestingNs(ns)
+}
+
+// marks testing namespaces with a label for future search and lookup.
+func labelTestingNs(ns string){
+	MustRun("kubectl label --overwrite ns", ns,testingNamespaceLabel+"=true")
+}
+
+// CleanupTestNamespaces finds the list of namespaces labeled as test namespaces and delegates
+// to cleanupNamespaces function.
+func CleanupTestNamespaces(){
+	nses := []string{}
+	EventuallyWithOffset(1, func() error {
+		labelQuery := testingNamespaceLabel+"=true"
+		out,err:=RunCommand("kubectl get ns -o custom-columns=:.metadata.name --no-headers=true", "-l", labelQuery)
+		if err != nil {
+			return err
+		}
+		nses= strings.Split(out,"\n")
+		return nil
+	}).Should(Succeed(), "while getting list of namespaces to clean up")
+	cleanupNamespaces(nses...)
+}
+
+// cleanupNamespaces does everything it can to delete the passed-in namespaces. It also uses very
 // high timeouts (30s) since this function is often called after HNC has just been reinstalled, and
 // it can take a while of HNC to start allowing changes to namespaces again.
 //
@@ -238,7 +275,7 @@ func RunCommand(cmdln ...string) (string, error) {
 // HNC doesn't put finalizers on namespaces themselves; it blocks namespace deletion by blocking
 // deletion of the objects in it, but if HNC is damaged or missing, this can result in namespaces
 // never being deleted without admin action.
-func CleanupNamespaces(nses ...string) {
+func cleanupNamespaces(nses ...string) {
 	const cleanupTimeout = 30
 
 	// Remove all objections HNC might have to deleting a namespace.
@@ -246,6 +283,9 @@ func CleanupNamespaces(nses ...string) {
 	for _, ns := range nses {
 		// Skip any namespace that doesn't actually exist. We only check once (e.g. no retries on
 		// errors) but reads are usually pretty reliable.
+
+		// TODO: This check should ideally be removed if we call this function only with a list of valid test namespaces
+		// found with the label search. But the RecoverHNC function has to call this with explicit namespaces for a specific test
 		if err := TryRunQuietly("kubectl get ns", ns); err != nil {
 			continue
 		}
@@ -332,13 +372,17 @@ func RecoverHNC() {
 		a = "recover-test-a"
 		b = "recover-test-b"
 	)
-	CleanupNamespaces(a, b)
+	// Need to explicitly call deleting these two namespaces because this function is called from the issues test
+	// ("Should allow deletion of namespaces with propagated objects that can't be removed - issue #1214", which tries
+	// to delete a specific namespace "child" . Cleaning up all namespaces using CleanupTestNamespaces will cause that
+	// test to fail.
+	cleanupNamespaces(a, b)
 	// Ensure validators work
 	mustRunWithTimeout(1, 30, "kubectl create ns", a)
 	// Ensure reconcilers work
 	mustRunWithTimeout(1, 30, "kubectl hns create", b, "-n", a)
 	mustRunWithTimeout(1, 30, "kubectl get ns", b)
-	CleanupNamespaces(a, b)
+	cleanupNamespaces(a, b)
 }
 
 func writeTempFile(cxt string) string {
--- a/incubator/hnc/test/e2e/issues_test.go
+++ b/incubator/hnc/test/e2e/issues_test.go
@@ -20,19 +20,18 @@ var _ = Describe("Issues", func() {
 	)
 
 	BeforeEach(func() {
-		CleanupNamespaces(nsParent, nsChild, nsSub1, nsSub2, nsSub1Sub1, nsSub2Sub1,
-			nsSubSub2, nsSubChild, nsSubSubChild)
+		CleanupTestNamespaces()
 	})
 
 	AfterEach(func() {
-		CleanupNamespaces(nsParent, nsChild, nsSub1, nsSub2, nsSub1Sub1, nsSub2Sub1,
-			nsSubSub2, nsSubChild, nsSubSubChild)
+		CleanupTestNamespaces()
 	})
 
 	It("Should not delete full namespace when a faulty anchor is deleted - issue #1149", func() {
 		// Setup
-		MustRun("kubectl create ns", nsParent)
-		MustRun("kubectl hns create", nsChild, "-n", nsParent)
+		CreateNamespace(nsParent)
+
+		CreateSubnamespace(nsChild, nsParent)
 
 		// Wait for subns
 		MustRun("kubectl describe ns", nsChild)
@@ -53,8 +52,8 @@ var _ = Describe("Issues", func() {
 	// like a good thing to test anyway.
 	It("Should delete full namespaces with propagated objects - issue #1130", func() {
 		// Set up the structure
-		MustRun("kubectl create ns", nsParent)
-		MustRun("kubectl create ns", nsChild)
+		CreateNamespace(nsParent)
+		CreateNamespace(nsChild)
 		MustRun("kubectl hns set", nsChild, "--parent", nsParent)
 		MustRun("kubectl create rolebinding admin-rb -n", nsParent,
 			"--clusterrole=admin --serviceaccount="+nsParent+":default")
@@ -68,8 +67,9 @@ var _ = Describe("Issues", func() {
 
 	It("Should delete subnamespaces with propagated objects - issue #1130", func() {
 		// Set up the structure
-		MustRun("kubectl create ns", nsParent)
-		MustRun("kubectl hns create", nsChild, "-n", nsParent)
+		CreateNamespace(nsParent)
+		CreateSubnamespace(nsChild, nsParent)
+
 		MustRun("kubectl create rolebinding admin-rb -n", nsParent,
 			"--clusterrole=admin --serviceaccount="+nsParent+":default")
 
@@ -82,9 +82,11 @@ var _ = Describe("Issues", func() {
 
 	It("Should set SubnamespaceAnchorMissing condition if the anchor is missing - issue #501", func() {
 		// Setting up a 3-level tree with 'parent' as the root
-		MustRun("kubectl create ns", nsParent)
+		CreateNamespace(nsParent)
+
 		// create a subnamespace without anchor by creating a full namespace with SubnamespaceOf annotation
-		MustRun("kubectl create ns", nsSub1)
+		CreateNamespace(nsSub1)
+
 		MustRun("kubectl hns set", nsSub1, "--parent", nsParent)
 		MustRun("kubectl annotate ns", nsSub1, "hnc.x-k8s.io/subnamespace-of="+nsParent)
 		// If the subnamespace doesn't allow cascadingDeletion and the anchor is missing in the parent namespace, it should have 'SubnamespaceAnchorMissing' condition while its descendants shoudn't have any conditions."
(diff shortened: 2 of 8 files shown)


incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,7 +232,49 @@ func RunCommand(cmdln ...string) (string, error) {
...
+	for k,v:= range testingNamespaceLabels{
+		labels=append(labels,fmt.Sprintf("%s=%s",k,v))
+	}
+	labelQuery := strings.Join(labels,",")
+	out,err:=RunCommand("kubectl get ns -o custom-columns=:.metadata.name --no-headers=true","-l",labelQuery)
+	if err !=nil{
+		GinkgoT().Log("Unable to get namespaces (failed): ", err)
+	}else{
+		nses:= strings.Split(out,"\n")
+		CleanupNamespaces(nses...)
```
adrianludwin, line 274: This should now be the only place that `CleanupNamespaces` is called, so either make it private (e.g. start with a lowercase letter - `cleanupNamespaces`) or just fold everything from that function into this one (assuming you've gotten rid of the indent).

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -246,6 +291,7 @@ func CleanupNamespaces(nses ...string) {
 	for _, ns := range nses {
 		// Skip any namespace that doesn't actually exist. We only check once (e.g. no retries on
 		// errors) but reads are usually pretty reliable.
+
```
adrianludwin, line 294: Was this intentional? This applies to the line 295-298, not to the whole block, so I wouldn't put a blank space here.

Also, if we've just finished searching for labels, we don't need this check at all.

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -20,19 +20,19 @@ var _ = Describe("Issues", func() {
...
 	})
 
 	It("Should not delete full namespace when a faulty anchor is deleted - issue #1149", func() {
 		// Setup
-		MustRun("kubectl create ns", nsParent)
-		MustRun("kubectl hns create", nsChild, "-n", nsParent)
+		CreateNamespace(nsParent)
+
+		CreateSubnamespace(nsChild,nsParent)
+
```
adrianludwin, line 35: I'd skip the extra newlines?

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -82,9 +83,12 @@ var _ = Describe("Issues", func() {
 
 	It("Should set SubnamespaceAnchorMissing condition if the anchor is missing - issue #501", func() {
 		// Setting up a 3-level tree with 'parent' as the root
-		MustRun("kubectl create ns", nsParent)
-		// create a subnamespace without anchor by creating a full namespace with SubnamespaceOf annotation
-		MustRun("kubectl create ns", nsSub1)
+		CreateNamespace(nsParent)
+
+		// create a subnamespace without anchor by crea
+		//ting a full namespace with SubnamespaceOf annotation
```
adrianludwin, line 89: Don't break a word ("crea" - "ting") with a newline. Better yet, just leave the existing comment as-is.

incubator/hnc/test/e2e/issues_test.go
```diff
@@ -125,18 +130,20 @@ var _ = Describe("Issues", func() {
...
-		MustRun("kubectl hns create", nsSub1Sub1, "-n", nsSub1)
-		MustRun("kubectl hns create", nsSub2Sub1, "-n", nsSub1)
+		CreateSubnamespace(nsSub1,nsParent)
+		CreateSubnamespace(nsSub1Sub1,nsSub1)
+		CreateSubnamespace(nsSub2Sub1,nsSub1)
 		// Creating the 2nd branch of subnamespaces
-		MustRun("kubectl hns create", nsSub2, "-n", nsParent)
-		MustRun("kubectl hns create", nsSubSub2, "-n", nsSub2)
+		CreateSubnamespace(nsSub2,nsParent)
+
```
adrianludwin, line 140: No extra newlines (here and elsewhere)

adrianludwin (commented):
This is a great cleanup - thanks! I've left a bunch of suggestions to tighten things up a bit more but this was exactly what I'd hoped for. Thanks for this contribution.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,7 +232,49 @@ func RunCommand(cmdln ...string) (string, error) {
 	stdout, err := cmd.CombinedOutput()
 	return string(stdout), err
 }
+// CreateNamespace creates the specified namespace with canned testing labels making it easier
+// to look up and delete later.
+func CreateNamespace( ns string)   {
```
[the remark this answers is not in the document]
venkatnsrinivasan: Will do.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,7 +232,49 @@ func RunCommand(cmdln ...string) (string, error) {
...
+	CreateNamespaceWithLabels(ns,"", testingNamespaceLabels)
+}
+// CreateSubnamespace creates the specified namespace in the parent namespace with canned testing labels making it easier
+// to look up and delete later.
+func CreateSubnamespace( ns string,parent string)   {
+	CreateNamespaceWithLabels(ns,parent, testingNamespaceLabels)
+}
+
+// CreateNamespaceWithLabels creates a namespace or a subnamespace (if a parent is specified) and marks the namespace
```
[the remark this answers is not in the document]
venkatnsrinivasan: Ok , sure. yes, currently only CreateNamespace  and CreateSubnamespace  use it but I figured if there might be future test cases where one might want to pass a list of labels and delete by those by labels it might be useful to leave these there .If you think its not needed , i will remove and simplify it.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,7 +232,49 @@ func RunCommand(cmdln ...string) (string, error) {
...
+// for a function to cleanup a specific namespace like CleanupNamespaces
+func CleanupTestNamespaces(){
+	labels:= make([]string, 0)
+	for k,v:= range testingNamespaceLabels{
+		labels=append(labels,fmt.Sprintf("%s=%s",k,v))
+	}
+	labelQuery := strings.Join(labels,",")
+	out,err:=RunCommand("kubectl get ns -o custom-columns=:.metadata.name --no-headers=true","-l",labelQuery)
+	if err !=nil{
+		GinkgoT().Log("Unable to get namespaces (failed): ", err)
```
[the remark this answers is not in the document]
venkatnsrinivasan: Ok, will check.

incubator/hnc/pkg/testutils/testutils.go
venkatnsrinivasan: Yes, again the reason to leave it public was have future test case wanting to delete a specific namespace. If not envisioned i will make it private.

incubator/hnc/pkg/testutils/testutils.go
venkatnsrinivasan: Yes we dont need this block if its only called from CleanupTestNamespaces which search for labels. But someone does , it won't gurantee the the ns exists. I can remove this .

incubator/hnc/test/e2e/issues_test.go
venkatnsrinivasan: Sorry about the new lines, my search replace of the old code resulted in these.Will fix.

incubator/hnc/test/e2e/issues_test.go
venkatnsrinivasan: Will do.

incubator/hnc/test/e2e/issues_test.go
venkatnsrinivasan: Will do.

incubator/hnc/test/e2e/issues_test.go
adrianludwin: np!

incubator/hnc/pkg/testutils/testutils.go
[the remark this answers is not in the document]
adrianludwin: I wouldn't worry about future test cases for now - it's easy to refactor stuff out later if we need it, but until then, it will just confuse anyone looking at this code. I always go for "simplicity now" unless there's a compelling reason otherwise, and I don't see one in this case.

Thanks!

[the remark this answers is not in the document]
venkatnsrinivasan: Sounds good. 
Thanks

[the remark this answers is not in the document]
venkatnsrinivasan: I have made the relevant changes . Please check .

incubator/hnc/pkg/testutils/testutils.go
venkatnsrinivasan: I tried removing the check for the existing namespaces as it is not relevant if we filter by labels anyway as you pointed out. But there are some connected tests which fail, so i restored it back. There is a function RecoverHNC which creates and deletes specific namespaces only. Deleting all namespaces causes to delete a namespace called 'child' in a previous test causing all tests to fail. Let me know if you have any suggestions to fix it or if it can be tracked seperately

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -15,6 +15,7 @@ import (
 // The time that Eventually() will keep retrying until timeout
 // we use 5 seconds here because some tests require deleting a namespace, and shorter time period might not be enough
 const eventuallyTimeout = 5
+const testingNamespaceLabel = "hnc.x-k8s.io/testNamespace"
```
adrianludwin, line 18: Comment what this constant is for?

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,23 +230,60 @@ func RunCommand(cmdln ...string) (string, error) {
 	stdout, err := cmd.CombinedOutput()
 	return string(stdout), err
 }
+// CreateNamespace creates the specified namespace with canned testing labels making it easier
```
adrianludwin, line 233: Newline after the last function and before the next comment

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,23 +230,60 @@ func RunCommand(cmdln ...string) (string, error) {
...
+// to look up and delete later.
+func CreateNamespace( ns string)   {
+	createNamespaceWithLabels(ns,"",eventuallyTimeout)
+}
+// CreateSubnamespace creates the specified namespace in the parent namespace with canned testing labels making it easier
```
adrianludwin, line 238: Newline before here

incubator/hnc/pkg/testutils/testutils.go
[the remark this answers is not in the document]
adrianludwin: This seems like a lot of complexity, and I _think_ it's just driven by the needs of `RecoverHNC`, right? I think I'd prefer that `CreateNamespace` and `CreateSubnamespace` should just be two-line functions, and let `RecoverHNC` do everything manually (since it doesn't even need the labels in the first place).

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,23 +230,60 @@ func RunCommand(cmdln ...string) (string, error) {
...
+	}else{
+		mustRunWithTimeout(1,timeout, "kubectl hns create", ns, "-n", parent)
+	}
+	MustRun("kubectl label --overwrite ns", ns,testingNamespaceLabel+"=true")
+}
+// CleanupTestNamespaces finds the list of namespaces labeled as test namespaces and delegates
+// to cleanupNamespaces function.
+func CleanupTestNamespaces(){
+	labelQuery := testingNamespaceLabel+"=true"
+	EventuallyWithOffset(1, func() string {
```
adrianludwin, line 258: I generally don't like calling functions with `Eventually` blocks from inside other `Eventually` blocks, especially since you then need to mess with the timeouts to make everything work. They also end up being very brittle as a single failure can cause the _entire_ block to repeat.

How about something like this:

```
nses := []string{}
EventuallyWithOffset(1, func() error {
  labelQuery := ...
  out, err := RunCommand(...)
  if err != nil {
    return err
  }
  nses = strings.Split(out, "\n")
  return nil
}).Should(Succeed(), "while getting list of namespaces to clean up)

cleanupNamespaces(nses)
```

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -229,23 +230,60 @@ func RunCommand(cmdln ...string) (string, error) {
...
+		out,err:=RunCommand("kubectl get ns -o custom-columns=:.metadata.name --no-headers=true","-l",labelQuery)
```
adrianludwin, line 259: Did you run `gofmt` after this line? Or run `make fmt` from the HNC root directory. Our coding standard is to have a space after every comma.

incubator/hnc/pkg/testutils/testutils.go
adrianludwin: sgtm

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -332,13 +370,17 @@ func RecoverHNC() {
...
 	)
-	CleanupNamespaces(a, b)
+	// Need to explicitly call deleting these two namespaces because this function is called from the issues test
+	// ("Should allow deletion of namespaces with propagated objects that can't be removed - issue #1214", which tries
+	// to delete a specific namespace "child" . Cleaning up all namespaces using CleanupTestNamespaces will cause that
+	// test to fail.
+	cleanupNamespaces(a,b)
 	// Ensure validators work
-	mustRunWithTimeout(1, 30, "kubectl create ns", a)
+	createNamespaceWithLabels(a,"",30)
```
adrianludwin, line 379: Since you're deleting these namespaces by name, not label, you don't have to use the `createNamespaceWithLabels` function - you can just go back to the old way.

incubator/hnc/pkg/testutils/testutils.go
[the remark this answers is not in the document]
venkatnsrinivasan: The whole timeout parameter is driven by the RecoverHNC yes. I had it like you suggested before without touching RecoverHNC but then figured for consistency to use this common function. But yes, i like the idea of reverting RecoverHNC to the manual way and removing all this timeout complexity. I added the CreateNamespaceWithLabels common function since i didn't like repeating the line 
	MustRun("kubectl label --overwrite ns", ns,testingNamespaceLabel+"=true")
in both the functions.

incubator/hnc/pkg/testutils/testutils.go
venkatnsrinivasan: Yes, will revert

incubator/hnc/pkg/testutils/testutils.go
venkatnsrinivasan: I ran go fmt before the commit. Will check again.

incubator/hnc/pkg/testutils/testutils.go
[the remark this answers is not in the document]
adrianludwin: I understand wanting to avoid repetition, but in this case I think having two "regular" functions call a common implementation makes the logic harder to understand. For one line, I think copy and paste is fine - at most I could imagine a utility function that you call like `labelNS(ns, testingNamespaceLabel, "true")` (eg factor out the common part, as opposed to wrapping a superset of functionality in a simpler interface) but personally I wouldn't bother with that. You can if you want though.

incubator/hnc/pkg/testutils/testutils.go
```diff
@@ -332,13 +372,17 @@ func RecoverHNC() {
 		a = "recover-test-a"
 		b = "recover-test-b"
 	)
-	CleanupNamespaces(a, b)
+	// Need to explicitly call deleting these two namespaces because this function is called from the issues test
+	// ("Should allow deletion of namespaces with propagated objects that can't be removed - issue #1214", which tries
+	// to delete a specific namespace "child" . Cleaning up all namespaces using CleanupTestNamespaces will cause that
+	// test to fail.
+	cleanupNamespaces(a,b)
```
adrianludwin, line 379: Space after comma

adrianludwin (commented):
lgtm with one last nit. Thanks for going through this review process, I know it can be long and difficult!

/ok-to-test

Before submitting, can you please take the following last steps:
* Squash all your commits into one commit (lmk if you need help with this) and force-submit it to this branch so we only have one commit per PR.
* Run `make fmt` one last time
* Reconfirm that you've re-run the e2e tests with this change and everything's working (we have one flaky test right now - #1285  - so it's ok if that one fails for you).

Thanks!

incubator/hnc/pkg/testutils/testutils.go
venkatnsrinivasan: Fixed and pushed single commit. Let me know if its good.

adrianludwin (commented):
/lgtm
/approve

Thanks for this!

Issue #1241, opened 2020-10-27
HNC: detect namespace deletion without a controller


adrianludwin:
#1236 checks whether a namespace is being deleted in the reconciler, then uses that data in the object validator. Ideally, the object validator would just do the check itself, but it doesn't have a client. I'm sure there's a way to do this which would allow us to get rid of this coupling.

/good-first-issue

tsunny:
@adrianludwin - I would like to contribute to this issue. Could you please assign this to me?

Thanks,
Sunny

adrianludwin:
Hi Sunny, are you a member of kubernetes-sigs? If not, I can't formally
assign it to you but either way, please feel free to go ahead and start
working on it!

frbimo:
@adrianludwin can i take over this?

tsunny:
@frbimo - Please take over. Sorry, I have been busy with other stuff.

adrianludwin:
Sorry I missed this, yes please go ahead!

PR #1284, opened 2020-11-21
Detect namespace deletion in object validator

tsunny:
Using k8s client in object validator to detect if the
namespace is being deleted. This decouples object validator from
hierarchy_config reconciler

Because fake k8s client-go was needed, all the fake dependencies were
pulled

Testing:
- make test
- make deploy && make test-e2e

Issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/1241

### DIFF
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -8,9 +8,12 @@ import (
 
 	"github.com/go-logr/logr"
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -136,7 +139,7 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 		return allow("source object")
 	}
 	// This is a propagated object.
-	return o.handleInherited(op, newSource, oldSource, inst, oldInst)
+	return o.handleInherited(ctx, op, newSource, oldSource, inst, oldInst)
 }
 
 func validateSelectorUniqueness(inst, oldInst *unstructured.Unstructured) string {
@@ -199,7 +202,7 @@ func validateNoneSelectorChange(inst, oldInst *unstructured.Unstructured) error 
 	return err
 }
 
-func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSource string, inst, oldInst *unstructured.Unstructured) admission.Response {
+func (o *Object) handleInherited(ctx context.Context, op admissionv1beta1.Operation, newSource, oldSource string, inst, oldInst *unstructured.Unstructured) admission.Response {
 	// Propagated objects cannot be created or deleted (except by the HNC SA, but the HNC SA
 	// never gets this far in the validation). They *can* have their statuses updated, so
 	// if this is an update, make sure that the canonical form of the object hasn't changed.
@@ -208,7 +211,7 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
 		return deny(metav1.StatusReasonForbidden, "Cannot create objects with the label \""+api.LabelInheritedFrom+"\"")
 
 	case admissionv1beta1.Delete:
-		if o.isDeletingNS(oldInst) {
+		if o.isDeletingNS(ctx, oldInst) {
 			// There are few things more irritating in (K8s) life than having some stupid controller stop
 			// your namespace from being deleted. If there's an object in here and we've decided that the
 			// namespace should be deleted, then don't block anything!
@@ -244,16 +247,17 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
 	return deny(metav1.StatusReasonInternalError, "unknown operation: "+string(op))
 }
 
-// isDeletingNS returns true if the namespace of the object is already being deleted, based on the
-// in-memory forest.
-func (o *Object) isDeletingNS(inst *unstructured.Unstructured) bool {
-	o.Forest.Lock()
-	defer o.Forest.Unlock()
-	ns := o.Forest.Get(inst.GetNamespace())
-	if ns == nil {
-		return false
+// isDeletingNS returns true if the namespace of the object is already being deleted.
+func (o *Object) isDeletingNS(ctx context.Context, inst *unstructured.Unstructured) bool {
+	ns := &corev1.Namespace{}
+	nnm := types.NamespacedName{Name: inst.GetNamespace()}
+	err := o.client.Get(ctx, nnm, ns)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false
+		}
 	}
-	return ns.IsDeleting
+	return true
 }
 
 // hasConflict checks if there's any conflicting objects in the descendants. Returns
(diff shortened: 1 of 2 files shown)


incubator/hnc/internal/validators/object.go
```diff
@@ -34,10 +35,11 @@ const (
 // +kubebuilder:webhook:path=/validate-objects,mutating=false,failurePolicy=ignore,groups="*",resources="*",verbs=create;update;delete,versions="*",name=objects.hnc.x-k8s.io
 
 type Object struct {
-	Log     logr.Logger
-	Forest  *forest.Forest
-	client  client.Client
```
adrianludwin, line 39: You should be able to use this client instead of the raw K8s client, which means you don't have to import anything new or add any of that extra plumbing.

incubator/hnc/internal/validators/object.go
```diff
@@ -244,16 +246,19 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
...
-// in-memory forest.
-func (o *Object) isDeletingNS(inst *unstructured.Unstructured) bool {
-	o.Forest.Lock()
-	defer o.Forest.Unlock()
-	ns := o.Forest.Get(inst.GetNamespace())
-	if ns == nil {
+// isDeletingNS returns true if the namespace of the object is already being deleted
+func (o *Object) isDeletingNS(ctx context.Context, inst *unstructured.Unstructured) bool {
+
+	ns, err := o.K8sClient.CoreV1().Namespaces().Get(ctx, inst.GetNamespace(), metav1.GetOptions{})
```
adrianludwin, line 252: Using the existing client, this should be something like:

```
ns := &corev1.Namespace{}
nnm := metav1.NamespacedName{Name: inst.GetNamespace()}
err := o.client.Get(nnm, ns)
```

But I'm not 100% sure about all the package names and stuff.

It's probably also ok to skip this check if `o.client` is nil, which will be the case in most of the unit tests.

incubator/hnc/internal/validators/object.go
```diff
@@ -244,16 +246,19 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
...
+	if err != nil && strings.HasSuffix(err.Error(), "not found") {
```
adrianludwin, line 253: Use `IsNotFound`, e.g.: https://github.com/kubernetes-sigs/multi-tenancy/blob/2f6efd498c7eb6b39cbaf3a59b6c60879153c352/incubator/hnc/internal/reconcilers/hierarchy_config.go#L107

incubator/hnc/internal/validators/object.go
```diff
@@ -244,16 +246,19 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
...
+		// Get returns an error if the namespace in not found
 		return false
```
adrianludwin, line 255: What should it mean if the namespace of the object isn't found?

incubator/hnc/internal/validators/object_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/fake"
```
adrianludwin, line 12: You should be able to fake the controller-runtime client fairly easily since it's an interface. The definition is here: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Client. Obviously the only function you need to implement is `Get`.

tsunny: Thanks @adrianludwin for the review. I will work on these and update the PR.

incubator/hnc/internal/validators/object.go
tsunny: Updated in latest commit

incubator/hnc/internal/validators/object.go
tsunny: Updated in latest commit

incubator/hnc/internal/validators/object_test.go
tsunny: Updated in latest commit

incubator/hnc/internal/validators/object.go
tsunny: Updated in latest commit

incubator/hnc/internal/validators/object.go
```diff
@@ -244,16 +247,17 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
...
-// isDeletingNS returns true if the namespace of the object is already being deleted, based on the
-// in-memory forest.
-func (o *Object) isDeletingNS(inst *unstructured.Unstructured) bool {
-	o.Forest.Lock()
-	defer o.Forest.Unlock()
-	ns := o.Forest.Get(inst.GetNamespace())
-	if ns == nil {
-		return false
+// isDeletingNS returns true if the namespace of the object is already being deleted.
+func (o *Object) isDeletingNS(ctx context.Context, inst *unstructured.Unstructured) bool {
```
adrianludwin, line 251: Now that `isDeletingNS` can fail, I'd prefer it to return an error. That way, if there's a transient error, it will be returned all the way up to the user.

incubator/hnc/internal/validators/object.go
```diff
@@ -244,16 +247,17 @@ func (o *Object) handleInherited(op admissionv1beta1.Operation, newSource, oldSo
...
+	ns := &corev1.Namespace{}
+	nnm := types.NamespacedName{Name: inst.GetNamespace()}
+	err := o.client.Get(ctx, nnm, ns)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return false
+		}
```
adrianludwin, line 258: Now that I think about it, IsNotFound will return true if the namespace was _previously_ deleted, not if it's _being_ deleted. But if the namespace is already gone, there won't be any objects in it so this function will never be called.

So I think that if there are any errors returned from the client, we should just return it directly to the caller without caring what it is.

The way to tell if the namespace is actually _being_ deleted is `if !ns.GetDeletionTimestamp().IsZero()`. We set the deletion timestamp when we _start_ deleting an object.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -273,8 +277,9 @@ func TestUserChanges(t *testing.T) {
 			},
 		},
 	}, {
-		name: "Deny deletions of propagated objects",
-		fail: true,
+		name:       "Deny deletions of propagated objects",
+		fail:       true,
+		isDeleting: true,
```
adrianludwin, line 282: This is actually backwards - if the namespace containing the object is being deleted, we should _allow_ the propagated object to be deleted too (since we're not trying to block namespace deletion). BTW this implies that we should have two variants of this test - one with the NS being deleted, and one without.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -493,6 +500,59 @@ func TestUserChanges(t *testing.T) {
 	}
 }
 
+// fakeClient implements a fake client (https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Client).
+type fakeClient struct {
```
adrianludwin, line 504: Make `fakeNSClient` since (as mentioned above) it has to actually return a namespace object?

adrianludwin (commented):
Sorry, I should have noticed this last time, but we want to know if the namespace is _being_ deleted, not if it was _previously_ deleted. That will change a bunch of the logic, which I've noted in more detailed comments below.

Also, the purpose of this change is to get rid of a bunch of unneeded code, specifically the `IsDeleting` field (https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/forest/namespace.go#L55), which is only used here. Can you please delete that field and everything associated with maintaining it? Thanks!

adrianludwin:
Hi @tsunny , are you still planning to submit this? I'm hoping to do another release of HNC in about two weeks and it would be cool if we can get this in. Thanks!

rjbez17:
@adrianludwin  correct me if I'm wrong, but I think the recent merges of https://github.com/kubernetes-sigs/multi-tenancy/pull/1355 makes this PR a duplicate?

adrianludwin:
This was replaced by #1355 

/close

PR #1287, opened 2020-11-23
[VirtualCluster] patch to correct the CRDs generated by controller-gen and use them in demo

brightzheng100:
Two major fixes:
1. Patch in `Makefile` to correct the CRDs generated by the `controller-gen` tool.
2. Use the generated CRDs, instead of temporary ones, in the demo doc.


Fei-Guo:
Thanks for resolving this annoying problem!

/lgtm
/approve

PR #1288, opened 2020-11-24
[VC] Remove dependency on k8s.io/kubernetes

munnerz:
This is a fairly chunky PR, but is mostly mechanical and does not change actual functionality.

It is split into two:

1) removing the dependency on `k8s.io/kubernetes` from all parts of the 'main' codebase - this has involved copying a number of packages into this repository instead of depending on k8s.io/kubernetes directly.

2) moving the `pkg/vn-agent/server` unit tests into their own submodule, as these have a *huge* list of imports of `k8s.io/kubernetes` packages (e.g. port forwarding) used to test the functionality of the vn-agent.

This will allow external consumers to import the `virtualcluster` module *without* then transitively depending on `k8s.io/kubernetes`, which is problematic for many people 😅.

Long term, I think changing that specific set of test files to *not* make use of `k8s.io/kubernetes` internal things would be the best path of action, and there may already be on-going efforts to make the areas of the codebase used here part of a k8s.io staging repository.

/cc @christopherhein @jackkleeman


Fei-Guo:
Thanks for cleaning these up!

/lgtm
/approve

PR #1291, opened 2020-11-26
Fix MustNotRun() can pass without error bug

yiqigao217:
Replace `Eventually()` with `Expect()` in `MustNotRun()` so that commands like
`kubectl create` won't pass for the first time and then eventually fail
without throwing any errors.

Tested by `make test-e2e` (22/32 passed, 10/32 skipped).


adrianludwin (commented):
/lgtm
/approve

PR #1294, opened 2020-12-01
[VC] syncer: support readiness gate condition

zhuangqh:
In most cases, the source of truth is super pod status, because super master actually
own the physical resources, status is reported by **real** kubelet.
Besides, kubernetes allows user-defined conditions, such as pod readiness gate, it will
report readiness state to pod conditions. In other words, the source of truth for user-defined
pod conditions should be tenant side controller, only the condition reported by kubelet should
keep consistent with super.
 It is not recommended that webhook in super append other readiness gate to pod. we left them unchanged
in super, don't sync them upward to tenant side.


christopherhein:
This seems good to me, I haven't personally hit this issue but it looks like you are covering most of the testcases. 👍

PR #1296, opened 2020-12-02
Don't delete the managedBy annotation

adrianludwin:
Prior to this change, HNC removes the v1a1 managedBy annotation, but if
it's being set by an external manager (which, after all, is the only
reason it exists), this will create a war between the manager and HNC.
Ironically, that's exactly what this annotation is trying to prevent.

This change simply removes the code to "clean up" the old annotation.
It should be the responsibility of the external tool to stop setting it.

This change will *not* be cherrypicked into the master branch since
we're removing all the v1a1 conversion code anyway.

Tested: I had to make this change in order to include HNC v0.6.0 in the
latest version of Hierarchy Controller (the GKE product that includes
HNC); this change simply upstreams that fix. Without this change, I
verified that as part of HierarchyController, the resourceVersions of
the namespaces synced by ConfigSync increased rapidly, while with this
change, they remained stable.

Also updated the conversion tests to remove a bunch of flakiness. One
test (ensuring that an object whose source is deleted during conversion
isn't touched) was still flaky in my last full run; I added a sleep to
fix it but I'm not going to rerun the whole 45min test (the test usually
passed in isolation anyway).

### DIFF
--- a/incubator/hnc/test/conversion/conversion_test.go
+++ b/incubator/hnc/test/conversion/conversion_test.go
@@ -16,7 +16,7 @@ var (
 )
 
 const (
-	certsReadyTime = 20
+	certsReadyTime = 30
 	// Some reconciliation may take longer so we have it as 7 seconds, e.g. removing
 	// v1alpha1 from CRD status.storedVersions after CRD conversion because it can
 	// be removed only if all the v1alpha1 CRs are reconciled and converted to v1alpha2.
@@ -94,8 +94,8 @@ var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
 		// Convert
 		setupV1alpha2()
 
-		// Only the new value should be present
-		FieldShouldNotContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managedBy:foo")
+		// The old annotation isn't removed and the new value should be unchanged
+		FieldShouldContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managedBy:foo")
 		FieldShouldContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managed-by:bar")
 	})
 
@@ -108,8 +108,7 @@ var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
 		MustRun("kubectl annotate ns", nsA, "hnc.x-k8s.io/managed-by=bar")
 		MustRun("kubectl annotate ns", nsA, "hnc.x-k8s.io/managedBy=foo") // deprecated in v0.6
 
-		// Verify that the deprecated version is removed
-		FieldShouldNotContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managedBy:foo")
+		// Verify that the deprecated version is not copied to the current version
 		FieldShouldContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managed-by:bar")
 	})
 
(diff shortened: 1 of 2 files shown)


adrianludwin:
Not sure if I'll bother to release a v0.6.1 since I doubt anyone but Hierarchy Controller uses managed-by, but let's at least check this into the tree so that if we do have a release, this fix will make it in.

/assign @yiqigao217

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -94,8 +94,8 @@ var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
 		// Convert
 		setupV1alpha2()
 
-		// Only the new value should be present
-		FieldShouldNotContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managedBy:foo")
+		// The new value should be unchanged
```
yiqigao217, line 97: Maybe add "The old annotation is not removed and the new ....."

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -108,8 +108,8 @@ var _ = Describe("Conversion from v1alpha1 to v1alpha2", func() {
 		MustRun("kubectl annotate ns", nsA, "hnc.x-k8s.io/managed-by=bar")
 		MustRun("kubectl annotate ns", nsA, "hnc.x-k8s.io/managedBy=foo") // deprecated in v0.6
 
-		// Verify that the deprecated version is removed
-		FieldShouldNotContain("ns", "", nsA, ".metadata.annotations", "hnc.x-k8s.io/managedBy:foo")
+		// Verify that the deprecated version is not copied to the current version
+		time.Sleep(1 * time.Second)
```
yiqigao217, line 112: Why we need this 1 second now?

incubator/hnc/test/conversion/conversion_test.go
```diff
@@ -584,6 +589,9 @@ func verifyConversion() {
 	checkCRDVersionInField("v1alpha2", ".spec.versions", true)
 	checkCRDVersionInField("v1alpha1", ".status.storedVersions", false)
 	checkCRDVersionInField("v1alpha2", ".status.storedVersions", true)
+
+	// Make sure HNC is up and running again
+	ensureVWHReady()
```
yiqigao217, line 594: Is this not needed but just to be safer?

yiqigao217 (commented):
>This change will *not* be cherrypicked into the master branch since
we're removing all the v1a1 conversion code anyway.

Do you mean we will remove all v1a1 anyway so this change will be removed then anyway?

yiqigao217:
> > This change will _not_ be cherrypicked into the master branch since
> > we're removing all the v1a1 conversion code anyway.
> 
> Do you mean we will remove all v1a1 anyway so this change will be removed then anyway?

Ah saw it in #1297 specifically.

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: With flaky tests it's hard to tell. After I added this, the amount of flakiness did seem to go down considerably.

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: Hmm, I thought the next function didn't do retries but clearly it did. Removed.

incubator/hnc/test/conversion/conversion_test.go
adrianludwin: Done

PR #1298, opened 2020-12-02
Ensure consistent logging for all validators

### LINKED ISSUE #1138: HNC: validators should log which user is trying to perform an invalid operation
See #1130 . The only log we get when the attempt to delete an object is denied is:

```
{
  "level":"info",
  "ts":1600883338.467028,
  "logger":"validators.Object",
  "msg":"Handled",
  "nm":"",
  "resource":{"group":"rbac.authorization.k8s.io","version":"v1","resource":"rolebindings"},
  "nnm":"child",
  "op":"DELETE",
  "allowed":false,
  "code":403,
  "reason":"Forbidden",
  "message":"Cannot delete object propagated from namespace \"parent\""}
```

Note that this doesn't say _who_ is trying to perform the operation (in this case, the K8s namespace controller). It would be very useful to see that in the logs in the future, both for HNC bugs or to see if someone's trying to evade HNC incorrectly.

/good-first-issue

adrianludwin:
Add the username and operation for all logs.

Tested: deployed with verbose logging and ensured the username and op
were visible for all operations.

Fixes #1138

### DIFF
--- a/incubator/hnc/internal/validators/anchor.go
+++ b/incubator/hnc/internal/validators/anchor.go
@@ -39,7 +39,7 @@ type anchorRequest struct {
 
 // Handle implements the validation webhook.
 func (v *Anchor) Handle(ctx context.Context, req admission.Request) admission.Response {
-	log := v.Log.WithValues("Namespace", req.Namespace, "Name", req.Name)
+	log := v.Log.WithValues("ns", req.Namespace, "nm", req.Name, "op", req.Operation, "user", req.UserInfo.Username)
 	// Early exit since the HNC SA can do whatever it wants.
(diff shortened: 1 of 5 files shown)


yiqigao217 (commented):
Just 1 question - is it okay to have user info in the logs?

adrianludwin:
> Just 1 question - is it okay to have user info in the logs?

Yes - logs already contain highly sensitive information, such as the names of namespaces and objects. Access to logs is typically highly restricted as well.

_Metrics_ are often less restricted so you wouldn't want to put PII into them. But only the cluster admin(s) should be able to access the logs.

yiqigao217:
Thanks for the explanation!

/lgtm

PR #1299, opened 2020-12-02
Exclude kube-node-lease from HNC

### LINKED ISSUE #1229: HNC: Exclude kube-node-lease from HNC
Noticed that we're creating hierarchy configs in this namespace but not the other `kube-` namespaces

adrianludwin:
We shouldn't allow subnamespaces or nontrivial hierarchies in
kube-node-lease, just like it's not allowed in kube-system and a few
others.

Tested: without this change, can create a subns in kube-node-lease; with
this change, I get the appropriate error message. Not adding an
automated test for this since it would just duplicate the contents of
the default_config.go file. All existing e2e tests pass.

Fixes #1229

### DIFF
--- a/incubator/hnc/internal/config/default_config.go
+++ b/incubator/hnc/internal/config/default_config.go
@@ -5,8 +5,9 @@ package config
 // exclude some default namespaces with constantly changing objects.
 // TODO make the exclusion configurable - https://github.com/kubernetes-sigs/multi-tenancy/issues/374
 var EX = map[string]bool{
-	"kube-system":  true,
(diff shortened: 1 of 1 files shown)


yiqigao217 (commented):
/lgtm

Just a question -  when will `kube-node-lease` ns be created? When network policy is enabled?

adrianludwin:
> /lgtm
> 
> Just a question - when will `kube-node-lease` ns be created? When network policy is enabled?

No, AFAIK it's part of any base Kubernetes installation. I think it's used to keep track of the nodes themselves.

PR #1300, opened 2020-12-02
Support unpropagated annotations

adrianludwin:
External systems like ConfigSync use labels and annotations to
understand which objects they should control, just like HNC does. If HNC
creates propagated objects with those labels and annotations, the
external system thinks it owns the objects and can try to modify them
(e.g. delete them - it didn't think they should be created in the first
place so it doesn't know why HNC is creating them).

This commit makes two changes:

* It sets the standard `app.kubernetes.io/managed-by` label on all
  propagated objects. This is very similar to the
  `hnc.x-k8s.io/managed-by` annotation we respect on the namespaces; we
  should probably switch to using this in a future change (Config Sync
  currently sets both).
* It adds an UnpropagatedAnnotations field to the HNCConfiguration
  object. Any annotation listed in this field will not be propagated.
  Config Sync doesn't currently need any unpropagated *labels* but these
  could easily be added if necessary.

Both of these changes are required to get HNC to play well with Config
Sync, with the following three annotations marked as unpropagated (I
believe only the fist one is strictly necessary):

  - configmanagement.gke.io/managed
  - configmanagement.gke.io/declared-config
  - configmanagement.gke.io/token

Tested: since this feature is designed to work as part of an end-to-end
system, most of my testing was end-to-end (apart from the added tests).
Here's what I tested with Config Sync and a simple unstructured repo
with HNC `HierarchyConfiguration` objects checked in:

* I verified that `app.kubernetes.io/managed-by` label is set correctly
  on a few propagated objects.
* In the log, I verified that the HNC webhooks were constantly rejecting
  "delete" requests from Config Sync.
* I added the three annotations shown above to the HNC Config object,
  and verified in the log that the object reconcilers were asked to full
  resync all objects, and that all source objects were indeed resynced,
  and that the requests from Config Sync stopped immediately after the
  objects were updated.
* I verified by hand that none of the three annotations appeared on a
  couple of propagated objects.
* I removed the three annotations from the HNCConfiguration and verified
  that the annotations reappeared on the objects and that Config Sync
  was once again fighting HNC. I then re-added them a final time and
  everything went back to normal.

Any further end-to-end testing will be done on the Hierarchy Controller
side, not in OSS HNC.


adrianludwin:
/hold

I haven't actually added the (fairly basic) integ tests for this yet, but while I'm doing that, I wanted to get feedback on what I have so far.

/assign @rjbez17 
/assign @yiqigao217 
/cc @GinnyJI 

Ryan, any concerns with this?

rjbez17:
> It sets the standard app.kubernetes.io/managed-by label on all
propagated objects. This is very similar to the
hnc.x-k8s.io/managed-by annotation we respect on the namespaces; we
should probably switch to using this in a future change (Config Sync
currently sets both).

Helm and other orchestration tools also set the `app.kubernetes.io/managed-by` label. Would HNC ever change the original (parent) resources labels? May be helpful to have some docs around how/why HNC sets this and more importantly when it will NOT set this. Otherwise HNC may cause issues on upgrades/rollbacks of resources.

rjbez17:
/lgtm 

other than the one question/doc request

adrianludwin:
> Helm and other orchestration tools also set the `app.kubernetes.io/managed-by` label. Would HNC ever change the original (parent) resources labels? May be helpful to have some docs around how/why HNC sets this and more importantly when it will NOT set this. Otherwise HNC may cause issues on upgrades/rollbacks of resources.

No, HNC never modifies any existing objects. It will only set this label on the resources it creates itself. So if Helm created a policy in a parent namespace and that policy object was propagated to the child, the child object would say it was managed by HNC (which I believe is correct), while the parent object would be managed by Helm (also correct). Of course, if Helm removed the parent object, HNC would respond by removing the propagated copy.

Does this make sense? Where you would you recommend documenting this?

adrianludwin:
Tests added

/hold cancel

rjbez17:
> > Helm and other orchestration tools also set the `app.kubernetes.io/managed-by` label. Would HNC ever change the original (parent) resources labels? May be helpful to have some docs around how/why HNC sets this and more importantly when it will NOT set this. Otherwise HNC may cause issues on upgrades/rollbacks of resources.
> 
> No, HNC never modifies any existing objects. It will only set this label on the resources it creates itself. So if Helm created a policy in a parent namespace and that policy object was propagated to the child, the child object would say it was managed by HNC (which I believe is correct), while the parent object would be managed by Helm (also correct). Of course, if Helm removed the parent object, HNC would respond by removing the propagated copy.
> 
> Does this make sense? Where you would you recommend documenting this?

Yeah that was my understanding/assumption, but figured it'd be worth calling out in the Readme?

adrianludwin:
> Yeah that was my understanding/assumption, but figured it'd be worth calling out in the Readme?

Seems a bit detailed for a README - how about a new section in the Concepts part of the doc? We can document _all_ our annotations/labels there.

rjbez17:
> Seems a bit detailed for a README - how about a new section in the Concepts part of the doc? We can document _all_ our annotations/labels there.

I like that idea. It can probably happen outside of this PR, I created an issue here: https://github.com/kubernetes-sigs/multi-tenancy/issues/1301

PR #1303, opened 2020-12-03
[VC] fix: vn-agent socket fd leak

zhuangqh:
don't renew transport object in each handler, reuse it
or it may keep a live connection.

related to https://github.com/golang/go/issues/24719

### DIFF
--- a/incubator/virtualcluster/pkg/vn-agent/server/route.go
+++ b/incubator/virtualcluster/pkg/vn-agent/server/route.go
@@ -17,16 +17,11 @@ limitations under the License.
 package server
 
 import (
-	"crypto/tls"
 	"net/http"
-	"net/url"
 
 	"github.com/emicklei/go-restful"
-
 	"k8s.io/apimachinery/pkg/util/httpstream"
 	"k8s.io/apimachinery/pkg/util/proxy"
-	"k8s.io/client-go/rest"
-	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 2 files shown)


incubator/virtualcluster/pkg/vn-agent/server/route.go
```diff
@@ -119,90 +114,43 @@ func (s *Server) InstallHandlers() {
...
 
-		transport = &http.Transport{
-			TLSClientConfig: &tls.Config{
-				RootCAs: caCrtPool,
-			},
-		}
+		// mutate the request, i.e., replacing the dst, add bearer token header
+		req.Request.URL.Host = s.superAPIServerAddress.Host
+		req.Request.URL.Scheme = "https"
+		req.Request.Header.Add("Authorization", "Bearer "+s.restConfig.BearerToken)
```
christopherhein, line 150: Do all auth providers use bearer tokens for auth?

incubator/virtualcluster/pkg/vn-agent/server/server.go
```diff
@@ -51,5 +57,35 @@ func NewServer(cfg *config.Config) (*Server, error) {
 	}
 
 	server.InstallHandlers()
+
+	if server.config.KubeletClientCert != nil {
+		server.transport = &http.Transport{
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
```
christopherhein, line 64: Why do we skip the verification here?

incubator/virtualcluster/pkg/vn-agent/server/route.go
Fei-Guo: The token is got from the InClusterConfig generated from the Pod service account/secret. Hence I assume it is valid in most of the cases.

christopherhein: @Fei-Guo awesome, thanks!

charleszheng44: Bearer token is not the only way used by Kubernetes to authenticate HTTP requests, but it should work here. Do you mean super apiserver may disable the bearer-token authentication?

charleszheng44 (approved):
LGTM

PR #1304, opened 2020-12-04
Update performance test and add README

yiqigao217:
The performance test is still working from last change, but this commit
replaces the configmaps created in the namespaces with roles and
rolebindings because they are default to propagate by HNCConfig.

Add README to provide more info of what's tested and how to run it.


incubator/hnc/scripts/performance/README.md
```diff
@@ -0,0 +1,59 @@
...
+- "--enable-internal-cert-management"
+- "--cert-restart-on-secret-refresh"
+- "--enable-test-log"        ####### <<<<---- Add this line
+```
+
+Run `make manifests` to update the manifest to include the flag.
+
+Now you can run performance test under the HNC root directory by:
+```
+.../incubator/hnc$. scripts/performance/test.sh
```
adrianludwin, line 43: I'd just start with the dollar sign, because everyone will have their shell configured a different way:

```
$ scripts/performance/test.sh
```

Also - what's that dot?

incubator/hnc/scripts/performance/README.md
```diff
@@ -0,0 +1,59 @@
...
+Current time : 2020-12-04T15:42:22
+Controllers start working time : 2020-12-04T15:42:14
+Controllers finish working time : 2020-12-04T15:42:19
+Controllers working time for HNC startup: 5s
+Total HierConfig reconciles: 33
+Total Object reconciles: 217
+Total HierConfig writes: 21
+Total Namespace writes: 21
+Total Object writes: 50
+```
```
adrianludwin, line 59: Maybe add a section on how you're measuring memory. You only know how to do this on GKE so you can just say something like "On GKE, ..."

adrianludwin (commented):
lgtm after a few small nits

incubator/hnc/scripts/performance/README.md
yiqigao217: I would get `Permission denied` without prefixing the script with `. `.

adrianludwin: Have you made it executable? `chmod +x path/to/script`.

rjbez17: If I had to guess, I assume you need to do `chomd +x scripts/performance/test.sh` to fix the no dot issue.

incubator/hnc/scripts/performance/test.sh
```diff
@@ -33,3 +33,6 @@ do
   kubectl -n hnc-system delete deployment hnc-controller-manager
   . ./scripts/performance/start-controllers.sh
 done
+
+# Clean up topologies
+. ./scripts/performance/clean-up-topologies.sh
```
rjbez17, line 38: using the `.` here means your're sourcing the file instead of calling it. The subtle difference there means that you could get clashes in variables when sourcing multiple files since your bringing them all into the main context of `test.sh`. If that isn't desired, I don't think we want it sourced. You probably just need to make these scripts executable to not have to source them.

yiqigao217: Thanks for the explanation! Will make them in to executable.

incubator/hnc/scripts/performance/README.md
yiqigao217: Ah thank you both, no. Made all scripts executable now!

rjbez17:
/lgtm
/approve

/hold 

for @adrianludwin

PR #1305, opened 2020-12-04
Create client for checking CRD deletion

### LINKED ISSUE #1285: HNC: subnamespaces are deleted because we don't realize CRDs are being deleted
The "should not delete subnamespace" test in `delete_crd_tests.go` is flaky and fails maybe 20% of the time. This is fairly important because this is our only line of defense when uninstalling HNC - when you delete the CRDs, the subnamespaces are _not_ supposed to be deleted.

According to the logs, HNC simply fails to notice that the CRDs are being deleted. A subsequent reconcile about 0.1s later _does_ notice that the CRDs are being deleted, but by then it's too late - we've already started deleting the subnamespace. In other words, there's a race condition.

I'm asking some apiserver experts if there's a better way to solve this.

GinnyJI:
The default client uses cache, and therefore, it might fail to notice if
a CR is deleted. We instead use our own client who doesn't use the
cache.

Tested: make test-e2e

Fix #1285

### DIFF
--- a/incubator/hnc/internal/reconcilers/setup.go
+++ b/incubator/hnc/internal/reconcilers/setup.go
@@ -9,11 +9,21 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 )
+
+// FakeDeleteCRDClient is a "fake" client used for testing only
+type FakeDeleteCRDClient struct{}
+
+// FakeDeleteCRDClient doesn't return any err on Get() because none of the reconciler test performs CRD deletion
+func (f FakeDeleteCRDClient) Get(context.Context, types.NamespacedName, runtime.Object) error {
+	return nil
+}
 
 var crds = map[string]bool{
 	"hncconfigurations.hnc.x-k8s.io":       true,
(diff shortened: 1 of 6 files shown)


adrianludwin:
Just a few nits but otherwise lgtm

incubator/hnc/internal/reconcilers/setup.go
```diff
@@ -9,25 +9,51 @@ import (
...
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
 	api "sigs.k8s.io/multi-tenancy/incubator/hnc/api/v1alpha2"
 	"sigs.k8s.io/multi-tenancy/incubator/hnc/internal/forest"
 )
 
+type FakeDeleteCRDClient struct{}
```
adrianludwin, line 20: Add some documentation?

incubator/hnc/internal/reconcilers/setup.go
```diff
@@ -9,25 +9,51 @@ import (
...
+
+func (f FakeDeleteCRDClient) Get(context.Context, types.NamespacedName, runtime.Object) error {
```
adrianludwin, line 22: Maybe document why this is the correct behaviour for the fake client?

incubator/hnc/internal/reconcilers/setup.go
```diff
@@ -9,25 +9,51 @@ import (
...
+	return nil
+}
+
 var crds = map[string]bool{
 	"hncconfigurations.hnc.x-k8s.io":       true,
 	"subnamespaceanchors.hnc.x-k8s.io":     true,
 	"hierarchyconfigurations.hnc.x-k8s.io": true,
 }
 
+var deleteCRDClient interface {
```
adrianludwin, line 32: I had to think about this a bit - it might be nice to define this in two parts, the interface and the variable.

```
type interface deleteCRDClientType { ... }
var deleteCRDClient deleteCRDClientType
```

with a bit more commenting for both of them.

incubator/hnc/internal/reconcilers/setup.go
```diff
@@ -9,25 +9,51 @@ import (
...
 	hcChan := make(chan event.GenericEvent)
 	anchorChan := make(chan event.GenericEvent)
 
+	// Create uncached client for CRD deletion check
+	if !useFakeClient {
+		// declaring err variable here to make sure we are using global variable deleteCRDClient later
+		var err error
+		deleteCRDClient, err = client.New(config.GetConfigOrDie(), client.Options{
+			Scheme: mgr.GetScheme(),
+			Mapper: mgr.GetRESTMapper(),
```
adrianludwin, line 49: We weren't sure if this was actually needed - maybe add a comment that says we don't know, or maybe test taking this out and see if it still works :D (though that might be overkill).

incubator/hnc/internal/reconcilers/setup.go
```diff
@@ -9,25 +9,51 @@ import (
...
 //
 // This function is called both from main.go as well as from the integ tests.
-func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int) error {
+func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int, useFakeClient bool) error {
 	hcChan := make(chan event.GenericEvent)
 	anchorChan := make(chan event.GenericEvent)
 
+	// Create uncached client for CRD deletion check
+	if !useFakeClient {
+		// declaring err variable here to make sure we are using global variable deleteCRDClient later
```
adrianludwin, line 45: I'm not sure you need this comment, it's a fairly common Go idiom to declare err outside of an if statement when needed.

adrianludwin (commented):
Whoops, forgot to include my actual comments last week! Here they are - a few nits still but mainly lgtm.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 

Hey Ryan, since we can't reliably reproduce this problem, we don't have a test proving that it's fixed, but using an uncached client seems like the right thing to do in this case. Does this sound reasonable to you?

incubator/hnc/internal/reconcilers/setup.go
```diff
@@ -9,25 +9,57 @@ import (
...
+}
+
+// deleteCRDClient is an uncached client for checking CRD deletion
+var deleteCRDClient deleteCRDClientType
+
 // Create creates all reconcilers.
 //
 // This function is called both from main.go as well as from the integ tests.
-func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int) error {
+func Create(mgr ctrl.Manager, f *forest.Forest, maxReconciles int, useFakeClient bool) error {
```
rjbez17, line 45: Since the client is created 100% by variables/info available outside this function, wouldn't it make more sense to just pass the client into this function instead of using a special bool to create a fake client or not?

adrianludwin: It's a good point, although in this context I'm not sure it's worth the added complexity. The interface and global variable would still have to be declared in this file, while the clients that implement it would now be in cmd/manager/main.go and internal/reconcilers/suite_test.go, so the logic would be fairly scattered. The way @GinnyJI has written it, pretty much everything you need to know is in the same place, and I think that simplicity is worthwhile.

If we start adding a bunch more options here, then I'd be open to refactoring it. wdyt?

rjbez17: Yeah, that sounds good.

PR #1306, opened 2020-12-04
Increase memory limits and requests and update docs

yiqigao217:
Increase memory limits to 300M and requests to 150M since we store more
info, e.g. source objects, in the forest now. About 700 namespaces with
10 propagatable objects in each namespace would use about 200M memory
from performance tests.

Add memory usage in user-guide faq docs.

### DIFF
--- a/incubator/hnc/config/manager/manager.yaml
+++ b/incubator/hnc/config/manager/manager.yaml
@@ -49,8 +49,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 100Mi
+            memory: 300Mi
           requests:
             cpu: 100m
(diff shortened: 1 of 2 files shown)


rjbez17:
Do we have any docs anywhere describing how an end user would determine the recommended limit/request sizes?

rjbez17:
This lgtm other than the one question about docs.

adrianludwin:
Discussed with @yiqigao217 - she's going to add our performance and memory testing results to the FAQ, and also increase the limits slightly as these limits are pretty tight for at least one known real usecase. We're going to go with 150MB request and 300MB limit.

PR #1308, opened 2020-12-07
Tool: map abstract to hierarchical namespaces

adrianludwin:
This is an experimental tool to inspect a directory structure containing
abstract namespaces (from GKE Config Sync or Anthos Config Management)
and add HNC objects so the directory can be used with `kubectl apply -f`
and recreate the directory structure using hierarchical namespaces.

Tested: ran the tool as described in the readme on the examples/simple-1
directory structure and verified that, when applied to a cluster, all
the expected hierarchical relationships were created.


adrianludwin (commented):
/assign @yiqigao217 
/assign @rjbez17 

Hey Ryan, we have a customer who's interested in switching between GKE Config Sync "abstract namespaces" (the inspiration for HNC) and HNC itself. Any concerns if we leave put this into our repo and see if anyone else is interested in it?

incubator/hnc/tools/ans2hns/README.md
```diff
@@ -0,0 +1,80 @@
...
+  directory structure if they're missing/incorrect.
+
+This tool does _not_ currently understand multiple repos or [namespace
+selectors](https://cloud.google.com/kubernetes-engine/docs/add-on/config-sync/concepts/namespace-inheritance#excluding_namespaces_from_inheritance).
+Namespace selectors must be replaced by HNC Exceptions (link to come when
+published) by hand.
+
+## Running the tool
+
+Ensure you have `kpt` installed and available in your path.
```
rjbez17, line 41: nit: adding a url to `kpt` would be useful (I had never heard of it).

https://googlecontainertools.github.io/kpt/

rjbez17 (commented):
I don't have reason to not be ok with this. However, it may be worthwhile in the future to have a more robust approach to migrations to HNC from other options and natively running large clusters (how can they adopt HNC, probably mostly docs).

rjbez17:
lgtm after the one nit.

incubator/hnc/tools/ans2hns/README.md
adrianludwin: Good call, done.

adrianludwin:
> I don't have reason to not be ok with this. However, it may be worthwhile in the future to have a more robust approach to migrations to HNC from other options and natively running large clusters (how can they adopt HNC, probably mostly docs).

Yeah I'm not sure if this is the first example in a pattern, or a one-off that we'll never use again. I think that we can revisit if we start to see other similar usecases - e.g. we could build this into `kubectl-hns` or write a more complete migration guide. sgty?

rjbez17:
> > I don't have reason to not be ok with this. However, it may be worthwhile in the future to have a more robust approach to migrations to HNC from other options and natively running large clusters (how can they adopt HNC, probably mostly docs).
> 
> Yeah I'm not sure if this is the first example in a pattern, or a one-off that we'll never use again. I think that we can revisit if we start to see other similar usecases - e.g. we could build this into `kubectl-hns` or write a more complete migration guide. sgty?

Agreed, I don't see any reason to preemptively doing this until we know there is a large need/request.

PR #1309, opened 2020-12-08
Generate event for exceptions

### LINKED ISSUE #1165: HNC: exceptions error handling
We should surface the potential errors in exceptions. Eg: when the annotation could not be parsed to selector.

GinnyJI:
If the reconciler cannot parse the selector annotation to a valid
selector, we record an event so that the user will get notified.

Tested:
```
$ k create ns a
$ k hns create a1 -n a
$ k delete validatingwebhookconfigurations.admissionregistration.k8s.io hnc-validating-webhook-configuration
$ k create role test -n a --verb=update --resource=deployments
$ k annotate role test -n a "propagate.hnc.x-k8s.io/none"="**foo"
$ k get event -n a
LAST SEEN   TYPE      REASON                OBJECT      MESSAGE
41s         Warning   CannotParseSelector   role/test   invalid propagate.hnc.x-k8s.io/none value: strconv.ParseBool: parsing "**foo": invalid syntax
```

Fix #1165

### DIFF
--- a/incubator/hnc/api/v1alpha2/hierarchy_types.go
+++ b/incubator/hnc/api/v1alpha2/hierarchy_types.go
@@ -85,8 +85,11 @@ const (
 	// EventCannotUpdate is for events when a namespace has an object that couldn't
 	// be propagated *into* this namespace - that is, it couldn't be created in
 	// the first place, or it couldn't be updated. The error message will point to
-	//the source namespace.
+	// the source namespace.
 	EventCannotUpdate string = "CannotUpdateObject"
+	// EventCannotGetSelector is for events when an object has annotations that cannot be
+	// parsed into a valid selector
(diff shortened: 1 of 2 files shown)


yiqigao217 (commented):
AFAIR there are more places to surface the error event than just 1 place?

adrianludwin:
> AFAIR there are more places to surface the error event than just 1 place?

Oh good point. Maybe this logic should be moved into `selectors.ShouldPropagate()` so that we don't miss another place like this?

GinnyJI:
> > AFAIR there are more places to surface the error event than just 1 place?
> 
> Oh good point. Maybe this logic should be moved into `selectors.ShouldPropagate()` so that we don't miss another place like this?

`selectors` doesn't have an `eventRecorder`. This is actually the only place that we call `shouldPropagate()` in the reconciler, the other places are in the validator

PR #1310, opened 2020-12-09
Set unpropagated annotations via command line args

adrianludwin:
The list of unpropagated annotations should very rarely be changed, and
controlling them via the HNCConfiguration object has a bunch of problems
(e.g. merging fields set by an external operator, like the Config Sync
operator, with user-specified fields). As a result, I think it makes
more sense to specify these as command-line params instead. This also
simplifies a bunch of logic.

Tested: verified that the annoations are removed if specified in the
command line args in the Deployment and that HNC plays well with Config
Sync as a result. Migrated the HNCConfiguration integ test to the object
reconciler instead.


adrianludwin:
/hold
/assign @rjbez17 
/assign @yiqigao217 

Let's not submit this just yet, but wdyt of this change? Sorry to partially undo my last change but the prospect of automatically managing these fields in a K8s resource that's also human-editable didn't seem too appealing, and command-line args are _much_ easier for our usecase. Pity I didn't figure that out last week :P

incubator/hnc/cmd/manager/main.go
```diff
@@ -86,6 +88,8 @@ func main() {
 	flag.IntVar(&qps, "apiserver-qps-throttle", 50, "The maximum QPS to the API server.")
 	flag.BoolVar(&stats.SuppressObjectTags, "suppress-object-tags", true, "If true, suppresses the kinds of object metrics to reduce metric cardinality.")
 	flag.IntVar(&webhookServerPort, "webhook-server-port", 443, "The port that the webhook server serves at.")
+	uaArg := arrayArg{val: &config.UnpropagatedAnnotations}
+	flag.Var(&uaArg, "unpropagated-annotation", "An annotation that should not be propagated by HNC. May be specified multiple times.")
```
rjbez17, line 92: I think this description could be clarified a bit to indicate the annotations are stripped from propagated resources. It currently could also be read as (IMO at least), if this annotation exists, the resources will not be propagated.

It may also be worthwhile to indicate restarting HNC with different values will in fact affect already propagated resources.

rjbez17 (commented):
lgtm after the one nit

incubator/hnc/cmd/manager/main.go
adrianludwin: sg - I'll add proper documentation and a link.

yiqigao217: Not sure if it should just use plural in the description "Annotations that..", since the var is an array itself.

incubator/hnc/cmd/manager/main.go
```diff
@@ -199,3 +203,22 @@ func startControllers(mgr ctrl.Manager, certsCreated chan struct{}) {
 
 	setupLog.Info("All controllers started; setup complete")
 }
+
+// arrayArg is an arg that can be specified multiple times. It implements
+// https://golang.org/pkg/flag/#Value an is based on
+// https://stackoverflow.com/questions/28322997/how-to-get-a-list-of-values-into-a-flag-in-golang.
```
yiqigao217, line 209: Per "can be specified multiple times", does it mean people can both insert or delete annotation from the array? Seems there's no deletion behavior in the code?

s/an/and

incubator/hnc/cmd/manager/main.go
adrianludwin: Any _one_ use of the flag is singular - you can just use the flag multiple times, e.g.:

`--unpropagated-annotation=acme.com/foo --unpropagated-annotation=acme.com/bar`

So singular is correct IMO, since the arg to this flag is only a single annotation.

incubator/hnc/cmd/manager/main.go
adrianludwin: No, this is only specified during startup so there's no "delete" behaviour. It just means that you can have the flag repeated multiple times, instead of only having the flag once and making it comma-delimited.

yiqigao217: sg!

incubator/hnc/cmd/manager/main.go
yiqigao217: sg

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -529,6 +534,54 @@ var _ = Describe("Secret", func() {
...
+			annots := inst.GetAnnotations()
+			if annots["annot-a"] != "value-a" {
+				return fmt.Errorf("annot-a: want 'value-a', got %q", annots["annot-a"])
+			}
+			if annots["annot-b"] != "value-b" {
+				return fmt.Errorf("annot-b: want 'value-b', got %q", annots["annot-b"])
+			}
+			return nil
+		}).Should(Succeed(), "waiting for initial sync of foo-annot-role")
+		deleteObject(ctx, "roles", fooName, "foo-annot-role")
```
yiqigao217, line 560: Hmm, why can we delete object in the integration test?

Also, maybe add validation to confirm both the source and the copy are gone before creating new object with the same name, or consider creating object with a different name.

yiqigao217 (commented):
lgtm other than 1 last question

incubator/hnc/internal/reconcilers/object_test.go
adrianludwin: It's only namespaces that we can't delete - we've always been able to delete individual objects.

I modified `deleteObject` to only return when the object is fully gone (the source object, in this case). I don't think we need to wait for the copy to be deleted, because the test on line 577 will fail until it's gone - but it's in an Eventually block so we'll just try again until it's been replaced.

yiqigao217: sg.

yiqigao217:
/lgtm
/hold

for @rjbez17

incubator/hnc/cmd/manager/main.go
adrianludwin: Added new documentation in https://github.com/adrianludwin/multi-tenancy/blob/unprop-args/incubator/hnc/docs/user-guide/how-to.md#admin-cli-args and tweaked the descriptions as described.

adrianludwin:
/hold cancel

@rjbez17 lgty?

PR #1314, opened 2020-12-10
Generate full manifest last

adrianludwin:
The makefile calls kustomize twice in the manifests directory, and
leaves the last kustomization.yaml file there when it finishes. This
change swaps the order so that the more useful file (the one that
generates hnc-manager.yaml) is the one that's left so that developers
can play with it if necessary.

Tested: both yaml files are still created and the expected
kustomization.yaml file is left behind.


yiqigao217 (commented):
/lgtm
/hold cancel

PR #1316, opened 2020-12-11
Validate HNC selector annotation

### LINKED ISSUE #1206: HNC: validate `hnc.x-k8s.io` annotations
We should check for any unwanted `hnc.x-k8s.io` fields. For example, we are looking for `propagate.hnc.x-k8s.io/select` but NOT `propagate.hnc.x-k8s.io/selector`. We should prevent the user from setting the latter one.

GinnyJI:
As of right now, the only prefix we allow before `hnc.x-k8s.io` is
`propagate`, and the only suffix after this combination should be
`select`, `treeSelect` or `none`. Any other prefix should or other 
selector suffix be rejected.

Tested: make test; make test-e2e

Fix #1206

### DIFF
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -111,6 +111,10 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	// If the object wasn't and isn't inherited, we will check to see if the
 	// source can be created without causing any conflict.
 	if !oldInherited && !newInherited {
+		// check if there is any invalid HNC annotation
+		if msg := validateSelectorAnnot(inst); msg != "" {
+			return deny(metav1.StatusReasonBadRequest, msg)
+		}
 		// check selector format
 		// If this is a selector change, and the new selector is not valid, we'll deny this operation
 		if err := validateSelectorChange(inst, oldInst); err != nil {
@@ -137,6 +141,33 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	}
 	// This is a propagated object.
 	return o.handleInherited(op, newSource, oldSource, inst, oldInst)
+}
+
+func validateSelectorAnnot(inst *unstructured.Unstructured) string {
+	annots := inst.GetAnnotations()
+	for key, _ := range annots {
+		// for example: segs = ["propagate.hnc.x-k8s.io", "select"]
+		segs := strings.SplitN(key, "/", 2)
+		// for example: prefix = ["propagate", "hnc.x-k8s.io"]
+		prefix := strings.SplitN(segs[0], ".", 2)
+		if len(prefix) < 2 || prefix[1] != api.MetaGroup {
+			continue
+		}
+		msg := "invalid HNC exceptions annotation: %v, should be one of the following: " +
+			api.AnnotationSelector + "; " + api.AnnotationTreeSelector + "; " +
+			api.AnnotationNoneSelector
+		// If this annotation is part of HNC metagroup, we check if the prefix value is valid
+		if segs[0] != api.AnnotationPropagatePrefix {
+			return fmt.Sprintf(msg, key)
+		}
+		// check if the suffix is valid by checking the whole annotation key
+		if key != api.AnnotationSelector &&
+			key != api.AnnotationTreeSelector &&
+			key != api.AnnotationNoneSelector {
+			return fmt.Sprintf(msg, key)
+		}
+	}
+	return ""
 }
 
 func validateSelectorUniqueness(inst, oldInst *unstructured.Unstructured) string {
--- a/incubator/hnc/internal/validators/object_test.go
+++ b/incubator/hnc/internal/validators/object_test.go
@@ -309,6 +309,48 @@ func TestUserChanges(t *testing.T) {
 			},
 		},
 	}, {
+		name: "Deny creation of object with invalid HNC annotation prefix",
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						"invalid.hnc.x-k8s.io/select": "foo",
+					},
+				},
+			},
+		},
+	}, {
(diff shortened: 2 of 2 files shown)


adrianludwin:
> Tested: make test

I'd feel a bit better if you could run the e2e tests too, I think.

incubator/hnc/internal/validators/object.go
```diff
@@ -139,6 +143,30 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
 	return o.handleInherited(op, newSource, oldSource, inst, oldInst)
 }
 
+func validateSelectorAnnot(inst *unstructured.Unstructured) string {
+	annots := inst.GetAnnotations()
+	for key, _ := range annots {
+		// for example: segs = ["propagate.hnc.x-k8s.io", "select"]
+		segs := strings.SplitN(key, "/", 2)
+		// for example: prefix = ["propagate", "hnc.x-k8s.io"]
+		prefix := strings.SplitN(segs[0], ".", 2)
```
adrianludwin, line 152: Will this capture anything _without_ a prefix - e.g., `hnc.x-k8s.io/foo`?

incubator/hnc/internal/validators/object.go
```diff
@@ -139,6 +143,30 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
+		}
+		// If this annotation is part of HNC metagroup, we check if the prefix value is valid
+		if segs[0] != api.AnnotationPropagatePrefix {
+			return fmt.Sprintf("invalid HNC annotation: %v", key)
+		}
+		// check if the suffix is valid by checking the whole annotation key
+		if key != api.AnnotationSelector &&
+			key != api.AnnotationTreeSelector &&
+			key != api.AnnotationNoneSelector {
+			return fmt.Sprintf("invalid HNC annotation: %v", key)
```
adrianludwin, line 164: I think it would be more user-friendly to print out the valid options here.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -308,6 +308,20 @@ func TestUserChanges(t *testing.T) {
...
+	}, {
+		name: "Deny creation of object with invalid HNC annotation",
+		fail: true,
+		inst: &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "v1",
+				"kind":       "Pod",
+				"metadata": map[string]interface{}{
+					"annotations": map[string]interface{}{
+						"invalid.hnc.x-k8s.io/select": "foo",
```
adrianludwin, line 320: I'd try a few other example, e.g.:

```
propagate.hnc.x-k8s.io/slct // misspelled part after the slash
hnc.x-k8s.cio/foo // unknown annotation
```

incubator/hnc/internal/validators/object.go
GinnyJI: No it won't. This function is only to catch invalid selector annotations

incubator/hnc/internal/validators/object.go
```diff
@@ -139,6 +143,30 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op admissionv1beta
...
 }
 
+func validateSelectorAnnot(inst *unstructured.Unstructured) string {
+	annots := inst.GetAnnotations()
+	for key, _ := range annots {
+		// for example: segs = ["propagate.hnc.x-k8s.io", "select"]
+		segs := strings.SplitN(key, "/", 2)
+		// for example: prefix = ["propagate", "hnc.x-k8s.io"]
+		prefix := strings.SplitN(segs[0], ".", 2)
+		if prefix[1] != api.MetaGroup {
```
yiqigao217, line 153: Seems it would get `panic: runtime error: index out of range [1] with length 1` if the annotation `key` is a plain word without a `.` in it.

incubator/hnc/internal/validators/object_test.go
yiqigao217: +1 and one more plain word annotation key? e.g. "plainword".

incubator/hnc/internal/validators/object.go
GinnyJI: I thought we were only checking the exceptions annotation here, so I'll only print valid exceptions annotations, is that correct?

incubator/hnc/internal/validators/object_test.go
GinnyJI: @yiqigao217 do you mean `plainword.hnc.x-k8s.io/select`?

incubator/hnc/internal/validators/object.go
adrianludwin: Discussed offline - this is more complicated to add so let's leave it as a possible future enhancement.

incubator/hnc/internal/validators/object_test.go
adrianludwin: Agreed that the second example (`hnc.x-k8s.io/foo`) is not a part of this change. My apologies for misspelling it in the original exampl!

GinnyJI: done the testcase with `plainword`

GinnyJI:
> > Tested: make test
> 
> I'd feel a bit better if you could run the e2e tests too, I think. `make test` doesn't actually have any of the validators running - at most, it tests some of the internal logic of the validators, but not that the validators themselves actually work.

Passed all E2E tests

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 

Nice!

adrianludwin:
> > I'd feel a bit better if you could run the e2e tests too, I think. `make test` doesn't actually have any of the validators running - at most, it tests some of the internal logic of the validators, but not that the validators themselves actually work.
> 
> Passed all E2E tests

If you have a chance can you add this to the commit message for posterity? But it's not critical if Ryan approves this first.

GinnyJI:
> > > I'd feel a bit better if you could run the e2e tests too, I think. `make test` doesn't actually have any of the validators running - at most, it tests some of the internal logic of the validators, but not that the validators themselves actually work.
> > 
> > 
> > Passed all E2E tests
> 
> If you have a chance can you add this to the commit message for posterity? But it's not critical if Ryan approves this first.

done

adrianludwin:
(I won't start on v0.7 until this is in at this point)

PR #1331, opened 2020-12-22
[VC] Add featuregates cmdline support in syncer

Fei-Guo:
We will enhance syncer to support various new features such as enabling a super cluster pool to serve multiple tenants. 

This change adds feature-gate command line support to the syncer.


christopherhein:
Potentially consider using the component base feature flag libraries. Otherwise looks good! ￼￼ https://github.com/kubernetes/component-base/tree/master/featuregate

PR #1332, opened 2020-12-23
[VC] Identify supercluster via configmap

Fei-Guo:
When supporting multiple super masters, the syncer needs to identify the super cluster it is serving for to decide if an object marked with super cluster owner in the tenant master should be synced or not. 

Since K8s does not provide a reliable UUID for a cluster, we leverage a configmap created by the super cluster administrator to identify the super cluster. The configmap is stored in kube-system namespace and uses the predefined name (we can make it configurable later if needed). The syncer will not verify the ID and relies on the administrator to guarantee its uniqueness. 

This change also renames the feature name for better readability.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/syncer.go
+++ b/incubator/virtualcluster/pkg/syncer/syncer.go
@@ -17,8 +17,10 @@ limitations under the License.
 package syncer
 
 import (
+	"context"
 	"fmt"
 	"net/http"
+	"os"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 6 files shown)


incubator/virtualcluster/pkg/syncer/syncer.go
```diff
@@ -50,15 +53,23 @@ import (
...
 )
 
 var (
 	numHealthCluster   uint64
 	numUnHealthCluster uint64
+	SuperClusterID     string
+)
+
+const (
+	SuperClusterInfoCfgMap = "superclusterinfo"
```
zhuangqh, line 66: what about `super-cluster-info`

PR #1340, opened 2021-01-13
[VC] Add supercluster reconciling loop

Fei-Guo:
This change adds supercluster reconciling loop. The implementation is very similar to that of virtualcluster reconciling loop.

We probably can avoid some duplicated code by introducing interfaces but the gain is moderate since we only have two types of cluster for now. 

In next change, I will refactor the cluster interface so that it can support both VirtualCluster and Cluster CRDs.

### DIFF
--- a/incubator/virtualcluster/experiment/pkg/scheduler/scheduler.go
+++ b/incubator/virtualcluster/experiment/pkg/scheduler/scheduler.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


incubator/virtualcluster/experiment/pkg/scheduler/scheduler.go
```diff
@@ -157,10 +210,14 @@ func (s *Scheduler) Run(stopChan <-chan struct{}) {
 			return
 		}
 
-		klog.V(5).Infof("starting scheduler workers")
+		klog.Infof("starting scheduler virtualcluster workers")
 		for i := 0; i < s.virtualClusterWorkers; i++ {
 			go wait.Until(s.virtualClusterWorkerRun, 1*time.Second, stopChan)
 		}
+		klog.Infof("starting scheduler virtualcluster workers")
```
zhuangqh, line 217: split into two go routine. waitForCache+start worker for each goroutine.
and you are missing `defer s.superClusterQueue.ShutDown() ` here

Fei-Guo: Good catch! Fixed.

PR #1342, opened 2021-01-14
[VC] Make Cluster and MultiClusterController generic

Fei-Guo:
This change removes vc specific method (i.e. GetSpec, and GetClusterDomain) from Cluster and MultiClusterController interfaces.  Instead, a GetObject method is added to retrieve the CRD obj. The type casting is done in the caller side when accessing the CRD obj. 

Now Cluster and MultiClusterController interfaces are decoupled with VC crd, and the same code framework can be used to manage multiple super clusters.


zhuangqh:
good job 👍

PR #1346, opened 2021-01-18
[VC] Fix a field name conflict in Cluster

Fei-Guo:
The Cache Option has a field name of "Namespace" which is conflict with the Cluster "Namespace" field. Hence the VC namespace name is mistakenly used in the cache watch option. This change fixes the bug.


zhuangqh:
please make your commit message meaningful😄

PR #1349, opened 2021-01-21
[VC] Introduce basic structures (Pod, Namespace, Cluster) for scheduler cache

Fei-Guo:
This change introduces basic structures in scheduler cache. They are Pod, Namespace and Cluster.

The basic model is that a namespace quota is divided into multiple slices based on the quota slice size. The slices are assigned to different clusters. A Pod is assigned to a cluster based on the quota slices distribution of the corresponding namespace.  The cluster object will track the overall slice allocation.

All objects will be protected by a cache lock which will be introduced in future change hence we don't have per object lock.

### DIFF
--- a/incubator/virtualcluster/experiment/pkg/scheduler/cache/cluster.go
+++ b/incubator/virtualcluster/experiment/pkg/scheduler/cache/cluster.go
@@ -1,0 +1,158 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type Cluster struct {
+	name     string
+	labels   map[string]string
+	capacity v1.ResourceList
+
+	alloc  v1.ResourceList
+	slices map[string][]*Slice            // ns key -> slice array
+	pods   map[string]map[string]struct{} // ns key -> Pod map
+}
+
+func NewCluster(name string, labels map[string]string, capacity v1.ResourceList) *Cluster {
+	alloc := capacity.DeepCopy()
+
+	for k, v := range alloc {
+		v.Set(0)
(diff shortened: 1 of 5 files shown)


incubator/virtualcluster/experiment/pkg/scheduler/cache/cluster.go
```diff
@@ -0,0 +1,153 @@
...
+	for k, v := range c.labels {
+		labelcopy[k] = v
+	}
+
+	out := NewCluster(c.name, c.labels, c.capacity.DeepCopy())
+
+	slicesCopy := make(map[string][]*Slice)
+
+	for k, v := range c.slices {
+		s := make([]*Slice, len(v))
```
zhuangqh, line 65: s := make([]*Slice, 0, len(v))

incubator/virtualcluster/experiment/pkg/scheduler/cache/cluster.go
```diff
@@ -0,0 +1,153 @@
...
+		for _, each := range v {
+			s = append(s, each.DeepCopy())
+		}
+		slicesCopy[k] = s
+	}
+
+	podsCopy := make(map[string]map[string]struct{})
+	for k, v := range c.pods {
```
zhuangqh, line 73: podsCopy[k] = make(map[string]struct{})  or you got panic here

incubator/virtualcluster/experiment/pkg/scheduler/cache/cluster.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+	out.slices = slicesCopy
+	out.pods = podsCopy
+	return out
+}
+
+func (c *Cluster) AddNamespace(key string, slices []*Slice) error {
+	if _, ok := c.slices[key]; ok {
+		return fmt.Errorf("Namespace key %s is already in cluster %s, cannot add twice", key, c.name)
```
zhuangqh, line 85: https://github.com/golang/go/wiki/CodeReviewComments#error-strings
Error string should not be capitalized or end with punctuation

incubator/virtualcluster/experiment/pkg/scheduler/cache/cluster.go
```diff
@@ -0,0 +1,153 @@
...
+	c.alloc = allocCopy
+	delete(c.slices, key)
+	return nil
+}
+
+func (c *Cluster) AddPod(pod *Pod) error {
+	key := pod.GetNamespaceKey()
+	if pod.cluster != c.name {
+		return fmt.Errorf("%s's Pod %s should not be placed in cluster %v (scheduled to %v) ", pod.GetNamespaceKey(), pod.name, c.name, pod.cluster)
+	}
```
zhuangqh, line 142: check if c.pod[key] exists, or c.pod[key] = make(map[string]struct{}) first

Fei-Guo:
Thanks for your careful reviews. All comments are addressed and a new UT is added to test DeepCopy.

PR #1350, opened 2021-01-21
HNC produces double colons on error message

### LINKED ISSUE #1218: HNC: metav1.StatusReasonInvalid produce double colons when formatting error message
metav1.StatusReasonInvalid produces `The request is invalid: : ` which has repeated colons

frbimo:
fixes #1218

tests:
make test
make kind-deploy

### DIFF
--- a/incubator/hnc/internal/validators/hierarchy.go
+++ b/incubator/hnc/internal/validators/hierarchy.go
@@ -15,6 +15,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


adrianludwin (commented):
This lgtm but I'd like @yiqigao217 to take a look as well. Thanks for the change!

/assign @yiqigao217 
/approve

yiqigao217 (commented):
The func changes lgtm! But it seems you will also need to change the callers who has, e.g. `deny(invalid,msg)` to `denyInvalid()`?

adrianludwin:
Ah, good point!

frbimo:
> The func changes lgtm! But it seems you will also need to change the callers who has, e.g. `deny(invalid,msg)` to `denyInvalid()`?

hi @yiqigao217 , thanks for the review.
I want to say that since `Field` is related to `k8s.io/apimachinery/pkg/util/validation/field`, i thought its better for denyInvalid not to accept string.

I made some changes, ptal.

yiqigao217 (commented):
Nice change, thanks!
/lgtm
/hold
for @adrianludwin to have another look.

PR #1355, opened 2021-01-23
Changes detect namespace deletion  into object validator

### LINKED ISSUE #1241: HNC: detect namespace deletion without a controller

frbimo:
This PR changes detect namespace deletion that was done in reconciler 
into object validator by utilizing controller-runtime's client

fixes #1241 
supersede PR #1284 

test:
make test

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -216,9 +216,6 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	hadCrit := ns.HasLocalCritCondition()
 	ns.ClearConditions()
 
-	// Record whether the namespace is being deleted; this is useful for object validators.
-	ns.IsDeleting = !nsInst.DeletionTimestamp.IsZero()
-
 	// Set external tree labels in the forest if this is an external namespace.
 	r.syncExternalNamespace(log, nsInst, ns)
 
@@ -301,7 +298,7 @@ func (r *HierarchyConfigReconciler) syncSubnamespaceParent(log logr.Logger, inst
 	// We could also add an exception to allow K8s SAs to override the object validator (and we
 	// probably should), but this prevents us from getting into a war with K8s and is sufficient for
 	// v0.5.
-	if pnm != "" && ns.IsDeleting {
+	if pnm != "" && !nsInst.DeletionTimestamp.IsZero() {
 		log.V(1).Info("Subnamespace is being deleted; ignoring SubnamespaceOf annotation", "parent", inst.Spec.Parent, "annotation", pnm)
 		pnm = ""
 	}
--- a/incubator/hnc/internal/validators/object.go
+++ b/incubator/hnc/internal/validators/object.go
@@ -8,9 +8,12 @@ import (
 
 	"github.com/go-logr/logr"
 	k8sadm "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
@@ -140,7 +143,7 @@ func (o *Object) handle(ctx context.Context, log logr.Logger, op k8sadm.Operatio
 		return allow("source object")
 	}
 	// This is a propagated object.
-	return o.handleInherited(op, newSource, oldSource, inst, oldInst)
+	return o.handleInherited(ctx, op, newSource, oldSource, inst, oldInst)
 }
 
 func validateSelectorAnnot(inst *unstructured.Unstructured) string {
@@ -230,7 +233,7 @@ func validateNoneSelectorChange(inst, oldInst *unstructured.Unstructured) error 
 	return err
 }
 
-func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource string, inst, oldInst *unstructured.Unstructured) admission.Response {
+func (o *Object) handleInherited(ctx context.Context, op k8sadm.Operation, newSource, oldSource string, inst, oldInst *unstructured.Unstructured) admission.Response {
 	// Propagated objects cannot be created or deleted (except by the HNC SA, but the HNC SA
 	// never gets this far in the validation). They *can* have their statuses updated, so
 	// if this is an update, make sure that the canonical form of the object hasn't changed.
@@ -239,17 +242,19 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
 		return deny(metav1.StatusReasonForbidden, "Cannot create objects with the label \""+api.LabelInheritedFrom+"\"")
 
 	case k8sadm.Delete:
-		if o.isDeletingNS(oldInst) {
-			// There are few things more irritating in (K8s) life than having some stupid controller stop
-			// your namespace from being deleted. If there's an object in here and we've decided that the
-			// namespace should be deleted, then don't block anything!
-			//
-			// It's probably slightly safer to create a K8s client, actually load the namespace and check
-			// its deletion timestamp here. But it's much easier to just record this information in the
-			// namespace/HC reconciler and read it here.
-			return allow("allowing deletion of propagated object since namespace is being deleted")
+		// There are few things more irritating in (K8s) life than having some stupid controller stop
+		// your namespace from being deleted. If there's an object in here and we've decided that the
+		// namespace should be deleted, then don't block anything!
+		isDeleting, err := o.isDeletingNS(ctx, oldInst.GetNamespace())
+		if err != nil {
+			return deny(metav1.StatusReasonInternalError, "Cannot delete object propagated from namespace \""+oldSource+"\" with error: "+err.Error())
 		}
-		return deny(metav1.StatusReasonForbidden, "Cannot delete object propagated from namespace \""+oldSource+"\"")
+
+		if !isDeleting {
+			return deny(metav1.StatusReasonForbidden, "Cannot delete object propagated from namespace \""+oldSource+"\"")
+		}
+
+		return allow("allowing deletion of propagated object since namespace is being deleted")
 
 	case k8sadm.Update:
 		// If the values have changed, that's an illegal modification. This includes if the label is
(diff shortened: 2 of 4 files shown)


frbimo:
rebased as controller-runtime update #1367

incubator/hnc/internal/validators/object.go
```diff
@@ -275,16 +276,19 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
...
-// isDeletingNS returns true if the namespace of the object is already being deleted, based on the
-// in-memory forest.
-func (o *Object) isDeletingNS(inst *unstructured.Unstructured) bool {
-	o.Forest.Lock()
-	defer o.Forest.Unlock()
-	ns := o.Forest.Get(inst.GetNamespace())
-	if ns == nil {
-		return false
+// validateDeletingNS validates if the namespace of the object is already being deleted
+func (o *Object) validateDeletingNS(ctx context.Context, inst *unstructured.Unstructured) error {
```
adrianludwin, line 280: Errors are usually there to define exceptional cases, but in this case I wouldn't call "the namespace is not being deleted" an exception. So I think I'd go back to calling this `isDeletingNS` but allow it to return both a boolean and an error; we only return an error if the call to `Get` fails and we can't tell whether the namespace is being deleted or not.

incubator/hnc/internal/validators/object.go
```diff
@@ -275,16 +276,19 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
...
+	ns := &corev1.Namespace{}
+	nnm := types.NamespacedName{Name: inst.GetNamespace()}
+	if err := o.client.Get(ctx, nnm, ns); err != nil {
+		return err
```
adrianludwin, line 284: Wrap this error: `return fmt.Errorf("while determining whether namespace %q is being deleted: %w", inst.GetNamespace(), err)`

incubator/hnc/internal/validators/object.go
```diff
@@ -275,16 +276,19 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
...
-func (o *Object) isDeletingNS(inst *unstructured.Unstructured) bool {
-	o.Forest.Lock()
-	defer o.Forest.Unlock()
-	ns := o.Forest.Get(inst.GetNamespace())
-	if ns == nil {
-		return false
+// validateDeletingNS validates if the namespace of the object is already being deleted
+func (o *Object) validateDeletingNS(ctx context.Context, inst *unstructured.Unstructured) error {
+	ns := &corev1.Namespace{}
+	nnm := types.NamespacedName{Name: inst.GetNamespace()}
```
adrianludwin, line 282: We don't need `inst` for anything _except_ its namespace, so maybe this function should just accept the string name of the namespace.

incubator/hnc/internal/validators/object.go
adrianludwin: Also: I've seen some weird corner cases where the namespace can be deleted even before the objects are. So maybe let's treat not-found errors specially? For an example, see: https://github.com/kubernetes-sigs/multi-tenancy/blob/d7e8a4dbcf5e8e594956a0546c73b015841e1323/incubator/hnc/internal/validators/hierarchy.go#L465

incubator/hnc/internal/validators/object_test.go
```diff
@@ -273,8 +281,24 @@ func TestUserChanges(t *testing.T) {
 			},
 		},
 	}, {
-		name: "Deny deletions of propagated objects",
-		fail: true,
+		name:       "Deny deletions of propagated objects when namespace is not being deleted",
+		fail:       true,
+		isDeleting: false,
```
adrianludwin, line 286: `false` is the zero-value of `boolean` so I'd prefer to leave the out here.

incubator/hnc/internal/validators/object_test.go
```diff
@@ -273,8 +281,24 @@ func TestUserChanges(t *testing.T) {
...
+				"metadata": map[string]interface{}{
+					"labels": map[string]interface{}{
+						api.LabelInheritedFrom: "foo",
+					},
+				},
+			},
+		},
+	}, {
+		name:       "Allow deletions of propagated objects when namespace is being deleted",
+		fail:       false,
```
adrianludwin, line 300: Zero-value; I'd prefer to remove

incubator/hnc/internal/validators/object_test.go
```diff
@@ -563,6 +590,54 @@ func TestUserChanges(t *testing.T) {
...
+	if c.isDeleting {
+		ns.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
+	}
+
+	j, err := json.Marshal(ns)
+	if err != nil {
+		return err
+	}
+	decoder := scheme.Codecs.UniversalDecoder()
+	_, _, err = decoder.Decode(j, nil, obj)
```
adrianludwin, line 609: What's the purpose of having this? The test never looks at anything other than the deletion timestamp, right?

incubator/hnc/internal/validators/object.go
```diff
@@ -275,16 +276,19 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
...
+	if err := o.client.Get(ctx, nnm, ns); err != nil {
+		return err
 	}
-	return ns.IsDeleting
```
adrianludwin, line 287: Can you please remove all the code that _sets_ `ns.IsDeleting` as well, assuming this was the only place that read it?

adrianludwin (commented):
Sorry I missed this PR when you first submitted it!

incubator/hnc/internal/validators/object_test.go
frbimo: Yes, and it will decode return value (with deletionTimestamp) into input pointer `obj`, which in function `Get()` is an interface. This is used to create `fakeClient` in controller-runtime https://github.com/kubernetes-sigs/controller-runtime/blob/v0.8.1/pkg/client/fake/client.go#L275-L281

incubator/hnc/internal/validators/object_test.go
frbimo: afaik, is explicit better than implicit?

adrianludwin: In most of our tests, we leave out the zero values, because otherwise every time you add a new flag for just one test, you have to go update all the other tests. This ends up making the tests harder to read IMO instead of just emphasizing the parts of the test that are important.

We tend not to rely on zero values as much in the real code, but tests are a bit of a different story.

incubator/hnc/internal/validators/object_test.go
adrianludwin: Given that we know that `obj` is actually a pointer to a namespace object in all cases that we care about, can we just cast it?

```
ns := obj.(*corev1.Namespace)
ns.DeletionTimestamp = <whatever>
```

incubator/hnc/internal/validators/object.go
frbimo: can you help to check whether these [lines](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/hierarchy_config.go#L295-L307)  should be removed? im not quite sure what its doing.

frbimo: will re-push after this get confirmed

adrianludwin: Those lines are still needed, but you can just check `nsInst.DeletionTimestamp.IsZero()` instead (check out line 220; this function is called on line 228).

adrianludwin: ... fyi, what it's doing is creating a subnamespace. Here's the flow:

1. User creates a `SubnamespaceAnchor` in the parent
2. HNC's anchor reconciler notices the anchor in the parent, and creates the subnamespace with an annotation pointing to the parent, but it doesn't create the `HierarchyConfiguration` object pointing back to the parent. This allows the anchor reconciler to be atomic (e.g. it only needs to create one object in a single call).
3. HNC's hierarchy config reconciler notices the namespace with the annotation, and updates the `HierarchyConfiguration` to match it.

This code is trying to suppress step 3 from happening when the namespace is being deleted.

frbimo: got it

frbimo:
ptal, in case i missed something

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -216,9 +216,6 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	hadCrit := ns.HasLocalCritCondition()
 	ns.ClearConditions()
 
-	// Record whether the namespace is being deleted; this is useful for object validators.
-	ns.IsDeleting = !nsInst.DeletionTimestamp.IsZero()
-
```
adrianludwin, line 221: Can you please delete the field as well? https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/forest/namespace.go#L55

And you also have to update the code in `syncSubnamespaceParent` if you're removing the code that sets the field.

incubator/hnc/internal/validators/object.go
```diff
@@ -239,17 +242,19 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
...
-			// It's probably slightly safer to create a K8s client, actually load the namespace and check
-			// its deletion timestamp here. But it's much easier to just record this information in the
-			// namespace/HC reconciler and read it here.
-			return allow("allowing deletion of propagated object since namespace is being deleted")
+		// There are few things more irritating in (K8s) life than having some stupid controller stop
+		// your namespace from being deleted. If there's an object in here and we've decided that the
+		// namespace should be deleted, then don't block anything!
+		isDeleting, err := o.isDeletingNS(ctx, oldInst.GetNamespace())
+		if err != nil {
+			return deny(metav1.StatusReasonForbidden, "Cannot delete object propagated from namespace \""+oldSource+"\" with error: "+err.Error())
```
adrianludwin, line 250: We're not really forbidding the change - it's more like we couldn't validate whether it was legal or not. Maybe `StatusReasonInternalError` would be a more appropriate error here (HTTP 500).

incubator/hnc/internal/validators/object.go
```diff
@@ -275,16 +280,22 @@ func (o *Object) handleInherited(op k8sadm.Operation, newSource, oldSource strin
...
-	defer o.Forest.Unlock()
-	ns := o.Forest.Get(inst.GetNamespace())
-	if ns == nil {
-		return false
+// validateDeletingNS validates if the namespace of the object is already being deleted
+func (o *Object) isDeletingNS(ctx context.Context, ns string) (bool, error) {
+	nsObj := &corev1.Namespace{}
+	nnm := types.NamespacedName{Name: ns}
+	if err := o.client.Get(ctx, nnm, nsObj); err != nil {
+		if errors.IsNotFound(err) {
```
adrianludwin, line 288: Maybe comment: this should never happen, but if for some bizarre reason the namespace appears to be deleted before the object, we should allow the object to be deleted too.

Perhaps logging something would be nice here as well, but I don't feel too strongly about that.

adrianludwin (commented):
/ok-to-test

incubator/hnc/internal/reconcilers/hierarchy_config.go
frbimo: i was about to do it before you suggested not to remove [this](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/reconcilers/hierarchy_config.go#L295-L307)  inside `syncSubnamespaceParent`
If i removed https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/internal/forest/namespace.go#L55, then `syncSubnamespaceParent` will have error since it uses `IsDeleting`.

or am i miss something?

incubator/hnc/internal/validators/object.go
frbimo: comment added

incubator/hnc/internal/reconcilers/hierarchy_config.go
adrianludwin: Sorry I wasn't clear. In [this comment](https://github.com/kubernetes-sigs/multi-tenancy/pull/1355#discussion_r567532923), I meant that you should replace the check for `ns.IsDeleting` with a check of `nsInst.DeletionTimestamp`, since they have exactly the same information. We need all the _other_ lines in that block of code, but we don't need `ns.IsDeleting` since we have a good alternative.

Does that make sense?

frbimo: ah i get it. Thanks for clarifying.

frbimo:
request final review @adrianludwin

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17 

Nice work! Ryan, can you ptal?

rjbez17:
/lgtm
/approve
/hold cancel

Thanks for this!

PR #1356, opened 2021-01-23
MTB: changed flag input format

phoenixking25:
`kubectl-mtb run -n test1,test --as nody,bob`

Right now hardcoded for 2 tenants but can be made dynamic, will do when we will get that requirement as it will lead to a major change in generic way of writing code of a benchmark.

/assign @JimBugwadia

### DIFF
--- a/benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
+++ b/benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
@@ -130,18 +130,28 @@ func removeBenchmarksWithIDs(ids []string) {
 
 // Validation of the flag inputs
 func validateFlags(cmd *cobra.Command) error {
-	benchmarkRunOptions.Tenant, _ = cmd.Flags().GetString("as")
-	if benchmarkRunOptions.Tenant == "" {
-		return fmt.Errorf("username must be set via --as")
+	tenants, _ := cmd.Flags().GetStringSlice("as")
+	tenantNamespaces, _ := cmd.Flags().GetStringSlice("namespace")
+
+	if len(tenants) < 1 || len(tenantNamespaces) < 1 {
+		return fmt.Errorf("please provide tenant or tenant namespace")
 	}
 
-	benchmarkRunOptions.TenantNamespace, _ = cmd.Flags().GetString("namespace")
-	if benchmarkRunOptions.TenantNamespace == "" {
-		return fmt.Errorf("tenant namespace must be set via --namespace or -n")
+	if len(tenants) != len(tenantNamespaces) {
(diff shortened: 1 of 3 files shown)


benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
```diff
@@ -130,18 +130,28 @@ func removeBenchmarksWithIDs(ids []string) {
...
 // Validation of the flag inputs
 func validateFlags(cmd *cobra.Command) error {
-	benchmarkRunOptions.Tenant, _ = cmd.Flags().GetString("as")
-	if benchmarkRunOptions.Tenant == "" {
-		return fmt.Errorf("username must be set via --as")
+	tenants, _ := cmd.Flags().GetStringSlice("as")
+	tenantNamespaces, _ := cmd.Flags().GetStringSlice("namespace")
+
+	if len(tenants) < 1 || len(tenantNamespaces) < 1 {
+		return fmt.Errorf("please provide tenant or tenant namespace")
```
JimBugwadia, line 137: change to, "user and namespace required"

benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
```diff
@@ -130,18 +130,28 @@ func removeBenchmarksWithIDs(ids []string) {
...
+	}
+
+	if len(tenants) != len(tenantNamespaces) {
+		return fmt.Errorf("please provide same number of tenant and tenantNamespaces")
```
JimBugwadia, line 141: change to, "user and namespace counts must be equal"

benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
```diff
@@ -130,18 +130,28 @@ func removeBenchmarksWithIDs(ids []string) {
...
 	}
 
-	benchmarkRunOptions.TenantNamespace, _ = cmd.Flags().GetString("namespace")
-	if benchmarkRunOptions.TenantNamespace == "" {
-		return fmt.Errorf("tenant namespace must be set via --namespace or -n")
+	if len(tenants) > 2 || len(tenantNamespaces) > 2 {
+		return fmt.Errorf("support for more than 2 tenants is not available yet")
```
JimBugwadia, line 145: change to, "user and namespace counts cannot exceed 2"

JimBugwadia (changes requested):
I've suggested a few changes with the help text. 

It would also be good to declare the flags and provide help text for the user, so they can view it using "--help / -h" and understand how to use the flags. We can do that as a separate issue, if you prefer.

JimBugwadia (approved):

PR #1361, opened 2021-01-26
add necessary helper functions and changes to support CRD sync

weiling61:
This PR is to support CRD synchronization using current vc-syncer infrastructure:

1) add GetSpec() in ClusterInterface to retrieve vc restful config from cluster
2) add Syncer Name flag to distinguish syncer "LockObjectName" in case user wants to have a separate CRD syncer from vc-syncer
3) remove scheme object lookup code to align with latest PR #1363 

More details are in: https://github.com/kubernetes-sigs/multi-tenancy/issues/1315


incubator/virtualcluster/cmd/syncer/app/options/options.go
```diff
@@ -98,6 +100,7 @@ func (o *ResourceSyncerOptions) Flags() cliflag.NamedFlagSets {
 	fs := fss.FlagSet("server")
 	fs.StringVar(&o.SuperMaster, "super-master", o.SuperMaster, "The address of the super master Kubernetes API server (overrides any value in super-master-kubeconfig).")
 	fs.StringVar(&o.ComponentConfig.ClientConnection.Kubeconfig, "super-master-kubeconfig", o.ComponentConfig.ClientConnection.Kubeconfig, "Path to kubeconfig file with authorization and master location information.")
+	fs.StringVar(&ComponentName, "componentname", ComponentName, "The component name (default vc-syncer).")
```
christopherhein, line 103: We should add this as a field on the `ResourceSyncerOptions` instead of using a global `var`

christopherhein (commented):
Nice work, as discussed on the CAPN call, an example would be helpful or documentation explaining how to use this.

incubator/virtualcluster/pkg/util/mccontroller/helper.go
```diff
@@ -0,0 +1,133 @@
+/*
```
Fei-Guo, line 1: Is it possible to remove "CR" from all variable and method names?

It is a good abstraction for both build-in and CR objs and having a "CR" substring in the names looks a little confusing.

Fei-Guo:
Btw, this change has a conflict with #1363.

weiling61 (commented):
@christopherhein @Fei-Guo Can you review the code again. I solved the conflicts with #1363. Thanks

PR #1363, opened 2021-01-26
[VC] syncer: using generic scheme register instead of hand writing

zhuangqh:
- syncer: using generic scheme register instead of hand writing
- syncer: abstract controller initial interface

### DIFF
--- a/incubator/virtualcluster/pkg/util/mccontroller/mccontroller.go
+++ b/incubator/virtualcluster/pkg/util/mccontroller/mccontroller.go
@@ -26,9 +26,6 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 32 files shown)


incubator/virtualcluster/pkg/util/mccontroller/mccontroller.go
```diff
@@ -230,7 +225,7 @@ func (c *MultiClusterController) Get(clusterName, namespace, name string) (inter
 	if cluster == nil {
 		return nil, errors.NewClusterNotFound(clusterName)
 	}
-	instance := getTargetObject(c.objectType)
+	instance := c.objectType.DeepCopyObject()
```
zhuangqh, line 228: deepcopy here is faster than reflect.New one
objectType is a zero value

Fei-Guo:
Nice change!

/approve
/lgtm

PR #1364, opened 2021-01-26
MTB: Default deny net conn 

phoenixking25:


rjbez17:
/assign @JimBugwadia 

This lgtm in general but @JimBugwadia should take a look.

rjbez17:
@JimBugwadia @phoenixking25 is this still relevant?

PR #1374, opened 2021-02-04
[VC] pluggable syncer

zhuangqh:
1. syncer: add BaseResourceSyncer for inheritance
2. syncer: introduce plugin mechanism


incubator/virtualcluster/pkg/util/plugin/context.go
```diff
@@ -0,0 +1,34 @@
+package plugin
```
Fei-Guo, line 1: Add file header.

incubator/virtualcluster/pkg/util/plugin/context.go
```diff
@@ -0,0 +1,34 @@
...
+type InitContext struct {
+	Context    context.Context
+	Config     interface{}
+	Client     clientset.Interface
+	Informer   informers.SharedInformerFactory
+	VCClient   vcclient.Interface
+	VCInformer vcinformers.VirtualClusterInformer
+}
+
+// Plugin represents an initialized plugin, used with an init context.
```
Fei-Guo, line 23: Should you move Plugin struct to plugin.go?

Fei-Guo (commented):
Overall, excellent change!  There are some nits listed below.

One thing is that /pkg/syncer/resource/register.go  can be removed with this change.

PR #1381, opened 2021-02-09
[VC] Add cache dump functionality

Fei-Guo:
This change adds dumping cache functionality, mainly used for debugging. For now, the cache state will be dumped every minute, which is subject to change in the future. 

New dump structures are introduced which are the same as internal cache objects except that all fields are exported. 
Then we can leverage json marshal capability to make "pretty print" for the cache content.

### DIFF
--- a/incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
+++ b/incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
@@ -18,6 +18,7 @@ package cache
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
```diff
@@ -334,3 +334,19 @@ func (c *schedulerCache) RemoveProvision(clustername, key string) error {
...
 	return clusterState.RemoveProvision(key)
 }
+
+func (c *schedulerCache) Dump() string {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	out := "-- Dump Super Clusters --"
+	for k, v := range c.clusters {
+		out = fmt.Sprintf("%s\n%s: %s", out, k, v.Dump())
```
zhuangqh, line 344: try using `strings.Builder` for better performance in this case
builder.Write([]byte)

Issue #1377, opened 2021-02-05
[VC] TODO list for multiple super cluster support


Fei-Guo:
The remaining tasks for syncer controller include: 

- [ ] Support selective consistency checks in the patroller based on the tenant namespace placement.
- [ ] Remove the super cluster namespace if the super cluster is removed from the tenant namespace placement.
- [ ] Ensure a tenant Pod can start when its namespace is scheduled in a different root cluster compared to that of the tenant default namespace.

The remaining tasks for the scheduler include:
- [ ] Add tenant resources reconciling logic (for Pod, resource quota, and namespace)
- [ ] Add super cluster resource reconciling logic (for namespace and node)
- [ ] Add basic scheduler framework to support a simple first fit algorithm
- [ ] Support resizing tenant namespace quota
- [ ] Support rescheduling to tolerate super cluster failures
- [ ] Advanced namespace slice scheduling policy (such as affinity) 

@zhuangqh

PR #1384, opened 2021-02-10
[VC] Add health check and update super cluster capacity periodically

Fei-Guo:
This change adds health check for the virtual clusters and super clusters. Some basic metrics are added to the scheduler.
We leverage the super cluster health check to update the super cluster capacity in case the capacity changes due to adding/removing nodes or node errors. We may tune this later if a single node failure needs to be notified to the scheduler immediately.


zhuangqh:
link #1377

PR #1389, opened 2021-02-17
VC: selective consistency ns checks 

zhuangqh:

link #1377

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/namespace/checker.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/namespace/checker.go
@@ -28,8 +28,12 @@ import (
 	"k8s.io/klog"
 
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/constants"
+	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/conversion"
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/metrics"
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/patrol/differ"
+	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/util"
+	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/util/featuregate"
+	utilconstants "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/util/constants"
 )
 
 func (c *controller) StartPatrol(stopCh <-chan struct{}) error {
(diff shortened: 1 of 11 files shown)


incubator/virtualcluster/pkg/syncer/resources/namespace/checker.go
```diff
@@ -76,98 +79,90 @@ func (c *controller) shouldBeGabageCollected(ns *v1.Namespace) bool {
...
-		for _, clusterName := range clusterNames {
-			wg.Add(1)
-			go func(clusterName string) {
-				defer wg.Done()
-				c.checkNamespacesOfTenantCluster(clusterName)
-			}(clusterName)
-		}
-		wg.Wait()
-	} else {
-		klog.Infof("tenant masters has no clusters, still check pNamespace for gc purpose.")
```
Fei-Guo, line 94: This is a behavior change, we should continue to check pNamespace for gc purpose when no cluster exists.

incubator/virtualcluster/pkg/syncer/util/helper.go
```diff
@@ -36,3 +40,21 @@ func GetVirtualClusterObject(mc *mc.MultiClusterController, clustername string)
 
 	return vc, nil
 }
+
+func IsObjectScheduleToCluster(obj metav1.Object, clusterID string) error {
```
Fei-Guo, line 44: I suggest to rename it to "IsNamespaceScheduledToCluster". The logic in the method only applies to Namespace.

incubator/virtualcluster/pkg/syncer/resources/namespace/checker.go
```diff
@@ -76,98 +79,90 @@ func (c *controller) shouldBeGabageCollected(ns *v1.Namespace) bool {
...
+		c.deleteNamespace(p)
+	}
 
-		if err != nil {
-			klog.Errorf("error getting pNamespace %s from super master cache: %v", targetNamespace, err)
-		}
+	vSet.Difference(pSet, differ.FilteringHandler{
+		Handler: d,
+		FilterFunc: func(obj differ.ClusterObject) bool {
+			if obj.OwnerCluster == "" && obj.GetAnnotations()[constants.LabelVCRootNS] == "true" {
```
Fei-Guo, line 152: Is it ok to remove the obj.OwnerCluster == "" condition?

incubator/virtualcluster/pkg/syncer/util/helper.go
zhuangqh: updated

incubator/virtualcluster/pkg/syncer/resources/namespace/checker.go
zhuangqh: just skip vObj

incubator/virtualcluster/pkg/syncer/resources/namespace/checker.go
zhuangqh: you are right

PR #1390, opened 2021-02-18
[VC] Implement major workflows in SchedulerEngine

Fei-Guo:
This change has refactored the interfaces in SchedulerEngine to satisfy the tenant namespace reconciler. It seems that we don't need the rollback mechanism if the scheduling result is failed to be updated to tenant namespace. Instead, we can retry later while keeping the placement in the scheduler cache. 

This change also implements the major workflows in the SchedulerEngine interfaces. The only missing part is to call the scheduling algorithm. The ScheduleNamespace interface covers the following scenarios: 
- Schedule for a new namespace (with quota)
- Schedule extra slices if a namespace's quota is increased.
- Determine the new placement if a namespace's quota is reduced.
- Reschedule if the placement result is modified (e.g., remove the stale schedule to a cluster that is unhealthy)

### DIFF
--- a/incubator/virtualcluster/experiment/pkg/scheduler/algorithm/types.go
+++ b/incubator/virtualcluster/experiment/pkg/scheduler/algorithm/types.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


christopherhein:
This looks good to me, but I haven't gone through a lot of this new code so I'll let someone else `lgtm` it. nice work!

PR #1396, opened 2021-02-18
Fix Docker-in-Docker and image name

adrianludwin:
This change uses the krte (KIND RunTime Environment) image as a base,
since it knows how to run Docker-in-Docker even when the Prow node
doesn't have Docker installed on it already (e.g. because it uses
containerd, I think).

Also sets the HNC_REGISTRY env var explicitly so it doesn't pick up the
default from gcloud in the image.

Tested: Without this change, HNC periodics are failing; with this
change, everything passes.

### DIFF
--- a/incubator/hnc/hack/prow-e2e/Dockerfile
+++ b/incubator/hnc/hack/prow-e2e/Dockerfile
@@ -1,26 +1,34 @@
-FROM golang:1.14.2-alpine
+# krte includes everything needed to run Docker-in-Docker, as well as cleaning
+# it up cleanly. Note that Prow uses containerd so typically, no Docker daemon
+# will be running on the node without using the contents of this image!
+#
+# This includes Go 1.15.7 and kubectl 1.17. UPGRADE TO A NEW VERSION WHEN WE
+# STOP SUPPORTING K8s 1.17 as that seems like a reasonable time. At that point,
+# I'd just upgrade to the latest image that works; we don't have to increment
+# one by one. Honestly I doubt the exact version of the base image makes too
+# much of a difference.
+#
+# Please update the version of Kind at the same time you update the version of
+# KRTE, if applicable!
+#
+# Source: https://github.com/kubernetes/test-infra/tree/master/images/krte
+FROM gcr.io/k8s-testimages/krte:v20210129-3799a64-master
 WORKDIR /repo
 
-# Since go modules isn't enabled by default.
+# I'm not sure if the following two env vars are needed for HNC; they're probably not. GO111MODULE=on
+# should be redundant outside of GOPATH (which we are) on any post-1.13 version of Go (which we have);
+# I don't *think* that we need CGO either. Feel free to try out removing them when you're next playing
+# around with this image.
 ENV GO111MODULE=on
-
-# Build static binaries; otherwise go test complains.
 ENV CGO_ENABLED=0
 
-# Install all basic dependencies. Bash is there in case you need to debug.
-# Curl, Docker, git and make are not included in the golang image.
-RUN apk add --no-cache bash curl docker git make
-#  bash curl docker gcc git jq make openssh-client
-
-# Install Kind
+# Install Kind. Feel free to update this version whenever you're updating the
+# KRTE base image too.
 RUN go get sigs.k8s.io/kind@v0.9.0
 
-# Install kubectl and make sure it's available in the PATH.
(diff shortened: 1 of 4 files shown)


adrianludwin:
First passing test: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/hnc-periodic-test/1362223411924832256

incubator/hnc/hack/prow-e2e/Dockerfile
```diff
@@ -7,20 +14,14 @@ ENV GO111MODULE=on
...
-#  bash curl docker gcc git jq make openssh-client
-
 # Install Kind
 RUN go get sigs.k8s.io/kind@v0.9.0
 
-# Install kubectl and make sure it's available in the PATH.
-RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/`curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt`/bin/linux/amd64/kubectl
-RUN chmod +x ./kubectl
-RUN mv ./kubectl /bin
-
```
yiqigao217, line 22: Why don't we need this any more? Or is it moved to somewhere else?

incubator/hnc/hack/prow-e2e/run-e2e.sh
```diff
@@ -1,62 +1,4 @@
...
-# Note that this won't run if the tests fail - see above. But we may as well
-# *try* to clean up after ourselves.
-kind delete clusters ${CLUSTERNAME}
-
-echo "Finished at $(date +%Y-%m-%d\ %H:%M:%S)"
-end_time="$(date -u +%s)"
-elapsed="$(($end_time-$start_time))"
-echo "Script took $elapsed seconds"
+# Wrapper introduced so we don't need to modify the Prow config during testing.
+# Remove this when the command below works reliably..
```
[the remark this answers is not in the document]
adrianludwin: I'll clean that up in a bit. Basically the Prow config is hardcoded to call `run-e2e.sh`, but probably should have been `wrapper.sh run-e2e.sh`. Rather than try to change the Prow config, I just made this script a wrapper to call `wrapper.sh run-e2e-tests.sh`.

I've already got a PR to update the Prow config too, at which point I can just delete this file.

incubator/hnc/hack/prow-e2e/Dockerfile
adrianludwin: `kubectl` is now included in the base image (`krte` vs `golang`).

incubator/hnc/hack/prow-e2e/Dockerfile
```diff
@@ -1,4 +1,11 @@
-FROM golang:1.14.2-alpine
+# krte includes everything needed to run Docker-in-Docker, as well as cleaning
+# it up cleanly. Note that Prow uses containerd so typically, no Docker daemon
+# will be running on the node without using the contents of this image!
+#
+# This includes Go 1.15.7 and kubectl 1.17
+#
+# Source: https://github.com/kubernetes/test-infra/tree/master/images/krte
+FROM gcr.io/k8s-testimages/krte:v20210129-3799a64-master
```
rjbez17, line 8: Do we need any docs around tracking the proper image to use? As in, how/when would I know we should change this?

Alternatively, it looks like there are a few tracking tags existing (although using tracking tags can cause other headaches):

- latest-master
- latest-1.17
- latest-1.18
- ...

rjbez17:
lgtm but the other PR merged so you may want to remove that script in this pr or not. I'll leave it to you.

incubator/hnc/hack/prow-e2e/Dockerfile
adrianludwin: I decided to stick to the existing one but added a bunch of docs in the Dockerfile and README about when to upgrade to a later version. I doubt that we're very sensitive to the exact version but I do think it makes sense to pin it just in case.

adrianludwin:
> lgtm but the other PR merged so you may want to remove that script in this pr or not. I'll leave it to you.

Removed. I can't re-test in less than 48h but once this is submitted, the postsubmits should test it right away.

yiqigao217 (commented):
/lgtm
/hold cancel

yiqigao217:
/hold cancel

redo the cancel hold since it was initially changed from hold in my comment.

PR #1395, opened 2021-02-18
[VC] Adding super cluster networking feature

christopherhein:
This adds a new featuregate - `featuregate.SuperClusterServiceNetwork` which when turned on will use the superclusters tenant control plane service `ClusterIP` for configuring the `KUBERNETES_SERVICE_HOST` environment variable. This allows controllers deployed into virtualclusters to route to the correct control plane without needing the forked CoreDNS.

Ref #1394 

Signed-off-by: Chris Hein <me@chrishein.com>

Issue #1394, opened 2021-02-18
[VC] Support Allowing the SuperCluster to manage ClusterIPs


christopherhein:
This is an overarching issue that will track being able to support using the supercluster as a proxy for supporting the tenant control planes Service `ClusterIP`.

- [x] #1395 Configure the `KUBERNETES_SERVICE_HOST` to use tenant control plane `apiserver-svc`
- [ ] Configure the syncer to allow services to be "adopted" from the super cluster
- [ ] Build tenant webhook endpoint to create services

christopherhein:
/close

We'll revisit this at a later point.

PR #1397, opened 2021-02-18
[VC] Support adopting supercluster services

christopherhein:
This extends #1395 to support adopting services that could be created in the super cluster, this is gated by the feature flag introduced in #1395 and expects that the super cluster service has a non-upward synced annotation set to `transparency.tenancy.x-k8s.io/super.adoptable`.

Following this, I'll be implementing a webhook that all the tenant control planes could be configured to for mutating Service objects prior to saving in etcd.

Blocked by #1395 
Ref #1394 

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/constants/constants.go
+++ b/incubator/virtualcluster/pkg/syncer/constants/constants.go
@@ -65,6 +65,10 @@ const (
 	// PublicObjectKey is a label key which marks the super master object that should be populated to every tenant master.
 	PublicObjectKey = "tenancy.x-k8s.io/super.public"
 
+	// AdoptableObjectKey is a label key that marks super master object as reclaimable via setting UID
+	// this uses the transparency. subdomain to force this label to not sync to the tenant
+	AdoptableObjectKey = "transparency.tenancy.x-k8s.io/super.adoptable"
+
 	LabelVirtualNode = "tenancy.x-k8s.io/virtualnode"
 
 	// DefaultvNodeGCGracePeriod is the grace period of time before deleting an orphan vNode in tenant master.
--- a/incubator/virtualcluster/pkg/syncer/resources/service/checker.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/service/checker.go
@@ -21,12 +21,15 @@ import (
 	"fmt"
 	"sync/atomic"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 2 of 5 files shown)


Fei-Guo:
Can you rebase? Thanks

christopherhein:
> Can you rebase? Thanks

@Fei-Guo done, thanks!

Fei-Guo:
When pService is created externally, but the vService has not been created, the Patroller will remove the pService.
https://github.com/kubernetes-sigs/multi-tenancy/blob/dc6670c75479ee15b1deaa39bd41811f92b78cea/incubator/virtualcluster/pkg/syncer/resources/service/checker.go#L88

If the following events happen sequentially: pService is created-> a vService is created -> pService is attached with UID -> vService is deleted -> vService with the same name is created again.

You should make sure webhook always deletes the super service first otherwise you will have nonempty UID mismatch.

Also, it is not clear if vService is deleted, should pService be deleted by the syncer or not? If syncer needs to delete it, you may have two actors (webhook and syncer) that delete pService at the same time which may cause race.

christopherhein:
This is a good call out @Fei-Guo I'm not catching the instance where the patroller runs in the middle between when the admission webhook from a tenant would create the Service and when the Service is finished being created in the tenant cluster. Within there we have a condition which we'd want to still catch where the service admission failed and the super still had a newly created service that it would be deleted, this was being caught by https://github.com/kubernetes-sigs/multi-tenancy/blob/dc6670c75479ee15b1deaa39bd41811f92b78cea/incubator/virtualcluster/pkg/syncer/resources/service/checker.go#L87 but there could be a race which happens.

Regarding garbage collection, once the UID is set on the object it's supposed to resume the existing flow allowing the syncer to handle the object updates & deletions. This is cause we only expect a service to be adoptable when the UID is `""`. and the first time the syncer gets the first update action it _should_ set the UID allowing the alternative flow to work.

christopherhein:
Updated to check before deletion cause of vService not existing @Fei-Guo.

Fei-Guo:
@christopherhein  You still have not addressed the case where pService does not exist but vService exists (which can happen if pService is deleted for whatever reason). In current code, the patroller will requeue the vService and a new pService is going to be created.

Fei-Guo:
Mutatewebhook workflow:   CreatePSerivce() -> Change service spec; then vService is created.
                                                                T0                       T1                                     T2
It seems that patroller may be triggered between T0 and T2.

christopherhein:
@Fei-Guo I believe I have covered this case now, instead of immediately deleting any Service we put it into a backoff loop which will keep checking the service status and once it becomes adopted it will remove the deletion if by the end of the back off it still hasn't been adopted then we delete the object. To combat any major latencies we also now will recreate the service against the tenant control plane if the super cluster service is deleted. 

For the vService existing but the pService not (in the case something deleted it in the cluster) we take the vService object deepcopy the contents and clear the uid and other metadata then delete the vService and immediately try to recreate it checking if the uid matches while it backs off creating the service this way retriggering the admissions lifecycle once the webhook is added. (this part is still to come in another PR)

christopherhein:
/close

I've decided to do this differently, we're able to make changes easier now that vc-syncer is more like a framework.

PR #1398, opened 2021-02-18
Attempt to fix HNC postsubmits

adrianludwin:
Our first run of the HNC postsubmits failed, most likely because the
workdir is modified by Prow (based on the messages I'm seeing). This
change updates the Dockerfile and test script to _not_ assume a specific
workdir, and also allows Prow to set up the repo when appropriate.

Tested: script runs locally but there's no good way for me to test on
Prow. As soon as this is merged, the postsubmits themselves will tell me
if I got this right or not.

### DIFF
--- a/incubator/hnc/hack/prow-e2e/Dockerfile
+++ b/incubator/hnc/hack/prow-e2e/Dockerfile
@@ -13,7 +13,13 @@
 #
 # Source: https://github.com/kubernetes/test-infra/tree/master/images/krte
 FROM gcr.io/k8s-testimages/krte:v20210129-3799a64-master
-WORKDIR /repo
+
+# For postsubmits, Prow will start the test in its own workdir (e.g.
+# /home/prow/go/src/sigs.k8s.io/multi-tenancy, though we shouldn't depend on
+# that). For periodics, I don't think Prow overrides the workdir. Either way,
+# set up the workdir in a place that makes sense for the periodics, but also
+# allows us to easily find the test script even if it's changed.
+WORKDIR /start
 
 # I'm not sure if the following two env vars are needed for HNC; they're probably not. GO111MODULE=on
 # should be redundant outside of GOPATH (which we are) on any post-1.13 version of Go (which we have);
@@ -29,6 +35,3 @@ RUN go get sigs.k8s.io/kind@v0.9.0
 # Add the actual script and set it as the default command for this image.
 COPY run-e2e-tests.sh run-e2e-tests.sh
 RUN chmod +x ./run-e2e-tests.sh
-
-# This needs to be re-specified in Prow
-CMD ["wrapper.sh", "./run-e2e-tests.sh"]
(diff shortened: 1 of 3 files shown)


adrianludwin:
This will be ready to go as soon as https://github.com/kubernetes/test-infra/pull/20906 is merged.

incubator/hnc/hack/prow-e2e/Dockerfile
```diff
@@ -13,7 +13,13 @@
 #
 # Source: https://github.com/kubernetes/test-infra/tree/master/images/krte
 FROM gcr.io/k8s-testimages/krte:v20210129-3799a64-master
-WORKDIR /repo
+
+# For postsubmits, Prow will start the test in its own workingdir (e.g.
+# /home/prow/go/src/sigs.k8s.io/multi-tenancy, though we shouldn't depend on
+# that). For periodics, I don't this prow overrides the workingdir. Either way,
```
rjbez17, line 19: nit: I don't this

incubator/hnc/hack/prow-e2e/Dockerfile
```diff
@@ -29,6 +35,3 @@ RUN go get sigs.k8s.io/kind@v0.9.0
 # Add the actual script and set it as the default command for this image.
 COPY run-e2e-tests.sh run-e2e-tests.sh
 RUN chmod +x ./run-e2e-tests.sh
-
-# This needs to be re-specified in Prow
-CMD ["wrapper.sh", "./run-e2e-tests.sh"]
```
rjbez17, line 34: Do we not want any entrypoint/cmd?

adrianludwin: The base image already has the entrypoint set to `wrapper.sh`, and IMO there's no point setting a default command of `run-e2e-tests.sh` since Prow requires that you override it anyway.

incubator/hnc/hack/prow-e2e/Dockerfile
adrianludwin: Fixed ("I don't think")

adrianludwin:
/assign @yiqigao217 
/assign @GinnyJI 

I think anyone can approve this on their own

adrianludwin:
Addendum: this still [fails](https://storage.googleapis.com/kubernetes-jenkins/logs/hnc-postsubmit-test/1364256723145068544/build-log.txt) but with an easier-to-resolve error message:

```
wrapper.sh] [TEST] Running Test Command: `/start/run-e2e-tests.sh` ...

Starting at 2021-02-23 16:53:25

Repo already exists in /home/prow/go/src/sigs.k8s.io/multi-tenancy

Creating KIND cluster 'hnc-postsubmit-20816' and setting kubectl context
/start/run-e2e-tests.sh: line 31: kind: command not found
```

That shouldn't be hard to fix, I'll have another look soon.

PR #1399, opened 2021-02-19
[VC] Update SchedulerCache when vc or cluster CRs are added/removed

Fei-Guo:
This change handles the cache updates when VC or Cluster CRs are added or removed. 

Since we have shadow cluster objects to handle super cluster offline, the shadow cluster may exist even after the super cluster CR is removed. A periodic checker is added to clean up old shadow clusters.

### DIFF
--- a/incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
+++ b/incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
@@ -20,8 +20,11 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/experiment/pkg/scheduler/constants"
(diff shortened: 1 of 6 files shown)


incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
```diff
@@ -34,19 +37,64 @@ type schedulerCache struct {
...
+	nsToDelete := make([]*Namespace, 0)
+
+	for _, v := range c.namespaces {
+		if v.owner == n {
+			nsToDelete = append(nsToDelete, v)
+		}
+	}
+
+	for _, each := range nsToDelete {
+		if err := c.removeNamespaceWithoutLock(each); err != nil {
```
zhuangqh, line 92: should rollback here?

incubator/virtualcluster/experiment/pkg/scheduler/cache/cache.go
```diff
@@ -34,19 +37,64 @@ type schedulerCache struct {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.tenants[n] = struct{}{}
+}
+
+func (c *schedulerCache) RemoveTenant(n string) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
 
+	nsToDelete := make([]*Namespace, 0)
```
zhuangqh, line 83: simply  `var nsToDelete []*Namespace` is better. this is a nil pointer while make create a zero value

Fei-Guo:
@zhuangqh All comments are addressed in the latest update. Thanks

PR #1408, opened 2021-02-23
[HNC] Add envvar printout for debugging

adrianludwin:
The postsubmits are now failing because Kind isn't in the path. Print
out all the env vars so we can figure out why.

Tested: image runs locally and prints out env. Image has already been
pushed to the registry.


adrianludwin:
/assign @rjbez17 
/assign @yiqigao217 
/assign @GinnyJI 

Hi may I please get a quick approval for this so I can debug the postsubmit failures?

PR #1412, opened 2021-02-24
Get rid of special-purpose Prow image

adrianludwin:
This change removes the now-unneeded image used for Prow periodic and
postsubmit tests. All that's left is the prow-run-e2e.sh script, a
Makefile to test it, and a lot of documentation that was moved from
other locations. The Prow configs were already pointing at this new copy
of the script so the existing configs should continue to work.

Tested: 'make prow-test' successfully starts the e2e tests on Kind; I
waited for the first two to pass but not the rest of them. Submitting
this PR will result in the postsubmits being called, while the periodics
are currently running daily, so any problems will pop up soon.


rjbez17:
/lgtm
/approve


🥇

PR #1418, opened 2021-03-03
[VC] fix: only check known cluster in differ

zhuangqh:
if the cluster cache is not loaded, vSet is missing them,
but pSet object is exists, differ will delete the pObj.
left the redundant object gc by ns checker.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/patrol/differ/handler.go
+++ b/incubator/virtualcluster/pkg/syncer/patrol/differ/handler.go
@@ -86,19 +86,19 @@ func (h FilteringHandler) OnDelete(obj ClusterObject) {
 	h.Handler.OnDelete(obj)
 }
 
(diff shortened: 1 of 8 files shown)


incubator/virtualcluster/pkg/syncer/patrol/differ/handler.go
```diff
@@ -86,19 +86,20 @@ func (h FilteringHandler) OnDelete(obj ClusterObject) {
 	h.Handler.OnDelete(obj)
 }
 
-func DefaultDifferFilter(blockedClusterSet sets.String) func(obj ClusterObject) bool {
+func DefaultDifferFilter(knownClusters []string) func(obj ClusterObject) bool {
```
zhuangqh, line 89: left the input parameter the buildin type `[]string`
`sets` is the Second-class citizens

Issue #1388, opened 2021-02-16
Proposal: CRD Synchronization in Multi-tenancy Syncer


weiling61:
# CRD Synchronization in Multi-tenancy Syncer

## Objective

The key objectives of this document are:

* Install annotated super cluster CRDs in all tenants virtual clusters.
* Dynamically build CR scheme and CR clients.
* Construct an example CR syncer using current syncer infrastructure, and keep the CR syncer code separate.
* CR Syncer cannot be registered dynamically. It should be built together with rest of multi-tenancy syncer.

## Overview

In this design, a customized syncer that supports both custom resources (CRs) and standard K8S resources will be built based on multi-tenant syncer. Proprietary CR syncers, main and resource type selection logics are kept in a private repo. Current Multi-tenant Syncer code used as libraries.

<img width="1498" alt="Screen Shot 2021-02-16 at 10 43 15 AM" src="https://user-images.githubusercontent.com/50972190/108107345-f924f280-7043-11eb-9c48-d4ae62549f30.png">

The CR syncer code will be located at “multi-tenancy/incubator/virtualcluster/example” ?

A new config option will be introduced to enable CRD Syncer in multi-tenancy syncer code.

## CRD Installation

CRDs with annotation: [tenancy.x-k8s.io/super.public](http://tenancy.x-k8s.io/super.public) will be installed  into tenant’s virtual cluster as soon as  the virtual cluster is ready.
VC-syncer will keep monitoring new CRD in super cluster and deploy a copy to all tenant virtual clusters.

## CR Scheme Building

Controller runtime scheme builder is used to add CR scheme into existing scheme singleton. 
Dynamic Scheme modification is allowed.

## CR Syncer Construction and Registration

Current vc-syncer requires a dedicated resource syncer component to be registered  at startup.

CR specific syncer components, including CR controller, CR DWS, and CR patroller, need to be compiled together with existing multi-tenant syncer code. No dynamic CR syncer generation and registeration will be allowed.

## CR Client Construction

CR client is constructed during CR resource controller creation using Controller runtime client.

## Resource Type Selection

Resource Syncer initialization module is responsible to initialize (register) some or all existing resource synchronization features for a customized syncer.

## CR Fake Client

CR relies on controller runtime fake client to perform tests.

[...]

PR #1423, opened 2021-03-09
Export checkDWObjectMetaEquality for custom resource syncer (Add an example custom resource syncer)

weiling61:
This PR gives an example of custom resource syncer. The proposed design is at: https://github.com/kubernetes-sigs/multi-tenancy/issues/1388 (will generate a readme later)

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/conversion/equality.go
+++ b/incubator/virtualcluster/pkg/syncer/conversion/equality.go
@@ -49,7 +49,7 @@ func Equality(syncerConfig *config.SyncerConfiguration, vc *v1alpha1.VirtualClus
 // notes: we only care about the metadata and pod spec update.
 func (e vcEquality) CheckPodEquality(pPod, vPod *v1.Pod) *v1.Pod {
 	var updatedPod *v1.Pod
-	updatedMeta := e.checkDWObjectMetaEquality(&pPod.ObjectMeta, &vPod.ObjectMeta)
+	updatedMeta := e.CheckDWObjectMetaEquality(&pPod.ObjectMeta, &vPod.ObjectMeta)
 	if updatedMeta != nil {
 		if updatedPod == nil {
 			updatedPod = pPod.DeepCopy()
@@ -114,7 +114,7 @@ func CheckDWPodConditionEquality(pPod, vPod *v1.Pod) *v1.PodStatus {
 	return nil
 }
 
-// checkDWObjectMetaEquality check whether super master object meta and virtual object meta
+// CheckDWObjectMetaEquality check whether super master object meta and virtual object meta
 // are logically equal. The source of truth is virtual object.
 // Reference to ObjectMetaUpdateValidation: https://github.com/kubernetes/kubernetes/blob/release-1.15/staging/src/k8s.io/apimachinery/pkg/api/validation/objectmeta.go#L227
 // Mutable fields:
@@ -126,7 +126,7 @@ func CheckDWPodConditionEquality(pPod, vPod *v1.Pod) *v1.PodStatus {
 // - finalizers: ignore. finalizer is observed by tenant controller.
 // - clusterName
 // - managedFields: ignore. observed by tenant. https://kubernetes.io/docs/reference/using-api/api-concepts/#field-management
-func (e vcEquality) checkDWObjectMetaEquality(pObj, vObj *metav1.ObjectMeta) *metav1.ObjectMeta {
+func (e vcEquality) CheckDWObjectMetaEquality(pObj, vObj *metav1.ObjectMeta) *metav1.ObjectMeta {
 	var updatedObj *metav1.ObjectMeta
 	if pObj.GenerateName != vObj.GenerateName {
 		if updatedObj == nil {
@@ -448,7 +448,7 @@ func (e vcEquality) checkInt64Equality(pObj, vObj *int64) (*int64, bool) {
 // are logically equal. The source of truth is virtual object.
 func (e vcEquality) CheckConfigMapEquality(pObj, vObj *v1.ConfigMap) *v1.ConfigMap {
 	var updated *v1.ConfigMap
-	updatedMeta := e.checkDWObjectMetaEquality(&pObj.ObjectMeta, &vObj.ObjectMeta)
+	updatedMeta := e.CheckDWObjectMetaEquality(&pObj.ObjectMeta, &vObj.ObjectMeta)
 	if updatedMeta != nil {
 		if updated == nil {
 			updated = pObj.DeepCopy()
@@ -523,7 +523,7 @@ func (e vcEquality) CheckSecretEquality(pObj, vObj *v1.Secret) *v1.Secret {
 	}
 
 	var updated *v1.Secret
-	updatedMeta := e.checkDWObjectMetaEquality(&pObj.ObjectMeta, &vObj.ObjectMeta)
+	updatedMeta := e.CheckDWObjectMetaEquality(&pObj.ObjectMeta, &vObj.ObjectMeta)
 	if updatedMeta != nil {
 		if updated == nil {
 			updated = pObj.DeepCopy()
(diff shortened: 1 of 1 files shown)


incubator/virtualcluster/example/cr-syncer/api/cr-project/clientset/v1alpha1/api.go
```diff
@@ -0,0 +1,39 @@
+package v1alpha1
```
christopherhein, line 1: So that we don't have to copy over types and clients as well as have references to non-k8s owned projects we could we use the crds/crs from https://github.com/kubernetes/sample-controller/ specifically `Foos`, then we can just use those by reference and not need to maintain the generated code.

incubator/virtualcluster/example/cr-syncer/cmd/syncer/builtins.go
```diff
@@ -0,0 +1,32 @@
...
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/event"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/namespace"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/node"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/persistentvolume"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/persistentvolumeclaim"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/pod"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/secret"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/service"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/serviceaccount"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/storageclass"
```
christopherhein, line 31: If we don't include the builtins does this work as a singular syncer for just the custom resources?

incubator/virtualcluster/example/cr-syncer/cmd/syncer/builtins_extra.go
```diff
@@ -0,0 +1,24 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/ingress"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/priorityclass"
+	_ "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/resources/crd"
```
christopherhein, line 22: Since these are builtins maybe the should be included in the other file if the builtins are necessary

incubator/virtualcluster/example/cr-syncer/pkg/syncer/resource/project/controller.go
```diff
@@ -0,0 +1,290 @@
...
+
+var (
+	SchemeBuilder      runtime.SchemeBuilder
+	localSchemeBuilder = &SchemeBuilder
+)
+
+
+func init() {
+	localSchemeBuilder.Register(addKnownTypes)
+	localSchemeBuilder.AddToScheme(scheme.Scheme)
```
christopherhein, line 62: If you use the sample controller project crs/crds we can just use the builtin exported scheme funcs and not need to maintain the versions - https://github.com/kubernetes/sample-controller/blob/master/pkg/apis/samplecontroller/v1alpha1/register.go#L40-L45

incubator/virtualcluster/example/cr-syncer/pkg/syncer/resource/project/controller.go
```diff
@@ -0,0 +1,290 @@
...
+	return c, nil
+}
+
+func createSuperProjectCRD(config *config.SyncerConfiguration) error {
+	apiextensionsClient, err := apiextensionsclientset.NewForConfig(config.RestConfig)
+	if err != nil {
+		return err
+	}
+
+	_, err = apiextensionsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(context.TODO(), "projects.example.example.com", metav1.GetOptions{})
```
christopherhein, line 200: Shouldn't this be done by the CRD syncer and having a CRD with the `super.public` annotation from the previous PR?

incubator/virtualcluster/example/cr-syncer/cmd/syncer/builtins.go
```diff
@@ -0,0 +1,32 @@
...
+)
```
zhuangqh, line 32: unnecessary to import them? because we import the `"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/cmd/syncer/app"` pkg?

zhuangqh (commented):
please group your imports and do gofmt

incubator/virtualcluster/example/cr-syncer/cmd/syncer/builtins.go
weiling61: Yes all these are optional. If included, this syncer can perform additional resource synchronization

incubator/virtualcluster/example/cr-syncer/cmd/syncer/builtins_extra.go
weiling61: Sure, can add another building

incubator/virtualcluster/example/cr-syncer/pkg/syncer/resource/project/controller.go
weiling61: agree, will remove

incubator/virtualcluster/example/cr-syncer/pkg/syncer/resource/project/controller.go
weiling61: Yes, this is just to add a Skeleton CRD in super cluster that is required by this CR syncer.. Without this, corresponding CR controller can not be initialized properly.
In most case, the CRD should have been deployed in super cluster. (as user decide to enable this CR syncer). Therefore we just do a get first and deploy the CRD in case it is not found.

incubator/virtualcluster/example/cr-syncer/cmd/syncer/builtins.go
weiling61: I see. I will do the change.

incubator/virtualcluster/example/cr-syncer/pkg/syncer/resource/project/controller.go
christopherhein: I would remove this and put this as a part of the docs to `kubectl apply -f` a CRD file into the persons local cluster when trying to learn from this example. Having this in here could confuse someone into thinking the syncer needed to manage the CRDs for the super cluster too

weiling61:
Move all example code into a doc. For this PR, only purpose is to export  'checkDWObjectMetaEquality' for convinence

zhuangqh (changes requested):
please update your commit message to match this pr

weiling61:
@christopherhein @Fei-Guo removed example cr syncer code as discussed. This PR is just to export one method in conversion lib

christopherhein:
I think you still need to update the commit message still to reflect these changes.

it currently says:
> add example cr syncer …
>
> add an example custom resource 'project' syncer
> 
> remove not fully tested build system
> 
> based on talk with Fei, will remove the buildable file and replace with readme

weiling61:
Updated commit msg

PR #1425, opened 2021-03-09
HNC: redirect stdlib logger output through zap

### LINKED ISSUE #1424: HNC: non json logs
With logging output in JSON, I was expecting all of the log output to actually be in JSON but it appears some of them are not captured, see below.

I believe these can be wrapped in json by redirecting the standard library's default logger output

```
{"level":"error","ts":1614332075.2494884,"logger":"controller","msg":"Reconciler error","reconcilerGroup":"hnc.x-k8s.io","reconcilerKind":"HNCConfiguration","controller":"hncconfiguration","name":"config","namespace":"","error":"Internal error occurred: failed calling webhook \"hncconfigurations.hnc.x-k8s.io\": Post https://hnc-webhook-service.hnc-system.svc:443/validate-hnc-x-k8s-io-v1alpha2-hncconfigurations?timeout=30s: no endpoints available for service \"hnc-webhook-service\""}
2021/02/26 12:23:22 http: TLS handshake error from 172.16.0.6:39660: EOF
2021/02/26 21:02:15 http: TLS handshake error from 172.16.0.6:47842: EOF
2021/02/26 21:33:31 Failed to export to Stackdriver: rpc error: code = DeadlineExceeded desc = One or more TimeSeries could not be written: Deadline expired before operation could complete.: timeSeries[0-6,8]
{"level":"info","ts":1614375267.5580504,"logger":"cert-rotation","msg":"no cert refresh needed"}
2021/02/26 21:34:31 Failed to export to Stackdriver: rpc error: code = DeadlineExceeded desc = One or more TimeSeries could not be written: Deadline expired before operation could complete.: timeSeries[0-8]
```

seankhliao:
Captures logs from the stdlib logger and logs them at `error` level

eg Stackdriver error in #1424 now looks like:
```
{"level":"error","ts":1615328526.2069242,"logger":"setup","msg":"Could not create Stackdriver exporter","error":"stackdriver: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/workspace/vendor/github.com/go-logr/zapr/zapr.go:132\nmain.main\n\t/workspace/cmd/manager/main.go:119\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203"}
```

Fixes #1424


incubator/hnc/cmd/manager/main.go
```diff
@@ -135,7 +137,9 @@ func main() {
 	if debugLogs {
 		logLevel = zapcore.DebugLevel
 	}
-	log := zap.New(zap.Level(logLevel), zap.StacktraceLevel(zapcore.PanicLevel))
+	rlog := zap.NewRaw(zap.Level(logLevel), zap.StacktraceLevel(zapcore.PanicLevel))
```
adrianludwin, line 140: I was a bit confused by the name, maybe `rawLog`?

What's the difference between a raw and regular logger? It might be worth a comment about why you're creating it in two different steps (e.g. just update the code, you don't need to answer this question in GitHub).

incubator/hnc/cmd/manager/main.go
```diff
@@ -135,7 +137,9 @@ func main() {
...
+	stdzap.RedirectStdLogAt(rlog, zapcore.ErrorLevel)
```
adrianludwin, line 141: Why `ErrorLevel`? Do we know that they'll all be errors?

adrianludwin (commented):
/ok-to-test

Nice, thanks for this change! Just a couple of questions and minor suggestions.

incubator/hnc/cmd/manager/main.go
seankhliao: I know the stdlib should only be calling `log` if it has no other way of returning an error, like http.Server before a handler is invoked. I'm assuming the same applies to the other libraries in use

adrianludwin: Thanks - can you please add that reasoning as a comment in the code? Then I'll approve.

PR #1426, opened 2021-03-10
[VC] support command options `--tail` for `kubectl logs`

charleszheng44:
support running `--tails=<num-lines>` command line options for `kubectl logs`


zhuangqh:
we had better to support all of the kubelet server parameter. can support all of the params of `kubectl logs`

christopherhein:
Maybe we can keep these all separate? we apparently were all looking at the same changes, maybe we can merge this and continue to add others along the way? 

I say this as I found we didn't support `--follow` yesterday and just submitted another PR for that. https://github.com/kubernetes-sigs/multi-tenancy/pull/1427

christopherhein:
Either that or you can close #1427 and add the follow flags to this PR.

charleszheng44:
I agree with @christopherhein. Let's merge this PR and #1427  now, and I will add the rest options for log/exec in these two days.

PR #1430, opened 2021-03-10
[VC] Use event sinker in mccontroller

Fei-Guo:
This change uses client-go EventCorrelator to filter/throttle the same/similar events sent to the tenant control plane. 

This change does not use the client-go EventBroadcaster since it uses a heavy weight event dispatch mechanism and it has to be created for each tenant control plane, which means the event cache (a fixed size LRU cache) has to be duplicated for every tenant control plane. 

In this implementation, one event cache is used by all tenant control planes with a default 4K entries.  Later, we can make the cache size configurable if there are too many virtual clusters. 

The scheduler uses the event sinker to notify the tenant control plane about the scheduling result. For example:

```
kubectl get event -A --kubeconfig=vc-sample-1.kubeconfig
NAMESPACE   LAST SEEN   TYPE     REASON      OBJECT      MESSAGE
test        24s         Normal   Scheduled   namespace   Successfully schedule namespace test with placement {"root-cluster-00001":1}
```


incubator/virtualcluster/pkg/util/record/event.go
```diff
@@ -0,0 +1,68 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package record
+
+import (
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/record"
```
zhuangqh, line 22: distinguish with k8s record package? such as  `clientgorecord "k8s.io/client-go/tools/record"`

PR #1432, opened 2021-03-12
[VC] Adding vn-agent deployment feature

christopherhein:
These changes make it so that you can use the vn-agent as a Kubernetes deployment when not using the direct to Kubelet proxy mechanism. All you need to do is deploy the vn-agent as a `Deployment`, then you create a new `ClusterIP` Service (this could potentially also be a LoadBalancer later to support the experimental scheduler with a couple tweaks) once that is setup you deploy the syncer with the flags `--vn-agent-namespace` and `--vn-agent-service-name` and it will fetch the service clusterIP at node creation to register with the virtual node resource.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/apis/config/types.go
+++ b/incubator/virtualcluster/pkg/syncer/apis/config/types.go
@@ -54,6 +54,10 @@ type SyncerConfiguration struct {
 	// VNAgentPort defines the port that the VN Agent is running on per host
 	VNAgentPort int32
 
+	// VNAgentNamespacedName defines the namespace/name of the VN Agent Kubernetes
+	// service, this is used for feature VNodeProviderService.
+	VNAgentNamespacedName string
+
 	// FeatureGates enabled by the user.
 	FeatureGates map[string]bool
 
--- a/incubator/virtualcluster/pkg/syncer/vnode/provider/provider.go
+++ b/incubator/virtualcluster/pkg/syncer/vnode/provider/provider.go
@@ -1,0 +1,27 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package provider
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// NodeProvider is the interface used for registering the node address.
+type VirtualNodeProvider interface {
+	GetNodeDaemonEndpoints(node *v1.Node) (v1.NodeDaemonEndpoints, error)
+	GetNodeAddress(node *v1.Node) ([]v1.NodeAddress, error)
+}
--- a/incubator/virtualcluster/pkg/syncer/vnode/service/provider.go
+++ b/incubator/virtualcluster/pkg/syncer/vnode/service/provider.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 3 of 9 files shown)


Fei-Guo:
High level thoughts:

1) You probably only need to introduce one config option called vnagentservicefullname with "ns/name" as the format to simplify the interface.
2) I would suggest to write a new provider instead of changing the default provider. This deployment mode has quite a few side effects in terms of performance and user experiences (now all vnodes show the same IP which may cause problem if any operator uses the node ip). I'd suggest to have the implementation as isolated as possible to avoid user confusion. 
You can use feature gate to directly switch to the new provider.

christopherhein:
> High level thoughts:
> 
> 1. You probably only need to introduce one config option called vnagentservicefullname with "ns/name" as the format to simplify the interface.
> 2. I would suggest to write a new provider instead of changing the default provider. This deployment mode has quite a few side effects in terms of performance and user experiences (now all vnodes show the same IP which may cause problem if any operator uses the node ip). I'd suggest to have the implementation as isolated as possible to avoid user confusion.
>    You can use feature gate to directly switch to the new provider.

Sounds good, I agree great idea.

christopherhein:
@Fei-Guo updated as a provider. Thanks for the feedback.

incubator/virtualcluster/pkg/syncer/vnode/service/provider.go
```diff
@@ -0,0 +1,66 @@
...
+
+type provider struct {
+	vnAgentPort          int32
+	vnAgentNamespaceName string
+	client               clientset.Interface
+}
+
+var _ vnodeprovider.VirtualNodeProvider = &provider{}
+
+func NewVirtualNodeProvider(vnAgentPort int32, vnAgentNamespaceName string, client clientset.Interface) vnodeprovider.VirtualNodeProvider {
```
Fei-Guo, line 37: How about NewServiceVirtualNodeProvider? to compared to NewNativeVirtualNodeProvider.

incubator/virtualcluster/pkg/syncer/apis/config/types.go
```diff
@@ -54,6 +54,10 @@ type SyncerConfiguration struct {
 	// VNAgentPort defines the port that the VN Agent is running on per host
 	VNAgentPort int32
 
+	// VNAgentNamespaceName defines the namespace/name of the VN Agent Kubernetes
+	// service, this is used for feature VNodeProviderService.
+	VNAgentNamespaceName string
```
Fei-Guo, line 59: How about name it as VNAgentNamespacedName?

incubator/virtualcluster/pkg/syncer/vnode/provider/provider.go
```diff
@@ -0,0 +1,27 @@
...
+*/
+
+package provider
+
+import (
+	v1 "k8s.io/api/core/v1"
+)
+
+// NodeProvider is the interface used for registering the node address.
+type VirtualNodeProvider interface {
```
Fei-Guo, line 24: You may need to remove the same definition from syncer/vnode/vnode.go

Fei-Guo (commented):
Overall LG. A few nits.

christopherhein:
Thanks @Fei-Guo for that feedback, all updated.

xiaonancc77:
Now incubator/virtualcluster/pkg/vn-agent/server/route.go 
req.Request.URL.Host = s.config.KubeletServerHost (127.0.0.1)

How make vn-agent find kubelet ip?

christopherhein:
@xiaonancc77 are you using the vn-agent deployment or daemonset route? This PR added a new feature gate that allowed you to use the VN agent as a Deployment as long as you set the `--feature-gate= VNodeProviderService=true` on your vc-syncer it will use this, by default the vn-agent is deployed as a DaemonSet.

xiaonancc77:
@christopherhein  I want to use the vn-agent by deployment route. When I set the --feature-gate=VNodeProviederService=true on vn-syncer, kube-apiserver can get the node address is the service cluster ip of the vn-agent, But vn-agent how can get the address of kubelet?

christopherhein:
> @christopherhein I want to use the vn-agent by deployment route. When I set the --feature-gate=VNodeProviederService=true on vn-syncer, kube-apiserver can get the node address is the service cluster ip of the vn-agent, But vn-agent how can get the address of kubelet?

I see what your question is now, so it works in the relatively less optimal route where instead of connecting the vn-agent directly to each kubelet it will use the super cluster kube-apiserver to proxy so where a normal vn-agent deployment routes like this  user -> tenant apiserver -> vn-agent -> kubelet -> pod when you deploy as a deployment it uses user -> tenant apiserver -> vn-agent -> super kube-apiserver -> kubelet -> pod. 

So you need to make sure the vn-agent service account has the rbac rules to allow it to handle all the proxying requests and not configure the vn-agent with the kubelet certificates

xiaonancc77:
@christopherhein Thanks very much!

xiaonancc77:
@christopherhein 
I have a very confused question. I don't know whether vn-agent is a cluster with tenant apiserver or a cluster with super kube-apiserver? 
If it is a cluster with super kube-apiserver, tenant apierver cannot access vn-agent through ClusterIP of its service.
If vn-agent and tenant apiserver are a cluster, how does vn-agent access upper apiserver by InClusterConfig?

Fei-Guo:
> If it is a cluster with the super kube-apiserver, the tenant apierver cannot access the vn-agent by ClusterIP of its service.

vn-agent is a super cluster pod. The tenant apiserver should be able to access the serivce because it is also a super cluster Pod in current model.  The raw cluster ip should ok.

christopherhein:
Yup, exactly as @Fei-Guo put it.

xiaonancc77:
@Fei-Guo @christopherhein  Thanks for your help.

Issue #1254, opened 2020-11-03
HNC: restrict recreate-on-failed-update semantics to Rolebindings


adrianludwin:
HNC will delete and recreate any object that it can't update; we did this to support modifying rolebindings, but it applies to any type. However, I just realized that if another object has an ownerReference pointing to the object we're deleting, it could be deleted as well, which isn't what we want. This is almost certainly not a problem for rolebindings but could be a problem for other arbitrary types.

Therefore, I think we should restrict this feature to rolebindings, and alert the user for all other failures so they can resolve it manually.

PR #1433, opened 2021-03-12
Restrict recreate-on-failed-update semantics to Rolebindings

GinnyJI:
In #1091, we update HNC to delete and recreate any object that it can't update. However, if another object has an ownerReference pointing to the object we're deleting, it could be deleted as well, which isn't desirable. So we're restricting the recreate-on-failed-update logic to RBs only.

Tested: passed e2e:rolebinding_test.go

Fix: #1254

### DIFF
--- a/incubator/hnc/internal/reconcilers/object.go
+++ b/incubator/hnc/internal/reconcilers/object.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


yiqigao217 (commented):
/lgtm
/hold
For @adrianludwin to approve.

incubator/hnc/internal/reconcilers/object.go
```diff
@@ -588,14 +588,13 @@ func (r *ObjectReconciler) writeObject(ctx context.Context, log logr.Logger, ins
...
-		// type. If we find out later that this assumption is not true, we can update the
-		// logic here to only deal with RoleBinding.
+		// We don't apply this logic to other objects because if another object has an
+		// ownerReference pointing to the object we're deleting, it could be deleted as
+		// well, which is undesirable.
 
 		// The error type is 'Invalid' after I tested it out with different error types
 		// from https://godoc.org/k8s.io/apimachinery/pkg/api/errors
-		if err != nil && errors.IsInvalid(err) {
+		if err != nil && errors.IsInvalid(err) && inst.GetKind() == "RoleBinding" {
```
adrianludwin, line 597: Technically, we should check the API group as well (`rbac.authorization.k8s.io`) since any group could have something called "RoleBinding" in it. Unlikely yes, but we may as well be precise.

yiqigao217 (commented):
/lgtm

adrianludwin (commented):
/lgtm
/approve

PR #1440, opened 2021-03-17
Add reinstall kubebuilder info in test and README

### LINKED ISSUE #1371: HNC: Migrate to v1 apis of CRDs, Mutating WH, and Admission WH (v1beta1 deprecation)
These APIs have moved to GA and the beta versions are now deprecated. They will be completed remove in kubernetes 1.22.

A quick search of the repo showed that we are using the v1beta1 apis in a few places. We also need to keep in mind that if we completely remove the usage of v1beta1 we change our supported versions to the version of kubernetes where v1 is introduced (1.16).

More info here: https://groups.google.com/g/kubernetes-sig-api-machinery/c/OmC8vB7oiGM/m/ZH3qv-0sDwAJ?pli=1

yiqigao217:
Fixes #1371

If you have older version of kubebuilder installed locally on your
computer, the envtest would load older version of etcd and
kube-apiserver from your /usr/local/kubebuilder/bin for tests. Since we
just migrated from v1beta1 to v1 for CRDs, K8s < v1.16 would not know
v1/CRD. Thus the tests would all fail.

Add reinstall kubebuilder info in `make test` and README for developers.

Failed tests look like this - https://gist.github.com/yiqigao217/9394c2aadaa515e82184684a005187af
Passed tests look like this - https://gist.github.com/yiqigao217/d8d097c99df54997915c57fda17f96bb

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -93,6 +93,10 @@ all: test docker-build
 # Run tests
 test: build
 	./hack/crd_patches/ensure-all-served-for-envtest.sh
+	@echo
+	@echo "If tests fail due to no matches for kind \"CustomResourceDefinition\" in version \"apiextensions.k8s.io/v1\","
+	@echo "please remove the old kubebuilder and reinstall it - https://book.kubebuilder.io/quick-start.html#installation"
+	@echo
 	go test ${DIRS} -coverprofile cover.out
 
 # Builds all binaries (manager and kubectl) and manifests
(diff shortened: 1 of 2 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

Once this is merged, the migration to v1 for CRDs and WHs is done.

adrianludwin (commented):
/lgtm
/approve
/hold
/assign @rjbez17

PR #1442, opened 2021-03-18
[VC] add support for all kubectl logs options

charleszheng44:
Add support for all kubectl logs options

### DIFF
--- a/incubator/virtualcluster/pkg/vn-agent/server/translate.go
+++ b/incubator/virtualcluster/pkg/vn-agent/server/translate.go
@@ -48,10 +48,6 @@ func translateRawQuery(req *restful.Request, containerName string) {
 		switch k {
 		case "command":
 			query.Add("command", v[0])
-		case "follow":
-			if v[0] == "true" {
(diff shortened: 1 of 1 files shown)


incubator/virtualcluster/pkg/vn-agent/server/translate.go
```diff
@@ -74,11 +70,17 @@ func translateRawQuery(req *restful.Request, containerName string) {
 				query.Add("stdout", "false")
 			}
 		case "tailLines":
+			// for log option --tail
 			klog.V(5).Infof("the tail parameter is %s", v[0])
 			if v[0] != "all" {
 				// "all" is the same as omitting tail
-				query["tailLines"] = []string{v[0]}
+				query.Add("tailLines", v[0])
```
Fei-Guo, line 77: Can we use query.Add(k, v[0]) as well?

charleszheng44: Yes. I just found out that "tailLines" can be handled in the same way as other options

Issue #1443, opened 2021-03-22
HNC: Should not allow creating subns with excluded namespace names, e.g. `kube-system`


yiqigao217:
While testing configurable excluded namespaces, I found that we may want to add these two new rules in our valiators:
```
#####  Should deny creating subnamespaces with excluded namespace names, but got: #####
$ kubectl hns create kube-public -n a
Successfully created "kube-public" subnamespace anchor in "a" namespace

##### Should output a better message, but got: #####
$ kubectl hns set kube-node-lease --parent a
Setting the parent of kube-node-lease to a

Could not update the hierarchical configuration of kube-node-lease.
Reason: admission webhook "hierarchyconfigurations.hnc.x-k8s.io" denied the request: HNC has not reconciled namespace "kube-node-lease" yet - please try again in a few moments.
```

adrianludwin:
+100

yiqigao217:
Will also look into conditions if the webhook is bypassed. Currently when
* Setting a non-excluded namespace as child of excluded namespace, we get:
```
$ kubectl hns set a --parent kube-system
Setting the parent of a to kube-system
Succesfully updated 1 property of the hierarchical configuration of a

$ k hns tree kube-system
kube-system

$ k hns tree a
a (1)

Conditions:
1) ActivitiesHalted (ParentMissing): Parent "kube-system" does not exist
```
* Setting an excluded namespace to a non-excluded namespace, we get nothing but the HC object is created.
* Creating subns in excluded namespace with an existing name, we get `Conflict` in `subns.status`.
* Creating subns in exlcuded namespace with a new name, we get `Forbidden` in `subns.status`.
* Creating subns using excluded namespace name in a non-excluded namespace, we get `Forbidden` in `subns.status`.

We need to make the first two conditions more explicit and specific for excluded namespaces.

yiqigao217:
My proposal for the above **webhook-bypassed** cases are

*  Setting a non-excluded namespace as child of excluded namespace - The HC singleton will get `ActivitiesHalted (ParentMissing): Parent "kube-system" is an excluded namespace` condition;
*  Setting an excluded namespace as a child of a non-excluded namespace - Our reconciler will remove the HC singleton directly instead of adding a new condition for it (say `BadConfiguration (NamespaceExcluded): "kube-system" is an excluded namespace and should not have a hierarchyconfiguration object`);
* All 3 subns status mentioned above remain the same.

[...]

PR #1444, opened 2021-03-23
Make excluded namespace configurable

### LINKED ISSUE #374: HNC: make controller exclusions configurable
Currently, we hardcoded to exclude namespaces like `kube-system` or `cert-manager` for object controllers to reconcile objects.

Such exclusions need to be configurable.

yiqigao217:
Fixes #374, #1023.
Part of #1443.
See design https://bit.ly/hnc-excluded-namespaces.

Add `excluded-namespace` container args in the
config/manager/manager.yaml to allow user defined excluded namespaces.
The default excluded namespaces are kube-system, kube-public, hnc-system
and kube-node-lease. Add excluded-namespace label to hnc-system by
default to avoid deadlock (between cert-rotator writing secret vs object
webhook failing CLOSE) when installing HNC. Let users to add
excluded-namespaces to other namespaces.

Replace the hard-coded `EX` excluded namespaces.

Add webhook patch only to the object validator with namespaceSelector to
filter excluded namespaces. Add comments on why not patching other
validators. Make object validator fail CLOSE now.

Patch object webhook to only apply on "namespaced" scope objects, so
that cert-rotator can insert certs to VWHConfiguration and namespaced
objects are indeed what we care in HNC.

Add webhook rules:
1) Do not allow creating/updating non-excluded namespace with excluded
label;
2) Do not allow excluded namespaces to be created as a subnamespace;
3) Do not allow setting an excluded namespace as a child.

Make hierarchyConfiguration reconciler remove excluded-namespace label
on non-excluded namespaces.

Add excluded-namespace label to user-guide/concepts.md doc.

Tested by make test and make test-e2e.

### DIFF
--- a/incubator/hnc/cmd/manager/main.go
+++ b/incubator/hnc/cmd/manager/main.go
@@ -52,17 +52,19 @@ var (
 )
 
 var (
-	metricsAddr            string
-	maxReconciles          int
-	enableLeaderElection   bool
-	leaderElectionId       string
-	novalidation           bool
-	debugLogs              bool
-	testLog                bool
-	internalCert           bool
-	qps                    int
-	webhookServerPort      int
-	restartOnSecretRefresh bool
+	metricsAddr             string
+	maxReconciles           int
+	enableLeaderElection    bool
+	leaderElectionId        string
+	novalidation            bool
+	debugLogs               bool
+	testLog                 bool
+	internalCert            bool
+	qps                     int
+	webhookServerPort       int
+	restartOnSecretRefresh  bool
+	unpropagatedAnnotations arrayArg
+	excludedNamespaces      arrayArg
 )
 
 func init() {
@@ -91,10 +93,16 @@ func main() {
 	flag.IntVar(&qps, "apiserver-qps-throttle", 50, "The maximum QPS to the API server. See the user guide for more information.")
 	flag.BoolVar(&stats.SuppressObjectTags, "suppress-object-tags", true, "If true, suppresses the kinds of object metrics to reduce metric cardinality. See the user guide for more information.")
 	flag.IntVar(&webhookServerPort, "webhook-server-port", 443, "The port that the webhook server serves at.")
-	uaArg := arrayArg{val: &config.UnpropagatedAnnotations}
-	flag.Var(&uaArg, "unpropagated-annotation", "An annotation that, if present, will be stripped out of any propagated copies of an object. May be specified multiple times, with each instance specifying one annotation. See the user guide for more information.")
+	flag.Var(&unpropagatedAnnotations, "unpropagated-annotation", "An annotation that, if present, will be stripped out of any propagated copies of an object. May be specified multiple times, with each instance specifying one annotation. See the user guide for more information.")
+	flag.Var(&excludedNamespaces, "excluded-namespace", "A namespace that, if present, will be excluded from HNC management. May be specified multiple times, with each instance specifying one namespace. See the user guide for more information.")
 	flag.BoolVar(&restartOnSecretRefresh, "cert-restart-on-secret-refresh", false, "Kills the process when secrets are refreshed so that the pod can be restarted (secrets take up to 60s to be updated by running pods)")
 	flag.Parse()
+	// Assign the array args to the configuration variables after the args are parsed.
+	config.UnpropagatedAnnotations = unpropagatedAnnotations
+	config.ExcludedNamespaces = make(map[string]bool)
+	for _, exn := range excludedNamespaces {
+		config.ExcludedNamespaces[exn] = true
+	}
 
 	// Enable OpenCensus exporters to export metrics
 	// to Stackdriver Monitoring.
@@ -218,18 +226,13 @@ func startControllers(mgr ctrl.Manager, certsCreated chan struct{}) {
 // arrayArg is an arg that can be specified multiple times. It implements
 // https://golang.org/pkg/flag/#Value an is based on
 // https://stackoverflow.com/questions/28322997/how-to-get-a-list-of-values-into-a-flag-in-golang.
-type arrayArg struct {
-	val *[]string
-}
+type arrayArg []string
 
-func (a *arrayArg) String() string {
-	if a == nil || a.val == nil {
-		return ""
-	}
-	return strings.Join(*a.val, ",")
+func (a arrayArg) String() string {
+	return strings.Join(a, ", ")
 }
 
 func (a *arrayArg) Set(val string) error {
-	*a.val = append(*a.val, val)
+	*a = append(*a, val)
 	return nil
 }
--- a/incubator/hnc/config/webhook/webhook_patch.yaml
+++ b/incubator/hnc/config/webhook/webhook_patch.yaml
@@ -7,6 +7,35 @@ webhooks:
 - name: objects.hnc.x-k8s.io
   timeoutSeconds: 2
   sideEffects: None
+  # We only filter out excluded namespaces from this object validator so that
+  # when HNC (webhook service specifically) is down, operations in the excluded
+  # namespaces won't be affected. Validators on HNC CRs are not filtered because
+  # they are supposed to prevent abuse of HNC CRs in excluded namespaces.
+  # Namespace validator is not filtered to prevent abuse of excluded namespace.
+  # Unfortunately, this means that when HNC is down, we will block updates on all
+  # namespaces, even "excluded" ones, but anyone who can update namespaces like
+  # `kube-system` should likely be able to delete the VWHConfiguration to make
+  # the updates.
+  namespaceSelector:
+    matchExpressions:
+    - key: hnc.x-k8s.io/excluded-namespace
+      operator: DoesNotExist
+  rules:
+  # This overwrites the rules specified in the object validator to patch object
+  # scope of `namespaced` since there's no kubebuilder marker for `scope`.
+  # There's no way to just patch "scope: Namespaced" to the first rule, since
+  # `rules` takes a list of rules and we can only patch the entire `rules`.
+  - apiGroups:
+    - '*'
+    apiVersions:
+    - '*'
+    operations:
+    - CREATE
+    - UPDATE
+    - DELETE
+    resources:
+    - '*'
+    scope: "Namespaced"
 - name: subnamespaceanchors.hnc.x-k8s.io
   sideEffects: None
 - name: hierarchyconfigurations.hnc.x-k8s.io
--- a/incubator/hnc/docs/user-guide/concepts.md
+++ b/incubator/hnc/docs/user-guide/concepts.md
@@ -498,7 +498,6 @@ objects, in addition to using the custom resources it defines.
 These annotations may be added to any namespaced object to define exceptions to
 propagation rules. More information to come.
 
-
 #### hnc.x-k8s.io/managed-by (annotation on namespaces)
 
 This annotation is mainly designed for use by external products such as GKE
@@ -529,6 +528,25 @@ namespace if this annotation already exists. The two are mutually exclusive.
 
 We are considering replacing this with the standard
 `app.kubernetes.io/managed-by` label in the future.
+
+<a name="excluded-namespace-label">
+
+#### hnc.x-k8s.io/excluded-namespace (label on namespaces)
+
+This label should be added to namespaces such as `kube-system` and `kube-public`
+so that HNC's validating webhook cannot accidentally prevent operations in these
+namespaces and block critical cluster operations.
+
+Before installing HNC, users can customize the excluded namespace list in the HNC
+deployment with a container arg called `excluded-namespace` in
+`config/manager/manager.yaml` and then set this label on the excluded namespaces.
+Setting this label on namespaces that are not
+listed in the HNC deployment as an `excluded-namespace` is not allowed.
+
+As of March 2021, the default excluded namespaces listed in [config/manager/manager.yaml](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/config/manager/manager.yaml)
+are `kube-system`, `kube-public`, `hnc-system` and
+`kube-node-lease`. HNC adds this label to `hnc-system` namespace by default, so
+users will have to add this label to other excluded namespaces manually.
 
 <a name="admin-labels-set">
 
--- a/incubator/hnc/internal/reconcilers/anchor.go
+++ b/incubator/hnc/internal/reconcilers/anchor.go
@@ -75,9 +75,9 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// Report "Forbidden" state and early exit if the namespace is not allowed to have subnamespaces
 	// but has bypassed the webhook and successfully created the anchor. Forbidden anchors won't have
 	// finalizers.
-	// TODO refactor/split the EX map for 1) reconciler exclusion and 2) subnamespaces exclusion
-	// purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
-	if config.EX[pnm] {
+	// TODO refactor/split the ExcludedNamespaces map for 1) reconciler exclusion and 2) subnamespaces exclusion
+	//  purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
+	if config.ExcludedNamespaces[pnm] {
 		inst.Status.State = api.Forbidden
 		return ctrl.Result{}, r.writeInstance(ctx, log, inst)
 	}
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -85,7 +85,7 @@ type HierarchyConfigReconciler struct {
 
 // Reconcile sets up some basic variables and then calls the business logic.
 func (r *HierarchyConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	if config.EX[req.Namespace] {
+	if config.ExcludedNamespaces[req.Namespace] {
 		return ctrl.Result{}, nil
 	}
 
@@ -112,6 +112,10 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 		return err
 	}
 	origNS := nsInst.DeepCopy()
+
+	// Remove excluded namespace label if there's one on the namespace since
+	// the excluded namespaces should be already skipped earlier.
+	r.removeExcludedNamespaceLabel(log, nsInst)
 
 	// Get singleton from apiserver. If it doesn't exist, initialize one.
 	inst, deletingCRD, err := r.getSingleton(ctx, nm)
@@ -168,6 +172,17 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, nm strin
 		ns.UnsetExists()
 		log.Info("Namespace has been deleted")
 	}
+}
+
+// removeExcludedNamespaceLabel removes the `hnc.x-k8s.io/excluded-namespace`
+// label from the namespace.
+func (r *HierarchyConfigReconciler) removeExcludedNamespaceLabel(log logr.Logger, nsInst *corev1.Namespace) {
+	lbs := nsInst.Labels
+	if _, found := lbs[api.LabelExcludedNamespace]; found {
+		log.Info("Illegal excluded-namespace label found; removing")
+		delete(lbs, api.LabelExcludedNamespace)
+	}
+	nsInst.SetLabels(lbs)
 }
 
 // updateFinalizers ensures that the HC can't be deleted if there are any subnamespace anchors in
--- a/incubator/hnc/internal/validators/anchor.go
+++ b/incubator/hnc/internal/validators/anchor.go
@@ -79,8 +79,13 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
 	switch req.op {
 	case k8sadm.Create:
 		// Can't create subnamespaces in excluded namespaces
-		if config.EX[pnm] {
-			msg := fmt.Sprintf("The namespace %s is not allowed to create subnamespaces. Please create subnamespaces in a different namespace.", pnm)
+		if config.ExcludedNamespaces[pnm] {
+			msg := fmt.Sprintf("Cannot create a subnamespace in the excluded namespace %q", pnm)
+			return deny(metav1.StatusReasonForbidden, msg)
+		}
+		// Can't create subnamespaces using excluded namespace names
+		if config.ExcludedNamespaces[cnm] {
+			msg := fmt.Sprintf("Cannot create a subnamespace using the excluded namespace name %q", cnm)
 			return deny(metav1.StatusReasonForbidden, msg)
 		}
 
@@ -89,7 +94,7 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
 		if cns.Exists() {
 			childIsMissingAnchor := (cns.Parent().Name() == pnm && cns.IsSub)
 			if !childIsMissingAnchor {
-				msg := fmt.Sprintf("The requested namespace %s already exists. Please use a different name.", cnm)
+				msg := fmt.Sprintf("Cannot create a subnamespace using an existing namespace name %q", cnm)
 				return deny(metav1.StatusReasonConflict, msg)
 			}
 		}
--- a/incubator/hnc/internal/validators/hierarchy.go
+++ b/incubator/hnc/internal/validators/hierarchy.go
@@ -119,6 +119,15 @@ func (v *Hierarchy) handle(ctx context.Context, log logr.Logger, req *request) a
 		return allow("HNC SA")
 	}
 
+	if config.ExcludedNamespaces[req.hc.Namespace] {
+		reason := fmt.Sprintf("Cannot set the excluded namespace %q as a child of another namespace", req.hc.Namespace)
+		return deny(metav1.StatusReasonForbidden, reason)
+	}
+	if config.ExcludedNamespaces[req.hc.Spec.Parent] {
+		reason := fmt.Sprintf("Cannot set the parent to the excluded namespace %q", req.hc.Spec.Parent)
+		return deny(metav1.StatusReasonForbidden, reason)
+	}
+
 	// Do all checks that require holding the in-memory lock. Generate a list of server checks we
 	// should perform once the lock is released.
 	serverChecks, resp := v.checkForest(log, req.hc)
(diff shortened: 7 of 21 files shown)


yiqigao217:
/assign @adrianludwin 
/assign @rjbez17 

(See commit message) I was not able to think of a good e2e test case for this, so I did a manual test. Please let me know if you have any sugguestions, thanks!

incubator/hnc/config/webhook/webhook_patch.yaml
```diff
@@ -7,6 +7,11 @@ webhooks:
 - name: objects.hnc.x-k8s.io
   timeoutSeconds: 2
   sideEffects: None
+  namespaceSelector:
+    matchExpressions:
+    - key: hnc.x-k8s.io/excluded-namespace
+      operator: NotIn
+      values: ["true"]
```
adrianludwin, line 14: Maybe use `DoesNotExist` as the operator? See https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -77,7 +77,7 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// finalizers.
 	// TODO refactor/split the EX map for 1) reconciler exclusion and 2) subnamespaces exclusion
 	// purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
```
adrianludwin, line 79: Update?

incubator/hnc/cmd/manager/main.go
```diff
@@ -93,6 +93,8 @@ func main() {
 	flag.IntVar(&webhookServerPort, "webhook-server-port", 443, "The port that the webhook server serves at.")
 	uaArg := arrayArg{val: &config.UnpropagatedAnnotations}
 	flag.Var(&uaArg, "unpropagated-annotation", "An annotation that, if present, will be stripped out of any propagated copies of an object. May be specified multiple times, with each instance specifying one annotation. See the user guide for more information.")
+	exnArg := mapArg{val: &config.ExcludedNamespaces}
```
adrianludwin, line 96: I feel like I'd rather just have a single arrayArg and then transfer the values to the config objects, e.g.:

```
config.UnpropagatedAnnotations = uaArg.Get()
for _, ex := range exnArg.Get() {
  config.ExcludedNamespaces[ex] = true
}
```

Seems more Go-like with less magic.

incubator/hnc/config/webhook/webhook_patch.yaml
```diff
@@ -7,6 +7,11 @@ webhooks:
 - name: objects.hnc.x-k8s.io
   timeoutSeconds: 2
   sideEffects: None
+  namespaceSelector:
```
adrianludwin, line 10: Should we put this on the namespace webhook too? If not, why not? Maybe comment why this only applies to certain namespaces here.

incubator/hnc/internal/reconcilers/suite_test.go
```diff
@@ -69,6 +70,14 @@ func TestAPIs(t *testing.T) {
...
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+	// Set excluded namespaces for testing.
+	config.ExcludedNamespaces = map[string]bool{
+		"kube-system":     true,
+		"kube-public":     true,
+		"hnc-system":      true,
+		"kube-node-lease": true,
+	}
+
```
adrianludwin, line 80: Do we need this full list for testing?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -99,6 +106,16 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
 	return allow("")
 }
 
+func (v *Namespace) illegalExcludedNamespaceLabel(req *nsRequest) admission.Response {
+	for l := range req.ns.Labels {
+		if l == api.LabelExcludedNamespace && !config.ExcludedNamespaces[req.ns.Name] {
+			msg := fmt.Sprintf("The namespace contains an illegal label %q, which should only be applied on the excluded namespaces set in the deployment configuration", api.LabelExcludedNamespace)
```
adrianludwin, line 112: This is a very low-level explanation of what's gone wrong. Maybe add some documentation and then reference that in the error message? I.e. don't call it an "illegal label," talk about the problem domain (e.g. "You can't exclude this namespace using the %q label from HNC because ..." or something like that).

incubator/hnc/config/webhook/webhook_patch.yaml
yiqigao217: Seems there are only `in`, `notin`, or `exists` according to your link https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement? So `notin` is the best option?

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: This one was rotten and you mentioned it's not important to fix, so I left it there in case we want to revisit it in the future? It's about splitting the excluded namespaces for reconcilers or self-serving specific in the future if users want different list of namespaces. Deleting this comment will lose the anchor of the issue in the code, WDYT?

incubator/hnc/config/webhook/webhook_patch.yaml
adrianludwin: I think that's what's available on the command line. Look at the "Resources that support set-based requirements" section for what's available through the API; that's where I saw `DoesNotExist`.

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: Ok, maybe just update the name in that case ("EX map")

incubator/hnc/config/webhook/webhook_patch.yaml
yiqigao217: Ah right, no I should put it on namespace validator too. Initially, I added the `namespaceSelector` to all validators and realized for HNC CRDs, e.g. `hc` or `anchor`, we should not filter excluded namespaces since we want to deny setting excluded namespaces in the hierarchy. So I removed the `namespaceSelector` on them, but I should keep it on the namespace validator (just tested it that if not, updates to the e.g. `kube-system` namespace will also get denied when webhook service is down and filtering it out won't affect blocking the excluded namespaces to be part of the hierarchy).

Will add it back on the namespace validtor.

yiqigao217: Ah the issue is that if I have it filtered, it won't block user adding `hnc.x-k8s.io/excluded-namespace=true` to other non-excluded namespaces. Let me look into how to work it around.

yiqigao217: Discussed offline with @adrianludwin, here's an example if having `hnc.x-k8s.io/excluded-namespace` label value `NotIn` `true` on the namespace validator:
```
$ k label ns default hnc.x-k8s.io/excluded-namespace=false
Error from server (Forbidden): admission webhook "namespaces.hnc.x-k8s.io" denied the request: The namespace contains an illegal label "hnc.x-k8s.io/excluded-namespace", which should only be applied on the excluded namespaces set in the deployment configuration

$ k label ns default hnc.x-k8s.io/excluded-namespace=true
namespace/default labeled
```
, that mismatching label is denied by the webhook but the matching label is exempt from validation.

I will keep the `namespaceSelector` for namespace validator and remove the label validation. On the other hand, remove the excluded-namespace label from the non-excluded namespaces in our reconcilers.

adrianludwin: sgtm. That's really unfortunate behaviour IMO in K8s but it is what it is!

yiqigao217: Added comment for only adding `namespaceSelector` on object validator but not on namespace validator:
```
  # We only filter out excluded namespaces from this object validator so that
  # when HNC (webhook service specifically) is down, operations in the excluded
  # namespaces won't be affected. Validators on HNC CRs are not filtered because
  # they are supposed to prevent abuse of HNC CRs in excluded namespaces.
  # Namespace validator is not filtered to prevent abuse of excluded namespace.
  # When HNC is down, updates on excluded namespace are forbidden, but those who
  # can update namespaces like `kube-system` should likely be able to delete the
  # VWHConfiguration to make the updates
```

incubator/hnc/internal/reconcilers/suite_test.go
yiqigao217: Removed this central list and only set the value before each test when necessary.

incubator/hnc/internal/validators/namespace.go
yiqigao217: documented and linked.

yiqigao217:
Please note in this PR, I added the `hnc.x-k8s.io/excluded-namespace` label to `hnc-system` namespace by default, so that HNC installation could work, because otherwise there will be a deadlock between cert-rotator writing secret VS object webhook failing CLOSE to block it. See PR message.

I will update the design doc to reflect it if it SGTY @adrianludwin @rjbez17 .

Additionally the e2e tests always pass on a clean cluster but may be flaky on problematic cluster (with resource stuck on terminating status). I will look further into it tomorrow.

incubator/hnc/docs/user-guide/concepts.md
```diff
@@ -530,6 +529,18 @@ namespace if this annotation already exists. The two are mutually exclusive.
 We are considering replacing this with the standard
 `app.kubernetes.io/managed-by` label in the future.
 
+#### hnc.x-k8s.io/excluded-namespace (label on namespaces)
+
+This label is supposed to be added to the excluded namespaces such as
+`kube-system`, `kube-public`, and `hnc-system` so that when HNC validating
+webhook services are down, operations in these excluded namespaces won't be
+affected.
```
adrianludwin, line 537: Suggestion:

This label should be added to namespaces such as `kube-system` and `kube-public` so that HNC's validating webhook cannot accidentally prevent operations in these namespaces and block critical cluster operations.

incubator/hnc/docs/user-guide/concepts.md
```diff
@@ -530,6 +529,18 @@ namespace if this annotation already exists. The two are mutually exclusive.
...
+
```
adrianludwin, line 538: State what the defaults are (as of March 2021, in case it changes) and maybe link to the relevant file and line. Also mention that HNC manifests exclude `hnc-system` by default, but that you need to set up the rest yourself.

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -75,9 +75,9 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// Report "Forbidden" state and early exit if the namespace is not allowed to have subnamespaces
 	// but has bypassed the webhook and successfully created the anchor. Forbidden anchors won't have
 	// finalizers.
-	// TODO refactor/split the EX map for 1) reconciler exclusion and 2) subnamespaces exclusion
-	// purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
-	if config.EX[pnm] {
+	// TODO refactor/split the ExcludedNamespaces map for 1) reconciler exclusion and 2) subnamespaces exclusion
+	//  purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
```
adrianludwin, line 79: Extra space at the beginning of the line

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -170,6 +174,22 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, nm strin
 	}
 }
 
+// removeExcludedNamespaceLabel removes the `hnc.x-k8s.io/excluded-namespace`
+// label from the namespace.
+func (r *HierarchyConfigReconciler) removeExcludedNamespaceLabel(log logr.Logger, nsInst *corev1.Namespace) {
+	old := nsInst.Labels
+	new := map[string]string{}
```
adrianludwin, line 181: Why not `delete(old, api.LabelExcludedNamespace)`?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -170,6 +174,22 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, nm strin
...
+	for l, v := range old {
+		if l == api.LabelExcludedNamespace {
+			msg := fmt.Sprintf("The %q label is removed since this is not an excluded namespace specified in the deployment configuration", api.LabelExcludedNamespace)
+			log.Info(msg)
```
adrianludwin, line 185: Use structured logging, not fmt

incubator/hnc/internal/validators/anchor.go
```diff
@@ -79,8 +79,13 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
...
 	case k8sadm.Create:
 		// Can't create subnamespaces in excluded namespaces
-		if config.EX[pnm] {
-			msg := fmt.Sprintf("The namespace %s is not allowed to create subnamespaces. Please create subnamespaces in a different namespace.", pnm)
+		if config.ExcludedNamespaces[pnm] {
+			msg := fmt.Sprintf("The namespace %q is not allowed to create subnamespaces. Please create subnamespaces in a different namespace.", pnm)
+			return deny(metav1.StatusReasonForbidden, msg)
+		}
+		// Can't create subnamespaces using excluded namespace names
+		if config.ExcludedNamespaces[cnm] {
```
adrianludwin, line 87: Do we not have a check that you can't create an anchor for a namespace that already exists (unless that namespace is a subns that's lost its anchor)?

incubator/hnc/internal/validators/hierarchy.go
```diff
@@ -119,6 +119,15 @@ func (v *Hierarchy) handle(ctx context.Context, log logr.Logger, req *request) a
...
 
+	if config.ExcludedNamespaces[req.hc.Namespace] {
+		reason := fmt.Sprintf("Cannot set the excluded namespace %q as a child of another namespace", req.hc.Namespace)
+		return deny(metav1.StatusReasonForbidden, reason)
+	}
+	if config.ExcludedNamespaces[req.hc.Spec.Parent] {
+		reason := fmt.Sprintf("Cannot set the parent to the excluded namespace %q", req.hc.Spec.Parent)
+		return deny(metav1.StatusReasonForbidden, reason)
+	}
+
```
adrianludwin, line 130: Should these cause conditions if the webhook is bypassed?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -99,6 +106,16 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
 	return allow("")
 }
 
+func (v *Namespace) illegalExcludedNamespaceLabel(req *nsRequest) admission.Response {
+	for l := range req.ns.Labels {
+		if l == api.LabelExcludedNamespace && !config.ExcludedNamespaces[req.ns.Name] {
+			msg := fmt.Sprintf("You cannot exclude this namespace using the %q label. See https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/concepts.md#hncx-k8sioexcluded-namespace-label-on-namespaces for detail.", api.LabelExcludedNamespace)
```
adrianludwin, line 112: That HTML anchor (everything after the '#') looks brittle and likely to change. Maybe put an explicit anchor into the file and refer to that?

incubator/hnc/internal/validators/namespace.go
```diff
@@ -99,6 +106,16 @@ func (v *Namespace) handle(req *nsRequest) admission.Response {
 	return allow("")
 }
 
+func (v *Namespace) illegalExcludedNamespaceLabel(req *nsRequest) admission.Response {
+	for l := range req.ns.Labels {
+		if l == api.LabelExcludedNamespace && !config.ExcludedNamespaces[req.ns.Name] {
```
adrianludwin, line 111: What about if that namespace _already_ has the label? Should we block further changes to that namespace until the label is deleted?

incubator/hnc/config/webhook/webhook_patch.yaml
```diff
@@ -7,6 +7,32 @@ webhooks:
...
+    - '*'
+    apiVersions:
+    - '*'
+    operations:
+    - CREATE
+    - UPDATE
+    - DELETE
+    resources:
+    - '*'
+    scope: "Namespaced"
```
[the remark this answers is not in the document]
yiqigao217: Updated the comment to
```
  # This overwrites the rules specified in the object validator to patch object
  # scope of `namespaced` since there's no kubebuilder marker for `scope`.
  # There's no way to just patch "scope: Namespaced" to the first rule, since
  # `rules` takes a list of rules and we can only patch the entire `rules`.
```

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: This indention is on purpose for IntelliJ to highlight the TODO paragraph since this line is part of the TODO. Otherwise only the first line of TODO will be highlighted.

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Ah right updated.

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: removed since using `delete()`.

incubator/hnc/internal/validators/anchor.go
yiqigao217: We do, but we use forest `cns.Exists()` right below this block. The `setExist()` is in HC reconciler but only when the namespace is not excluded (there's an early exit at the top of `Reconcile()` for excluded namespaces).

incubator/hnc/internal/validators/hierarchy.go
yiqigao217: Details added in https://github.com/kubernetes-sigs/multi-tenancy/issues/1443#issuecomment-811384761. I will mark this PR is part of #1443 instead of fixing it and work on it (standardize the condition) in a next PR.

incubator/hnc/internal/validators/namespace.go
yiqigao217: Yes, I updated the message to say "Please remove the %q label. See http://... for detail".

incubator/hnc/internal/reconcilers/anchor.go
adrianludwin: Ack, thanks

incubator/hnc/internal/validators/namespace.go
adrianludwin: I'm not a huge fan of this. I feel like webhooks should only validate the _changes_ in a resource, not what's already there. For example, imagine an automated process that's trying to make some unrelated change; we shouldn't block them. And we shouldn't have to ask the user to remove the label given that we've got a reconciler that can do it automatically.

incubator/hnc/config/webhook/webhook_patch.yaml
```diff
@@ -7,6 +7,32 @@ webhooks:
 - name: objects.hnc.x-k8s.io
   timeoutSeconds: 2
   sideEffects: None
+  # We only filter out excluded namespaces from this object validator so that
+  # when HNC (webhook service specifically) is down, operations in the excluded
+  # namespaces won't be affected. Validators on HNC CRs are not filtered because
+  # they are supposed to prevent abuse of HNC CRs in excluded namespaces.
+  # Namespace validator is not filtered to prevent abuse of excluded namespace.
+  # When HNC is down, updates on excluded namespace are forbidden, but those who
```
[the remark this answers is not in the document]
adrianludwin: Oh, I see. Maybe emphasize this a bit more? E.g.

> Unfortunately, this means that when HNC is down, we will block updates on _all_ namespaces, even "excluded" ones, but anyone who can update namespaces like `kube-system` should ...

incubator/hnc/docs/user-guide/concepts.md
```diff
@@ -530,6 +529,25 @@ namespace if this annotation already exists. The two are mutually exclusive.
...
+namespaces and block critical cluster operations.
+
+Before installing HNC, users can customize the excluded namespace list in the HNC
+deployment with a container arg called `excluded-namespace` in
+`config/manager/manager.yaml` and then set this label on the excluded namespaces.
+Setting this label on namespaces that are not
+listed in the HNC deployment as an `excluded-namespace` is not allowed.
+
+As of March 2021, the default excluded namespaces listed in
+`config/manager/manager.yaml` are `kube-system`, `kube-public`, `hnc-system` and
```
adrianludwin, line 547: You can actually link to the file here if you like

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -170,6 +174,14 @@ func (r *HierarchyConfigReconciler) onMissingNamespace(log logr.Logger, nm strin
 	}
 }
 
+// removeExcludedNamespaceLabel removes the `hnc.x-k8s.io/excluded-namespace`
+// label from the namespace.
+func (r *HierarchyConfigReconciler) removeExcludedNamespaceLabel(nsInst *corev1.Namespace) {
+	lbs := nsInst.Labels
```
adrianludwin, line 180: It's not a bad idea to log something here, since this is definitely unusual.

```
if _, found := lbs[api.LabelExcludedNamespace]; found {
  log.Info("Illegal excluded-namespace label found; removing")
}
```

incubator/hnc/internal/validators/anchor.go
```diff
@@ -79,8 +79,13 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
 	switch req.op {
 	case k8sadm.Create:
 		// Can't create subnamespaces in excluded namespaces
-		if config.EX[pnm] {
-			msg := fmt.Sprintf("The namespace %s is not allowed to create subnamespaces. Please create subnamespaces in a different namespace.", pnm)
+		if config.ExcludedNamespaces[pnm] {
+			msg := fmt.Sprintf("The namespace %q is not allowed to create subnamespaces. Please create subnamespaces in a different namespace.", pnm)
```
adrianludwin, line 83: I know this isn't a new message for this PR, but... the namespace %q isn't creating the namespace, the user is :) Also the second sentence doesn't really add a lot, but it's more important now that the list is configurable to give users an idea of what the problem is.

Your new error messages below are really good ("Cannot set the excluded namespace %q as a child of another namespace") so maybe follow that pattern here?

incubator/hnc/internal/validators/anchor.go
adrianludwin: Maybe worth a comment?

incubator/hnc/internal/validators/anchor.go
```diff
@@ -79,8 +79,13 @@ func (v *Anchor) handle(req *anchorRequest) admission.Response {
...
+			return deny(metav1.StatusReasonForbidden, msg)
+		}
+		// Can't create subnamespaces using excluded namespace names
+		if config.ExcludedNamespaces[cnm] {
+			msg := fmt.Sprintf("%q is not allowed to be created as a subnamespaces. Please use a different name for the subnamespace.", cnm)
```
adrianludwin, line 88: Maybe: "The namespace %q is excluded from HNC and cannot be created as a subnamespace."

adrianludwin (commented):
lgtm after these last few minor fixes

/cc @rjbez17

incubator/hnc/internal/validators/namespace.go
yiqigao217: sg. Changed the message back to "You cannot exclude this namespace using the %q label. See..for detail"
and commented here with:
```
			// Note: this only blocks the request if it has a newly added illegal
			// excluded-namespace label because existing illegal excluded-namespace
			// label should have already been removed by our reconciler. For example,
			// even when the VWHConfiguration is removed, adding the label to a non-
			// excluded namespace would pass but the label is immediately removed; when
			// the VWHConfiguration is there but the reconcilers are down, any request
			// gets denied anyway.
```

incubator/hnc/internal/validators/anchor.go
yiqigao217: Updated all three here.

incubator/hnc/internal/validators/anchor.go
yiqigao217: Updated to:
"Cannot create a subnamespace in the excluded namespace %q"
"Cannot create a subnamespace using the excluded namespace name %q",
"Cannot create a subnamespace using an existing namespace name %q"

yiqigao217:
FYI the doc is updated in the "Detailed Design" section. PTAL thanks!
https://bit.ly/hnc-excluded-namespaces

incubator/hnc/docs/user-guide/concepts.md
```diff
@@ -530,6 +529,25 @@ namespace if this annotation already exists. The two are mutually exclusive.
...
+so that HNC's validating webhook cannot accidentally prevent operations in these
+namespaces and block critical cluster operations.
+
+Before installing HNC, users can customize the excluded namespace list in the HNC
+deployment with a container arg called `excluded-namespace` in
+`config/manager/manager.yaml` and then set this label on the excluded namespaces.
+Setting this label on namespaces that are not
+listed in the HNC deployment as an `excluded-namespace` is not allowed.
+
+As of March 2021, the default excluded namespaces listed in [config/manager/manager.yaml](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/config/manager/manager.yaml)
```
rjbez17, line 546: I kind of feel this should be called out in the installation section or linked from their. As it's a pre/post installation action that is somewhat hidden away but also highly recommended for users to do.

rjbez17 (commented):
lgtm after the one question

incubator/hnc/docs/user-guide/concepts.md
adrianludwin: Yeah, it should probably be mentioned [here](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/how-to.md#install-or-upgrade-hnc-on-a-cluster), but it will be more important to call it out in the release notes, which is where the latest installation instructions are.

yiqigao217: Filed https://github.com/kubernetes-sigs/multi-tenancy/issues/1466 to add excluded namespace configuration to v0.8 release notes and the installation instruction.

yiqigao217:
/hold cancel
Asked Ryan offline.

PR #1452, opened 2021-03-30
[VC] Revise SuperClusterPooling demo

Fei-Guo:
This change revises the demo md file for a better description.


zhuangqh:
good writing!

PR #1456, opened 2021-03-31
[VC] Fix vNodeGC and Pod checker bugs when enabling SuperClusterPooling feature

Fei-Guo:
This change fixes two problems when enabling SuperClusterPooling feature.

1) The Pod checker needs to exclude vPod that is not scheduled to the syncer's cluster;
2) The vNodeGC module should not check the vNode that does not belong to the syncer's cluster;

The checkers of other resources should be fixed as well, which will be done later.

### DIFF
--- a/incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
+++ b/incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
@@ -36,6 +36,8 @@ import (
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/metrics"
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/patrol/differ"
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/util"
+	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/syncer/util/featuregate"
+	utilconstants "sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/util/constants"
 	"sigs.k8s.io/multi-tenancy/incubator/virtualcluster/pkg/util/reconciler"
 )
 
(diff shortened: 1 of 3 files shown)


incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
```diff
@@ -163,6 +165,12 @@ func (c *controller) PatrollerDo() {
 		}
 		vList := listObj.(*v1.PodList)
 		for i := range vList.Items {
+			if featuregate.DefaultFeatureGate.Enabled(featuregate.SuperClusterPooling) {
+				cname, ok := vList.Items[i].GetAnnotations()[utilconstants.LabelScheduledCluster]
```
christopherhein, line 169: Is is label or annotation?

incubator/virtualcluster/pkg/syncer/vnode/vnode.go
```diff
@@ -66,6 +67,11 @@ func NewVirtualNode(provider provider.VirtualNodeProvider, node *v1.Node) (vnode
 	labels := map[string]string{
 		constants.LabelVirtualNode: "true",
 	}
+
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.SuperClusterPooling) {
```
christopherhein, line 71: Do you for see this needing anything else specific to pooling? for pooling it must be assumed that all nodes are addressable to the meta tenants, would there be plans other things we need to add later to support additional features.

incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
Fei-Guo: Pod scheduling result is in annotation. Maybe the const name looks a little confusing.

incubator/virtualcluster/pkg/syncer/vnode/vnode.go
Fei-Guo: For vnode object, this is the only change I found necessary so far.  Note that it is not necessary that all nodes are addressable to the meta cluster. For example, in edge computing, the nodes and apiserver can be in different networks under the the help of a tunnel service to connect kubelet and apiserver.  Here it is necessary that meta cluster can address the super apiserver, which is already ensured by the syncer since it cannot be started otherwise.

incubator/virtualcluster/pkg/syncer/resources/pod/checker.go
christopherhein: yeah, that's what tripped me up.

incubator/virtualcluster/pkg/syncer/vnode/vnode.go
christopherhein: Oh, that's awesome.

PR #1457, opened 2021-04-01
Do not remove/update un-Labled resources in virtual cluster

weiling61:
do not remove/update unlabled (PublicObjectKey!=true) resources in virtual cluster

/assign @christopherhein @Fei-Guo


christopherhein:
Just a note for later, let's make sure we add a couple test cases to catch these additions.

PR #1472, opened 2021-04-13
Document excluded namespace configuration only available after v0.8

yiqigao217:


incubator/hnc/docs/user-guide/concepts.md
```diff
@@ -533,6 +533,8 @@ We are considering replacing this with the standard
 
 #### hnc.x-k8s.io/excluded-namespace (label on namespaces)
 
+***Excluded namespaces configuration is only available in HNC v0.8 and later***
```
adrianludwin, line 536: I think italics alone would have been fine but I won't hold this up :) Feel free to remove the hold if you like.

adrianludwin (commented):
/approve
/lgtm
/hold

Feel free to remove the hold if you like the bold italics. To my eyes, it looks a bit too strong (e.g. stronger than the L4 header) and regular italics would be fine but up to you

incubator/hnc/docs/user-guide/concepts.md
yiqigao217: I'm using the same format as https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/concepts.md#exceptions-and-propagation-control in the same page. So I will remove the hold :)

yiqigao217:
/hold cancel

FYI using the same format as https://github.com/kubernetes-sigs/multi-tenancy/blob/master/incubator/hnc/docs/user-guide/concepts.md#exceptions-and-propagation-control on the same page.

adrianludwin:
sgtm, thanks

Issue #1480, opened 2021-04-16
HNC: e2e tests intermittent failure due to no excluded-namespace label on `kube-system` ns, etc.


yiqigao217:
After #1444 is merged on April 6, our [testgrid](https://testgrid.k8s.io/wg-multi-tenancy-hnc#periodic-e2e-tests) showed 3/10 test failure as of today.

The errors are `no endpoints available` for webhook services. It turned out that we didn't add `hnc.x-k8s.io/excluded-namespace` label to namespaces `kube-system`, `kube-public`, `kube-node-lease` before running e2e tests.
(`hnc-system` get labelled by HNC but the other 3 excluded-namespaces are required to be labelled by users.)

We need to:
* Update our `test-e2e` Makefile target to verify excluded-namespace labels first before running e2e tests and early exit if they are not properly set (just like verifying `HNC_REPAIR` value before test).
* Update testgrid to add the excluded-namespace labels first.
* Add a troubleshooting doc in HNC for "webhook service `no endpoints available`" issue that they should check if the excluded-namespace labels are added properly.

yiqigao217:
The periodic test still [failed](https://testgrid.k8s.io/wg-multi-tenancy-hnc#periodic-e2e-tests).

The error was:
```
Ensuring all system namespaces are excluded from HNC...
Error from server (InternalError): Internal error occurred: failed calling webhook "namespaces.hnc.x-k8s.io": Post "https://hnc-webhook-service.hnc-system.svc:443/validate-v1-namespace?timeout=10s": dial tcp 10.98.73.157:443: connect: connection refused
make: *** [Makefile:268: exclude-system-namespaces] Error 1
```

This is due to the VWHConfiguration is already there before we label the other system namespaces.

yiqigao217:
For the new periodic test failure, I manually did the same commands as `hack/prow-run-e2e.sh` in a kind cluster. I noticed that sometimes even though our HNC pod is `Running` and the certs are ready in the VWHC, WH can still get
```
Error from server (InternalError): Internal error occurred: failed calling webhook "namespaces.hnc.x-k8s.io": Post "https://hnc-webhook-service.hnc-system.svc:443/validate-v1-namespace?timeout=10s": dial tcp 10.96.19.17:443: connect: connection refused
```
[...]

PR #1481, opened 2021-04-16
Ensure excluded-namespaces are properly configured before e2e tests

yiqigao217:
Part of #1480 

Force adding excluded-namespace labels to excluded namespaces before
running the e2e test so that it won't break the cluster. E.g. HNC object
webhook denies any object changes in namespaces without this label, it
could break `kube-system` namespace if the label is missing.

Add `test-e2e-batch` target and factor out the messages before e2e
tests for regular `test-e2e` and `test-e2e-batch`. Add link to
excluded-namespace concept.

Tested by `make test-e2e` and `make test-e2e-batch N=3`.

Prow config is unchanged because in `hack/prow-run-e2e.sh` for
postsubmit and periodic e2e tests, it's already using `test-e2e` target.

The output for `make test-e2e-batch N=10` looks like [this](https://gist.github.com/yiqigao217/d4b04bfe70afadae0fa31460e54752c5).

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -247,7 +247,28 @@ kind-deploy:
 # Go test suite has a 10m timeout, and the flag disables that timeout (as of
 # July 2020, these tests take ~15m and that number is expected to grow).
 .PHONY: test-e2e
-test-e2e:
+test-e2e: exclude-system-namespaces
+	go clean -testcache
+	go test -v -timeout 0 ./test/e2e/...
+
+# This batch test will run e2e tests N times on the current cluster the user
+# deployed (either kind or a kubernetes cluster), e.g. "make test-e2e-batch N=10"
+.PHONY: test-e2e-batch
+test-e2e-batch: exclude-system-namespaces
+	number=1 ; while [[ $$number -le $N ]] ; do \
+		echo $$number ; \
+    ((number = number + 1)) ; \
+		go clean -testcache ; \
+		go test -v -timeout 0 ./test/e2e/... ; \
+	done
+
+exclude-system-namespaces:
+	@echo
+	@echo "Ensuring all system namespaces are excluded from HNC..."
+	@kubectl label ns hnc-system hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@kubectl label ns kube-node-lease hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@kubectl label ns kube-public hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@kubectl label ns kube-system hnc.x-k8s.io/excluded-namespace=true --overwrite
 	@echo
 	@echo "If these tests fail due to the webhook not being ready, wait 30s and try again. Note that webhooks can take up to 30s to become ready"
 	@echo "after HNC is first deployed in a cluster."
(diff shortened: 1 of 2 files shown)


incubator/hnc/Makefile
```diff
@@ -247,7 +247,35 @@ kind-deploy:
...
+.PHONY: test-e2e-batch
+test-e2e-batch: before-e2e
+	number=1 ; while [[ $$number -le $N ]] ; do \
+		echo $$number ; \
+    ((number = number + 1)) ; \
+		go clean -testcache ; \
+		go test -v -timeout 0 ./test/e2e/... ; \
+	done
+
+before-e2e:
```
adrianludwin, line 265: Better to give this a more descriptive name, like `exclude-system-namespaces`

incubator/hnc/Makefile
```diff
@@ -247,7 +247,35 @@ kind-deploy:
...
+	@echo
+	kubectl label ns hnc-system hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	kubectl label ns kube-node-lease hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	kubectl label ns kube-public hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	kubectl label ns kube-system hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	@echo "If you see the above excluded namespaces are just labeled, it means your HNC installation was not complete."
```
adrianludwin, line 277: What does this mean? The first time you run this, you _should_ expect to see the namespaces being labeled, right?

adrianludwin: Regular users shouldn't be using the makefile so I wouldn't worry about "installing" HNC, so you can assume that this will only be read by developers who are willing to open up the makefile. E.g. you could do something like:

```
@echo Ensuring all system namespaces are excluded from HNC
@kubectl label ns ...
...
```

yiqigao217:
Hi @adrianludwin this is ready for another review.

BTW since we now force adding excluded-namespaces labels to system namespaces in the e2e test, do I still need to change the testgrid? Is it using `make test-e2e` target directly?

incubator/hnc/Makefile
```diff
@@ -247,7 +247,32 @@ kind-deploy:
...
+	@echo "Ensuring all system namespaces are excluded from HNC..."
+	@echo
+	kubectl label ns hnc-system hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	kubectl label ns kube-node-lease hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	kubectl label ns kube-public hnc.x-k8s.io/excluded-namespace=true --overwrite
+	@echo
+	kubectl label ns kube-system hnc.x-k8s.io/excluded-namespace=true --overwrite
 	@echo
```
adrianludwin, line 276: Any reason you wanted to keep this so verbose?

yiqigao217: Ah now I can remove the spaces. They were there for easier to read for whether each label is just added.

yiqigao217: Updated it to make these commands silent (with @).

yiqigao217:
> BTW since we now force adding excluded-namespaces labels to system namespaces in the e2e test, do I still need to change the testgrid? Is it using `make test-e2e` target directly?

`hack/prow-run-e2e.sh` used in postsubmit and periodic e2e tests is using `test-e2e` target, so no change is needed there.

adrianludwin (commented):
/lgtm
/approve

PR #1483, opened 2021-04-19
Handle bypassed webhook rules on excluded namespaces

### LINKED ISSUE #1443: HNC: Should not allow creating subns with excluded namespace names, e.g. `kube-system`

yiqigao217:
Fixes #1443

Remove any HNC CRs (hierarchyconfiguration and anchor) from the excluded
namespaces if they are successfully created bypassing the webhook.

Set `ActivitiesHalted: IllegalParent` condition if a namespace sets an
excluded namespace as a parent, e.g.
```
$ kubectl hns describe a
Hierarchy configuration for namespace a
  Parent: kube-system
  Children:
  - d (subnamespace)
  Conditions:
  - ActivitiesHalted (IllegalParent): Parent "kube-system" is an excluded namespace
```

Subnamespace anchor using an excluded namespace name would get
`Forbidden` state. Update the `kubectl hns describe` to show:
```
$ kubectl hns describe parent
Hierarchy configuration for namespace parent
  No parent
  Children:
  - d (Conflict subnamespace)
  - e (subnamespace)
  - f
  - kube-system (Forbidden subnamespace)
  No conditions
```

Tested by `make test` since there's no wehbook in the integration tests. Passed `make test-e2e`.

### DIFF
--- a/incubator/hnc/internal/kubectl/describe.go
+++ b/incubator/hnc/internal/kubectl/describe.go
@@ -39,7 +39,7 @@ var describeCmd = &cobra.Command{
 		nnm := args[0]
 		fmt.Printf("Hierarchy configuration for namespace %s\n", nnm)
 		hier := client.getHierarchy(nnm)
-		anms := client.getAnchorNames(nnm)
+		as := client.getAnchorStatus(nnm)
 
 		// Parent
 		if hier.Spec.Parent != "" {
@@ -53,11 +53,11 @@ var describeCmd = &cobra.Command{
 		for _, cn := range hier.Status.Children {
 			childrenAndStatus[cn] = ""
 		}
-		for _, cn := range anms {
+		for cn, st := range as {
 			if _, ok := childrenAndStatus[cn]; ok {
 				childrenAndStatus[cn] = "subnamespace"
 			} else {
-				childrenAndStatus[cn] = "missing subnamespace"
+				childrenAndStatus[cn] = st + " subnamespace"
 			}
 		}
 		if len(childrenAndStatus) > 0 {
--- a/incubator/hnc/internal/reconcilers/anchor.go
+++ b/incubator/hnc/internal/reconcilers/anchor.go
@@ -18,6 +18,7 @@ package reconcilers
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
@@ -72,12 +73,21 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
-	// Report "Forbidden" state and early exit if the namespace is not allowed to have subnamespaces
-	// but has bypassed the webhook and successfully created the anchor. Forbidden anchors won't have
-	// finalizers.
-	// TODO refactor/split the ExcludedNamespaces map for 1) reconciler exclusion and 2) subnamespaces exclusion
-	//  purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
+	// Always delete anchor (and any other HNC CRs) in the excluded namespaces and
+	// early exit.
 	if config.ExcludedNamespaces[pnm] {
+		// Since the anchors in the excluded namespaces are never synced by HNC,
+		// there are no finalizers on the anchors that we can delete them without
+		// removing the finalizers first.
+		log.Info("Deleting anchor in an excluded namespace")
+		return ctrl.Result{}, r.deleteInstance(ctx, log, inst)
+	}
+
+	// Report "Forbidden" state and early exit if the anchor name is an excluded
+	// namespace that should not be created as a subnamespace, but the webhook has
+	// been bypassed and the anchor has been successfully created. Forbidden
+	// anchors won't have finalizers.
+	if config.ExcludedNamespaces[nm] {
 		inst.Status.State = api.Forbidden
 		return ctrl.Result{}, r.writeInstance(ctx, log, inst)
 	}
@@ -331,6 +341,15 @@ func (r *AnchorReconciler) writeInstance(ctx context.Context, log logr.Logger, i
 			log.Error(err, "while updating on apiserver")
 			return err
 		}
+	}
+	return nil
+}
+
+// deleteInstance deletes the anchor instance. Note: Make sure there's no
+// finalizers on the instance before calling this function.
+func (r *AnchorReconciler) deleteInstance(ctx context.Context, log logr.Logger, inst *api.SubnamespaceAnchor) error {
+	if err := r.Delete(ctx, inst); err != nil {
+		return fmt.Errorf("while deleting on apiserver: %w", err)
 	}
 	return nil
 }
--- a/incubator/hnc/internal/reconcilers/anchor_test.go
+++ b/incubator/hnc/internal/reconcilers/anchor_test.go
@@ -22,6 +22,7 @@ var _ = Describe("Anchor", func() {
 	BeforeEach(func() {
 		fooName = createNS(ctx, "foo")
 		barName = createNSName("bar")
+		config.ExcludedNamespaces = nil
 	})
 
 	It("should create an subnamespace and update the hierarchy according to the anchor", func() {
@@ -50,11 +51,18 @@ var _ = Describe("Anchor", func() {
 		Eventually(getAnchorState(ctx, fooName, barName)).Should(Equal(api.Ok))
 	})
 
-	It("should set the anchor.status.state to Forbidden if the parent is not allowed to have subnamespaces", func() {
+	It("should remove the anchor in an excluded namespace", func() {
 		config.ExcludedNamespaces = map[string]bool{"kube-system": true}
 		kube_system_anchor_bar := newAnchor(barName, "kube-system")
 		updateAnchor(ctx, kube_system_anchor_bar)
-		Eventually(getAnchorState(ctx, "kube-system", barName)).Should(Equal(api.Forbidden))
+		Eventually(canGetAnchor(ctx, barName, "kube-system")).Should(Equal(false))
+	})
+
+	It("should set the anchor.status.state to Forbidden if the subnamespace is an excluded namespace", func() {
+		config.ExcludedNamespaces = map[string]bool{"kube-system": true}
+		foo_anchor_kube_system := newAnchor("kube-system", fooName)
+		updateAnchor(ctx, foo_anchor_kube_system)
+		Eventually(getAnchorState(ctx, fooName, "kube-system")).Should(Equal(api.Forbidden))
 	})
 
 	It("should set the anchor.status.state to Conflict if a namespace of the same name already exists", func() {
@@ -120,6 +128,17 @@ func getAnchorWithOffset(offset int, ctx context.Context, pnm, nm string) *api.S
 		return k8sClient.Get(ctx, nsn, anchor)
 	}).Should(Succeed())
 	return anchor
+}
+
+func canGetAnchor(ctx context.Context, pnm, nm string) func() bool {
+	return func() bool {
+		nsn := types.NamespacedName{Name: nm, Namespace: pnm}
+		anchor := &api.SubnamespaceAnchor{}
+		if err := k8sClient.Get(ctx, nsn, anchor); err != nil {
+			return false
+		}
+		return true
+	}
 }
 
 func updateAnchor(ctx context.Context, anchor *api.SubnamespaceAnchor) {
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -85,16 +85,20 @@ type HierarchyConfigReconciler struct {
 
 // Reconcile sets up some basic variables and then calls the business logic.
 func (r *HierarchyConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	if config.ExcludedNamespaces[req.Namespace] {
-		return ctrl.Result{}, nil
+	ns := req.NamespacedName.Namespace
+	log := loggerWithRID(r.Log).WithValues("ns", ns)
+
+	// Always delete hierarchyconfiguration (and any other HNC CRs) in the
+	// excluded namespaces and early exit.
+	if config.ExcludedNamespaces[ns] {
+		// Since singletons in the excluded namespaces are never synced by HNC, there
+		// are no finalizers on the singletons that we can delete them without
+		// removing the finalizers first.
+		return ctrl.Result{}, r.deleteSingletonIfExists(ctx, log, ns)
 	}
 
 	stats.StartHierConfigReconcile()
 	defer stats.StopHierConfigReconcile()
-
-	ns := req.NamespacedName.Namespace
-
-	log := loggerWithRID(r.Log).WithValues("ns", ns)
 
 	return ctrl.Result{}, r.reconcile(ctx, log, ns)
 }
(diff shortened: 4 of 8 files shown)


incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -72,12 +72,23 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		// removing the finalizers first.
+		log.Info("Will delete subnamespaceanchor since the namespace is an excluded namespace")
+		return ctrl.Result{}, r.deleteInstance(ctx, log, inst)
+	}
+
+	// Report "Forbidden" state and early exit if the anchor name is an excluded
+	// namespace that should not be created as a subnamespace, but the webhook has
+	// been bypassed and the anchor has been successfully created. Forbidden
+	// anchors won't have finalizers.
 	// TODO refactor/split the ExcludedNamespaces map for 1) reconciler exclusion and 2) subnamespaces exclusion
```
adrianludwin, line 89: Still a TODO?

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -335,6 +346,16 @@ func (r *AnchorReconciler) writeInstance(ctx context.Context, log logr.Logger, i
 	return nil
 }
 
+// deleteInstance deletes the anchor instance. Note: Make sure there's no
+// finalizers on the instance before calling this function.
+func (r *AnchorReconciler) deleteInstance(ctx context.Context, log logr.Logger, inst *api.SubnamespaceAnchor) error {
+	if err := r.Delete(ctx, inst); err != nil {
+		log.Error(err, "while deleting on apiserver")
+		return err
```
adrianludwin, line 354: We generally don't both log and return errors, right? Usually we do one or the other? An error that's returned to controller-runtime gets logged automatically, so we don't need to log it here.

Maybe: `return fmt.Errorf("while deleting on apiserver: %w", err)`?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -373,7 +377,10 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
 
 	// Sync this namespace with its current parent.
 	curParent := r.Forest.Get(inst.Spec.Parent)
-	if curParent != nil && !curParent.Exists() {
+	if config.ExcludedNamespaces[inst.Spec.Parent] {
+		log.Info("The parent is an excluded namespace that should not be set as a parent", "parent", inst.Spec.Parent)
+		ns.SetCondition(api.ConditionBadConfiguration, api.ReasonIllegalParent, fmt.Sprintf("Parent %q is an excluded namespace. HNC will treat this namespace as if it has no parent.", inst.Spec.Parent))
```
adrianludwin, line 382: Why is this BadConfiguration and not ActivitiesHalted?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -581,6 +588,37 @@ func (r *HierarchyConfigReconciler) writeHierarchy(ctx context.Context, log logr
...
+	if err != nil {
+		return err
+	}
+
+	// Early exit if the singleton doesn't exist.
+	if inst.CreationTimestamp.IsZero() {
+		return nil
+	}
+
+	log.Info("Will delete hierarchyconfiguration since the namespace is an excluded namespace")
```
adrianludwin, line 605: "Deleting illegal HierarchyConfiguration in excluded namespace"

incubator/hnc/internal/reconcilers/hierarchy_config_test.go
```diff
@@ -31,6 +31,31 @@ var _ = Describe("Hierarchy", func() {
...
+		config.ExcludedNamespaces = map[string]bool{"kube-system": true}
+		exHier := newHierarchy("kube-system")
+		exHier.Spec.Parent = barName
+		updateHierarchy(ctx, exHier)
+
+		// Verify the hierarchyconfiguration singleton is deleted. Give apiserver
+		// 0.5 second here to be able to process the deletion of the singleton
+		// before we constantly retry getting the singleton with Get().
+		time.Sleep(500 * time.Millisecond)
+		Eventually(canGetHierarchy(ctx, "kube-system")).ShouldNot(Succeed())
```
adrianludwin, line 45: `canGetHierarchy` should return a function that `Eventually` can retry, not an answer that never changes (e.g. see `hasCondition`). Then you get get rid of `Sleep`.

As a general rule, use `Expect()` and not `Eventually()` if you're not truly retrying.

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: I was initially thinking the activities can still continue as if this namespace has no parent, but since `ParentMissing` condition is similar to this case and it didn't treat the namespace as if it doesn't have a parent, I guess it makes sense to have this one as `ActivitiesHalted` too. In that case, the descendants would have `ActivitiesHalted: AncestorHaltActivities` condition too, so that it's more explicit to unveil the issue to users, SGTY?

incubator/hnc/internal/kubectl/describe.go
```diff
@@ -57,7 +57,7 @@ var describeCmd = &cobra.Command{
 			if _, ok := childrenAndStatus[cn]; ok {
 				childrenAndStatus[cn] = "subnamespace"
 			} else {
-				childrenAndStatus[cn] = "missing subnamespace"
+				childrenAndStatus[cn] = "missing subnamespace; may be an excluded namespace"
```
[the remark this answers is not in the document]
yiqigao217: sg. Updated and it will look like:
```
$ kubectl hns describe parent
Hierarchy configuration for namespace parent
  No parent
  Children:
  - d (Conflict subnamespace)
  - e (subnamespace)
  - f
  - kube-system (Forbidden subnamespace)
  No conditions
```

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: Removed.

incubator/hnc/internal/reconcilers/anchor_test.go
```diff
@@ -50,11 +50,18 @@ var _ = Describe("Anchor", func() {
...
+	It("should remove the anchor in an excluded namespace", func() {
 		config.ExcludedNamespaces = map[string]bool{"kube-system": true}
 		kube_system_anchor_bar := newAnchor(barName, "kube-system")
 		updateAnchor(ctx, kube_system_anchor_bar)
-		Eventually(getAnchorState(ctx, "kube-system", barName)).Should(Equal(api.Forbidden))
+		Eventually(canGetAnchor(ctx, barName, "kube-system")).ShouldNot(Succeed())
+	})
+
+	It("should set the anchor.status.state to Forbidden if the subnamespace is an excluded namespace", func() {
+		config.ExcludedNamespaces = map[string]bool{"kube-system": true}
```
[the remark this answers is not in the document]
yiqigao217: Since `Ginkgo will normally run It blocks synchronously`, it's safer to have it set right above the test?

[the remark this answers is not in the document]
adrianludwin: I'm just wondering if we ever reset `config.ExcludedNamespaces` to its initial value between tests. Otherwise, the value we set in one test could persist to another test.

Maybe in the "before" block we could always reset it?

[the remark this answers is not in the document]
yiqigao217: Ah no it won't reset to initial value between tests, so the value set in one test could persist to another (if the `It` block is called before the other).

The other reason I prefer placing it in the `It` block is it's local/closer to the test case for readablitiy.

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -373,7 +377,10 @@ func (r *HierarchyConfigReconciler) syncParent(log logr.Logger, inst *api.Hierar
 
 	// Sync this namespace with its current parent.
 	curParent := r.Forest.Get(inst.Spec.Parent)
-	if curParent != nil && !curParent.Exists() {
+	if config.ExcludedNamespaces[inst.Spec.Parent] {
+		log.Info("The parent is an excluded namespace that should not be set as a parent", "parent", inst.Spec.Parent)
```
adrianludwin, line 381: Nit: you repeat the word "parent" three times here. Also, the log should record what HNC is doing/observing, e.g.:

`log.Info("Excluded namespace illegally set as parent", "parent", inst.Spec.Parent)`

Even better, for this and the message on line 385:

`log.Info("Setting ConditionActivitiesHalted: excluded namespace set as parent", "parent", inst.Spec.Parent)`

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -581,6 +588,37 @@ func (r *HierarchyConfigReconciler) writeHierarchy(ctx context.Context, log logr
...
+	}
+
+	log.Info("Deleting illegal HierarchyConfiguration in excluded namespace")
+
+	// If the CRD is being deleted, we don't need to delete it separately. It will
+	// be deleted with the CRD.
+	if deletingCRD {
+		return nil
+	}
+
```
adrianludwin, line 612: This is a bit odd. We say we're deleting the HC but we might actually not do anything. Imagine a case where the HC _isn't_ being deleted and the logs would look weird.

How about:

```
if deletingCRD {
  log.Info("Already deleting illegal HC in excluded namespace")
  return nil
}
log.Info("Deleting illegal...")
```

adrianludwin (commented):
lgtm after two minor nits and one question about resetting the excluded namespaces config in the tests.

incubator/hnc/internal/reconcilers/anchor_test.go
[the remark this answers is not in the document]
adrianludwin: _Setting_ the value in the `It` block sgtm. But maybe _clear_ it in the `BeforeEach` block so you know at the beginning of each test that it's starting in a known-good state?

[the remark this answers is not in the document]
yiqigao217: Maybe clear it in the `AfterEach`? Since in the `It` block, I set the value using a new map, not adding a new key/value pair to the existing map?

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Updated to `Setting Condition...`

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: sg. The first log, I used `Already deleting illegal HC in excluded namespace since the CRD is being deleted`.

incubator/hnc/internal/reconcilers/anchor_test.go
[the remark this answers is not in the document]
yiqigao217: Eventually, I decided to move the setting to `BeforeEach` from each `It` block, since `kube-system` is a well-know system(excluded) namespace and since the value is always set to `map[string]bool{"kube-system": true}`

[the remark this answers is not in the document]
adrianludwin: I actually do prefer it being explicit in the test. I'd prefer to see the following in `BeforeEach`:

`config.ExcludedNamespaces = nil`

That's all I'm going for...

incubator/hnc/internal/reconcilers/hierarchy_config.go
yiqigao217: Discussed offline, updated to `HC in excluded namespace is already being deleted` and `Deleting illegal HC in excluded namespace`

incubator/hnc/internal/reconcilers/anchor_test.go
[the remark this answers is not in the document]
yiqigao217: Done.

adrianludwin (commented):
/lgtm
/approve
/assign @rjbez17 
/hold

Hey Ryan, can you PTAL? No rush, I know there's a lot going on in your town right now.

incubator/hnc/internal/reconcilers/anchor.go
```diff
@@ -72,12 +73,21 @@ func (r *AnchorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
-	// Report "Forbidden" state and early exit if the namespace is not allowed to have subnamespaces
-	// but has bypassed the webhook and successfully created the anchor. Forbidden anchors won't have
-	// finalizers.
-	// TODO refactor/split the ExcludedNamespaces map for 1) reconciler exclusion and 2) subnamespaces exclusion
-	//  purposes. See issue: https://github.com/kubernetes-sigs/multi-tenancy/issues/495
+	// Always delete anchor (and any other HNC CRs) in the excluded namespaces and
+	// early exit.
 	if config.ExcludedNamespaces[pnm] {
+		// Since the anchors in the excluded namespaces are never synced by HNC,
```
rjbez17, line 79: Just to double check. In the scenario that we have an existing structure with multiple anchors, and the allowCascadingDelete is true, if we update HNC to now exclude the parent namespace, this will delete the anchor resources but NOT delete the children namespaces and they will be be parent-less, correct?

rjbez17:
lgtm other than the one question.

incubator/hnc/internal/reconcilers/anchor.go
yiqigao217: Ah I didn't think of this scenario.

I just tested this scenario, both the subnamespace and anchor were not deleted (before I manually remove the finalizer in the anchor).

I created 2 level subns in ns `ex`, then exclude the `ex` and redeploy HNC. After that, I got:
```
$ k hns tree ex
ex
└── [s] sub (1,2)
    └── [s] grand (3)

[s] indicates subnamespaces

Conditions:
1) BadConfiguration (SubnamespaceAnchorMissing): The anchor is missing in the parent namespace
2) ActivitiesHalted (IllegalParent): Parent "ex" is an excluded namespace
3) ActivitiesHalted (AncestorHaltActivities): Propagation paused in the subtree of "sub" due to ActivitiesHalted condition

$ k get subns -A
NAMESPACE   NAME    AGE
ex          sub     12m
sub         grand   8m53s
```

The `badconfiguration` condition was due to excluded namespaces are not synced in the forest (memory), so it's invisible when we check forest.

The anchor object was still there even though we tried deleting it here. The reason is that after redeploying HNC, this anchor reconciliation always early exit here so the finalizers are not removed.

I then manually delete the anchor finalizer, the anchor was gone and the subnamespace `sub` and `grand` were still there having the same condition.

So everything looks like the intended behavior (conditions reported, not deleting ns), except my manual removal of anchor finalizers in the new excluded namespace. I will add this guidance in the excluded-namespace configuration doc. Does it sgty?

rjbez17: Awesome yep sgtm.

PR #1485, opened 2021-04-22
Exclude namespaces before deploying HNC

### LINKED ISSUE #1480: HNC: e2e tests intermittent failure due to no excluded-namespace label on `kube-system` ns, etc.

yiqigao217:
Fixes #1480

Exclude system namespaces `kube-system`, `kube-public`,
`kube-node-lease`before deploying HNC in `deploy` target in Makefile.

Tested by running `./incubator/hnc/hack/prow-run-e2e.sh`.

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -180,12 +180,23 @@ controller-gen:
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config.
 #
-# We only delete and redeploy the deployment, and nothing else, because a)
-# deleting the CRDs will cause all the existing hierarchy configs to be wiped
-# away and b) if we don't delete the deployment, a new image won't be pulled
-# unless the tag changes.
+# We only delete the deployment and the validatingwebhookconfiguration if they
+# exist before applying the manifest, because
+# a) deleting the CRDs will cause all the existing CRs to be wiped away;
+# b) if not deleting the deployment, a new image won't be pulled unless the tag changes;
+# c) if not deleting the validatingwebhookconfiguration, we cannot label
+#    namespaces to exclude them if the HNC pod is already in a bad state.
+#
+# Then we ensure the system namespaces are excluded before we deploy HNC. This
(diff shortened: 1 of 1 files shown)


adrianludwin (commented):
Should we:

a) Remove the labels from `make test-e2e`?
b) Add the labelling in `make deploy`?

The second one might work better here as well.

yiqigao217:
> Should we:
> 
> a) Remove the labels from `make test-e2e`?
> b) Add the labelling in `make deploy`?
> 
> The second one might work better here as well.

Good point! Yes, I added # 2 instead. As for # 1, I left it as is, in case anyone deploys HNC by applying manifests. SGTY?

adrianludwin:
> Good point! Yes, I added # 2 instead. As for # 1, I left it as is, in case anyone deploys HNC by applying manifests. SGTY?

Good call, but maybe comment that in the makefile so someone like me doesn't try to delete it :)

yiqigao217:
SG. @adrianludwin PTAL, thanks!

adrianludwin (commented):
/lgtm

Issue #1295, opened 2020-12-01
HNC: document that some E2E tests cannot run with Hierarchy Controller


GinnyJI:
Before running E2E test suite, Hierarchy Controller has to be uninstalled first, otherwise any test that needs HNC_REPAIR will cause problems. I don't know if and how we should document this.

adrianludwin:
We already print a big warning if HNC_REPAIR is _not_ set. Maybe if
HNC_REPAIR _is_ set, we need to print a big warning saying "Make sure HNC
isn't being managed by an operator (like ACM or Config Sync) since we're
about to make changes to it!" followed by a 5s delay so that you can ctrl-c
your way out of it?

adrianludwin:
Fixed by #1487
/close

PR #1487, opened 2021-04-26
Update cert-rotator to 0.2.0; tweak test targets

### LINKED ISSUE #1484: HNC: Update cert-controller
Cer-controller updated its WH support from v1beta1 to v1 (https://github.com/open-policy-agent/cert-controller/pull/34). HNC has already updated to v1 for WHs (https://github.com/kubernetes-sigs/multi-tenancy/pull/1434) but our cert-controller still supports v1beta1:
https://github.com/kubernetes-sigs/multi-tenancy/blob/a6253034e7d197089b6d3003c459c5f394355dd7/incubator/hnc/vendor/github.com/open-policy-agent/cert-controller/pkg/rotator/rotator.go#L72-L74

It's currently working because the builtin type is still serving both v1beta1 and v1.

However after K8s v1.22, v1beta1 WH api will be [unserved](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-22). We need to upgrade our cert-controller so HNC won't break.

adrianludwin:
This change uses an "official" cert-rotator release for the first time,
and includes support for v1 of all K8s APIs.

It also introduces the "test-smoke" target that only runs the quickstart
tests, to allow for faster local development. This allows developers to
rely more heavily on the postsubmit and periodic tests we worked so hard
to set up for HNC.

Finally, it clarifies some of the warnings around HNC_REPAIR.

Tested: make test-smoke passes. Tried running 'make test-e2e' with and
without HNC_REPAIR set to verify that the warnings look correct in both
cases.

Fixes #1484, #1295


incubator/hnc/Makefile
```diff
@@ -289,14 +296,27 @@ exclude-system-namespaces:
...
 	@echo "HNC_REPAIR IS NOT SET. CRITICAL TESTS WILL BE SKIPPED."
-	@echo "You have 5s to hit Ctrl-C to cancel and try again."
-	@echo "******************************************************"
 	@echo
-	@sleep 5
+	@echo "If HNC is installed via an operator (e.g. GKE Hierarchy"
+	@echo "Controller), this is probably what you want to do. Otherwise,"
+	@echo "consider setting HNC_REPAIR to run the full set of tests."
+else
+	@echo "HNC_REPAIR IS SET; ENSURE YOU HAVE PERMISSION TO DELETE HNC."
```
yiqigao217, line 309: Why `ensure you have permission to delete HNC` here?

yiqigao217 (commented):
One question otherwise lgtm.

incubator/hnc/Makefile
adrianludwin: That's basically what HNC_REPAIR tests do - they delete components of HNC and repair them. Maybe "modify HNC configs" would be a better description here?

adrianludwin: Changed to "modify HNC configs."

yiqigao217: Ah I see, it's simply for running e2e test cases. I misunderstood with the scenario that HNC is installed by operator that you will clean it up by deleting HNC.

BTW just confirming, for the case of using operator with `HNC_REPAIR` not set, the reinstalling HNC will be trigger immediately by the operator when HNC is deleted and the missing `HNC_REPAIR` won't cause other e2e test failure, right?

adrianludwin: If HNC_REPAIR isn't set, we just skip all the tests that need us to break and/or reinstall HNC.

PR #1488, opened 2021-04-27
Sync object propagation on newly labeled namespaces

### LINKED ISSUE #1448: HNC: `propagate.hnc.x-k8s.io/select` won't trigger object propagation on newly labelled namespaces
In a namespace hierarchy like this:
```
a
├── [s] b
└── [s] c

[s] indicates subnamespaces
```
If a role in the namespace `a` has
```
    annotations:
      propagate.hnc.x-k8s.io/select: score.networkpolicy/allow-shared-ingress
```
, it won't be propagated into `b` or `c` if they don't have the `score.networkpolicy/allow-shared-ingress` label on the namespace.

The issue is that if we now add `score.networkpolicy/allow-shared-ingress` label to `b` or `c`, the object should be propagated to these namespaces. However currently, the object propagation is not triggered unless there's an update on the source object.

We should make sure the object propagation is triggered when the label is added/removed from a child namespace.

yiqigao217:
Detect namespace label changes. If the labels are updated, ensure
syncing all the objects (including non-existing yet but possible
propagated objects) in the namespace, so that we will not miss any
propagation label changes that should trigger propagation or
unpropagation.

Add integration tests in reconciler/object_test.go.

Tested manually that labeling and unlabeling a child with the propagate
label won't do anything before the fix, but triggers propagation and
unpropagation accordingly after the fix. Also tested with `make test`,
passed after the fix ant failed before the fix.

### DIFF
--- a/incubator/hnc/internal/reconcilers/hierarchy_config.go
+++ b/incubator/hnc/internal/reconcilers/hierarchy_config.go
@@ -144,13 +144,13 @@ func (r *HierarchyConfigReconciler) reconcile(ctx context.Context, log logr.Logg
 	r.updateFinalizers(ctx, log, inst, nsInst, anms)
 
 	// Sync the Hierarchy singleton with the in-memory forest.
-	initial := r.syncWithForest(log, nsInst, inst, deletingCRD, anms)
+	needUpdateObjects := r.syncWithForest(log, nsInst, inst, deletingCRD, anms)
 
 	// Write back if anything's changed. Early-exit if we just write back exactly what we had and this
 	// isn't the first time we're syncing.
 	updated, err := r.writeInstances(ctx, log, origHC, inst, origNS, nsInst)
-	updated = updated || initial
-	if !updated || err != nil {
+	needUpdateObjects = updated || needUpdateObjects
+	if !needUpdateObjects || err != nil {
 		return err
 	}
 
@@ -223,7 +223,8 @@ func (r *HierarchyConfigReconciler) updateFinalizers(ctx context.Context, log lo
 // guarded by the forest mutex, which means that none of the other namespaces being reconciled will
 // be able to proceed until this one is finished. While the results of the reconiliation may not be
 // fully written back to the apiserver yet, each namespace is reconciled in isolation (apart from
-// the in-memory forest) so this is fine.
+// the in-memory forest) so this is fine. Return true, if the namespace is just synced or the
+// namespace labels are changed that requires updating all objects in the namespaces.
 func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *corev1.Namespace, inst *api.HierarchyConfiguration, deletingCRD bool, anms []string) bool {
 	r.Forest.Lock()
 	defer r.Forest.Unlock()
@@ -257,9 +258,9 @@ func (r *HierarchyConfigReconciler) syncWithForest(log logr.Logger, nsInst *core
 	r.syncConditions(log, inst, ns, deletingCRD, hadCrit)
 
 	// Sync the tree labels. This should go last since it can depend on the conditions.
-	r.syncLabel(log, nsInst, ns)
+	nsCustomerLabelUpdated := r.syncLabel(log, nsInst, ns)
 
-	return initial
+	return initial || nsCustomerLabelUpdated
 }
 
 // syncExternalNamespace sets external tree labels to the namespace in the forest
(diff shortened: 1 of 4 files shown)


yiqigao217:
/hold
until the issue in the integration test is figured out.

@adrianludwin the newly-added integration test always passed, no matter before or after the fix. I guess the source object is somehow reconciled very frequently in the integration test environment, so it can sync propagated objects in children frequent enough that we cannot reproduce the result in #1448 before the fix. I'm not sure what else I can try to fix this integration test issue.

You can search for `???????` in the code to see two of my comments in the code.

adrianludwin:
> ******????? (There's issue with the newly-added integration tests that
> they passed after the fix but also passed before the fix)

Did you mean to submit this?

incubator/hnc/internal/reconcilers/hierarchy_config.go
```diff
@@ -220,7 +222,7 @@ func (r *HierarchyConfigReconciler) updateFinalizers(ctx context.Context, log lo
 // be able to proceed until this one is finished. While the results of the reconiliation may not be
 // fully written back to the apiserver yet, each namespace is reconciled in isolation (apart from
 // the in-memory forest) so this is fine.
```
adrianludwin, line 224: Document return values? Do we really need two or do they just get "or'd" together?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -231,6 +231,118 @@ var _ = Describe("Exceptions", func() {
 			})
 		}
 	})
+
+	Context("Update the descendant namespaces after 'select' exception annotation is set", func() {
```
adrianludwin, line 235: Add the issue number in the context (I've just submitted a change to make it easy to run the tests for a single issue)

adrianludwin (commented):
I don't have any great ideas about why this isn't working - I'd just add a ton of log messages so you can see exactly what gets called and when and try to debug it that way (make sure you only submit the most essential messages though). E.g. with a reasonable sleep of 10s (which should show up nicely in the logs) you should be pretty sure that HNC is quiet, and so you'll be able to see why it's not behaving as you expect.

adrianludwin:
Maybe there's something special about adding a label via `kubectl` as opposed to updating the entire namespace via the client? I hope not, but it's possible.

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -231,6 +231,118 @@ var _ = Describe("Exceptions", func() {
...
+				})
+				Eventually(hasObject(ctx, api.RoleResource, names[nolabelchild], "testrole")).Should(BeFalse(), "When propagating testrole to %s", names[nolabelchild])
+				Consistently(hasObject(ctx, api.RoleResource, names[labeledchild], "testrole")).Should(BeTrue(), "When propagating testrole to %s", names[labeledchild])
+
+				// ?????????????
+				// For issue 1448, we may need to add a sleep to ensure the source object
+				// reconciliation is done and not enqueued elsewhere, but even with 360s
+				// sleep, the test still passed before the fix for 1448 (manual tests before
+				// the fix failed).
+				//time.Sleep(60 * time.Second)
```
adrianludwin, line 312: Maybe the sleep should be _after_ the updates, not before?

incubator/hnc/internal/reconcilers/object_test.go
```diff
@@ -231,6 +231,118 @@ var _ = Describe("Exceptions", func() {
...
+
+				// Add the label to the namespace if the value is not empty.
+				if tc.toLabel != "" {
+					ns := getNamespace(ctx, names[tc.toLabel])
+					ns.SetLabels(map[string]string{label: "true"})
+					updateNamespace(ctx, ns)
+				}
```
yiqigao217, line 319: Ah figured out the issue! It's the problem in setting labels. Here it resets the labels, including the tree labels! That's why the parent namespace still gets all object synced (because the ns is considered as `updated` even before the fix).

I will update the logic for label and unlabel here and resubmit.

incubator/hnc/internal/reconcilers/object_test.go
yiqigao217: I added the issue number to the `It` block since 2 out 5 test cases are for issue #1448 and the `HNC_FOCUS` is on e2e tests (integration and unit tests are fast so no focus test is needed).

adrianludwin: Oh right :)

adrianludwin (commented):
/lgtm
/hold
/assign @rjbez17

Issue #772, opened 2020-05-27
HNC: cannot propagate "admin" rolebindings anymore


adrianludwin:
This used to work, but I think a newer version of K8s has a `deletecollections` verb that we didn't cover.

PR #1489, opened 2021-04-27
Support cluster-admin propagation and update docs

### LINKED ISSUE #1311: Give HNC permission to all verbs on all resources
`user "system:serviceaccount:hnc-system:default" (groups=["system:serviceaccounts" "system:serviceaccounts:hnc-system" "system:authenticated"]) is attempting to grant RBAC permissions not currently held`

When a sub-namespace is created under a parent namespace containing RoleBindings that are attached to a Role that have permissions such as:

```
- apiGroups:
  - '*'
  resources:
  - 'persistentvolumeclaims'
  verbs:
  - '*'
```

The hnc-controller will not have enough permissions to propogate the rolebinding. Note the key problem here is that the hnc-manager-role does not have `verbs: '*'` permissions itself. 

I don't suspect adding this permission is desirable, and this error has given me some roles to review, but I do think it would be desirable to note somewhere in the documentation or installation guide that this might be among the first issues a new user might come across if they've got some quite permissive roles hanging around, created by themselves or through third-party controller installs. 

Thank you for your work on this project!

adrianludwin:
Rather than continuing to play whack-a-mole with the list of verbs HNC
can propagate, this change allows it to perform all verbs on all
resources. This is equivalent to `cluster-admin` so I've updated the
docs accordingly (see also issue #1311).

I also noticed that the docs referred to K8s v1.15, which is no longer
supported, so I updated them to v1.16.

Finally, this change adds the `HNC_FOCUS` makefile var, allowing you to
say something like:

```
HNC_FOCUS=772 make test
```

which only runs the e2e tests with "772" in the title.

Tested: All quickstart e2e tests pass. Updated the test for issue #772
and verified that it failed without the other changes in this commit,
and passed with them.

Fixes #1311

### DIFF
--- a/incubator/hnc/Makefile
+++ b/incubator/hnc/Makefile
@@ -257,10 +257,19 @@ kind-deploy:
 # Note the `-timeout 0` that's passed to the `go test` command - by default, a
 # Go test suite has a 10m timeout, and the flag disables that timeout (as of
 # July 2020, these tests take ~15m and that number is expected to grow).
+#
+# To focus on specific tests, use the HNC_FOCUS var as follows:
+#
+# HNC_FOCUS=772 make test-e2e # only runs the test for issue 772
+# HNC_FOCUS=Quickstart make test-e2e # Runs all tests in the "Quickstart" Describe block
 .PHONY: test-e2e
 test-e2e: exclude-system-namespaces
 	go clean -testcache
+ifndef HNC_FOCUS
 	go test -v -timeout 0 ./test/e2e/...
+else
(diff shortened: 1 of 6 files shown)


adrianludwin:
Preview the doc changes here:

* [FAQ](https://github.com/adrianludwin/multi-tenancy/blob/rbac/incubator/hnc/docs/user-guide/faq.md#what-are-hncs-minimum-requirements)
* [How-to](https://github.com/adrianludwin/multi-tenancy/blob/rbac/incubator/hnc/docs/user-guide/how-to.md#admin-install)

yiqigao217 (commented):
/lgtm

Issue #1492, opened 2021-04-27
HNC: find a new way to test unpropagatable objects


adrianludwin:
Since #1489, HNC has been able to propagate `cluster-admin` rolebindings and therefore the test for #328, #605 and #771 have broken. We need to find another way to stop HNC from propagating these objects, possibly by having some special-purpose test webhook or something.

adrianludwin:
Another possible way to do it:
* Create a Role in the parent and forbid it from being propagated
* Create a RoleBinding in the parent

Usually, RBs cannot be created with the Roles they refer to existing first, so this *should* cause an error.

adrianludwin:
Replaced by https://github.com/kubernetes-sigs/hierarchical-namespaces/issues/13

/close

PR #1493, opened 2021-04-27
Disable broken test

adrianludwin:
This test assumed that HNC was not able to propagate cluster-admin
rolebindings, but this just changed so the test is now failing.

Issue #1492 tracks re-enabling this, but for now we'll disable.


adrianludwin:
/assign @rjbez17 
/assign @yiqigao217 

May I get a quick lgtm on this one?

yiqigao217 (commented):
/lgtm

PR #1499, opened 2021-04-29
Increase HNC release timeout to 30m (master branch)

adrianludwin:
Apparently the default limit is 10m and we've now gone over that (we
were averaging around 7m, I wonder if GCB now uses a slower machine
since little has changed on our side). Anyway, 30m should be enough for
the foreseeable future.


yiqigao217 (commented):
/lgtm

PR #1522, opened 2021-12-04
Allow the AlwaysPullImages admission plugin to satisfy benchmark

### LINKED ISSUE #1516: AlwaysPullImages admission plugin to satisfy benchmark
The "Require always imagePullPolicy" benchmark appears to try to run a pod with an ImagePullPolicy of "Never": https://github.com/kubernetes-sigs/multi-tenancy/blob/master/benchmarks/kubectl-mtb/test/benchmarks/require_always_pull_image/require_always_pull_image.go#L40

My cluster will allow that container to run, but I have the AlwaysPullImages admission plugin enabled which will change the field to "Always" before it starts: https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages

That plugin is recommended by the CIS Benchmarks for Kubernetes, so it may be common: https://workbench.cisecurity.org/sections/639538/recommendations/1047690 (requires login)

Is it possible for the benchmark to detect that?

mac-chaffee:
Fixes #1516 

This PR checks the kube-apiserver pod Command to determine if the [AlwaysPullImages admission plugin](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#alwayspullimages) is enabled. If it is, that is sufficient to satisfy the require_always_pull_image benchmark.

The Admission plugin works by allowing users to create pods with whatever imagePullPolicy they want, but it just modifies the value to "Always" before the pod starts up. As a result, without this PR, the benchmark fails unless you have some other validatingwebhook like gatekeeper enforcing the policy.

### DIFF
--- a/benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
+++ b/benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
@@ -160,7 +160,7 @@ func validateFlags(cmd *cobra.Command) error {
 
 	_, err = benchmarkRunOptions.ClusterAdminClient.CoreV1().Namespaces().Get(context.TODO(), benchmarkRunOptions.TenantNamespace, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("invalid namespace")
+		return err
 	}
 
 	return nil
--- a/benchmarks/kubectl-mtb/test/benchmarks/require_always_pull_image/require_always_pull_image.go
+++ b/benchmarks/kubectl-mtb/test/benchmarks/require_always_pull_image/require_always_pull_image.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/multi-tenancy/benchmarks/kubectl-mtb/bundle/box"
 	"sigs.k8s.io/multi-tenancy/benchmarks/kubectl-mtb/pkg/benchmark"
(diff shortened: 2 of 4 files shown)


benchmarks/kubectl-mtb/test/benchmarks/require_always_pull_image/require_always_pull_image.go
```diff
@@ -56,6 +66,23 @@ var b = &benchmark.Benchmark{
 	},
 }
 
+func isAlwaysPullImagesAdmissionPluginEnabled(options types.RunOptions) (bool, error) {
```
JimBugwadia, line 69: @mac-chaffee - wondering if checking the apiserver args is the best option. 

If the AlwaysPullImages admission control is enabled it will update the pod `imagePullPolicy` and we can simply check the value. Can we change the check to compare the results, if the operation that creates a pod with `ImagePullPolicy: Never` (or IfNotPresent) is allowed?

Besides, the check will only work for clusters where the `kube-apiserver` is accessible as a pod. That seeems typical, but not sure if its guaranteed.

JimBugwadia (changes requested):

benchmarks/kubectl-mtb/test/benchmarks/require_always_pull_image/require_always_pull_image.go
mac-chaffee: You're right, I forget people can run kube-apiserver as anything other than a mirrored pod like how kubeadm does it.

And yes you're right that the `metav1.DryRunAll` means the admission control plugins do take effect, and you can check the returned Pod object to see if any fields were mutated.

benchmarks/kubectl-mtb/internal/kubectl-mtb/run.go
```diff
@@ -160,7 +160,7 @@ func validateFlags(cmd *cobra.Command) error {
 
 	_, err = benchmarkRunOptions.ClusterAdminClient.CoreV1().Namespaces().Get(context.TODO(), benchmarkRunOptions.TenantNamespace, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("invalid namespace")
+		return err
```
mac-chaffee, line 163: If you have no connectivity to the cluster, this is the first API call made which means you get a confusing "invalid namespace" error instead of the true error, which is something like "i/o timeout".

Even when the namespace doesn't exist, the message is "namespace 'foo' not found"

JimBugwadia (approved):

